PR #8, opened 2023-11-14
fix: Handle import for both pydantic v1 and v2 in generate_response.py

IlyaTorch:
The fix is related to [this comment](https://github.com/vellum-ai/vellum-client-python/issues/7#issuecomment-1783016451)


noanflaherty (approved):
Thank you for your help with this!

PR #20, opened 2024-04-17
Create custom wrappers in our sdks for easily executing external evals

dvargas92495:
Used [this notion](https://www.notion.so/vellum-ai/Externally-Evaluated-Entities-Triggering-Test-Suites-as-Code-9d64e00548d94326a5fe169385038a8b?pvs=4#2340b17b8d0b4188855a887eced69a64) doc as the basis for the test + implementation. The first commit is a test, the second commit is the implementation.

Unfortunately, in the pursuit of this implementation, I discovered several blockers that we will need to start resolving. I will inline them all in the files as well as list them below. They include:
- [x] adding `requests_mock` to the repos dependencies
    - OR alternatively, making an active decision to run these tests end-to-end live 👀 
- [x] Align on method for getting test suite stuff by `name`
- [x] Align on method for mapping test case input values by name
- [x] Align on method for sending outputs from external execution
- [x] Expose external exec config


src/vellum/lib/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .test_suite import VellumTestSuite
```
dvargasfuertes: I was getting `Import ".test_suite" could not be resolved` warnings here, even though it's exactly what we do in other directories... any insight into what I'm doing wrong here?

src/vellum/lib/test_suite.py
dvargasfuertes, line 1: Please review the test file first to ensure we are aligned on end user behavior

tests/lib/test_test_suite.py
```diff
@@ -0,0 +1,39 @@
...
+    # AND a vellum test suite setup to handle that external eval
+    # - Note that with TestSuites as code, we could define the test suite inline
+    example_test_suite = VellumTestSuite("example_test_suite")
+
+    # AND Vellum successfully evaluates the external execution
+    test_suite_run_id = uuid4()
+    requests_mock.post(url="https://api.vellum.ai/v1/test_suite_runs")
+    requests_mock.get(
+        url=f"https://api.vellum.ai/v1/test_suite_runs/{test_suite_run_id}/executions?expand[0]=results.metric_results.metric_definition",
+    )
```
dvargasfuertes: blocking decision 1 - do we want to use requests_mock to mock our api or hit Vellum directly?

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,174 @@
...
+    def _ensure_metric_results_are_loaded(self):
+        if self.test_suite_run_executions is not None:
+            return
+
+        self._refresh_test_suite_run_state()
+
+        while self.test_suite_run_state in {TestSuiteRunState.QUEUED, TestSuiteRunState.RUNNING}:
+            time.sleep(1)
+            self._refresh_test_suite_run_state()
+            # TODO: Max retries?
```
dvargasfuertes: Do we want our poll timeout and retry limit to be configurable? what should defaults be?

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+
+class VellumTestSuite:
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.client = Vellum(
+            api_key=os.environ.get("VELLUM_API_KEY"),
+        )
+        self.test_suite_id = self._get_test_suite_id_by_name(name)
+
+    def _get_test_suite_id_by_name(self, name: str) -> str:
```
dvargasfuertes: blocking decision 2 - we don't have a way to get test suite stuff by name

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+        return uuid4()
+
+    def run_external(self, external_evaluator: Callable[..., dict[str, ValidOutput]]) -> VellumTestSuiteResults:
+        test_cases = self.client.test_suites.list_test_suite_test_cases(id=self.test_suite_id)
+        executions: list[ExternalTestCaseExecution] = []
+
+        # TODO - we need either:
+        # - Retrieve input_variables[i].key from test_suites
+        # - OR add a `name` field to `TestCaseVariableValue`
+        input_variable_name_by_id = {}
```
dvargasfuertes: blocking decision 3 - we don't have a way to map test_case input values to the names they use

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+            for external_output_key, external_output_value in external_outputs.items():
+                if isinstance(external_output_value, str):
+                    # Doesn't currently work - requires NamedTestCaseVariableValueRequest_String
+                    # Same for all other options below
+                    outputs.append(
+                        NamedTestCaseStringVariableValueRequest(
+                            value=external_output_value,
+                            name=external_output_key,
+                        )
+                    )
```
dvargasfuertes: blocking decision 4 - either we 
- accept `NamedTestCaseVariableValueRequest_String`
- work with fern to make `NamedTestCaseStringVariableValueRequest` work (I believe they posted an update to make this possible when combined with our true polymorphic dataclass work)
- push most of this to the backend as another way to send vellum values

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+
+                elif isinstance(external_output_value, dict):
+                    outputs.append(
+                        NamedTestCaseJsonVariableValueRequest(
+                            value=external_output_value,
+                            name=external_output_key,
+                        )
+                    )
+
+                # TODO: Handle the rest of the types
```
dvargasfuertes: highlighting this - the reason why we do the type/value paradigm is that it's hard to infer between list[SearchResult] and list[ChatMessage] for example. of JSON vs VellumError vs FunctionCall. My concern with the `dict` approach is that some if/elif inference chain will spread throughout our client sdk, making it harder to add new types. I propose we either:
- change the return type of `external_evaluator` to be a list of `NamedTestCaseVariableValue` instead of a dict
- change our API to accept either the list or dict, running the inference in a centralized location in our backend
- Stomach the concerns I listed above

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+            )
+
+        test_suite_run = self.client.test_suite_runs.create(
+            test_suite_id=self.test_suite_id,
+            # TODO: Also doesn't work yet for the same reason as variables above
+            exec_config=TestSuiteRunExternalExecConfigRequest(
+                data=TestSuiteRunExternalExecConfigDataRequest(
+                    executions=executions,
+                ),
+            ),
```
dvargasfuertes: blocking 5 - no decision here, I just forgot to re-expose `TestSuiteRunExternalExecConfigRequest` in our api client.

I do want to take this opportunity to demonstrate why I lean away from `type/data` and more for `type/[...attrs]` - the consumer needs to import two dataclasses that say basically the same thing for an unclear value add

tests/lib/test_test_suite.py
noanflaherty: Let's use `requests_mock`. We should add this as a dev dependency so that it's not installed by default when someone installs the sdk for actual use.

tests/lib/test_test_suite.py
```diff
@@ -0,0 +1,38 @@
...
+
+# Get started with writing tests with pytest at https://docs.pytest.org
+def test_vellum_test_suite__external__basic(requests_mock: RequestsMocker) -> None:
+    """Verify that the Vellum test suite could execute on External executions."""
+
+    # GIVEN an external execution
+    def mock_execute_thing(input_a: str, input_b: dict[str, Any]) -> dict[str, Any]:
+        """This could be the invocation of a Prompt, Langchain chain, etc."""
+
+        return {"output_a": "Example string output", "output_b": {"key": "value"}}
```
noanflaherty: Doesn't have to be in V1, but I think that for this executable function, we should support this input/output interface, but should also support:

```
def mock_execute_thing(inputs: list[CompiledVellumVariableValue]) -> list[CompiledVellumVariableValue]:
        pass
```

tests/lib/test_test_suite.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # AND a vellum test suite setup to handle that external eval
+    # - Note that with TestSuites as code, we could define the test suite inline
+    example_test_suite = VellumTestSuite("example_test_suite")
```
noanflaherty: We should support both this as well as
```
example_test_suite = VellumTestSuite("<test-suite-uuid>")
```

tests/lib/test_test_suite.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # AND Vellum successfully evaluates the external execution
+    test_suite_run_id = uuid4()
+    requests_mock.post(url="https://api.vellum.ai/v1/test_suite_runs")
+    requests_mock.get(
+        url=f"https://api.vellum.ai/v1/test_suite_runs/{test_suite_run_id}/executions?expand[0]=results.metric_results.metric_definition",
+    )
+
+    # WHEN the test suite is run
+    results = example_test_suite.run_external(mock_execute_thing)
```
noanflaherty: There's an open Q in my mind whether we should have both a sync and async version of these helper classes, just as the fern-generated API client has both. I think for now, only using sync is fine.

src/vellum/lib/test_suite.py
noanflaherty: Yeah I'm thinking something like:

_DEFAULT_POLLING_INTERVAL_MS = 500
_DEFAULT_MAX_POLLING_DURATION_MS = 10000

```
class VellumTestSuiteResults:
    def __init__(
        self,
        id: str,
        *, 
        polling_interval: int = _DEFAULT_POLLING_INTERVAL_MS,
        max_polling_duration: _DEFAULT_POLLING_INTERVAL_MS
    ) -> None:
```

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class VellumTestSuiteResults:
+    def __init__(self, id: str) -> None:
+        self.id = id
+        self.client = Vellum(
+            api_key=os.environ.get("VELLUM_API_KEY"),
+        )
+        self.test_suite_run_state = TestSuiteRunState.QUEUED
+        self.test_suite_run_executions: list[TestSuiteRunExecution] | None = None
```
noanflaherty: For something like this, I prefer to set to private vars, like `self._test_suite_run_state` and then expose properties for consumers to have read-access of them:

```
@property
def test_suite_run_state(self):
    return self._test_suite_run_state
```

This makes it clear that consumers shouldn't be setting the state of these variables.

noanflaherty: Note the property can be skipped if we don't think there's a good use-case for consumers to need this info.

src/vellum/lib/test_suite.py
noanflaherty: Max retries would be derived from these two values

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+
+        self._refresh_test_suite_run_state()
+
+        while self.test_suite_run_state in {TestSuiteRunState.QUEUED, TestSuiteRunState.RUNNING}:
+            time.sleep(1)
+            self._refresh_test_suite_run_state()
+            # TODO: Max retries?
+
+        if self.test_suite_run_state == TestSuiteRunState.FAILED:
+            raise Exception("Test suite run failed")
```
noanflaherty: We should implement custom exception classes and raise them instead of generic Exceptions

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if self.test_suite_run_state == TestSuiteRunState.CANCELLED:
+            raise Exception("Test suite run was cancelled")
+
+        response = self.client.test_suite_runs.list_executions(
+            self.id, expand=["results.metric_results.metric_definition"]
+        )
+        self.test_suite_run_executions = response.results
+
+    def get_metric_outputs(
```
noanflaherty: Prefer to have a class's public methods above its private methods for easy discovery

src/vellum/lib/test_suite.py
dvargasfuertes: the original intention was for these to be private

src/vellum/lib/test_suite.py
dvargasfuertes: I like this better, we won't even need max retries at that point

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+        self, metric_identifier: str | None = None, output_identifier: str | None = None
+    ) -> list[list[TestSuiteRunMetricOutput]]:
+        self._ensure_metric_results_are_loaded()
+
+        all_metric_results = [execution.metric_results for execution in self.test_suite_run_executions]
+
+        filtered_metric_outputs = [
```
noanflaherty: We should perform validation such that if you provide a metric identifier that's not unique based on `metric_id`, we raise an exception. For example, if you passed in `metric_definition_label`, but that Metric Definition was used twice in the Test Suite, we should raise an exception.

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+                if not output_identifier
+                or metric_output.name == output_identifier
+            ]
+            for metric_output_execution in filtered_metric_outputs
+        ]
+
+        return filtered_metric_filtered_outputs
+
+
+ValidOutput = Union[
```
noanflaherty: We might want to make a top-level `test_suites` folder within `lib`. This way we can have:
```
/lib
    /test_suites
        types.py
        constants.py
        core.py  # Open to name suggestions on this one

```

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+    float,
+    dict[str, Any],
+    list[SearchResult],
+    list[ChatMessage],
+    VellumError,
+]
+
+
+class VellumTestSuite:
+    def __init__(self, name: str) -> None:
```
noanflaherty: This should support name or id

src/vellum/lib/test_suite.py
noanflaherty: Let's add support for getting all this by name. Here's an example of how we support name or id in a retrieve endpoint: https://github.com/vellum-ai/vellum/blob/e4fa05b44525c8d14074e5adc33b471899bb67d3/django/app/api/document_indexes/views.py#L228

However, let's do that later and for now, unblock this PR by only support a test suite's id as the input.

src/vellum/lib/test_suite.py
noanflaherty: I had a feeling this would come up. We should derive name (or make it expandable) in the response serializers.

src/vellum/lib/test_suite.py
noanflaherty: I don't quite follow this, but let's can chat irl.

src/vellum/lib/test_suite.py
noanflaherty: I'm down to start by changing the return type of `external_evaluator` and seeing how we feel after implementing the test cases for it and using it.

src/vellum/lib/test_suite.py
dvargasfuertes: how about just `__init__.py` instead of `core`?

noanflaherty: It's generally frowned upon to put implementation code in `__init__.py`. it's preferred to put it in a named file and then import into `__init__.py` to expose for public consumption.

tests/lib/test_test_suite.py
```diff
@@ -0,0 +1,39 @@
+from typing import Any
+from uuid import uuid4
+from ...src.vellum.lib import VellumTestSuite
+import pytest
+from requests_mock import Mocker as RequestsMocker
+
+
+# Get started with writing tests with pytest at https://docs.pytest.org
```
siddseethepalli:
```suggestion
```

tests/lib/test_test_suite.py
```diff
@@ -0,0 +1,39 @@
...
+@pytest.mark.skip(reason="This test is not yet implemented")
+def test_vellum_test_suite__external__basic(requests_mock: RequestsMocker) -> None:
+    """Verify that the Vellum test suite could execute on External executions."""
```
siddseethepalli:
```suggestion
    """Verify that the Vellum test suite can execute on external executions."""
```

src/vellum/lib/test_suite.py
siddseethepalli: In the spirit of dev friendliness, I'm a big fan of shorter variable names if we can do so without introducing ambiguity:
`test_suite_run_state -> state`
`test_suite_run_executions -> executions`

siddseethepalli: Let's prefix everything here with `_` and add properties for whatever we want to actually expose

src/vellum/lib/test_suite.py
siddseethepalli: `_DEFAULT_MAX_POLLING_DURATION_MS` should probably be pretty high like `30*60*1000`

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def _ensure_metric_results_are_loaded(self):
+        if self.test_suite_run_executions is not None:
+            return
+
+        self._refresh_test_suite_run_state()
+
+        while self.test_suite_run_state in {TestSuiteRunState.QUEUED, TestSuiteRunState.RUNNING}:
+            time.sleep(1)
+            self._refresh_test_suite_run_state()
```
siddseethepalli: [Aesthetic Opinion] another way to write this that avoids two references to `_refresh_test_suite_run_state`
```suggestion
        while True:
            self._refresh_test_suite_run_state()
            if self.test_suite_run_state not in {TestSuiteRunState.QUEUED, TestSuiteRunState.RUNNING}:
                break
            time.sleep(1)
```

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+        self.id = id
+        self.client = Vellum(
+            api_key=os.environ.get("VELLUM_API_KEY"),
+        )
+        self.test_suite_run_state = TestSuiteRunState.QUEUED
+        self.test_suite_run_executions: list[TestSuiteRunExecution] | None = None
+
+    def _refresh_test_suite_run_state(self):
+        test_suite_run = self.client.test_suite_runs.retrieve(self.id)
+        self.test_suite_run_state = test_suite_run.state
```
siddseethepalli: Should we just save the whole response (i.e. `_refresh_test_suite_run`) since we're fetching it anyway? Want to avoid a situation where in the future we make extra calls to the API because we fetch in two different places for two different reasons

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+class VellumTestSuiteResults:
+    def __init__(self, id: str) -> None:
+        self.id = id
+        self.client = Vellum(
+            api_key=os.environ.get("VELLUM_API_KEY"),
+        )
+        self.test_suite_run_state = TestSuiteRunState.QUEUED
+        self.test_suite_run_executions: list[TestSuiteRunExecution] | None = None
+
+    def _refresh_test_suite_run_state(self):
```
siddseethepalli:
```suggestion
    def _refresh_test_suite_run_state(self) -> None:
```

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+        test_suite_run = self.client.test_suite_runs.retrieve(self.id)
+        self.test_suite_run_state = test_suite_run.state
+
+    def _ensure_metric_results_are_loaded(self):
```
siddseethepalli:
```suggestion
    def _ensure_metric_results_are_loaded(self) -> None:
```

siddseethepalli: Since `_ensure_metric_results_are_loaded` will always be followed by accessing them (why would you load ensure they're loaded otherwise?), does it make more sense to change this to `_get_metric_results` and actually return the results?

siddseethepalli: [nit] we use `executions` and `results` interchangeably - can we be more rigorous? this should be `executions` not `results` right?

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+        self.name = name
+        self.client = Vellum(
+            api_key=os.environ.get("VELLUM_API_KEY"),
+        )
+        self.test_suite_id = self._get_test_suite_id_by_name(name)
+
+    def _get_test_suite_id_by_name(self, name: str) -> str:
+        # TODO: We need to EITHER:
+        # - Allow for all of these methods to be called with `name`
+        # - OR, we need to expose a get test suite id by name method and save the `id`
```
siddseethepalli: Ideally we don't need to perform an additional API call because the user specified a `name` instead of an `id`

src/vellum/lib/test_suite.py
```diff
@@ -0,0 +1,175 @@
...
+        return uuid4()
+
+    def run_external(self, external_evaluator: Callable[..., dict[str, ValidOutput]]) -> VellumTestSuiteResults:
```
siddseethepalli: Can we pls call this something other than `external_evaluator` - when I read this I thought the evaluation was going to happen externally instead of the execution

tests/lib/test_test_suite.py
dvargasfuertes: wow didn't even know we supported both

src/vellum/lib/test_suite.py
dvargasfuertes: I wanted `_state` to be top level on `self` bc we could then initialize to a non-nullable value and avoid checking nullability everywhere. I'll rename the method to `_refresh_test_suite_run` so that if there are other fields we want to cache, we could do it here

src/vellum/lib/test_suite.py
dvargasfuertes: yea gonna rename

src/vellum/lib/test_suite.py
dvargasfuertes: while True's scare me but I'll switch it

src/vellum/lib/test_suite.py
dvargasfuertes: I'm thinking `resources` to be consistent with fern - I plan to reexport them out of __init__ anyway

src/vellum/lib/test_suite.py
dvargasfuertes: do we not want to allow users to be able to say "get me all of my metric results that use exact match"?

src/vellum/__init__.py
```diff
@@ -510,7 +510,7 @@
     workflow_deployments,
 )
 from .environment import VellumEnvironment
-from .version import __version__
+# from .version import __version__
```
dvargasfuertes: I have tests passing except for two blockers:
- I used the wrong request mocking library. Fix here: https://github.com/vellum-ai/vellum-client-generator/pull/225
- I needed to comment out this line, which seems to try to resolve the package version? gonna ping fern

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+            metric_result.outputs
+            for metric_result_execution in all_metric_results
+            for metric_result in metric_result_execution
+            if not metric_identifier
+            or (metric_result.metric_id == metric_identifier)
+            or (metric_result.metric_label == metric_identifier)
+            or (metric_result.metric_definition and metric_result.metric_definition.id == metric_identifier)
+            or (metric_result.metric_definition and metric_result.metric_definition.name == metric_identifier)
+            or (metric_result.metric_definition and metric_result.metric_definition.label == metric_identifier)
+        ]
```
dvargasfuertes: my response to an outdated comment requesting that we throw non-unique validation errors:

> do we not want to allow users to be able to say "get me all of my metric results that use exact match"?

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+            api_key=os.environ.get("VELLUM_API_KEY"),
+        )
+        self._state = TestSuiteRunState.QUEUED
+        self._executions: list[TestSuiteRunExecution] | None = None
+        self._polling_interval = polling_inteval
+        self._max_polling_duration = max_polling_duration
+
+    def get_metric_outputs(
+        self, metric_identifier: str | None = None, output_identifier: str | None = None
+    ) -> list[TestSuiteRunMetricOutput]:
```
dvargasfuertes: I'm having trouble deciding whether to return a flat list here, a nested list, or a list of a new object that points to test cast id and list of metric outputs. I suppose I'm having trouble visualize how this gets use. Particularly bc of [this comment](https://www.notion.so/Externally-Evaluated-Entities-Triggering-Test-Suites-as-Code-9d64e00548d94326a5fe169385038a8b?d=5ca8116b0e554c22aacf3508777011c9&pvs=4#f8a7d38a0de040868b121a07858532ac) I left

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+    TestSuiteRunExternalExecConfigRequest,
+    TestSuiteRunExternalExecConfigDataRequest,
+    ExternalTestCaseExecution,
+)
+
+
+class VellumTestSuiteResults:
+    def __init__(
+        self,
+        id: str,
```
noanflaherty: It's include to me what this is an id of and `id` is a keyword in python. Let's avoid using an be explicit

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+        polling_inteval: int = DEFAULT_POLLING_INTERVAL_MS,
+        max_polling_duration: int = DEFAULT_MAX_POLLING_DURATION_MS,
+    ) -> None:
+        self.id = id
+        self.client = Vellum(
```
noanflaherty: These should be private vars

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+            api_key=os.environ.get("VELLUM_API_KEY"),
```
noanflaherty: Idk that we should be so opinionated about how to get this API key. I'd suggest making `client` an optional kwarg to the `__init__` method and default to this behavior if none is passed.

src/vellum/lib/test_suites/resources.py
noanflaherty: Blocking w/o a convo: I think we should have helpers that do both. One two get a specific output value, one to get any output values that meet some filter criteria. I'd argue the former is more important.

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+        filtered_metric_filtered_outputs = [
+            metric_output
+            for metric_output_execution in filtered_metric_outputs
+            for metric_output in metric_output_execution
+            if not output_identifier or metric_output.name == output_identifier
+        ]
+
+        return filtered_metric_filtered_outputs
+
+    def _refresh_test_suite_run(self):
```
noanflaherty: nit: The current naming made me think we were going to re-run the TSR

```suggestion
    def _refresh_test_suite_run_state(self):
```

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+        return self._executions
+
+
+class VellumTestSuite:
+    def __init__(
+        self,
+        id: str | None = None,
+        # TODO: Support APIs that query by name
+        # https://app.shortcut.com/vellum/story/2689
+        # name: str | None = None,
```
noanflaherty: 1/5 nit: Avoid adding TODOs in SDK code.

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+            self.id, expand=["results.metric_results.metric_definition", "results.metric_results.metric_label"]
+        )
+        self._executions = response.results
+        return self._executions
+
+
+class VellumTestSuite:
+    def __init__(
+        self,
+        id: str | None = None,
```
noanflaherty: Same feedback on `id`.

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+        # name: str | None = None,
+    ) -> None:
+        self.client = Vellum(
+            api_key=os.environ.get("VELLUM_API_KEY"),
+        )
+        self._id = id
+
+    def run_external(
+        self, executable: Callable[[list[TestCaseVariableValue]], list[NamedTestCaseVariableValueRequest]]
+    ) -> VellumTestSuiteResults:
```
noanflaherty: We should have docstrings for any public method

tests/lib/test_test_suites.py
```diff
@@ -0,0 +1,126 @@
...
+from src.vellum.types import (
+    TestCaseVariableValue_String,
+    TestCaseVariableValue_Json,
+    NamedTestCaseVariableValueRequest_String,
+    NamedTestCaseVariableValueRequest_Json,
+    TestCaseVariableValue,
+)
+from pytest_httpx import HTTPXMock
+
+# TODO: We should try to resolve this warning:
```
noanflaherty: Same feedback on TODOs

tests/lib/test_test_suites.py
```diff
@@ -0,0 +1,126 @@
...
+# src/vellum/types/test_case_variable_value.py:15
+#   /Users/dvargas/vellum-ai/vellum-client-python/src/vellum/types/test_case_variable_value.py:15: \
+#   PytestCollectionWarning: cannot collect test class 'TestCaseVariableValue_String' because it has a __init__ constructor (from: tests/lib/test_test_suites.py)
+#     class TestCaseVariableValue_String(TestCaseStringVariableValue):
+
+
+def test_vellum_test_suite__external__basic(httpx_mock: HTTPXMock) -> None:
```
noanflaherty: Love including tests like this!

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,134 @@
...
+        # https://app.shortcut.com/vellum/story/2689
+        # name: str | None = None,
+    ) -> None:
+        self.client = Vellum(
+            api_key=os.environ.get("VELLUM_API_KEY"),
+        )
+        self._id = id
+
+    def run_external(
+        self, executable: Callable[[list[TestCaseVariableValue]], list[NamedTestCaseVariableValueRequest]]
```
noanflaherty: I think this is the right call signature for this initial PR, but I would love to _also_ support a call signature that uses primitive python types and under the hood, use a common util function we define somewhere to "serialize" / "deserialize" into our strict types.

dvargasfuertes: Yea I have ideas there

dvargas92495 (commented):
Two last Fern blockers before merging:
- vellum-client-generator is still failing the pytest-httpx pip install
- ensuring tests run in CI without version commented out

Deep has mentioned he'll be online soon to help. Ready for another round of review otherwise

src/vellum/lib/test_suites/constants.py
```diff
@@ -0,0 +1,2 @@
+DEFAULT_POLLING_INTERVAL_MS = 500
+DEFAULT_MAX_POLLING_DURATION_MS = 30 * 60 * 1000
```
noanflaherty:
```suggestion
DEFAULT_POLLING_INTERVAL_MS = 500  # Half a second
DEFAULT_MAX_POLLING_DURATION_MS = 30 * 60 * 1000  # Half an hour
```

src/vellum/lib/utils/paginator.py
```diff
@@ -0,0 +1,25 @@
...
+
+@dataclass
+class PaginatedResults(Generic[Result]):
+    count: int
+    results: list[Result]
+
+
+def get_all_results(
+    paginated_api: Callable[[int, int | None], PaginatedResults[Result]], page_size: int | None = None
+) -> list[Result]:
```
noanflaherty: As a consumer of this, I'd much prefer a return type of:
```suggestion
def get_all_results(
    paginated_api: Callable[[int, int | None], PaginatedResults[Result]], page_size: int | None = None
) -> Generator[Result, None, None]:
```

This way, I'm not _forced_ to exhaust the paginator.

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+    TestSuiteRunExternalExecConfigDataRequest,
+    ExternalTestCaseExecution,
+)
+from src.vellum.lib.utils.paginator import PaginatedResults, get_all_results
+
+
+class VellumTestSuiteRunExecution(TestSuiteRunExecution):
+    @staticmethod
+    def from_api(execution: TestSuiteRunExecution) -> VellumTestSuiteRunExecution:
+        return VellumTestSuiteRunExecution(
```
noanflaherty: I've also seen this. Unclear to me if one is better than the other

```suggestion
    @classmethod
    def from_api(cls, execution: TestSuiteRunExecution) -> VellumTestSuiteRunExecution:
        return cls(
```

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+        self,
+        metric_identifier: str | None = None,
+        output_identifier: str | None = None,
+    ) -> TestSuiteRunMetricOutput:
+        metric_outputs = self.get_metric_outputs(metric_identifier)
+
+        filtered_metric_outputs = [
+            metric_output
+            for metric_output in metric_outputs
+            if not output_identifier or metric_output.name == output_identifier
```
noanflaherty: nit: Avoids edge case of falsy values
```suggestion
            if output_identifier is None or metric_output.name == output_identifier
```

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+            outputs=execution.outputs,
+            metric_results=execution.metric_results,
+        )
+
+    def get_metric_output(
+        self,
+        metric_identifier: str | None = None,
+        output_identifier: str | None = None,
+    ) -> TestSuiteRunMetricOutput:
+        metric_outputs = self.get_metric_outputs(metric_identifier)
```
noanflaherty:
```suggestion
    ) -> TestSuiteRunMetricOutput:
        """Retrieve a metric's output by specifying the info needed to uniquely identify that metric and output."""

        metric_outputs = self.get_metric_outputs(metric_identifier)
```

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+        return filtered_metric_outputs[0]
+
+    def get_metric_outputs(
+        self,
+        metric_identifier: str | None = None,
+    ) -> list[TestSuiteRunMetricOutput]:
+        filtered_metric_results = [
+            metric_output
+            for metric_output in self.metric_results
+            if not metric_identifier
```
noanflaherty:
```suggestion
            if metric_identifier is None
```

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+        if len(filtered_metric_outputs) > 1:
+            raise TestSuiteRunResultsException(f"Multiple metric outputs found with identifier: {output_identifier}")
+
+        return filtered_metric_outputs[0]
+
+    def get_metric_outputs(
+        self,
+        metric_identifier: str | None = None,
+    ) -> list[TestSuiteRunMetricOutput]:
+        filtered_metric_results = [
```
noanflaherty:
```suggestion
    ) -> list[TestSuiteRunMetricOutput]:
        """Return a metric's output across all executions by providing the info needed to uniquely identify that metric."""
    
        filtered_metric_results = [
```

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+            raise TestSuiteRunResultsException(f"Multiple metric results found with identifier: {metric_identifier}")
+
+        return filtered_metric_results[0].outputs
+
+
+class VellumTestSuiteRunResults:
+    def __init__(
+        self,
+        test_suite_run_id: str,
+        client: Vellum | None = None,
```
noanflaherty:
```suggestion
        test_suite_run_id: str,
        *,
        client: Vellum | None = None,
```

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+    ) -> PaginatedResults[TestSuiteRunExecution]:
+        response = self._client.test_suite_runs.list_executions(
+            self._test_suite_run_id,
+            offset=offset,
+            limit=limit,
+            expand=["results.metric_results.metric_definition", "results.metric_results.metric_label"],
+        )
+        return PaginatedResults(results=response.results, count=response.count)
+
+    def _get_test_suite_run_executions(self) -> list[VellumTestSuiteRunExecution]:
```
noanflaherty: This should return a Generator as well, following the feedback below.

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+            api_key=os.environ.get("VELLUM_API_KEY"),
+        )
+        self._state = TestSuiteRunState.QUEUED
+        self._executions: list[VellumTestSuiteRunExecution] | None = None
+        self._polling_interval = polling_inteval
+        self._max_polling_duration = max_polling_duration
+
+    def get_metric_outputs(
+        self, metric_identifier: str | None = None, output_identifier: str | None = None
+    ) -> list[TestSuiteRunMetricOutput]:
```
noanflaherty: This should return a Generator as well, following feedback from below.

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+        self._test_suite_id = test_suite_id
+
+    def run_external(
+        self, executable: Callable[[list[TestCaseVariableValue]], list[NamedTestCaseVariableValueRequest]]
+    ) -> VellumTestSuiteRunResults:
+        """
+        Runs this Vellum Test Suite on any executable function defined external to Vellum. Returns a results
+        wrapper that polls the generated Test Suite Run for the latest metric results.
+        """
```
noanflaherty:
```suggestion
        Runs this Vellum Test Suite on any executable function defined external to Vellum.
        
        Returns a results wrapper that polls the generated Test Suite Run for the latest metric results.
        """
```

src/vellum/lib/test_suites/resources.py
```diff
@@ -0,0 +1,184 @@
...
+            expand=["results.metric_results.metric_definition", "results.metric_results.metric_label"],
+        )
+        return PaginatedResults(results=response.results, count=response.count)
+
+    def _get_test_suite_run_executions(self) -> list[VellumTestSuiteRunExecution]:
+        if self._executions is not None:
+            return self._executions
+
+        start_time = time.time_ns()
+        while True:
```
noanflaherty, line 136: I'm thinking we should use logging with debug level to show each polling

src/vellum/lib/test_suites/resources.py
dvargasfuertes: Based on Github Copilot's response, I'm switching to classmethod:
Screenshot 2024-04-22 at 2 46 04 PM

noanflaherty (approved):

dvargas92495:
We're merging!

PR #22, opened 2024-04-25
Fix sdk enums

dvargas92495:
The new enum config means we use strings instead of Enum.VALUE now.

Note that this means we likely need to update our code snippets in the app


dvargas92495:
Gonna merge through to unblock release

noanflaherty (approved):
Setting aside that this is a breaking change, it’s a great dev exp improvement!

dvargas92495:
it's also going to allow us to more freely add new enums to our api. Like new VellumVariableType's, new ErrorCode's, etc

PR #23, opened 2024-04-25
Improve SDK Code & Add Example Notebook

noanflaherty:
This makes some improvements to our SDK that I came across while building out the included example notebook.


src/vellum/lib/utils/env.py
```diff
@@ -6,6 +6,8 @@
 def get_api_key() -> str:
     api_key = os.environ.get("VELLUM_API_KEY")
     if api_key is None:
-        raise VellumClientException("`VELLUM_API_KEY` environment variable id required to be set.")
-    
+        raise VellumClientException(
+            "`VELLUM_API_KEY` environment variable id required to be set."
```
dvargasfuertes: Oh I left a typo here. `is` instead of `id`

src/vellum/lib/test_suites/resources.py
```diff
@@ -116,44 +116,141 @@ class VellumTestSuiteRunResults:
...
+        return self._test_suite_run.state
+
+    @cached_property
+    def all_executions(self) -> list[VellumTestSuiteRunExecution]:
+        return list(self._get_test_suite_run_executions())
+
     def get_metric_outputs(
         self, metric_identifier: str | None = None, output_identifier: str | None = None
-    ) -> Generator[TestSuiteRunMetricOutput, None, None]:
+    ) -> List[TestSuiteRunMetricOutput]:
```
dvargasfuertes, line 145: [q] what changed your mind here?

src/vellum/lib/test_suites/resources.py
```diff
@@ -116,44 +116,141 @@ class VellumTestSuiteRunResults:
...
+        """Returns the values of a numeric metric output that match the given criteria."""
+
+        metric_outputs: list[TestSuiteRunMetricOutput_Number] = []
+
+        for output in self.get_metric_outputs(
+            metric_identifier=metric_identifier, output_identifier=output_identifier
+        ):
+            if output.type != "NUMBER":
+                raise TestSuiteRunResultsException(
+                    "Cannot calculate mean of non-numeric metric outputs"
```
dvargasfuertes: [nit] typo - this helper looks like is public and used generally, not just in `mean`

src/vellum/lib/test_suites/resources.py
```diff
@@ -116,44 +116,141 @@ class VellumTestSuiteRunResults:
...
         )
         self._polling_interval = polling_interval
         self._max_polling_duration = max_polling_duration
 
+    @property
+    def state(self) -> TestSuiteRunState:
+        return self._test_suite_run.state
+
+    @cached_property
+    def all_executions(self) -> list[VellumTestSuiteRunExecution]:
```
dvargasfuertes, line 140: [nit] this makes self._executions redundant right?

dvargas92495 (approved):

dvargas92495:
Oh make sure to add examples to fernignore

src/vellum/lib/test_suites/resources.py
noanflaherty: Once a generator is exhausted, it cannot be re-consumed. The lower-level methods should still be generators, but we should cache the list of results once they're cycled through (which is what the `cached_property` helps with) and allows you to iterate over the results as many times as you want.

src/vellum/lib/test_suites/resources.py
noanflaherty: That holds the generator, this holds the cached exhausted generator

.gitignore
```diff
@@ -2,3 +2,4 @@ dist/
 .mypy_cache/
 __pycache__/
 poetry.toml
+**/.ipynb_checkpoints
```
dvargasfuertes, line 5: fern is going to wipe this out too unless we add this to the fernignore

PR #84, opened 2024-10-14
Add Support for Using Test Suite Name

noanflaherty:
This updates our external evals hand-written lib code to support either a Test Suite's ID or Name in use with external evals.

It also updates the Jupyter notebook example accordingly.


dvargas92495 (approved):

peicodes (commented):
👍

PR #174, opened 2024-11-25
Add Codegen Directory from prototype to `ee/codegen`

dvargas92495:
This PR migrates our codegen logic from the prototype repo into `ee/codegen`. In order to enable that, we had to include the other following changes:
- Updating .gitignore with codegen related files
- Introducing `isort`, which is also being updated [upstream](https://github.com/vellum-ai/vellum-client-generator/pull/552)
- add a github action for running codegen in CI with the relevant fernignore
- The fixtures that were in `vellum_ee_tests/codegen_integration` were moved to `ee/codegen_integration`.
    - Note that I at first wanted to do `ee/tests/codegen_integration`, but we can't have multiple modules that start with `tests.`, so it conflicted with our root level `tests` dirs


noanflaherty (approved):

kirawoods:
🥳

PR #179, opened 2024-11-25
Implement Boilerplate for Generic Nodes

noanflaherty:
I'll soon be writing unit tests for our `graph` attribute. As part of this, I'm gonna want to keep my tests very basic and just work with custom/generic nodes, rather than instantiating specific nodes of specific types.

As such, I figured it's as good a time as any to introduce the boilerplate needed for generic nodes.

### DIFF
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -876,12 +876,7 @@ export function apiNodeFactory({
 
 export function genericNodeFactory(): GenericNode {
   const nodeData: GenericNode = {
-    id: "7e09927b-6d6f-4829-92c9-54e66bdcaf80",
     type: WorkflowNodeType.GENERIC,
-    data: {
-      label: "Generic Node",
-    },
-    inputs: [],
   };
   return nodeData;
 }
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -874,11 +874,13 @@ export function apiNodeFactory({
   return nodeData;
 }
 
-export function genericNodeFactory(): GenericNode {
+export function genericNodeFactory(
+  { name }: { name: string } = { name: "MyCustomNode" }
+): GenericNode {
   const nodeData: GenericNode = {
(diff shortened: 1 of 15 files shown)


ee/codegen/src/serializers/vellum.ts
```diff
@@ -1655,6 +1656,29 @@ export declare namespace ErrorNodeSerializer {
...
+> = objectSchema({
+  type: stringLiteralSchema("GENERIC"),
+  id: stringSchema(),
+  data: objectSchema({
+    label: stringSchema(),
+  }),
+  inputs: listSchema(NodeInputSerializer),
+  displayData: propertySchema("display_data", anySchema()),
+  definition: WorkflowNodeDefinitionSerializer.optional(),
+});
```
noanflaherty: I expect this to change a lot down the line, but starting with this for now.

NgoHarrison (approved):

peicodes (approved):

ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -872,3 +873,15 @@ export function ApiNodeFactory({
...
 }
+
+export function genericNodeFactory(): GenericNode {
+  const nodeData: GenericNode = {
+    id: "7e09927b-6d6f-4829-92c9-54e66bdcaf80",
+    type: WorkflowNodeType.GENERIC,
+    data: {
+      label: "Generic Node",
+    },
+    inputs: [],
```
dvargasfuertes: [q] do we need `id`, `data` and `inputs`?

dvargas92495 (approved):
Ideally we are able to only include fields that we perceive existing long term. If the goal is "to play nicely with the UI today", we'd be able to make a hard fork on the logic based on `type=='GENERIC'`, and go down a component tree that doesn't assume old fields. Once we migrate all old nodes to be generic nodes, we could then drop the `type`

noanflaherty:
I'm going to close this and reopen following Vargas's feedback. There's like other smaller PRs I can get in to ease us into this.

noanflaherty:
@dvargas92495 this is ready for re-review.

cc @peicodes

dvargas92495 (approved):
💥

PR #185, opened 2024-11-26
[SC-5637] Add stable condition ids

NgoHarrison:


ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
```diff
@@ -22,6 +22,10 @@ class ConditionalNodeDisplay(BaseConditionalNodeDisplay[ConditionalNode]):
             rhs=None,
         )
     ]
+    condition_ids = {
+        0: UUID("8d0d8b56-6c17-4684-9f16-45dd6ce23060"),
+        1: UUID("ea63ccd5-3fe3-4371-ba3c-6d3ec7ca2b60"),
+    }
```
noanflaherty: nit: If the keys are always gonna be consecutive integer values starting from 0, we might as well just use an array, no?

ee/vellum_ee/workflows/display/nodes/vellum/conditional_node.py
```diff
@@ -44,13 +44,15 @@ class RuleIdMap:
...
     rule_ids: ClassVar[List[RuleIdMap]]
+    condition_ids: ClassVar[Dict[int, UUID]]
 
     def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> JsonObject:
         node = self._node
         node_id = self.node_id
 
         node_inputs: List[NodeInput] = []
         source_handle_ids = self._get_source_handle_ids() or {}
+        condition_ids = self._get_condition_ids() or {}
```
noanflaherty: Assuming we go with my suggestion above, we should raise an exception if the user explicitly specified condition_ids on the class, but the length of the array they provided doesn't match the number of conditions

noanflaherty (approved):

ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
NgoHarrison: Good point will update

NgoHarrison:
@noanflaherty @peicodes    
So these changes helps with the root condition id but this id in the rule group still uses a hash. i can introduce another field like the condition_ids but this feels like overkill/redundant. Do you have ideas for another approach?
```
{
                    "id": condition_id,
                    "type": port._condition_type.value,
                    "source_handle_id": source_handle_id,
                    "data": {
                        "id": str(uuid4_from_hash(f"{condition_id}|rule_group")),
                        "rules": rules,
                        "combinator": "AND",
                        "negated": False,
                        "field_node_input_id": None,
                        "operator": None,
                        "value_node_input_id": None,
                    },
                }
```

peicodes:
@NgoHarrison Great catch! I think the issue is hinted at in this code:
```
            rules = condition_rule["rules"] if condition_rule["rules"] else [condition_rule]
```

We are treating the top-level rule group (in `data`) differently because we want to guarantee that the top-level is always a group. One option is to treat it the same.

I'm imagining something like this:
- Codegen generates `rule_ids` assuming that each condition will always serialize in such a way where the top-level entity for a condition is always a rule group and never a rule. 
- We split up the serialization into a few steps:
  1. Port conditions -> Vellum conditions, without adding the IDs
  2. When a vellum condition's top-level entity is NOT a group, wrap it in a group
  3. Use `rule_ids` to populate the resulting vellum conditions

I feel 3/5 that given the complexity of this option, we should instead consider updating the list we use for overriding condition IDs from a list of UUID to a list of dataclasses:

```
class ConditionId:
  id: UUID
  rule_group_id: UUID
```

It doesn't scratch the elegance itch the same way but I think will be more maintainable going forward. No sophisticated transformations or mappings.

PR #186, opened 2024-11-26
[Codegen] Add Support for Codegenning Workspace Secrets

noanflaherty:
Adds codegen support for when a node points to a workspace secret.


ee/codegen/src/generators/node-inputs/node-input-value-pointer-rules/execution-counter-pointer.ts
```diff
@@ -12,12 +12,10 @@ export class ExecutionCounterPointerRule extends BaseNodeInputValuePointerRule<E
...
     );
 
-    const inputRef = python.reference({
+    return python.reference({
       name: nodeContext.nodeClassName,
       modulePath: nodeContext.nodeModulePath,
       attribute: ["Execution", "counter"],
     });
-    this.addReference(inputRef);
-    return inputRef;
```
noanflaherty, line 21: We no longer have to manually add references because the base class does this with the return value of this function.

NgoHarrison (approved):

PR #188, opened 2024-11-26
[Codegen][SC-5269] Add Partial Codegen Support for Subworkflow Deployment Nodes

noanflaherty:
This adds partial codegen support for Subworkflow Deployment Nodes.

We still need to do some work around generating their outputs, but that'll come in a future PR.


kirawoods (commented):
🆒

PR #189, opened 2024-11-26
Fix Codegen Version Script & Increment

noanflaherty:
Fixes the script used to increment post versions, and also increments the version.


m-abboud (commented):
✅

PR #190, opened 2024-11-26
[Codegen] Fix e2e Support for Execution Counters

noanflaherty:
This PR:
1. Fixes a typo in codegen for execution counters (I had used `counter` instead of `count`)
2. Adds serialization support for `ExecutionCountReference` descriptors
3. Adds an e2e test feature an execution counter and a templating node. I had to add this test to the exclusion list for serialization due to unrelated issues regarding templating node serialization. We'll tackle those separately.

### DIFF
--- a/ee/codegen_integration/conftest.py
+++ b/ee/codegen_integration/conftest.py
@@ -37,6 +37,7 @@ def _get_fixture_paths(root: str) -> Tuple[str, str]:
             "simple_code_execution_node",
             # TODO: Remove once graph is fleshed out. Current the workflow is not pointing to any port branch
             "simple_conditional_node",
+            "simple_templating_node",
         }
     )
 )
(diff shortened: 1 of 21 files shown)


ee/codegen_integration/conftest.py
```diff
@@ -37,6 +37,7 @@ def _get_fixture_paths(root: str) -> Tuple[str, str]:
             "simple_code_execution_node",
             # TODO: Remove once graph is fleshed out. Current the workflow is not pointing to any port branch
             "simple_conditional_node",
+            "simple_templating_node",
```
noanflaherty, line 40: @peicodes I'll need your help removing this from the exclusion list after we've fixed templating node serialization issues.

peicodes (commented):
👍

PR #201, opened 2024-11-26
Introducing Parent Context

marlenc3:
Introducing Parent Context to events. 
Should default to None
Focusing on Nodes, Deployments, and Workflows (for now)

### DIFF
--- a/src/vellum/workflows/events/tests/test_event.py
+++ b/src/vellum/workflows/events/tests/test_event.py
@@ -90,10 +90,10 @@ module_root = name_parts[: name_parts.index("events")]
                     },
                 ),
                 parent=NodeParentContext(
-                    definition=MockNode,
+                    node_definition=MockNode,
                     span_id=UUID("123e4567-e89b-12d3-a456-426614174000"),
                     parent=WorkflowParentContext(
-                        definition=MockWorkflow,
+                        workflow_definition=MockWorkflow,
                         span_id=UUID("123e4567-e89b-12d3-a456-426614174000")
                     )
                 )
--- a/src/vellum/workflows/events/tests/test_event.py
+++ b/src/vellum/workflows/events/tests/test_event.py
@@ -115,12 +115,12 @@ module_root = name_parts[: name_parts.index("events")]
                     },
                 },
                 "parent": {
-                    "definition": {
+                    "node_definition": {
                         "name": "MockNode",
                         "module": module_root + ["events", "tests", "test_event"],
                     },
                     "parent": {
-                        "definition": {
+                        "workflow_definition": {
                             "name": "MockWorkflow",
                             "module": module_root + ["events", "tests", "test_event"],
                         },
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/events/types.py
```diff
@@ -44,9 +43,41 @@ def default_serializer(obj: Any) -> Any:
     )
 
 
+class BaseParentContext(UniversalBaseModel):
+    span_id: UUID
+    parent: 'Optional[ParentContext]' = None
```
dvargasfuertes: [nit] can quote just ParentContext

src/vellum/workflows/events/types.py
```diff
@@ -44,9 +43,41 @@ def default_serializer(obj: Any) -> Any:
...
+    deployment_history_item_id: UUID
+    workflow_version_id: UUID
+    release_tag_id: UUID
+    release_tag_name: str
+    external_id: Optional[str]
+
+
+class NodeParentContext(BaseParentContext):
+    node_id: UUID
+    class_name: str
```
dvargasfuertes: [blocking] I know in previous docs we had this shape. going forward, we should use `definition: Type[BaseNode]` and `definition: Type[BaseWorkflow]` instead and use the same serializer that our events use for those two

src/vellum/workflows/events/tests/test_event.py
```diff
@@ -73,6 +73,7 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
                         "foo": "bar",
                     },
                 },
+                "parent": None,
```
dvargasfuertes, line 77: [blocking] let's add a test somewhere here to ensure we serialize parent context correctly

marlenc3: 🫡

src/vellum/workflows/events/types.py
marlenc3: ohhhh, sweet 👍 so instead of node_id / class_name just use definition: Type[]

src/vellum/workflows/events/types.py
```diff
@@ -43,10 +46,48 @@ def default_serializer(obj: Any) -> Any:
...
+class BaseParentContext(UniversalBaseModel):
+    span_id: UUID
+    parent: Optional['ParentContext'] = None
+
+
+class DeploymentParentContext(BaseParentContext):
+    deployment_id: UUID
+    deployment_name: str
+    deployment_history_item_id: UUID
+    workflow_version_id: UUID
```
dvargasfuertes: [oos] I suspect that we may want to make this field agnostic to prompt or workflow, but am still not sure what to call it. Let's defer until we have to cross that road with execute prompt

dvargasfuertes: Gonna actually raise in sdk channel now in the meantime

dvargas92495 (approved):

dvargas92495 (approved):

dvargas92495:
Gonna merge in to include in 0.10.4

PR #213, opened 2024-11-27
Solve issue where subworkflow'd nodes had no access to api keys

dvargas92495:
I was seeing the basic rag workflow getting permission errors. I realized it was because the vellum api key wasn't being passed down from parent node to subworkflow'd nodes. This PR aims to solve this


marlenc3 (approved):
:fire:

PR #216, opened 2024-11-27
[SC-5638] Dynamically Drive Subworkflow Deployment Outputs

noanflaherty:
Now that codegen has access to the Vellum API, we can use it to generate dynamic outputs, such as those of Subworkflow Deployment Nodes.


ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
```diff
@@ -285,222 +285,6 @@ class PromptNode(InlinePromptNode):
 "
 `;
 
-exports[`InlinePromptNode > FUNCTION_DEFINITION block type > basic > getNodeDisplayFile for FUNCTION_DEFINITION block type 1`] = `
```
noanflaherty, line 288: Removes some snapshots whose tests were actually removed in a prior PR but `npm run test:update` hasn't been run since. We may want to add this to CI to fail if your PR causes a diff in snapshots.

ee/codegen/src/__test__/nodes/__snapshots__/subworkflow-deployment-node.test.ts.snap
```diff
@@ -13,6 +16,14 @@ class SubworkflowNodeDisplay(BaseSubworkflowDeploymentNodeDisplay[SubworkflowNod
...
     target_handle_id = UUID("f5e6bd33-527a-4ba6-8906-cd5e96a4321c")
     node_input_ids_by_name = {"input": UUID("f62b7511-dd69-4dff-a4fc-718a9db3ceba")}
+    output_display = {
+        SubworkflowNode.Outputs.output_1: NodeOutputDisplay(
+            id=UUID("1"), name="output-1"
+        ),
+        SubworkflowNode.Outputs.output_2: NodeOutputDisplay(
+            id=UUID("2"), name="output-2"
+        ),
+    }
```
noanflaherty, line 26: Shows that we successfully generate static id and name references for output variables.

ee/codegen/src/__test__/nodes/__snapshots__/subworkflow-deployment-node.test.ts.snap
```diff
@@ -31,8 +42,12 @@ exports[`SubworkflowDeploymentNode > basic > getNodeFile 1`] = `
 
 
 class SubworkflowNode(SubworkflowDeploymentNode):
-    deployment = "58171df8-cdf9-4d10-a9ed-22eaf545b22a"
+    deployment = "test-deployment"
```
noanflaherty, line 45: Shows that we use the API to also populate the name instead of id.

ee/codegen/src/__test__/nodes/__snapshots__/subworkflow-deployment-node.test.ts.snap
```diff
@@ -31,8 +42,12 @@ exports[`SubworkflowDeploymentNode > basic > getNodeFile 1`] = `
...
     release_tag = "LATEST"
     subworkflow_inputs = {"input": None}
+
+    class Outputs(SubworkflowDeploymentNode.Outputs):
+        output_1: str
+        output_2: float
```
noanflaherty, line 51: Shows that we now generate outputs and even handle the conversion of kebab-case names into valid  snake case python attribute names.

ee/codegen/src/context/node-context/create-node-context.ts
```diff
@@ -48,9 +50,19 @@ export async function createNodeContext(
           });
         }
         case "DEPLOYMENT": {
+          const { releaseTag, workflowDeploymentId } = subworkflowNodeData.data;
+          const workflowDeploymentHistoryItem =
+            await new WorkflowDeploymentsClient(
+              {}
+            ).workflowDeploymentHistoryItemRetrieve(
+              releaseTag,
+              workflowDeploymentId
```
dvargasfuertes, line 59: woah, kinda wild that fern generates it in this order

dvargas92495 (approved):

ee/codegen/src/context/node-context/create-node-context.ts
noanflaherty: Yeah 😕 the worst part is if they “fixed” it to be in the reverse order, it’d be a breaking logical change uncaught by type checks 😵‍💫

noanflaherty: Ah I suspect it's because our OpenAPI spec has it in this order: https://github.com/vellum-ai/vellum-client-generator/blob/18e4eb34c2e87b86a86f4af45ea7890485f833f5/fern/openapi/openapi.yml#L269

Given that these are new endpoints, I'm gonna try to make this breaking change asap before people start relying on them.

noanflaherty: Dang, it looks like drf spectacular generates the order of query parameters in the spec alphabetically, so the only way to auto-generate an OpenAPI spec with the desired order is to rename one of the parameters.

This means we have three options:
1. See if fern wants to make a change such that the order always follows how they're used in the url, rather than the order in which they're shown in the spec
2. Rename one of them
3. Create our own custom pre/post processing step to manually change the order.

dvargasfuertes: > it looks like drf spectacular generates the order of query parameters in the spec alphabetically

can we override this? I think fern following the order of the spec is "correct"

noanflaherty: Not that I can find.

PR #217, opened 2024-11-27
[Codegen] Add Codegen Support for Note Nodes

noanflaherty:
This adds workflow sdk and codegen support for Note Nodes.

Note that we won't yet successfully serialize Note Nodes, because they aren't present in the Workflow's Graph, but we'll at least codegen them. The follow up for serializing nodes not in the graph is [here](https://app.shortcut.com/vellum/story/5394).

### DIFF
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -19,6 +19,7 @@ import {
   MergeNode,
   GenericNode,
   SubworkflowNode,
+  NoteNode,
 } from "src/types/vellum";
 
 export function entrypointNodeDataFactory(): EntrypointNode {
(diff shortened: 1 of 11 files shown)


dvargas92495 (commented):
Before giving an initial review, @peicodes and I had discussed in the past codegenning note nodes _only_ as a Workflow Display attribute instead of an actual node. Note Nodes serve no purpose during execution and would love to prune out of that codegen. Wdyt?

noanflaherty:
If we do that, let's do it later. Doing what you suggest requires establishing of new patterns for that aren't worth forging a path for right now imo.

dvargas92495 (approved):

peicodes:
@noanflaherty would it be a lot of work to do what @dvargas92495 is suggesting? I'm imagining note nodes in vellum coming out as "note_displays" on the workflow display. And the vellum workflow displays can just iterate over it on serialization to produce note nodes

I can give it an attempt this wknd - I only have one serialization card left

noanflaherty:
@peicodes Vargas and I synced offline and I'm going to merge as is given our timeline. I view this as an easily reversible decision given that users are unlikely to import or reference the `NoteNode` class in any meaningful way. But yes, I agree the direction you're both advocating for is "better" and worth following up with when we have fewer time sensitive things in front of us.

noanflaherty:
For additional context, while the serialization side of this might be easy, the codegen side has additional implications that I don't want to block on.

PR #218, opened 2024-11-27
[Codegen] Fix Workflow Input Attr Name Generation

noanflaherty:
Correctly use our snake-case naming and improve that function too.


dvargas92495 (commented):
✅

PR #219, opened 2024-11-27
Flesh out implementation for TryNode wrapping a streaming node

dvargas92495:
This PR adds streaming support for TryNode's. It enables the p80 case of prompt nodes that have reject on error toggle on today.

This PR was blocking basic rag execution and has serialization implications as well

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -3,9 +3,6 @@ import inspect
 from uuid import UUID
 from typing import TYPE_CHECKING, Any, Dict, Generic, Optional, Type, TypeVar, get_args
 
-from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplay, PortDisplayOverrides
-from vellum_ee.workflows.display.utils.uuids import uuid4_from_hash
-from vellum_ee.workflows.display.vellum import CodeResourceDefinition, NodeDefinition
 from vellum.workflows.nodes.bases.base import BaseNode
 from vellum.workflows.nodes.utils import get_wrapped_node, has_wrapped_node
 from vellum.workflows.ports import Port
@@ -14,6 +11,9 @@ from vellum.workflows.types.core import JsonObject
 from vellum.workflows.types.generics import NodeType
 from vellum.workflows.types.utils import get_original_base
 from vellum.workflows.utils.names import pascal_to_title_case
+from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplay, PortDisplayOverrides
+from vellum_ee.workflows.display.utils.uuids import uuid4_from_hash
+from vellum_ee.workflows.display.vellum import CodeResourceDefinition, NodeDefinition
 
 if TYPE_CHECKING:
     from vellum_ee.workflows.display.types import WorkflowDisplayContext
@@ -31,7 +31,7 @@ class BaseNodeDisplay(Generic[NodeType]):
     def __init__(self, node: Type[NodeType]):
         self._node = node
 
-    def serialize(self, display_context: "WorkflowDisplayContext", **kwargs: Any) -> JsonObject:
+    def serialize(self, display_context: "WorkflowDisplayContext") -> JsonObject:
         raise NotImplementedError(f"Serialization for nodes of type {self._node.__name__} is not supported.")
 
     def get_definition(self) -> NodeDefinition:
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -31,7 +31,7 @@ class BaseNodeDisplay(Generic[NodeType]):
     def __init__(self, node: Type[NodeType]):
         self._node = node
 
-    def serialize(self, display_context: "WorkflowDisplayContext") -> JsonObject:
+    def serialize(self, display_context: "WorkflowDisplayContext", **kwargs: Any) -> JsonObject:
         raise NotImplementedError(f"Serialization for nodes of type {self._node.__name__} is not supported.")
 
     def get_definition(self) -> NodeDefinition:
--- a/ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
@@ -30,13 +30,10 @@ class BaseTryNodeDisplay(BaseNodeVellumDisplay[_TryNodeType], Generic[_TryNodeTy
         node_display_class = get_node_display_class(BaseNodeVellumDisplay, inner_node)
         node_display = node_display_class(inner_node)
 
-        serialized_node = node_display.serialize(display_context)
-
-        serialized_node_data = serialized_node.get("data")
-        if isinstance(serialized_node_data, dict):
-            serialized_node_data["error_output_id"] = str(
-                self.error_output_id or uuid4_from_hash(f"{node_display.node_id}|error_output_id")
-            )
+        serialized_node = node_display.serialize(
+            display_context,
+            error_output_id=self.error_output_id or uuid4_from_hash(f"{node_display.node_id}|error_output_id"),
+        )
 
         serialized_node_definition = serialized_node.get("definition")
         if isinstance(serialized_node_definition, dict):
--- a/ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
@@ -1,13 +1,13 @@
 from uuid import UUID
 from typing import Any, ClassVar, Generic, Optional, TypeVar
 
+from vellum.workflows.nodes.core.try_node.node import TryNode
+from vellum.workflows.nodes.utils import get_wrapped_node
+from vellum.workflows.types.core import JsonObject
 from vellum_ee.workflows.display.nodes.base_node_vellum_display import BaseNodeVellumDisplay
 from vellum_ee.workflows.display.nodes.get_node_display_class import get_node_display_class
 from vellum_ee.workflows.display.types import WorkflowDisplayContext
 from vellum_ee.workflows.display.utils.uuids import uuid4_from_hash
-from vellum.workflows.nodes.core.try_node.node import TryNode
-from vellum.workflows.nodes.utils import get_wrapped_node
-from vellum.workflows.types.core import JsonObject
 
 _TryNodeType = TypeVar("_TryNodeType", bound=TryNode)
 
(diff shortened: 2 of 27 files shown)


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -31,7 +31,7 @@ class BaseNodeDisplay(Generic[NodeType]):
     def __init__(self, node: Type[NodeType]):
         self._node = node
 
-    def serialize(self, display_context: "WorkflowDisplayContext", **kwargs: Any) -> JsonObject:
+    def serialize(self, display_context: "WorkflowDisplayContext") -> JsonObject:
```
dvargasfuertes: would be nice to keep this API simple - especially since the param that was injected seemed like it could be handled at the consumer (see TryNodeDisplay below)

ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
```diff
@@ -30,9 +30,24 @@ def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> J
...
-            display_context,
-            error_output_id=str(self.error_output_id or uuid4_from_hash(f"{node_display.node_id}|error_output_id")),
-        )
+        serialized_node = node_display.serialize(display_context)
+
+        serialized_node_data = serialized_node.get("data")
+        if isinstance(serialized_node_data, dict):
+            serialized_node_data["error_output_id"] = str(
+                self.error_output_id or uuid4_from_hash(f"{node_display.node_id}|error_output_id")
+            )
```
dvargasfuertes: Try nodes annotate their inner nodes. So, I think TryNodeDisplay should annotate its inner display, instead of passing error output id

ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
```diff
@@ -30,9 +30,24 @@ def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> J
...
+
+        serialized_node_definition = serialized_node.get("definition")
+        if isinstance(serialized_node_definition, dict):
```
dvargasfuertes, line 39: This is the main reason I wanted to pull error output id up - we are doing the same for definition

ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
```diff
@@ -30,9 +30,24 @@ def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> J
...
+            serialized_node_definition_module = serialized_node_definition.get("module")
+            if isinstance(serialized_node_definition_module, list):
+                serialized_node_definition_module.extend(
+                    [
+                        serialized_node_definition["name"],
+                        "<decorator>",
```
dvargasfuertes: More on this guy in `try_node.node` below

src/vellum/workflows/nodes/core/try_node/node.py
```diff
@@ -56,34 +56,60 @@ class TryNode(BaseNode[StateType], Generic[StateType], metaclass=_TryNodeMeta):
     class Outputs(BaseNode.Outputs):
         error: Optional[VellumError] = None
 
-    def run(self) -> Outputs:
+    def run(self) -> Iterator[BaseOutput]:
```
dvargasfuertes, line 62: rewrote this method to work more like subworkflow nodes, since we need to handle streaming

src/vellum/workflows/runner/runner.py
```diff
@@ -540,11 +550,15 @@ def _run_cancel_thread(self) -> None:
         )
 
     def stream(self) -> WorkflowEventStream:
-        background_thread = Thread(target=self._run_background_thread)
+        background_thread = Thread(
+            target=self._run_background_thread, name=f"{self.workflow.__class__.__name__}.background_thread"
```
dvargasfuertes, line 554: thread names help with debugging

src/vellum/workflows/nodes/core/try_node/node.py
```diff
@@ -101,11 +128,15 @@ class Subworkflow(BaseWorkflow):
...
-
+            WrappedNode = type(
+                cls.__name__,
+                (TryNode,),
+                {
+                    "__module__": f"{inner_cls.__module__}.{inner_cls.__name__}.{ADORNMENT_MODULE_NAME}",
+                    "on_error_code": _on_error_code,
+                    "subworkflow": Subworkflow,
+                },
+            )
```
dvargasfuertes: This is the most important block in the PR to review

First some python context. When Python initiates a class in local scope, they use this special `<locals>` module path. We are borrowing from that pattern to introduce `<adornment>`.

Currently, we were emitting the same `WrappedNode` dynamic class back to vellum, so our translator had no idea how to map it back to a node id. Therefore, we needed a dynamic class so that our workflow translator in Vellum knows how to map it back to the PromptNode. Instead of making the class name dynamic, we are making the module path dynamic and downstream of the original wrapped node in order to distinguish which original node was using reject on error toggle.

I chose the word adornment for a few reasons:
- consistency with the UI
- differentiates from "decorator" - not all python decorators return a new class, let alone a new node

Eager for feedback here both on approach and choice

src/vellum/workflows/workflows/base.py
```diff
@@ -363,3 +372,17 @@ def load_from_module(module_path: str) -> Type["BaseWorkflow"]:
...
+WorkflowExecutionPausedEvent.model_rebuild()
+WorkflowExecutionResumedEvent.model_rebuild()
+WorkflowExecutionStreamingEvent.model_rebuild()
+
+NodeExecutionInitiatedEvent.model_rebuild()
+NodeExecutionFulfilledEvent.model_rebuild()
+NodeExecutionRejectedEvent.model_rebuild()
+NodeExecutionPausedEvent.model_rebuild()
+NodeExecutionResumedEvent.model_rebuild()
+NodeExecutionStreamingEvent.model_rebuild()
```
dvargasfuertes, line 388: new parent context require pydantic rebuilds of both events and bodies

tests/helpers.py
```diff
@@ -0,0 +1,17 @@
+import sys
+import threading
+import time
+import traceback
+
+
+def debug_threads(wait_time: float = 1):
```
dvargasfuertes, line 10: neat helper that assisted me in figuring out why a test was hanging. Thanks cursor

tests/workflows/stream_try_node_annotation/tests/test_workflow.py
```diff
@@ -0,0 +1,72 @@
...
+
+    assert events[3].name == "workflow.execution.streaming"
+    assert events[3].output.name == "final_value"
+    assert events[3].output.is_initiated
+
+    assert events[4].name == "node.execution.streaming"
+    assert events[4].output.delta == "apple apple"
+
+    assert events[5].name == "workflow.execution.streaming"
+    assert events[5].output.delta == "apple apple"
```
dvargasfuertes, line 45: Shows that we will see prompt node deltas in the UI

dvargas92495 (commented):
CI is failing even though all of the tests pass locally. This PR is meaty, so I'm going to hope that resolving those tests will not result in huge diffs and request reviews now while I hunt down what's going on here

dvargas92495:
omg I was on python 3.13 locally somehow - once I switched to python 3.9 I was able to repro the test failure. This means that this is only an issue for 3.9... fun -.-

dvargas92495:
hell yea

src/vellum/workflows/nodes/core/try_node/node.py
```diff
@@ -101,11 +127,15 @@ class Subworkflow(BaseWorkflow):
                 class Outputs(inner_cls.Outputs):  # type: ignore[name-defined]
                     pass
 
-            class WrappedNode(TryNode[StateType]):
-                on_error_code = _on_error_code
-
-                subworkflow = Subworkflow
-
+            WrappedNode = type(
```
noanflaherty, line 138: nit: A comment explaining what going on here would be helpful

src/vellum/workflows/nodes/displayable/tests/test_inline_text_prompt_node.py
```diff
@@ -136,13 +136,13 @@ def generate_prompt_events(*args: Any, **kwargs: Any) -> Iterator[ExecutePromptE
...
-    # We need mypy support for annotations to remove these type ignores
-    # https://app.shortcut.com/vellum/story/4890
-    assert outputs.error == WacVellumError(  # type: ignore[attr-defined]
-        message="OpenAI failed",
-        code=VellumErrorCode.PROVIDER_ERROR,
-    )
-    assert outputs.text is UNDEF  # type: ignore[attr-defined]
+    assert BaseOutput(
+        name="error",
+        value=WacVellumError(
```
noanflaherty: [oos] We should rename `WacVellumError`

tests/helpers.py
noanflaherty: nit: a docstring would be helpful

ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
noanflaherty: [4/5] I think we can get away with it in this very specific case due to the fact that
1) We happen to have inner nodes whose serialized repr all _happen_ to model this the same way, and
2) These nodes will go away at some point

However, more philosophically, I think the previous pattern is more true to the intention here. TryNode shouldn't have to know anything about what it wraps or the intended data model of the inner node's serialized representation. This is why we did it the way we did. I think of this more like django's serializer context – we're allowing access to parent serializer data in inner serializers. We do this now by passing a kwarg, but could do it just as well with a concept of context. In either case, I feel 4/5 that TryNode shouldn't know/care about what it wrapss

noanflaherty (approved):
Approving to unblock, but I'm not sold on the removal of `error_output_id` as a kwarg.

ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
dvargasfuertes: I suppose error_output_id is node specific. definition isn't though - that will be universal. So proposal:
- some notion of context equivalent for error output id (or I'll just add back the kwarg if more than >15 min LoE)
- definition manipulation stays on the outside here

noanflaherty: agreed!

dvargasfuertes: I ended up adding back the kwarg - context on init doesn't work bc the error output id calculation needs the already initiated `node_display.node_id`.

Emotionally I still feel "an inner node doesn't need to know it's wrapped" more than "a wrapper node doesn't need to know what it's wrapping", particularly since the way the wrapper adds the `error` output itself is consistent across all nodes, but `error_output_id` itself should* be short lived anyway and defer to your domain DRI

dvargasfuertes: > TryNode shouldn't have to know anything about what it wraps or the intended data model of the inner node's serialized representation

I now disagree with this statement, also with a 4/5. The thing that a decorator wraps in python is defined as an input to the decorator. This means the decorator knows the type of the thing its wrapping and could discriminate and act accordingly. Not doing so is what I think is leading to a lot of adornment logic bleeding into base node components: https://github.com/vellum-ai/vellum-python-sdks/pull/655/files

PR #220, opened 2024-11-27
[Codegen] Handle Unspecified Workspace Secret

noanflaherty:
It's possible in the Vellum UI for a Node Input to need a Workspace Secret, but for the user to not yet have specified which secret should be used.

Codegen needs to be able to handle this case.

In the PR currently, we simply generate `None` as the value instead of a `VellumSecretReference`. This is what wSDK supports currently, but it does open an interesting question. How does serialization later know to serialize this value into a Workspace Secret pointer? It doesn't. It'll think it's a constant value string with a null value. Maybe this is fine? I have a reverted commit showing the direction of a second option that's maybe worth exploring.


dvargas92495:
> It's possible in the Vellum UI for a Node Input to need a Workspace Secret, but for the user to not yet have specified which secret should be used.

When do we do this? _Should_ we allow this?

dvargas92495 (approved):
cool with `None` over `VellumSecretReference(None)` for now

noanflaherty:
Synced offline – rolling with `None`

PR #221, opened 2024-11-27
Harrison/sc 4649/remove fixture exclusions

NgoHarrison:


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -92,6 +92,29 @@ def _get_explicit_node_display_attr(
         if node_display_attribute is None:
             return None
 
+        origin = get_origin(attribute_type)
```
NgoHarrison, line 95: Did this because I was getting this error:
```  def __subclasscheck__(self, cls):
>       raise TypeError("Subscripted generics cannot be used with"
                        " class and instance checks")
E       TypeError: Subscripted generics cannot be used with class and instance checks```

File: ee/codegen_integration/test_code_to_display.py [hunk 2]
```diff
@@ -33,4 +35,35 @@ def exclude_obj_callback(value, key):
         actual_serialized_workflow,
         exclude_obj_callback=exclude_obj_callback,
         significant_digits=6,
+        # This is for the input_variables order being out of order sometimes.
+        ignore_order=True
```
NgoHarrison, line 40: Input variables were sometimes in the wrong order but they did have the right values so this is a quick bandaid. If this shouldn't be here then I can create a ticket to explore why the inputs display would sometimes be out of order. At first glance I assumed it was because it's a dict but that doesn't quite make sense since others don't seem to have this problem like search node

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -98,6 +106,7 @@ export class WorkflowContext {
     this.parentNode = parentNode;
 
     this.sdkModulePathNames = generateSdkModulePaths(workflowsSdkModulePath);
+    this.workflowRawEdges = workflowRawEdges;
```
noanflaherty, line 109: nit 2/5: If we're adding edges to `WorkflowContext`, it'd be great to remove them from `Workflow.Args` since it's now kinda duplicative. Note that if we do this, there might be wonky stuff cuz I think I filter out edges in some places.

ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
```diff
@@ -39,6 +51,7 @@ class ConditionalNodeDisplay(BaseConditionalNodeDisplay[ConditionalNode]):
             "cf63d0ad-5e52-4031-a29f-922e7004cdd8"
         ),
     }
+    port_displays = {}
```
noanflaherty, line 54: [q] Do we expect this to be empty?

ee/codegen_integration/test_code_to_display.py
```diff
@@ -33,4 +35,35 @@ def exclude_obj_callback(value, key):
...
     )
+
+
+def _process_position_hook(key, value):
```
noanflaherty: Can we add type hints for this? I'm surprised mypy didn't fail it. cc @dvargas92495 for devexp.

ee/codegen/src/context/workflow-context/workflow-context.ts
noanflaherty: If you do decide to take this one, I suggest merging this as is and doing as a follow-up.

noanflaherty (approved):
This looks awesome! Nice work – I know it was a hard one

ee/codegen_integration/test_code_to_display.py
dvargasfuertes: hm I wonder if this directory is getting excluded from type checking. Adding to list

ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
NgoHarrison: Yes because I didn't pass in any edges for the workflow data factory

PR #222, opened 2024-11-27
Add codegen publish job and docs

dvargas92495:
Adding more docs on how our codegen package should publish, as well as a GHA to start to do this automatically.

### DIFF
--- a/.github/workflows/ci-codegen.yaml
+++ b/.github/workflows/ci-codegen.yaml
@@ -1,14 +1,30 @@
 name: ci-codegen
 
-on:
-  push:
-    paths:
-      - ".github/workflows/ci-codegen.yaml"
-      - "pyproject.toml"
-      - "ee/codegen/**"
-      - "ee/codegen_integration/**"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-codegen.yaml
```diff
@@ -35,18 +51,41 @@ jobs:
...
+
+      - name: Publish
+        run: npm publish
         working-directory: ee/codegen
+
+      - name: Notify on failure
+        if: failure()
+        run: |
+          echo "Publishing failed. Please check the logs and publish manually."
+          echo "Please see the ee/codegen/README.md for more information."
```
dvargasfuertes, line 91: Adding this step so that this action doesn't fail on main, but I can iterate on failures over time. I imagine that the first issue I'll run into will be around authentication of `gar-login`

dvargas92495:
Gonna keep moving through these quick, but will followup on any feedback left

noanflaherty (approved):

PR #223, opened 2024-11-27
[Codegen] Introduce Custom AstNode for Json

noanflaherty:
We need to be able to convert json serializable objects into python primitive values. This introduces a custom AstNode for this.


ee/codegen/src/__test__/__snapshots__/json.test.ts.snap
```diff
@@ -0,0 +1,90 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Json > write > should handle arrays 1`] = `
+"[1, "test", True]
```
dvargasfuertes, line 4: Nice it's the full json

dvargas92495 (commented):
✅

PR #224, opened 2024-11-27
[SC-5147 Correctly Generate JSON Values

noanflaherty:
Uses the new Json node introduced in https://github.com/vellum-ai/vellum-python-sdks/pull/223 everywhere where we should be generating valid json.

### DIFF
--- a/ee/codegen/src/generators/chat-message-content.ts
+++ b/ee/codegen/src/generators/chat-message-content.ts
@@ -8,6 +8,7 @@ import {
 } from "vellum-ai/api";
 
 import { VELLUM_CLIENT_MODULE_PATH } from "src/constants";
+import { Json } from "src/generators/json";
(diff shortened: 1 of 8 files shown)


ee/codegen/src/generators/chat-message-content.ts
```diff
@@ -119,9 +120,7 @@ export class ChatMessageContent extends AstNode {
         functionCallChatMessageContentValueArgs.push(
           new MethodArgument({
             name: "arguments",
-            value: python.codeBlock(
-              JSON.stringify(functionCallChatMessageContentValue.arguments)
-            ),
+            value: new Json(functionCallChatMessageContentValue.arguments),
```
dvargasfuertes, line 123: love seeing first class > `codeBlock`

dvargas92495 (approved):

PR #226, opened 2024-11-27
Clean Up API Nodes

noanflaherty:
Does everything needed to make sure that when you pull down an `APINode`, it generates valid code.

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
@@ -58,10 +58,10 @@ class APINodeDisplay(BaseAPINodeDisplay[APINode]):
 `;
 
 exports[`ApiNode > basic > getNodeFile 1`] = `
-"from vellum.workflows.nodes.displayable import ApiNode
+"from vellum.workflows.nodes.displayable import APINode as BaseAPINode
 
 
-class APINode(ApiNode):
+class APINode(BaseAPINode):
     pass
 "
 `;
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
```diff
@@ -58,10 +58,10 @@ class APINodeDisplay(BaseAPINodeDisplay[APINode]):
 `;
 
 exports[`ApiNode > basic > getNodeFile 1`] = `
-"from vellum.workflows.nodes.displayable import ApiNode
+"from vellum.workflows.nodes.displayable import APINode as BaseAPINode
```
noanflaherty, line 61: The actual base node had been renamed.

src/vellum/workflows/nodes/displayable/bases/api_node/node.py
```diff
@@ -26,11 +26,11 @@ class BaseAPINode(BaseNode, Generic[StateType]):
     url: str
     method: APIRequestMethod
     data: Optional[str] = None
-    json: Optional["JsonObject"] = None
+    json: Optional["Json"] = None
```
noanflaherty, line 29: Cleans up some types to what I claim to be more correct.

src/vellum/workflows/utils/vellum_variables.py
```diff
@@ -32,6 +33,17 @@ def primitive_type_to_vellum_variable_type(type_: Union[Type, BaseDescriptor]) -
...
+            # Matches the type found at vellum.workflows.utils.vellum_variables.Json
+            if types == [
+                bool,
+                int,
+                float,
+                str,
+                typing.List[typing.ForwardRef('Json')],  # type: ignore [misc]
+                typing.Dict[str, typing.ForwardRef('Json')],  # type: ignore [misc]
+            ]:
+                return "JSON"
```
noanflaherty, line 46: This is kinda funky, but it works. Without this, a test fails.

noanflaherty:
Open to post-push feedback on this! Merging so I can test e2e

src/vellum/workflows/utils/vellum_variables.py
dvargasfuertes: [oos] wondering if we should do a set comparison (and if `BaseDescriptor.types` should actually be a `set`...)

dvargas92495 (commented):
✅

PR #230, opened 2024-11-27
Remove subworkflow serialization exclusion

peicodes:
The process of doing this unveiled an issue with serializing inline subworkflows: https://app.shortcut.com/vellum/story/5660/fix-output-id-in-subworkflow-nodes

### DIFF
--- a/ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/nodes/subworkflow_node/__init__.py
+++ b/ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/nodes/subworkflow_node/__init__.py
@@ -1,7 +1,7 @@
 from uuid import UUID
 
 from vellum_ee.workflows.display.nodes import BaseInlineSubworkflowNodeDisplay
-from vellum_ee.workflows.display.nodes.types import PortDisplayOverrides
+from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplayOverrides
 from vellum_ee.workflows.display.vellum import NodeDisplayData, NodeDisplayPosition
 
 from ....nodes.subworkflow_node import SubworkflowNode
@@ -17,6 +17,9 @@ class SubworkflowNodeDisplay(BaseInlineSubworkflowNodeDisplay[SubworkflowNode]):
     node_input_ids_by_name = {}
     port_displays = {
         SubworkflowNode.Ports.default: PortDisplayOverrides(id=UUID("fa5c22bc-2499-43fa-880f-75fb20d0587f"))
+    }
+    output_display = {
+        SubworkflowNode.Outputs.final_output: NodeOutputDisplay(id=UUID("6ab3665f-881d-488b-9124-a6da40136c68"), name="final_output")
     }
     display_data = NodeDisplayData(
         position=NodeDisplayPosition(x=1991.684833859175, y=178.94753425793772), width=None, height=None
--- a/ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/nodes/subworkflow_node/__init__.py
+++ b/ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/nodes/subworkflow_node/__init__.py
@@ -5,6 +5,8 @@ from vellum_ee.workflows.display.nodes.types import PortDisplayOverrides
 from vellum_ee.workflows.display.vellum import NodeDisplayData, NodeDisplayPosition
 
 from ....nodes.subworkflow_node import SubworkflowNode
+from .nodes import *
+from .workflow import *
 
 
 class SubworkflowNodeDisplay(BaseInlineSubworkflowNodeDisplay[SubworkflowNode]):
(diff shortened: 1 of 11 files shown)


ee/vellum_ee/workflows/display/workflows/get_vellum_workflow_display_class.py
```diff
@@ -24,9 +24,4 @@ def get_workflow_display(
         except IndexError:
             return base_display_class(workflow_class)
 
-    if not issubclass(workflow_display_class, base_display_class):
```
peicodes, line 27: @dvargas92495 I had to remove this because there are instances where we want to retrieve a subworkflow's display class and that display class won't be a subclass of the outer workflow's display class.

I might be using this utility wrong though. Thoughts?

peicodes:
@noanflaherty @NgoHarrison  I made some changes to the display classes to ensure serialization generates the correct data. This has codegen implications. Could one of you help with this part?

ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/nodes/subworkflow_node/__init__.py
```diff
@@ -1,10 +1,12 @@
 from uuid import UUID
 
 from vellum_ee.workflows.display.nodes import BaseInlineSubworkflowNodeDisplay
-from vellum_ee.workflows.display.nodes.types import PortDisplayOverrides
+from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplayOverrides
 from vellum_ee.workflows.display.vellum import NodeDisplayData, NodeDisplayPosition
 
 from ....nodes.subworkflow_node import SubworkflowNode
+from .nodes import *
```
peicodes, line 8: @noanflaherty @NgoHarrison Codegen: We need to import the displays for the subworkflow to register the overrides

ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/nodes/subworkflow_node/__init__.py
```diff
@@ -16,6 +18,9 @@ class SubworkflowNodeDisplay(BaseInlineSubworkflowNodeDisplay[SubworkflowNode]):
     port_displays = {
         SubworkflowNode.Ports.default: PortDisplayOverrides(id=UUID("fa5c22bc-2499-43fa-880f-75fb20d0587f"))
     }
+    output_display = {
+        SubworkflowNode.Outputs.final_output: NodeOutputDisplay(id=UUID("6ab3665f-881d-488b-9124-a6da40136c68"), name="final_output")
```
peicodes, line 22: The display classes need to know what the subworkflow node's output IDs are.
This may be unnecessary if we fix the output display issue mentioned in the card in this PR's description, since this ID should be pulled from the workflow's outputs

ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/workflow.py
```diff
@@ -44,7 +44,7 @@ class WorkflowDisplay(VellumWorkflowDisplay[Workflow]):
     }
     output_displays = {
         Workflow.Outputs.final_output: WorkflowOutputVellumDisplayOverrides(
-            id=UUID("b38e08c7-904d-4f49-b8fb-56e1eff254d6"),
+            id=UUID("6ab3665f-881d-488b-9124-a6da40136c68"),
```
peicodes, line 47: This was the wrong ID

ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/nodes/subworkflow_node/__init__.py
noanflaherty: Yep, this indeed makes sense! In fact, we probably also need to include a `from .workflow import *` so that the inner workflow's display class is in the import path.

Please feel free to remove this case from codegen tests and cut a card in my epic to add back after I've added codegen support for this piece and any others this PR exposes.

ee/vellum_ee/workflows/display/workflows/get_vellum_workflow_display_class.py
noanflaherty: It's possible that with my suggestion above of including `from .workflow import *` this will go back to working without the need for this change. Not sure though.

noanflaherty (approved):
Love when we add a test and it finds bugs. Nice!

ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/workflow.py
peicodes: This is the ID of the workflow's output, which is defined in the JSON here:
```
  "output_variables": [
    {
      "id": "6ab3665f-881d-488b-9124-a6da40136c68",
      "key": "final-output",
      "type": "STRING"
    }
  ]
```

ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/nodes/subworkflow_node/__init__.py
peicodes: I have `from .workflow import *` just below this line - doesn't fix the subworkflow node output ID issue :/

ee/vellum_ee/workflows/display/workflows/get_vellum_workflow_display_class.py
dvargasfuertes: My understanding of the usage of this method was that `base_display_class` represents a common _base_ used throughout serialization, not strictly the exact display class the workflow uses.

However, if Noa's suggestion doesn't work out and removing this passes all tests with no obvious counterexample test to make, then fire away 👍

PR #231, opened 2024-11-28
 Node Parent Context

marlenc3:
Initial Pass of Parent Context inside of Workflow Runner! 
for now, passing in parent context, however @noanflaherty had a pretty cool idea to use contextmanager that we can then use like we do logging_context. Next pass will be attempting to have some sort of implementation of it, so that we can avoid parameter passing


tests/workflows/basic_parent_context/tests/test_basic_parent_context_workflow_runner.py
```diff
@@ -0,0 +1,31 @@
...
+    assert len(events) == 4
+
+    assert events[0].name == "workflow.execution.initiated"
+    assert events[0].parent is None
+
+    assert events[1].name == "node.execution.initiated"
+    parent_context = events[1].parent.model_dump() if events[1].parent else {}
+    assert parent_context.get('type') == 'WORKFLOW'
+    assert parent_context.get('parent') is None
+    assert parent_context.get('workflow_definition') is not None
```
dvargasfuertes, line 28: [nit] could we assert the contents of this guy?

dvargas92495 (approved):
Good stuff!

One piece of general feedback - it was unclear whether this PR was requesting review:
- Post on slack indicated "ready for review"
- `WIP` prefix in title indicated "not ready for review"
- PR in non-draft state indicated "ready for review"
- No reviewers tagged indicated "not ready for review"

Aligning these signals to a consistent direction one way or another will help reviewers know when it is or isn't ready for us to come take a look!

marlenc3:
> Good stuff!
> 
> One piece of general feedback - it was unclear whether this PR was requesting review:
> 
> * Post on slack indicated "ready for review"
> * `WIP` prefix in title indicated "not ready for review"
> * PR in non-draft state indicated "ready for review"
> * No reviewers tagged indicated "not ready for review"
> 
> Aligning these signals to a consistent direction one way or another will help reviewers know when it is or isn't ready for us to come take a look!

@dvargas92495 completely my bad! I should have been explicit in wanting just a quick 'this is the right path' 

that being said, new update available that's the new iteration on this!

tests/workflows/basic_parent_context/tests/test_basic_wrapper.py
```diff
@@ -0,0 +1,36 @@
...
+from tests.workflows.basic_parent_context.basic_workflow import StartNode
+
+
+class A:
+    def __init__(self):
+        self._span_id = uuid4()
+        self._context = None
+
+    @wrapper_execution_parent_context(
+        lambda self: NodeParentContext(node_definition=StartNode().__class__, span_id=self._span_id))
```
dvargasfuertes: [blocking] we should only initialize node's during the workflow runner. here, we could easily simplify to just `node_definition=StartNode`

src/vellum/core/context.py
```diff
@@ -0,0 +1,82 @@
+from contextlib import contextmanager
+from functools import wraps
+import inspect
+import threading
+from typing import Any, Callable, Dict, Iterator, TypeVar, Union, cast
+
+from vellum.workflows.events.types import ParentContext
+
+LoggingContext = Dict[str, Any]
```
dvargasfuertes: [nit] This should be `ExecutionContext` correct?

dvargasfuertes: Would also be nice stringly type this, instead of us calling `.get("parent_context")` below

src/vellum/core/context.py
```diff
@@ -0,0 +1,82 @@
...
+    """Context manager for handling execution context."""
+    prev_context = get_execution_context()
+    try:
+        set_execution_context({**prev_context, **kwargs})
+        yield
+    finally:
+        set_execution_context(prev_context)
+
+
+def wrapper_execution_parent_context(
```
dvargasfuertes: [blocking] I'm having trouble understanding what this annotation enables for us, especially with the type ignores below. I _think_ this is eventually meant to replace/combine with our existing [context](https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/state/context.py)?

It looks promising, but I'd love to chat through this pattern synchronously before we commit to it. unfortunately, I don't think I'd be able to until earliest Sunday.

In the meantime, I think we could split out commit 1 from this PR (basically, the minimal changes needed to pass this test: `tests/workflows/basic_parent_context/tests/test_basic_parent_context_workflow_runner.py`). Then we could leave the annotation stuff off for a separate PR until we can chat through it.

wdyt?

src/vellum/core/context.py
```diff
@@ -0,0 +1,82 @@
+from contextlib import contextmanager
```
dvargasfuertes: [blocking] we need to avoid putting stuff in `vellum/core` bc Fern auto generates into that repo and will cause conflicts. would suggest something within the `vellum/workflows` directory

src/vellum/core/context.py
marlenc3: Heyo! Let's do it. I'm also going to be hard to reach till Sunday. 

I'll split the PR, and harden some of the code here in this PR. 

Thanks for giving a look! Appreciate it!

dvargas92495 (approved):

dvargas92495 (approved):

kirawoods (commented):
✅

src/vellum/workflows/runner/runner.py
```diff
@@ -143,6 +145,11 @@ def _run_work_item(self, node: BaseNode[StateType], span_id: UUID) -> None:
                         node_definition=node.__class__,
                         inputs=node._inputs,
                     ),
+                    parent=WorkflowParentContext(
+                        span_id=span_id,
+                        workflow_definition=self.workflow.__class__,
+                        parent=self._parent_context
+                    )
```
siddseethepalli, line 152: Let's either use an execution context wrapper or DRY this out

siddseethepalli (commented):
✅

src/vellum/workflows/runner/runner.py
dvargasfuertes: we're working on an execution context wrapper in a separate branch

PR #232, opened 2024-11-28
[SC-4649] Remove code execution exclusion

NgoHarrison:


peicodes:
Should this also be removing code execution from `exclude_fixtures`?

ee/codegen/src/generators/nodes/code-execution-node.ts
```diff
@@ -135,10 +146,97 @@ export class CodeExecutionNode extends BaseSingleFileNode<
       })
     );
 
+    statements.push(
```
noanflaherty: nit: We typically put these fields at the top of the display class.

ee/codegen/src/generators/nodes/code-execution-node.ts
```diff
@@ -112,6 +113,16 @@ export class CodeExecutionNode extends BaseSingleFileNode<
...
 
+    const argId = this.nodeInputsByKey.get("arg")?.nodeInputData.id;
+    if (argId) {
+      statements.push(
+        python.field({
+          name: "arg_input_id",
+          initializer: python.TypeInstantiation.uuid(argId),
+        })
+      );
+    }
```
noanflaherty: [blocking] Code execution nodes have dynamic inputs for their arguments. Meaning, it's guaranteed for all code execution nodes to have a node input called `arg_input_id`. It just happens that this fixture has a single node input with this name.

Code execution nodes behave more similarly to inline prompt nodes or templating nodes, where they have a class attribute call `code_inputs` that defines a mapping of their inputs. We need to generate that. I suggest looking at templating node's `inputs` class attribute for the pattern we should use.

ee/codegen_integration/conftest.py
```diff
@@ -34,10 +36,29 @@ def _get_fixture_paths(root: str) -> Tuple[str, str]:
...
-            "simple_code_execution_node",
         }
     )
 )
 def code_to_display_fixture_paths(request) -> Tuple[str, str]:
     root = request.param
     return _get_fixture_paths(root)
+
+# Save the original `open`
+builtin_open = open
```
noanflaherty: It doesn't look like this is actually used?

ee/codegen_integration/conftest.py
```diff
@@ -34,10 +36,29 @@ def _get_fixture_paths(root: str) -> Tuple[str, str]:
...
+
+# Fixture that mocks open for code execution nodes
+@pytest.fixture
+def mock_open_code_execution_file():
```
noanflaherty: Given that this fixture is only really use for code execution nodes, it'd be nice to colocate it closer to where it's used.

ee/codegen_integration/fixtures/simple_code_execution_node/code/display/nodes/code_execution_node.py
```diff
@@ -1,22 +1,32 @@
...
-from vellum_ee.workflows.display.nodes.types import PortDisplayOverrides
+from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplayOverrides
 from vellum_ee.workflows.display.vellum import NodeDisplayData, NodeDisplayPosition
 
 from ...nodes.code_execution_node import CodeExecutionNode
 
 
 class CodeExecutionNodeDisplay(BaseCodeExecutionNodeDisplay[CodeExecutionNode]):
     code_input_id = UUID("19b05769-cee3-4659-80d1-66fcae4e27c3")
+    arg_input_id = UUID("78da07ee-cc77-445e-af85-f60ab4f7a59f")
```
noanflaherty: To go with my comment above, this should not be here and we should instead be generating a class attribute called something like `node_input_ids_by_name` similar to `BaseTemplatingNodeDisplay.input_ids_by_name`

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
```diff
@@ -2,9 +2,20 @@
 from typing import Union
 
 
+def get_project_root() -> str:
```
noanflaherty, line 5: Could you elaborate on why these code changes were needed?

noanflaherty (changes requested):

ee/codegen/src/generators/nodes/code-execution-node.ts
NgoHarrison: Updated

ee/codegen_integration/conftest.py
NgoHarrison: It's used a few lines below

ee/codegen_integration/conftest.py
NgoHarrison: Moved this to the bottom of the `test_code_to_display`

ee/codegen_integration/fixtures/simple_code_execution_node/code/display/nodes/code_execution_node.py
NgoHarrison: Updated

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
NgoHarrison: I was getting a file not found for the given path, I suspect it was because we changed directory structures when we moved from old prototype repo to here

noanflaherty (approved):

PR #243, opened 2024-11-28
remove guardrail node exclusion

peicodes:
@dvargas92495 I had to change the fixture because setting an output to `None` breaks things:

`node.Outputs.value.instance` on a `FinalOutput(FinalOutputNode)` will resolve to `FinalOutputNode.Outputsvalue` instead of `None`, which is what I would expect. If I set `value` to `lol` instead of `None`, it correctly resoles to `lol`.

### DIFF
--- a/ee/codegen_integration/fixtures/simple_guardrail_node/code/nodes/final_output.py
+++ b/ee/codegen_integration/fixtures/simple_guardrail_node/code/nodes/final_output.py
@@ -1,7 +1,9 @@
 from vellum.workflows.nodes.displayable import FinalOutputNode
 from vellum.workflows.state import BaseState
 
+from ..inputs import Inputs
 
-class FinalOutput(FinalOutputNode[BaseState, float]):
+
+class FinalOutput(FinalOutputNode[BaseState, str]):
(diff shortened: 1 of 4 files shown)


ee/codegen_integration/fixtures/simple_guardrail_node/code/nodes/final_output.py
```diff
@@ -1,7 +1,9 @@
...
 from vellum.workflows.state import BaseState
 
+from ..inputs import Inputs
 
-class FinalOutput(FinalOutputNode[BaseState, float]):
+
+class FinalOutput(FinalOutputNode[BaseState, str]):
     class Outputs(FinalOutputNode.Outputs):
-        value = None
+        value = Inputs.actual
```
peicodes: Try this with None and serialization will fail due to this value resolving to `FinalOutputNode.Outputs.value` instead of `None`

dvargasfuertes: [nit] we should point this to the `score` output of the guardrail node no? closer to an actual example of how this node would be used

dvargas92495 (approved):

PR #245, opened 2024-11-29
Mitigate some issues relating to poetry in path

dvargas92495:
The poetry installation script by default simply suggests how to add poetry to your `PATH` variable. While there's no way to edit your PATH in the parent context we could do some light improvements to make the initial onboarding experience smoother:
- We add to your ~/.zshrc or ~/.bashrc a command to add poetry to your PATH
- We instruct you to source after it's over
- We update install deps to use the exact path to poetry to allow `make setup` to run smoothly

### DIFF
--- a/scripts/install_poetry.py
+++ b/scripts/install_poetry.py
@@ -255,6 +255,20 @@ You can test that everything is set up by executing:
 `{test_command}`
 """
 
+POST_MESSAGE_RC_ADDED_PATH = """{poetry} ({version}) is installed now. Great!
+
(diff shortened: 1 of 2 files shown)


scripts/install_poetry.py
```diff
@@ -515,7 +529,7 @@ def run(self) -> int:
             version = self._path
         else:
             try:
-                version, current_version = self.get_version()
+                version = self.get_version()
```
dvargasfuertes, line 532: `current_version` was unused

noanflaherty (commented):
✅

PR #246, opened 2024-11-29
Fix dynamic output related regressions locally

dvargas92495:
There were two tests failing locally but not on CI. Relevant thread here: https://github.com/vellum-ai/vellum-python-sdks/pull/246#issue-2704951463

While I'm not sure why they weren't failing on CI (they def should've, later PR to investigate), I was able to determine that the root of the bug locally was due to the same reused `__annotations__` dict across subclasses. Check out the script reverted in the first commit for a reproducible mvp of the issue. Because we the same annotations dict was reused across classes, it would cause failures depending on the order that tests were run. It appears that this is less of an issue in later versions or python


src/vellum/workflows/nodes/core/templating_node/node.py
```diff
@@ -49,7 +49,11 @@ def __new__(mcs, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any
         if not isinstance(parent, _TemplatingNodeMeta):
             raise ValueError("TemplatingNode must be created with the TemplatingNodeMeta metaclass")
 
-        parent.__dict__["Outputs"].__annotations__["result"] = parent.get_output_type()
+        annotations = parent.__dict__["Outputs"].__annotations__
+        parent.__dict__["Outputs"].__annotations__ = {
+            **annotations,
+            "result": parent.get_output_type(),
+        }
```
dvargasfuertes, line 56: I spent some time trying to make this change higher up - in theory, every new Outputs class should be have its own `__annotations__` dict, similar to how 3.11 works. I'm deferring that for now into another ticket, as there were some other wacky errors that came up when I did so

pyproject.toml
```diff
@@ -74,6 +74,9 @@ types-requests = "2.32.0.20240712"
 [tool.pytest.ini_options]
 testpaths = [ "tests" ]
 asyncio_mode = "auto"
+filterwarnings = [
+    "ignore::pytest.PytestCollectionWarning"
+]
```
dvargasfuertes, line 79: removes the annoying Collection Warning spam that emits bc we have types that start with `Test*`

ee/codegen_integration/conftest.py
```diff
@@ -26,15 +26,20 @@ def _get_fixture_paths(root: str) -> Tuple[str, str]:
...
-    params=_get_fixtures(
-        # TODO: Remove exclusions on all of these fixtures
-        # https://app.shortcut.com/vellum/story/4649/remove-fixture-exclusions-for-serialization
-        exclude_fixtures={
-            "simple_map_node",
-            "simple_code_execution_node",
-        }
-    )
+    params=_fixture_paths,
+    ids=_fixture_ids,
```
dvargasfuertes, line 42: `ids` make it easier to run a single case within a parameterization

marlenc3:
🔥

noanflaherty (commented):
✅

PR #250, opened 2024-11-29
Add a setup fern step to the Makefile

dvargas92495:
Our `make test` command depends on `fern`, so let's make sure it's installed before instructing contributors to run tests


noanflaherty (commented):
✅

PR #251, opened 2024-11-29
Check which version of python poetry is running in CI

dvargas92495:
This is to help diagnose which python poetry is running in CI so that we know local devs are installing the right one during setup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ setup-poetry:
 
 # We use the full path to poetry to avoid any issues with the shell configuration from the setup-poetry step
 install-deps:
-	$(HOME)/.local/bin/poetry lock && $(HOME)/.local/bin/poetry install
+	$(HOME)/.local/bin/poetry env use 3.9 && $(HOME)/.local/bin/poetry lock && $(HOME)/.local/bin/poetry install
 
 setup-pre-commit:
 	pre-commit install \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ setup-poetry:
 
 # We use the full path to poetry to avoid any issues with the shell configuration from the setup-poetry step
 install-deps:
-	$(HOME)/.local/bin/poetry lock && $(HOME)/.local/bin/poetry install
+	$(HOME)/.local/bin/poetry env use 3.9 && $(HOME)/.local/bin/poetry lock && $(HOME)/.local/bin/poetry install
```
dvargasfuertes, line 17: Forces the contributor to use `python3.9` during poetry setup. `poetry env use` is similar to `nvm use`

scripts/install_poetry.py
```diff
@@ -938,6 +938,12 @@ def main():
     if args.uninstall or string_to_bool(os.getenv("POETRY_UNINSTALL", "0")):
         return installer.uninstall()
 
+    preexisting_venv = os.environ.get("VIRTUAL_ENV")
+    if preexisting_venv:
+        # We need to deactivate the existing virtual environment, if it exists, in case the contributor
+        # switching from elsewhere in the Vellum ecosystem
+        os.system("deactivate")
```
dvargasfuertes, line 945: during onboarding, @peicodes had his venv activated from vellum, so this will forcibly deactivated before continuing with the rest of setup

noanflaherty (commented):
✅

PR #252, opened 2024-11-29
WIP - Node Execution Context Wrapper

marlenc3:
*** TBD 
split from previous PR https://github.com/vellum-ai/vellum-python-sdks/pull/231
main goal is deciding if this is a pattern we want to use going forward, or pass parent context in another way, which would be either passing it through initializations or building it when we need it through the intital_state.meta.parent information

Loom soon:  first loom https://www.loom.com/share/2c6c053e24d044b7ab8e9144109a0a6b?sid=f8440dbe-9036-49ff-8d5c-fa9fdd23c302
second loom - context walkthrough: https://www.loom.com/share/e08f140d7f0542748815395cde700cfa?sid=bdd09a69-2c30-4e9e-873b-e2cd5930c233
pros cons: https://www.loom.com/share/aa02f27977b64e9e954c905b47490f8a?sid=8695d072-717f-40ec-a4fb-02f5149e068b


src/vellum/workflows/runner/runner.py
```diff
@@ -132,7 +133,14 @@ def _emit_event(self, event: WorkflowEvent) -> WorkflowEvent:
...
         return event
 
+    @wrapper_execution_parent_context(
+        lambda self: WorkflowParentContext(
+            workflow_definition=self.workflow.__class__,
+            span_id=self._initial_state.meta.span_id,
+        )
+    )
     def _run_work_item(self, node: BaseNode[StateType], span_id: UUID) -> None:
+        parent_context = get_execution_context().get('parent_context')
```
dvargasfuertes, line 143: discussion topic: doing something like `with get_execution_context() as parent_context:` here

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,82 @@
+from contextlib import contextmanager
+from functools import wraps
+import inspect
+import threading
+from typing import Any, Callable, Dict, Iterator, TypeVar, Union, cast
+
+from vellum.workflows.events.types import ParentContext
+
+LoggingContext = Dict[str, Any]
```
siddseethepalli, line 9: `ExecutionContext`?

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,82 @@
...
+_CONTEXT_KEY = "_execution_context"
+
+local = threading.local()
```
siddseethepalli, line 12: Let's rename this to `_local` to discourage importing it from elsewhere

src/vellum/workflows/context.py
siddseethepalli: Can we make this a dataclass?

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,82 @@
...
+
+F = TypeVar('F', bound=Callable[..., Any])
+T = TypeVar('T')
+P = TypeVar('P', bound=ParentContext)
+
+
+def get_execution_context() -> LoggingContext:
+    """Retrieve the current execution context."""
+    return getattr(local, _CONTEXT_KEY, {})
```
siddseethepalli, line 21: Once we make `ExecutionContext` a dataclass, I'm thinking it might be better to raise an exception if it's not set instead of returning something like `None`, so we don't have to do a null check every time we use it.

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def set_execution_context(context: LoggingContext) -> None:
```
siddseethepalli, line 24: If we don't intend for this to be used outside of this module (which I think is a good idea if we can get away with it) let's rename it to `_set_execution_context`

src/vellum/workflows/runner/runner.py
siddseethepalli: @dvargas92495 the downside of that is it adds additional layer of indentation
I'd be a big fan of being able to do `.parent_context` instead of `.get("parent_context")`

siddseethepalli (commented):
Seems pretty reasonable

PR #244, opened 2024-11-28
Remove exclusion on map nodes

peicodes:
This PR removes exclusions for map nodes. It also highlights some bugs we didn't know existed.

Follow ups from findings while doing this:
- https://app.shortcut.com/vellum/story/5666/fix-codegen-for-map-node
- https://app.shortcut.com/vellum/story/5667/fix-output-display-reference-on-map-nodes

PR #253, opened 2024-11-30
Harrison/sc 5667/fix output display

NgoHarrison:
This PR will merge into the parent branch and this ONLY fixes the output display issue for map nodes. The other issues that come from the map node display tests will be addressed in the parent PR here: https://github.com/vellum-ai/vellum-python-sdks/pull/244

### DIFF
--- a/ee/codegen/src/generators/nodes/map-node.ts
+++ b/ee/codegen/src/generators/nodes/map-node.ts
@@ -5,7 +5,11 @@ import { OUTPUTS_CLASS_NAME } from "src/constants";
 import { MapNodeContext } from "src/context/node-context/map-node";
 import { BaseNestedWorkflowNode } from "src/generators/nodes/bases/nested-workflow-base";
 import { WorkflowProjectGenerator } from "src/project";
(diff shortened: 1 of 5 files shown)


ee/codegen/src/generators/nodes/map-node.ts
```diff
@@ -142,6 +139,60 @@ export class MapNode extends BaseNestedWorkflowNode<
     });
   }
 
+  protected getOutputDisplay(): python.Field {
```
NgoHarrison, line 142: actual fix below

ee/codegen_integration/conftest.py
```diff
@@ -31,6 +31,8 @@ def _get_fixture_paths(root: str) -> Tuple[str, str]:
     # https://app.shortcut.com/vellum/story/4649/remove-fixture-exclusions-for-serialization
     exclude_fixtures={
         "simple_code_execution_node",
+        # TODO: Remove after all other map node fixes go in
```
NgoHarrison, line 34: Will remove in the parent PR / parent branch this PR is merging into. This is temporary to have CI happy

noanflaherty (commented):
✅

PR #255, opened 2024-11-30
Save invoked ports in Node events

dvargas92495:
Full context: https://vellum-ai.slack.com/archives/C06P13ABK0W/p1732033982824129

In the immediate term, we need a way to backfill a conditional node event's chosen source handle id, referenced [here](https://github.com/vellum-ai/vellum/blob/main/django/app/domain_models/types/workflows.py#L992). Long term, I do believe it will be helpful debugging/monitoring info for events to know exactly which ports were invoked on any given fulfilled or streaming event.

Will block this PR until review

### DIFF
--- a/src/vellum/workflows/events/node.py
+++ b/src/vellum/workflows/events/node.py
@@ -1,6 +1,7 @@
 from typing import Any, Dict, Generic, Iterable, List, Literal, Optional, Set, Type, Union
 
-from pydantic import field_serializer
+from pydantic import ConfigDict, SerializerFunctionWrapHandler, field_serializer, model_serializer
+from pydantic.main import IncEx
 
 from vellum.core.pydantic_utilities import UniversalBaseModel
 from vellum.workflows.errors import VellumError
@@ -20,6 +21,15 @@ class _BaseNodeExecutionBody(UniversalBaseModel):
     @field_serializer("node_definition")
     def serialize_node_definition(self, node_definition: Type, _info: Any) -> Dict[str, Any]:
         return serialize_type_encoder(node_definition)
+
+    # Couldn't get this to work with model_config.exclude_none or model_config.exclude_defaults
+    # so we're excluding null invoked_ports manually here for now
+    @model_serializer(mode="wrap", when_used="json")
+    def serialize_model(self, handler: SerializerFunctionWrapHandler) -> Any:
+        serialized = super().serialize_model(handler)  # type: ignore[call-arg, arg-type]
+        if "invoked_ports" in serialized and serialized["invoked_ports"] is None:
+            del serialized["invoked_ports"]
+        return serialized
 
 
 class _BaseNodeEvent(BaseEvent):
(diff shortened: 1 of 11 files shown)


src/vellum/workflows/events/node.py
```diff
@@ -21,6 +22,15 @@ class _BaseNodeExecutionBody(UniversalBaseModel):
     def serialize_node_definition(self, node_definition: Type, _info: Any) -> Dict[str, Any]:
         return serialize_type_encoder(node_definition)
 
+    # Couldn't get this to work with model_config.exclude_none or model_config.exclude_defaults
+    # so we're excluding null invoked_ports manually here for now
```
dvargasfuertes, line 26: p99 of streaming events won't have invoked ports emitted, so we want to exclude them from the dump

src/vellum/workflows/events/tests/test_event.py
```diff
@@ -233,14 +239,88 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
...
+                ),
+            ),
+            {
+                "id": "123e4567-e89b-12d3-a456-426614174000",
+                "api_version": "2024-10-25",
+                "timestamp": "2024-01-01T12:00:00",
+                "trace_id": "123e4567-e89b-12d3-a456-426614174000",
+                "span_id": "123e4567-e89b-12d3-a456-426614174000",
+                "name": "node.execution.streaming",
+                "body": {
```
dvargasfuertes, line 263: shows that we successfully skipped invoked ports

src/vellum/workflows/events/tests/test_event.py
```diff
@@ -233,14 +239,88 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
...
+                        "module": module_root + ["events", "tests", "test_event"],
+                    },
+                    "outputs": {
+                        "example": "foo",
+                    },
+                    "invoked_ports": [
+                        {
+                            "name": "default",
+                        }
+                    ],
```
dvargasfuertes, line 309: data model that will be emitted when there _is_ invoked ports. Opted to keep it simple for now, with room to expand if we choose to down the line

src/vellum/workflows/events/tests/test_event.py
```diff
@@ -233,14 +239,88 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
...
         "workflow.execution.streaming",
         "workflow.execution.fulfilled",
         "workflow.execution.rejected",
+        "node.execution.streaming",
+        "node.execution.fulfilled",
     ],
 )
 def test_event_serialization(event, expected_json):
-    assert not DeepDiff(json.loads(event.model_dump_json()), expected_json)
+    assert not DeepDiff(event.model_dump(mode="json"), expected_json)
```
dvargasfuertes, line 326: Updated to exactly match what we do in vembda: https://github.com/vellum-ai/vembda-service/blob/main/python-workflow-server/src/workflow_server/executor.py#L203

src/vellum/workflows/nodes/displayable/conditional_node/node.py
```diff
@@ -15,7 +15,7 @@ class ConditionalNode(BaseNode):
     """
 
     class Ports(NodePorts):
-        def __call__(self, outputs: BaseOutputs, state: BaseState) -> Iterable[Port]:
+        def __call__(self, outputs: BaseOutputs, state: BaseState) -> Set[Port]:
```
dvargasfuertes, line 18: Decided to switch from our normal pattern here to start returning `Set`s instead of `Iterable`, mostly bc we needed to define this field as a `Set` on the pydantic event instead of `Iterable`

noanflaherty (approved):
lgtm

PR #257, opened 2024-11-30
Add setup-node make task

dvargas92495:
Aligns the node version we use locally with the one in CI and ensures that the contributor has the right node version locally, installing `nvm` just in case if not already in view


noanflaherty (commented):
✅

PR #260, opened 2024-11-30
[Codegen][SC-5666][Part 1] Fix Codegen for Map Nodes

noanflaherty:
This gets us one step of the way closer to correctly codegenning Map Nodes losslessly.

The focus of this PR is to get the value of `inputs_display` correct.

### DIFF
--- a/ee/codegen/src/context/workflow-context/workflow-context.ts
+++ b/ee/codegen/src/context/workflow-context/workflow-context.ts
@@ -109,6 +109,29 @@ export class WorkflowContext {
     this.workflowRawEdges = workflowRawEdges;
   }
 
+  /* Create a new workflow context for a nested workflow from its parent */
+  public createNestedWorkflowContext({
+    workflowLabel,
+    parentNode,
+    workflowRawEdges,
+  }: {
+    workflowLabel: string;
+    parentNode: BaseNode<WorkflowDataNode, BaseNodeContext<WorkflowDataNode>>;
+    workflowRawEdges: WorkflowEdge[];
+  }) {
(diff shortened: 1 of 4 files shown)


ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -28,8 +28,8 @@ export declare namespace WorkflowContext {
     moduleName: string;
     workflowLabel?: string;
     workflowClassName?: string;
-    inputVariableContextsById?: InputVariableContextsById;
-    nodeContextsByNodeId?: NodeContextsByNodeId;
+    globalInputVariableContextsById?: InputVariableContextsById;
+    globalNodeContextsByNodeId?: NodeContextsByNodeId;
```
noanflaherty, line 32: Currently, we're only ever tracking _all_ node contexts and input variable contexts, across the whole stack of a workflow. This is causing us to include extra inputs where we shouldn't be.

The solution is tot rack both local and global values.

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -109,6 +118,29 @@ export class WorkflowContext {
     this.workflowRawEdges = workflowRawEdges;
   }
 
+  /* Create a new workflow context for a nested workflow from its parent */
```
noanflaherty, line 121: Introduces a helper for creating a child workflow context and moves/adapts a similar helper here from below.

NgoHarrison (approved):

peicodes:
👌

PR #261, opened 2024-12-01
[SC-5666][Part 2] Fix Codegen for Map Nodes

noanflaherty:
This fixes everything related to Map Node codegen such that we can stop excluding it from tests.

Note that in this PR, I tried exploring changing JSON vellum variable types to use our internal `Json` type instead of `Any`, but got mypy errors that we'd have to `type: ignore`. I may explore this once fern codegen supports generating comments to the right of code, but I've reverted this commit for now.

### DIFF
--- a/ee/codegen/src/generators/node-inputs/node-input-value-pointer-rules/input-variable-pointer.ts
+++ b/ee/codegen/src/generators/node-inputs/node-input-value-pointer-rules/input-variable-pointer.ts
@@ -3,7 +3,6 @@ import { python } from "@fern-api/python-ast";
 import { BaseNodeInputValuePointerRule } from "./base";
 
 import { InputVariablePointer } from "src/types/vellum";
-import { toSnakeCase } from "src/utils/casing";
 
 export class InputVariablePointerRule extends BaseNodeInputValuePointerRule<InputVariablePointer> {
   getAstNode(): python.Reference {
@@ -17,7 +16,7 @@ export class InputVariablePointerRule extends BaseNodeInputValuePointerRule<Inpu
     return python.reference({
       name: "Inputs",
       modulePath: inputVariableContext.modulePath,
-      attribute: [toSnakeCase(inputVariableContext.getInputVariableName())],
+      attribute: [inputVariableContext.getInputVariableName()],
     });
   }
 }
--- a/ee/codegen/src/generators/node-inputs/node-input-value-pointer-rules/node-input-value-pointer-rule.ts
+++ b/ee/codegen/src/generators/node-inputs/node-input-value-pointer-rules/node-input-value-pointer-rule.ts
@@ -21,7 +21,7 @@ export declare namespace NodeInputValuePointerRule {
 
 export class NodeInputValuePointerRule extends AstNode {
   private workflowContext: WorkflowContext;
-  private astNode: AstNode;
+  private astNode: BaseNodeInputValuePointerRule<NodeInputValuePointerRuleType>;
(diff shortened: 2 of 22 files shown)


ee/codegen/src/generators/node-inputs/node-input-value-pointer-rules/input-variable-pointer.ts
```diff
@@ -17,7 +16,7 @@ export class InputVariablePointerRule extends BaseNodeInputValuePointerRule<Inpu
     return python.reference({
       name: "Inputs",
       modulePath: inputVariableContext.modulePath,
-      attribute: [toSnakeCase(inputVariableContext.getInputVariableName())],
+      attribute: [inputVariableContext.getInputVariableName()],
```
noanflaherty, line 19: `getInputVariableName` already calls `toSnakeCase` within.,

ee/codegen/src/generators/node-inputs/node-input-value-pointer-rules/node-input-value-pointer-rule.ts
```diff
@@ -21,7 +21,7 @@ export declare namespace NodeInputValuePointerRule {
 
 export class NodeInputValuePointerRule extends AstNode {
   private workflowContext: WorkflowContext;
-  private astNode: AstNode;
+  private astNode: BaseNodeInputValuePointerRule<NodeInputValuePointerRuleType>;
```
noanflaherty, line 24: Just makes this type a bit more strict.

ee/codegen_integration/fixtures/simple_map_node/code/nodes/final_output.py
```diff
@@ -1,12 +1,11 @@
...
 
-from vellum import VellumValue
 from vellum.workflows.nodes.displayable import FinalOutputNode
 from vellum.workflows.state import BaseState
 
 from .map_node import MapNode
 
 
-class FinalOutput(FinalOutputNode[BaseState, List[VellumValue]]):
+class FinalOutput(FinalOutputNode[BaseState, Any]):
```
noanflaherty, line 9: This `Any` is expected since we use a JSON vellum variable type.

ee/codegen_integration/fixtures/simple_map_node/code/nodes/map_node/nodes/search_node.py
```diff
@@ -1,11 +1,11 @@
 from vellum import SearchFiltersRequest, SearchRequestOptionsRequest, SearchResultMergingRequest, SearchWeightsRequest
 from vellum.workflows.nodes.displayable import SearchNode as BaseSearchNode
 
-from ....nodes.map_node.inputs import Inputs
+from ..inputs import Inputs
```
noanflaherty, line 4: This now correctly points to the inner workflow

ee/codegen_integration/fixtures/simple_map_node/code/nodes/map_node/nodes/search_node.py
```diff
@@ -1,11 +1,11 @@
...
 
 
 class SearchNode(BaseSearchNode):
-    query = Inputs.items
+    query = Inputs.item
```
noanflaherty, line 8: The fixture was wrong and made for an invalid workflow. The query to the search should be the individual item. Update the ID below to reflect this.

NgoHarrison (approved):
✅

peicodes (commented):
👌

PR #262, opened 2024-12-01
Introduce AWAIT_ATTRIBUTES merge behavior

dvargas92495:
This PR introduces the AWAIT_ATTRIBUTES merge behavior - a merging strategy that waits to ensure that all attributes on a node are resolved before proceeding to execution. This aims to be a better default than our current two options:
- AWAIT_ALL as a default is undesirable in the case of conditionals since only one of the paths will only ever be invoked
- AWAIT_ANY as a default is undesirable in the case of parallel execution, since the two will race and execute before all dependencies are resolved

This PR doesn't yet make AWAIT_ATTRIBUTES the default strategy. I simply introduces and makes a series of changes that resolved other issues that were discovered during implementation including:
- Repeated executions of a node
- Streaming node pipeline executing a downstream node multiple times

There were also some higher level design questions that this implementation invoked, so going to temporarily leave this in draft while we solve those questions

EDIT: We solved and landed on a path forward for all of the AWAIT_ANY related questions. This PR can now focus on just introducing AWAIT_ATTRIBUTES. It doesn't yet pose it as a default - I'm kicking that for after sales release as well

### DIFF
--- a/src/vellum/workflows/descriptors/utils.py
+++ b/src/vellum/workflows/descriptors/utils.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, Optional, Sequence, Set, TypeVar, Union, cast, ove
 
 from pydantic import BaseModel
 
+from vellum.workflows.constants import UNDEF
 from vellum.workflows.descriptors.base import BaseDescriptor
 from vellum.workflows.state.base import BaseState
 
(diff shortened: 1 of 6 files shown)


siddseethepalli (commented):
nice

PR #266, opened 2024-12-01
[Codegen] Use Fern's New File Imports & Comments

noanflaherty:
The end goal is to get the display of inline subworkflow and map nodes working correctly. Doing so will require code comments and explicit "*" imports.

As a stepping stone to getting there, this PR simply refactors the code that's here currently to use the newly exposed fern functionality. This will be extended in an upcoming PR to actually perform the fix that's needed.

What's nice is that with this, we now use native fern ast functionality everywhere, rather than having to eject and use the `CodeBlock` AST Node.

### DIFF
--- a/ee/codegen/src/generators/base-persisted-file.ts
+++ b/ee/codegen/src/generators/base-persisted-file.ts
@@ -2,6 +2,8 @@ import { mkdir, writeFile } from "fs/promises";
 import path, { join } from "path";
 
 import { python } from "@fern-api/python-ast";
+import { Comment } from "@fern-api/python-ast/Comment";
+import { StarImport } from "@fern-api/python-ast/StarImport";
 import { AstNode } from "@fern-api/python-ast/core/AstNode";
 import { Writer } from "@fern-api/python-ast/core/Writer";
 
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/base-persisted-file.ts
```diff
@@ -29,6 +31,14 @@ export abstract class BasePersistedFile extends AstNode {
 
   protected abstract getFileStatements(): AstNode[] | undefined;
 
+  protected getFileImports(): StarImport[] | undefined {
+    return undefined;
+  }
+
+  protected getComments(): Comment[] | undefined {
+    return undefined;
+  }
```
noanflaherty, line 40: These can be overridden by child classes.

ee/codegen/src/project.ts
```diff
@@ -202,19 +204,23 @@ ${errors.slice(0, 3).map((err) => {
   }
 
   private generateRootInitFile(): InitFile {
```
noanflaherty, line 206: Here's where we do some refactoring to get rid of `CodeBlock`.

dvargas92495 (commented):
✅

PR #267, opened 2024-12-02
[Codegen] Fix Remaining Inline Subworkflow Node Codegen

noanflaherty:
This takes the work started in https://github.com/vellum-ai/vellum-python-sdks/pull/266 across the finish line.

Here we now include the extra comments and imports needed for Inline Subworkflow Node codegen to behave as expected.

### DIFF
--- a/ee/codegen/src/project.ts
+++ b/ee/codegen/src/project.ts
@@ -244,6 +244,17 @@ ${errors.slice(0, 3).map((err) => {
     const parentNode = this.workflowContext.parentNode;
     if (parentNode) {
       statements.push(...parentNode.generateNodeDisplayClasses());
+      comments.push(python.comment({ docs: "flake8: noqa: F401, F403" }));
+      imports.push(
+        python.starImport({
+          modulePath: [...parentNode.getNodeDisplayModulePath(), "nodes"],
+        })
+      );
+      imports.push(
+        python.starImport({
+          modulePath: [...parentNode.getNodeDisplayModulePath(), "workflow"],
+        })
+      );
     } else {
       comments.push(python.comment({ docs: "flake8: noqa: F401, F403" }));
       imports.push(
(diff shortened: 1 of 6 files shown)


ee/codegen/src/project.ts
```diff
@@ -244,6 +244,17 @@ ${errors.slice(0, 3).map((err) => {
...
+      imports.push(
+        python.starImport({
+          modulePath: [...parentNode.getNodeDisplayModulePath(), "nodes"],
+        })
+      );
+      imports.push(
+        python.starImport({
+          modulePath: [...parentNode.getNodeDisplayModulePath(), "workflow"],
+        })
+      );
```
noanflaherty, line 257: These two lines ensure we generate:
```python
from .nodes import *
from .workflow import *
```

ee/codegen_integration/fixtures/simple_inline_subworkflow_node/code/display/workflow.py
```diff
@@ -44,7 +44,7 @@ class WorkflowDisplay(VellumWorkflowDisplay[Workflow]):
     }
     output_displays = {
         Workflow.Outputs.final_output: WorkflowOutputVellumDisplayOverrides(
-            id=UUID("6ab3665f-881d-488b-9124-a6da40136c68"),
+            id=UUID("b38e08c7-904d-4f49-b8fb-56e1eff254d6"),
```
noanflaherty, line 47: The id in the fixture was wrong. Updating it generates the correct id here.

ee/codegen_integration/fixtures/simple_inline_subworkflow_node/display_data/simple_inline_subworkflow_node.json
```diff
@@ -378,7 +378,7 @@
   ],
   "output_variables": [
     {
-      "id": "6ab3665f-881d-488b-9124-a6da40136c68",
+      "id": "b38e08c7-904d-4f49-b8fb-56e1eff254d6",
```
noanflaherty, line 381: We were pointing to the output id of the terminal node within the subworkflow rather than the outer terminal node of the parent workflow. This was an easy mistake since this part of the fixture is hand-crafted.

PR #273, opened 2024-12-02
[SC-5696] Handle Adorned Node Definitions

noanflaherty:
Handles case where a Node Definition contains info about adornments.

### DIFF
--- a/ee/codegen/src/utils/paths.ts
+++ b/ee/codegen/src/utils/paths.ts
@@ -48,12 +48,27 @@ export function getGeneratedNodeModuleInfo({
   nodeDefinition: WorkflowNodeDefinition | undefined;
   nodeLabel: string;
 }): { moduleName: string; nodeClassName: string; modulePath: string[] } {
-  const moduleName =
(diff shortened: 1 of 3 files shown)


ee/codegen/src/utils/paths.ts
```diff
@@ -37,11 +39,42 @@ export function getGeneratedNodesModulePath(
...
+}): { moduleName: string; nodeClassName: string; modulePath: string[] } {
+  const modulePathLeaf =
+    nodeDefinition?.module?.[nodeDefinition.module.length - 1];
+
+  let moduleName: string;
+  let nodeClassName: string;
+
+  // In the case of adorned Nodes, we need to traverse the Adornment Node's definition to get
+  // info about the inner Node that it adorns.
+  if (modulePathLeaf && modulePathLeaf === "<adornment>") {
```
dvargasfuertes, line 61: [oos] In the future, we'll want to handle multiple adornments as well. can defer and ticket for now

dvargas92495 (approved):

PR #275, opened 2024-12-02
Harrison/sc 5663/fix guardrail codegen

NgoHarrison:


ee/codegen/src/context/node-context/guardrail-node.ts
```diff
@@ -1,13 +1,33 @@
...
+export declare namespace GuardrailNodeContext {
+  interface Args extends BaseNodeContext.Args<GuardrailNodeType> {
+    metricDefinitionsHistoryItem: MetricDefinitionHistoryItem;
+  }
+}
+
 export class GuardrailNodeContext extends BaseNodeContext<GuardrailNodeType> {
-  // TODO: Figure out a way to correctly get the node outputs from metric definitions
-  // https://app.shortcut.com/vellum/story/5348/figure-out-correct-way-to-handle-guardrailnode-output
+  public metricDefinitionsHistoryItem: MetricDefinitionHistoryItem;
```
noanflaherty: nit: seems like this could be private and readonly

ee/codegen/src/generators/nodes/guardrail-node.ts
```diff
@@ -97,6 +98,51 @@ export class GuardrailNode extends BaseSingleFileNode<
     return statements;
   }
 
+  protected getOutputDisplay(): python.Field | undefined {
+    const record = this.nodeContext.getNodeOutputNamesById();
+    Promise.resolve(record).then((record) => {
```
noanflaherty: nit: Now that everything is synchronous, we should be able to get rid of the promise resolve here.

ee/codegen/src/generators/nodes/guardrail-node.ts
```diff
@@ -97,6 +98,51 @@ export class GuardrailNode extends BaseSingleFileNode<
...
+      const scoreId = Object.keys(record).find(
+        (key) => record[key] === "score"
+      );
+      if (!scoreId) {
+        throw new Error(
+          "Score id is not found in the guardrail node output displays"
+        );
+      }
```
noanflaherty: Can we simply iterate over all outputs and create an output display for each, rather than specifically looking for just one called `score`? See subworkflow deployment node for an example.

NgoHarrison: Updated

noanflaherty (approved):

kirawoods (commented):
✅

PR #278, opened 2024-12-03
Resolve function block serialization

dvargas92495:
Function blocks needed to be split from the rest of the blocks during codegen.

Instead of handling this from within serialization, this PR simplified serialization to be just `old types(snake_case) -> old types(camelCase)`, and our codegen ast nodes purely operate on `old types(camelCase) -> new generated types`, removing the need for intermediate "new types"


NgoHarrison (approved):

noanflaherty (commented):
Huge! ✅

kirawoods (commented):
Nice!

PR #282, opened 2024-12-03
[Codegen][SC-5648] Stabilize Templating Node Output IDs

noanflaherty:
We weren't using output displays in Templating Nodes to ensure a stable ID reference. This is because `getOutputDisplay` wasn't made abstract and so children weren't forced to implement it, just optionally override it, and so it's easily missed.

This PR makes it abstract and then implements it for all children classes.


kirawoods (approved):
✅

PR #283, opened 2024-12-03
[Codegen][SC-5623] Use Workflow Definition to Power Class Name

noanflaherty:
If a Workflow has a definition, we should use its name as the generated Workflow's class name. Only if there isn't one should we fall back to the name "Workflow"


dvargas92495 (approved):
hell yea

PR #284, opened 2024-12-03
Remove legacy module option from pull

dvargas92495:
This option was originally added to support codegen differences between when this repo was in the prototype repo and the cutover to the public sdk. Now that we are fully in sdk land, we could remove the arg


noanflaherty (commented):
✅

PR #287, opened 2024-12-03
Uses gcp gha provider to authenticate action

dvargas92495:
Following up on Sidd's advice here: https://vellum-ai.slack.com/archives/C06P13ABK0W/p1733242314423369?thread_ts=1733242124.315989&cid=C06P13ABK0W

The hope is that once this is in, codegen versions could start publishing automatically

### DIFF
--- a/.github/workflows/ci-codegen.yaml
+++ b/.github/workflows/ci-codegen.yaml
@@ -75,16 +75,23 @@ jobs:
         run: npm run build
         working-directory: ee/codegen
 
+      - id: auth
+        uses: google-github-actions/auth@v1
+        with:
+          token_format: "access_token"
+          create_credentials_file: true
+          workload_identity_provider: "projects/585775334980/locations/global/workloadIdentityPools/github-pool/providers/github-actions-provider"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-codegen.yaml
```diff
@@ -75,16 +75,23 @@ jobs:
...
         run: npm run gar-login
         working-directory: ee/codegen
 
       - name: Publish
+        env:
+          GOOGLE_APPLICATION_CREDENTIALS: ${{ steps.auth.outputs.credentials_file_path }}
         run: npm publish
         working-directory: ee/codegen
-
-      - name: Notify on failure
```
dvargasfuertes, line 86: I was hoping this step would still pass the action, but it didn't. Since it's essentially useless and we'll be close to a fix, removing

siddseethepalli:
Did it work without this part? https://github.com/vellum-ai/vellum/pull/5833/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR23-R25

dvargas92495:
turns out codegen tests are failing on main, will slip this suggestion in just in case in the meantime

PR #289, opened 2024-12-03
Clarifying AWAIT ANY behavior

dvargas92495:
This PR introduces a simple PoC for AWAIT ANY nodes. it raises an important question: how many times should the `AwaitAnyNode` execute?

Workflows today:
We jump through a lot of hoops to attempt to make the answer 1. There is a multi conditional node in a loop test case where this breaks, which I can add as a second commit. It also requires the maintenance of execution state which we preview here: https://github.com/vellum-ai/vellum-python-sdks/pull/262/files#diff-525e0b968ebb2c6526f327466023e8c160b51b9adf37075237b5eaf6dd80353eR76-R77

SDK today:
The answer in the SDK currently is 2, sort of by accident. If we want to ensure the answer is 2, it will allow us to remove some of this execution state tracking. However, it _does_ mean that two nodes feeding into a node will invoke twice by default, which may not align with user expectation. It also means downstream references to this node's output (eg `AwaitAnyNode`), will operate under the most recent execution, as they always do (meaning in our test, `final_value == 2`)

EDIT: Current decisions:
- Implement our await any logic to be as close to workflows today as possible (this PR)
- Leave the test that would force the issue on automatic state forking pytest.mark.skip
- Defer solving that test until after sales release

### DIFF
--- a/tests/workflows/await_any_conditional_loops/workflow.py
+++ b/tests/workflows/await_any_conditional_loops/workflow.py
@@ -1,0 +1,88 @@
+import time
+
+from vellum.workflows.nodes.bases.base import BaseNode
+from vellum.workflows.ports.port import Port
+from vellum.workflows.state.base import BaseState
+from vellum.workflows.types.core import MergeBehavior
+from vellum.workflows.workflows.base import BaseWorkflow
+
+
+class StartNode(BaseNode):
+    pass
+
+
+class TopNode(BaseNode):
+    class Ports(BaseNode.Ports):
+        success = Port.on_if(StartNode.Execution.count.equals(1))
+        failure = Port.on_else()
+
+
+class TopDeadEndNode(BaseNode):
+    pass
+
+
+class BottomNode(BaseNode):
+    class Ports(BaseNode.Ports):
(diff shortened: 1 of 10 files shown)


dvargas92495:
Added a second commit showcasing the edge case referenced in the PR description for "Workflows Today"
- Workflows Today: Treats the second iteration of Bottom Node as if it was the first iteration
- Accidental SDK Today: Would not track iterations, would invoke as normal

tests/workflows/await_any_conditional_loops/workflow.py
```diff
@@ -0,0 +1,88 @@
...
+
+class AwaitAnyWithConditionalLoopsWorkflow(BaseWorkflow):
+    r"""
+    This is a trickier example of the AWAIT_ANY merge behavior.
+
+    Here's the Ascii diagram:
+
+                     TopDeadEndNode
+                     /
+              TopNode                       StartNode
```
dvargasfuertes, line 63: algorithm for auto-detecting this loop:
- Traverse the graph from StartNode
- When you detect a Loop who's target is in set of discovered nodes, automatic fork
    - "basically DFS"

PR #293, opened 2024-12-04
Add support for ImageVellumValue

kirawoods:


dvargas92495 (approved):
looks good to me!

ee/codegen/src/generators/vellum-variable-value.ts
```diff
@@ -178,6 +179,43 @@ class ErrorVellumValue extends AstNode {
...
+
+  public write(writer: Writer): void {
+    const arguments_ = [
+      python.methodArgument({
+        name: "src",
+        value: python.TypeInstantiation.str(this.value.src),
+      }),
+    ];
+
+    if (this.value.metadata !== undefined) {
```
noanflaherty: This helper from lodash is useful for checking if something is either null or undefined

```suggestion
    if (!isNil(this.value.metadata)) {
```

noanflaherty (approved):
Looks great!

PR #294, opened 2024-12-04
Support emitting of subworkflow events

dvargas92495:
This PR makes it such that events emitted by inline subworkflows (in this case, _specifically_ try nodes) can be emitted back through the `stream` API. This is something we _don't_ support in workflows today, because A) it's noisy and B) we are able to initialize the inline workflow emitter as part of the node.

We don't have access to B in Vembda, so we'll need to emit them out of Vembda and back to vellum in order to be A) stored in clickhouse and B) sent to the UI for display purposes. We solve the noisy problem in a separate PR where we introduced event filters - Vembda will need to use the `all_workflow_event_filter ` filter to allow them all to pass through

### DIFF
--- a/src/vellum/workflows/nodes/core/try_node/node.py
+++ b/src/vellum/workflows/nodes/core/try_node/node.py
@@ -8,7 +8,9 @@ from vellum.workflows.nodes.bases import BaseNode
 from vellum.workflows.nodes.bases.base import BaseNodeMeta
 from vellum.workflows.nodes.utils import ADORNMENT_MODULE_NAME
 from vellum.workflows.outputs.base import BaseOutput, BaseOutputs
+from vellum.workflows.state.context import WorkflowContext
 from vellum.workflows.types.generics import StateType
+from vellum.workflows.workflows.event_filters import all_workflow_event_filter
 
 if TYPE_CHECKING:
     from vellum.workflows import BaseWorkflow
@@ -48,7 +50,7 @@ class _TryNodeMeta(BaseNodeMeta):
         try:
             return super().__getattribute__(name)
         except AttributeError:
-            if name != "__wrapped_node__" and hasattr(cls, "__wrapped_node__"):
+            if name != "__wrapped_node__" and issubclass(cls, TryNode):
                 return getattr(cls.__wrapped_node__, name)
             raise
 
@@ -61,6 +63,7 @@ class TryNode(BaseNode[StateType], Generic[StateType], metaclass=_TryNodeMeta):
     subworkflow: Type["BaseWorkflow"] - The Subworkflow to execute
     """
 
+    __wrapped_node__: Optional[Type["BaseNode"]] = None
     on_error_code: Optional[VellumErrorCode] = None
     subworkflow: Type["BaseWorkflow"]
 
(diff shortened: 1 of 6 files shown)


src/vellum/workflows/nodes/core/try_node/node.py
```diff
@@ -70,15 +73,20 @@ class Outputs(BaseNode.Outputs):
     def run(self) -> Iterator[BaseOutput]:
         subworkflow = self.subworkflow(
             parent_state=self.state,
-            context=self._context,
+            context=WorkflowContext(
+                _vellum_client=self._context._vellum_client,
+            ),
```
dvargasfuertes, line 78: cc: @marlenc3 we likely will want to combine this `WorkflowContext` class with the `ExecutionContext` that we're using to pass `ParentContext` around.

This PR also left inline subworkflows and map nodes [oos] - could use your help in enabling it for those two use cases once this is in and parent context is online 🙏

src/vellum/workflows/nodes/core/try_node/node.py
```diff
@@ -70,15 +73,20 @@ class Outputs(BaseNode.Outputs):
...
+            event_filter=all_workflow_event_filter,
         )
-        subworkflow_stream = subworkflow.stream()
 
         outputs: Optional[BaseOutputs] = None
         exception: Optional[NodeException] = None
         fulfilled_output_names: Set[str] = set()
 
         for event in subworkflow_stream:
+            self._context._emit_subworkflow_event(event)
```
dvargasfuertes, line 89: This is the key piece that emits events from this subworkflow back up to the parent.

Could be controversial, since it means that nodes always has access to the main workflow event emitter and could emit whatever it wants, but hopefully the underscore'd private attributes are enough to say "don't do that"

tests/workflows/stream_try_node_annotation/tests/test_workflow.py
```diff
@@ -16,57 +16,103 @@ def test_workflow_stream__happy_path():
...
+
+    # workflow initiated events
+    workflow_initiated_events = [e for e in events if e.name == "workflow.execution.initiated"]
+    assert workflow_initiated_events[0].workflow_definition == StreamingTryExample
+    assert workflow_initiated_events[0].parent is None
+    assert workflow_initiated_events[1].workflow_definition == InnerWorkflow
+    # TODO: Try Node Parent Context - https://app.shortcut.com/vellum/story/5601
+    # assert workflow_initiated_events[1].parent is not None
+    # assert workflow_initiated_events[1].parent.type == "WORKFLOW_NODE"
+    # assert workflow_initiated_events[1].parent.node_definition == InnerNode
```
dvargasfuertes, line 35: cc: @marlenc3

tests/workflows/stream_try_node_annotation/tests/test_workflow.py
```diff
@@ -16,57 +16,103 @@ def test_workflow_stream__happy_path():
...
+    assert node_initiated_events[0].parent is not None
+    assert node_initiated_events[0].parent.type == "WORKFLOW"
+    assert node_initiated_events[0].parent.workflow_definition == StreamingTryExample
+    assert node_initiated_events[1].node_definition == WrappedNode
+    assert node_initiated_events[1].parent is not None
+    assert node_initiated_events[1].parent.type == "WORKFLOW"
+    assert node_initiated_events[1].parent.workflow_definition == InnerWorkflow
+    assert len(node_initiated_events) == 2
+
+    # inner node streaming events
```
dvargasfuertes, line 54: We can no longer guarantee order between all of these streaming events, so these filters make the assertions more deterministic

marlenc3 (approved):

PR #295, opened 2024-12-04
Introduce the event filter param into workflow streaming to replace

dvargas92495:
Workflows never needed to return subworkflow events as part of the API before because we had new emitters initialized within the node to capture events.

Now that everything runs in Vembda, we need to support the need for emitting events from inner subworkflows as part of the API. Before doing so (we started exploring it [here](https://github.com/vellum-ai/vellum-python-sdks/pull/294/files#diff-bf147bd0bb4a35ee966635944c240b29aeac1162f4f1e17c9d663b3ebda65559R34-R39)), we need to make our event filter more expressive so that our default behaviors still don't emit all of those events.

This PR focuses on replacing our param for listening to events from `event_types` to `event_filter`, so that we can say by default, we only return root workflow events

### DIFF
--- a/src/vellum/workflows/events/types.py
+++ b/src/vellum/workflows/events/types.py
@@ -15,11 +15,6 @@ if TYPE_CHECKING:
     from vellum.workflows.workflows.base import BaseWorkflow
 
 
-class WorkflowEventType(Enum):
-    NODE = "NODE"
-    WORKFLOW = "WORKFLOW"
-
-
 def default_datetime_factory() -> datetime:
     """
     Makes it possible to mock the datetime factory for testing.
@@ -49,7 +44,7 @@ def default_serializer(obj: Any) -> Any:
 
 class BaseParentContext(UniversalBaseModel):
     span_id: UUID
-    parent: Optional['ParentContext'] = None
+    parent: Optional["ParentContext"] = None
 
 
 class BaseDeploymentParentContext(BaseParentContext):
@@ -73,7 +68,7 @@ class PromptDeploymentParentContext(BaseDeploymentParentContext):
 
 class NodeParentContext(BaseParentContext):
     type: Literal["WORKFLOW_NODE"] = "WORKFLOW_NODE"
-    node_definition: Type['BaseNode']
+    node_definition: Type["BaseNode"]
 
     @field_serializer("node_definition")
     def serialize_node_definition(self, definition: Type, _info: Any) -> Dict[str, Any]:
@@ -82,7 +77,7 @@ class NodeParentContext(BaseParentContext):
 
 class WorkflowParentContext(BaseParentContext):
     type: Literal["WORKFLOW"] = "WORKFLOW"
-    workflow_definition: Type['BaseWorkflow']
+    workflow_definition: Type["BaseWorkflow"]
 
     @field_serializer("workflow_definition")
     def serialize_workflow_definition(self, definition: Type, _info: Any) -> Dict[str, Any]:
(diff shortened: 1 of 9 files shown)


src/vellum/workflows/workflows/base.py
```diff
@@ -255,9 +256,41 @@ def stream(
...
+            or event.name == "workflow.execution.fulfilled"
+            or event.name == "workflow.execution.rejected"
+            or event.name == "workflow.execution.paused"
+            or event.name == "workflow.execution.streaming"
+        ):
+            return event.workflow_definition == self.__class__
+
+        return False
+
+    def ROOT_EVENT_FILTER(self, event: WorkflowEvent) -> bool:
```
dvargasfuertes: the naming is weird, but I wanted to denote "public, easy to use filters for common use cases". I chose constant naming (all caps) for this but am open to alternatives

src/vellum/workflows/events/types.py
```diff
@@ -15,11 +15,6 @@
     from vellum.workflows.workflows.base import BaseWorkflow
 
 
-class WorkflowEventType(Enum):
```
dvargasfuertes, line 18: This filter allows us to remove this enum which we were anticipating for some time

src/vellum/workflows/events/types.py
```diff
@@ -49,7 +44,7 @@ def default_serializer(obj: Any) -> Any:
 
 class BaseParentContext(UniversalBaseModel):
     span_id: UUID
-    parent: Optional['ParentContext'] = None
+    parent: Optional["ParentContext"] = None
```
noanflaherty, line 47: [oos] looks like we need some linting in CI 😄

src/vellum/workflows/workflows/base.py
```diff
@@ -255,9 +255,41 @@ def stream(
             external_inputs=external_inputs,
             cancel_signal=cancel_signal,
         ).stream():
-            if WorkflowEventType(event.name.split(".")[0].upper()) in event_types:
+            if should_yield(event):
                 yield event
 
+    def DEFAULT_EVENT_FILTER(self, event: WorkflowEvent) -> bool:
```
noanflaherty: nit: We allow for a user pass in a callable of their own. This callable of theirs likely won't have access to `self` right? And will be more akin to a static method? If so, then it'd be nice for our default callable to be more analogous to that which a user might pass. i.e. not live on this class and just be static helper methods.

In this world especially (but also I think in general) I'd prefer to name these:
`def workflow_event_filter` (the fact that it's declared above in `or` implies that it's the default imo, so would rather the name be more indicative of what it actually does)

`def root_workflow_event_filter`

And then also incldue docstrings for each.

src/vellum/workflows/workflows/base.py
```diff
@@ -239,14 +239,14 @@ def run(
...
     def stream(
         self,
-        event_types: Optional[Set[WorkflowEventType]] = None,
+        event_filter: Optional[Callable[[WorkflowEvent], bool]] = None,
         inputs: Optional[WorkflowInputsType] = None,
         state: Optional[StateType] = None,
         entrypoint_nodes: Optional[RunFromNodeArg] = None,
         external_inputs: Optional[ExternalInputsArg] = None,
         cancel_signal: Optional[ThreadingEvent] = None,
     ) -> WorkflowEventStream:
```
noanflaherty: [oos] We should probably add a docstring for the public methods on this class.

noanflaherty (approved):

src/vellum/workflows/events/types.py
dvargasfuertes: that's the black/flake task I had on my list from yesterday haha

src/vellum/workflows/workflows/base.py
dvargasfuertes: gonna do this on a separate pass to request your review on copy

siddseethepalli (commented):
Nice!

PR #296, opened 2024-12-04
Setting up more parent context pass throughs 

marlenc3:
setting up pass through - new union type for event, should let us not have to worry about resetting up 'class' type 

goal for PR - 
be able to allow for vembda service to pass objects over to the sdk easily
allow for serialization to json and back to objects

once vembda is ready, we should be passing in previous parent context into our vembda run


src/vellum/workflows/events/types.py
```diff
@@ -73,20 +74,30 @@ class PromptDeploymentParentContext(BaseDeploymentParentContext):
...
 
 class WorkflowParentContext(BaseParentContext):
     type: Literal["WORKFLOW"] = "WORKFLOW"
-    workflow_definition: Type['BaseWorkflow']
+    workflow_definition: Union[Type["BaseWorkflow"], CodeResourceDefinition]
 
     @field_serializer("workflow_definition")
-    def serialize_workflow_definition(self, definition: Type, _info: Any) -> Dict[str, Any]:
-        return serialize_type_encoder(definition)
+    def serialize_workflow_definition(
```
marlenc3: should allow for both __class__ and the definition that is passed through to be handled correctly

src/vellum/workflows/events/utils.py
```diff
@@ -1,5 +1,62 @@
...
+    }
+    current_type_value = current_context_json.get("type", "DEFAULT")
+    current_type = type_dict.get(current_type_value, None)
+    if current_type:
+        temp = current_type(**current_context_json)
+        return temp
+    else:
+        return None
+
+
```
marlenc3: helper util to convert json to parent_context classes (ParentContext serializers does *not* like building)

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -99,12 +100,12 @@ def _localns(cls) -> Dict[str, Any]:
     def __getattribute__(cls, name: str) -> Any:
         attribute = super().__getattribute__(name)
         if (
-            name.startswith("_")
-            or inspect.isfunction(attribute)
-            or inspect.ismethod(attribute)
-            or is_nested_class(attribute, cls)
-            or isinstance(attribute, (property, cached_property))
-            or not issubclass(cls, BaseNode)
+                name.startswith("_")
```
marlenc3: a lot of this is formatting updates, but can revert/update my settings

src/vellum/workflows/events/types.py
```diff
@@ -68,20 +69,30 @@ class PromptDeploymentParentContext(BaseDeploymentParentContext):
 
 class NodeParentContext(BaseParentContext):
     type: Literal["WORKFLOW_NODE"] = "WORKFLOW_NODE"
-    node_definition: Type["BaseNode"]
+    node_definition: Union[Type["BaseNode"], CodeResourceDefinition]
```
dvargasfuertes: Oh boy. I _think_ I know why we need to do this but... would love to sync on this

dvargasfuertes: actually I don't know if we could avoid this...

src/vellum/workflows/events/types.py
```diff
@@ -9,6 +9,7 @@
 from vellum.core.pydantic_utilities import UniversalBaseModel
 from vellum.workflows.state.encoder import DefaultStateEncoder
 from vellum.workflows.types.utils import datetime_now
+from vellum_ee.workflows.display.vellum import CodeResourceDefinition
```
dvargasfuertes: [blocking] move into here

src/vellum/workflows/events/types.py
dvargasfuertes: discussed offline - yea we need `CodeResourceDefinition`. To make our lives easier, let's just commit to the latter

src/vellum/workflows/events/utils.py
```diff
@@ -1,5 +1,60 @@
...
+    NodeParentContext,
+    ParentContext,
+    PromptDeploymentParentContext,
+    WorkflowDeploymentParentContext,
+    WorkflowParentContext,
+)
 from vellum.workflows.events.workflow import WorkflowEvent
 
 
 def is_terminal_event(event: WorkflowEvent) -> bool:
```
dvargasfuertes: this guy is getting moved on rebase

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -273,7 +275,10 @@ class Execution(metaclass=_BaseNodeExecutionMeta):
         node_class: Type["BaseNode"]
         count: int
 
-    def __init__(self, *, state: Optional[StateType] = None, context: Optional[WorkflowContext] = None):
+    def __init__(
+            self, *, state: Optional[StateType] = None, context: Optional[WorkflowContext] = None,
+            parent_context: Optional[ParentContext] = None
```
dvargasfuertes: combine the two contexts if possible

src/vellum/workflows/events/types.py
```diff
@@ -42,9 +38,21 @@ def default_serializer(obj: Any) -> Any:
...
 
 
+class CodeResourceDefinition(UniversalBaseModel):
+    name: str
+    module: List[str]
+
+
+VellumCodeResourceDefinition = Annotated[
+    CodeResourceDefinition,
+    BeforeValidator(lambda d: (d if type(d) is dict else serialize_type_encoder(d))),
```
marlenc3, line 52: sets up __class__ to be encoded before trying to cast to CRD type

src/vellum/workflows/events/types.py
```diff
@@ -57,39 +65,41 @@ class BaseDeploymentParentContext(BaseParentContext):
...
-    type: Literal["WORKFLOW"] = "WORKFLOW"
-    workflow_definition: Type["BaseWorkflow"]
-
-    @field_serializer("workflow_definition")
-    def serialize_workflow_definition(self, definition: Type, _info: Any) -> Dict[str, Any]:
-        return serialize_type_encoder(definition)
+    type: Literal["WORKFLOW"]
+    workflow_definition: VellumCodeResourceDefinition
+
+
```
marlenc3, line 90: pydantic field magic ✨

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -31,9 +32,15 @@ def is_nested_class(nested: Any, parent: Type) -> bool:
         inspect.isclass(nested)
         # If a class is defined within a function, we don't consider it nested in the class defining that function
         # The example of this is a Subworkflow defined within TryNode.wrap()
-        and (len(nested.__qualname__.split(".")) < 2 or nested.__qualname__.split(".")[-2] != "<locals>")
```
marlenc3, line 34: only formatting changes 👍

src/vellum/workflows/state/context.py
```diff
@@ -3,15 +3,21 @@
...
 if TYPE_CHECKING:
     from vellum.workflows.events.workflow import WorkflowEvent
 
 
 class WorkflowContext:
-    def __init__(self, _vellum_client: Optional[Vellum] = None):
+    def __init__(
+        self,
+        _vellum_client: Optional[Vellum] = None,
+        _parent_context: Optional["ParentContext"] = None,
```
marlenc3: including parent context as WorkflowContext passthrough

tests/workflows/stream_try_node_annotation/tests/test_workflow.py
```diff
@@ -38,81 +41,140 @@ def test_workflow_stream__happy_path():
...
+    ].parent.workflow_definition == CodeResourceDefinition(
+        **serialize_type_encoder(StreamingTryExample)
+    )
     assert node_initiated_events[1].node_definition == WrappedNode
     assert node_initiated_events[1].parent is not None
     assert node_initiated_events[1].parent.type == "WORKFLOW"
-    assert node_initiated_events[1].parent.workflow_definition == InnerWorkflow
+    assert node_initiated_events[
+        1
+    ].parent.workflow_definition == CodeResourceDefinition(
```
marlenc3: updated checks

src/vellum/workflows/events/types.py
```diff
@@ -57,39 +65,41 @@ class BaseDeploymentParentContext(BaseParentContext):
...
+        WorkflowParentContext,
+        NodeParentContext,
+        WorkflowDeploymentParentContext,
+        PromptDeploymentParentContext,
+    ],
+    Field(discriminator="type"),
+]
 
+# Create the final type
+ParentContext = Optional[ParentContextUnion]
```
marlenc3: catch here is that parentcontext isn't quite a "type" now but a Field. but should be able to be added to models now and have type discrimination

marlenc3: https://docs.pydantic.dev/latest/concepts/unions/#nested-discriminated-unions

`
Cat = Annotated[Union[BlackCat, WhiteCat], Field(discriminator='color')]
`

src/vellum/workflows/events/types.py
dvargasfuertes: [q] Do we need this bc the dict being fed into pydantic could contain a `CodeResourceDefinition` instead of a `dict`? when does this happen?

src/vellum/workflows/events/types.py
```diff
@@ -57,39 +65,41 @@ class BaseDeploymentParentContext(BaseParentContext):
 
 
 class WorkflowDeploymentParentContext(BaseDeploymentParentContext):
-    type: Literal["WORKFLOW_RELEASE_TAG"] = "WORKFLOW_RELEASE_TAG"
```
dvargasfuertes: [blocking] why do we need to remove these defaults? would like to avoid passing the burden onto the consumer of these models

src/vellum/workflows/events/types.py
```diff
@@ -57,39 +65,41 @@ class BaseDeploymentParentContext(BaseParentContext):
...
+
+# Define the discriminated union
+ParentContextUnion = Annotated[
+    Union[
+        WorkflowParentContext,
+        NodeParentContext,
+        WorkflowDeploymentParentContext,
+        PromptDeploymentParentContext,
+    ],
+    Field(discriminator="type"),
```
dvargasfuertes, line 99: this looks cool

src/vellum/workflows/events/types.py
dvargasfuertes: [nit] I would remove this here and rename the above back to `ParentContext`, and use the `Optional` in the consumers.

The link you commented here is about _multiple_ discriminators, which this data model does not yet exhibit

src/vellum/workflows/state/context.py
dvargasfuertes: [nit] don't need the `"quotes"` here if the import above is normal and outside of TYPE_CHECKING

tests/workflows/basic_conditional_branch/tests/test_workflow.py
```diff
@@ -54,12 +54,14 @@ def test_stream_workflow__verify_invoked_ports():
     # WHEN we stream the Workflow
     stream = workflow.stream(
         inputs=Inputs(value=True),
-        event_filter=root_workflow_event_filter,
+        event_filter=all_workflow_event_filter,
```
dvargasfuertes: [blocking] why did we change these tests? would like to keep all of these to the most restrictive filter possible to still communicate the point, which in this case is still the root filter

tests/workflows/basic_parent_context/tests/test_basic_parent_context_workflow_runner.py
```diff
@@ -13,7 +17,7 @@ def test_run_workflow__happy_path():
 
 def test_stream_workflow__happy_path():
     workflow = TrivialWorkflow()
-    events = list(workflow.stream(event_filter=root_workflow_event_filter))
+    events = list(workflow.stream(event_filter=all_workflow_event_filter))
```
dvargasfuertes: [nit] even for these examples, we get no benefit from going from root -> all since the only difference is that the latter allows subworkflow events to pass on through

tests/workflows/stream_try_node_annotation/tests/test_workflow.py
```diff
@@ -38,81 +41,140 @@ def test_workflow_stream__happy_path():
...
+        ],
     }
     assert node_initiated_events[0].parent is not None
     assert node_initiated_events[0].parent.type == "WORKFLOW"
-    assert node_initiated_events[0].parent.workflow_definition == StreamingTryExample
+    assert node_initiated_events[
+        0
+    ].parent.workflow_definition == CodeResourceDefinition(
+        **serialize_type_encoder(StreamingTryExample)
+    )
```
dvargasfuertes: [nit] I could also see us supporting a static method on `CodeResourceDefinition` to handle this more natively, eg:

```python
CodeResourceDefinition.from(StreamingTryExample)
```

src/vellum/workflows/events/types.py
dvargasfuertes: after seeing that we do `Optional[ParentContext]` below and in other consumers anyway, I'm doubling down on this feedback

dvargas92495 (commented):
I like the direction! mostly nits and one main blocker around the type default initializers

src/vellum/workflows/events/types.py
marlenc3: This happens before validation! We could pass in an argument of both a dictionary hat's being passed in when instantiating via something like NodeParentContext( ** previous_parent_context ) and when creating a new context like NodeParentContext(node_definition=node.__class__ ... ) so we technically could also support someone directly using a "CodeResourceDefinition", but they'd still need to transform that __class__ . This happens before validation tho, so it'll still run through the validator.

src/vellum/workflows/events/types.py
marlenc3: At some step while trying to get this working the defaults were an issue, but fixed, and it should be fine now!

src/vellum/workflows/events/types.py
marlenc3: 🫡  should be up

tests/workflows/basic_conditional_branch/tests/test_workflow.py
marlenc3: these tests were breaking with the recent filter changes 🤔  change it back and update the test though.

tests/workflows/stream_try_node_annotation/tests/test_workflow.py
marlenc3: oh that's a great idea, and it'd probably be able to be used pre-validation

tests/workflows/basic_conditional_branch/tests/test_workflow.py
marlenc3: well, basically the test is testing the node events - but those don't get returned with the root event filter

dvargasfuertes: > well, basically the test is testing the node events - but those don't get returned with the root event filter

I would expect the node events _to_ come back with the root event filter. It's likely not bc of this equality here needs to be ubdated now that `parent.workflow_definition` is not a class: https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/workflows/event_filters.py#L47

dvargas92495 (approved):
We should fix the `root_workflow_event_filter` but otherwise looks great

tests/workflows/basic_conditional_branch/tests/test_workflow.py
marlenc3: oooooo, I'll add it in so I can get these tests passing with the root_event

marlenc3: thanks for catching that! 🔥

tests/workflows/stream_try_node_annotation/tests/test_workflow.py
```diff
@@ -25,94 +26,153 @@ def test_workflow_stream__happy_path():
     WrappedNode = InnerNode.__wrapped_node__
 
     # workflow initiated events
-    workflow_initiated_events = [e for e in events if e.name == "workflow.execution.initiated"]
+    workflow_initiated_events = [
+        e for e in events if e.name == "workflow.execution.initiated"
+    ]
     assert workflow_initiated_events[0].workflow_definition == StreamingTryExample
     assert workflow_initiated_events[0].parent is None
-    assert workflow_initiated_events[1].workflow_definition == InnerWorkflow
```
marlenc3, line 31: currently digging into why these failed

PR #297, opened 2024-12-04
Harrison/sc 5671/update conditional node to handle all variable inputs

NgoHarrison:
This PR addresses conditional node fixes. This came up from the faa_q_and_a_bot workflow which is included in fixture tests now but I can take it out if needed since this was just a bug bash test. Alternatively I could spin up a different PR to represent integration tests between our "simple" fixtures and then fixtures that represent real workflows


ee/codegen/src/generators/nodes/conditional-node.ts
```diff
@@ -24,9 +24,13 @@ export class ConditionalNode extends BaseSingleFileNode<
   protected getNodeClassBodyStatements(): AstNode[] {
     const statements: AstNode[] = [];
 
-    const inputVarIdsByRuleId = new Map<string, string>();
-    const constantValuesByRuleIds = new Map<string, VellumValue>();
-    this.constructMapsByRuleIds(inputVarIdsByRuleId, constantValuesByRuleIds);
+    const inputFieldKeysByRuleId = new Map<string, string>();
+    const valueInputKeysByRuleId = new Map<string, string>();
```
NgoHarrison, line 28: I ended up needing these two maps because the nodeInputsByKey had keys as the Map.key. I wanted an easy way to just access the values so I built a map from (ruleId -> input.key) and then use that key to search up the value instead of doing a reverse lookup which would have gotten kind of messy but can change on this approach. I am 1/5 on changing this for something better

ee/codegen/src/__test__/project.test.ts
```diff
@@ -48,6 +49,9 @@ describe("WorkflowProjectGenerator", () => {
           // TODO: Get Merge Node graph codegen working
           //    https://app.shortcut.com/vellum/story/5588
           // "simple_merge_node",
+
+          // Real customer workflows:
+          //"faa_q_and_a_bot",
```
noanflaherty: [1/5 opinion] I'd say we should probs exclude this fixture from this PR unless we're prepared to uncomment this and include as a test we test again

ee/codegen/src/utils/typing.ts
```diff
@@ -1,7 +1,27 @@
...
+
 export function isDefined<TValue>(value: TValue | undefined): value is TValue {
   return value !== undefined;
 }
 
 export function assertUnreachable(_: never): never {
   throw new Error("Didn't expect to get here");
 }
+
+export function isNilOrEmpty<T>(
```
noanflaherty, line 11: [nit] Some quick unit tests for this would be nice

ee/codegen_integration/fixtures/faa_q_and_a_bot/code/nodes/api_node.py
```diff
@@ -0,0 +1,5 @@
+from vellum.workflows.nodes.displayable import APINode as BaseAPINode
+
+
+class APINode(BaseAPINode):
+    pass
```
noanflaherty: I imagine this will get updated with your recent ApiNode PR

ee/codegen_integration/fixtures/faa_q_and_a_bot/code/workflow.py
```diff
@@ -0,0 +1,48 @@
...
+from .nodes.prompt_node_16 import PromptNode16
+from .nodes.prompt_node_18 import PromptNode18
+from .nodes.prompt_node_19 import PromptNode19
+from .nodes.subworkflow_node import SubworkflowNode
+from .nodes.templating_node import TemplatingNode
+from .nodes.templating_node_15 import TemplatingNode15
+
+
+class Workflow(BaseWorkflow[Inputs, BaseState]):
+    graph = (
```
noanflaherty, line 23: Super cool to see this get generated!

noanflaherty (approved):
🔥

ee/codegen/src/utils/typing.ts
NgoHarrison: I will make a follow up PR for this as a fast follow since the other functions in typing could probably use some unit tests as well

PR #298, opened 2024-12-04
Set up failing test and impl for graph attribute generation for merge nodes

dvargas92495:
This PR introduces the early innings of the new graph generation algorithm, which will be a breadth first search across edges, amending the AST as we traverse.

This PR _only_ strives to pass an initial bank of test cases (4), in order to show that the original ticket can be solved. I intend to continue solving the rest over a series of smaller PRs, each with its own test case, taking after what we support in execution land here: https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/graph/tests/test_graph.py

So as to not regress what we currently have in the interim, I introduced a "feature flag", so that we could build out graph gen across tests, and then cut over once they're all ready and passing.

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -3,8 +3,8 @@
 exports[`Workflow > write > graph > should be correct for a basic merge node case 1`] = `
 "from vellum.workflows import BaseWorkflow
 from .nodes.templating_node import TemplatingNode
-from .nodes.merge_node import MergeNode
 from .nodes.templating_node_2 import TemplatingNode2
+from .nodes.merge_node import MergeNode
 
 
 class TestWorkflow(BaseWorkflow):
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -1,5 +1,62 @@
...
+from .nodes.templating_node_2 import TemplatingNode2
+
+
+class TestWorkflow(BaseWorkflow):
+    graph = TemplatingNode >> TemplatingNode2
+
+    class Outputs(BaseWorkflow.Outputs):
+        pass
+"
+`;
```
dvargasfuertes, line 58: Is there a way to have two parameters of a test case match the same snapshot?

ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -560,19 +560,27 @@ export function promptDeploymentNodeDataFactory(): PromptNode {
 }
 
 export function templatingNodeFactory({
```
dvargasfuertes, line 562: how do we feel about introducing a genericNodeFactory or a baseNodeFactory for building out most of these tests? Had to edit this factory bc two templating nodes with the same id/handle ids/labels would throw errors

ee/codegen/src/generators/workflow.ts
```diff
@@ -636,14 +663,161 @@ export class Workflow {
...
+  private buildGraphAttribute(): AstNode | undefined {
+    if (!this.breadthFirstGraphGeneration) {
+      return this.buildGraphLegacy({
+        portContexts: this.entrypointPortContexts,
+      });
+    }
+
+    // This graph attribute generation is a breadth-first traversal of the WorkflowEdges
+    // starting at the entrypoint port edges. We process each one sequentially and incrementally
+    // build up the graph as we go.
```
dvargasfuertes, line 675: I'm defining all types and helper methods I need internal to this method for now, and happy to split out and reorg when ready. Also happy to reorg now if there are strong opinions

ee/codegen/src/generators/workflow.ts
```diff
@@ -636,14 +663,161 @@ export class Workflow {
...
+    type GraphPortReference = {
+      type: "port_reference";
+      reference: PortContext;
+    };
+    type GraphRightShift = {
+      type: "right_shift";
+      lhs: GraphMutableAst;
+      rhs: GraphMutableAst;
+    };
+    type GraphMutableAst =
```
dvargasfuertes, line 691: fern's ast nodes are immutable, and our algorithm involves an evolving graph attribute

ee/codegen/src/__test__/helpers/node-data-factories.ts
kirawoods: Seems like a good idea to me!

kirawoods (approved):

noanflaherty (approved):
approach seems solid to me

PR #300, opened 2024-12-04
Kirawoods/sc 5661/codegen additional vellum value types search results

kirawoods:


dvargas92495 (approved):

ee/codegen/src/generators/vellum-variable-value.ts
```diff
@@ -255,6 +256,62 @@ class AudioVellumValue extends AstNode {
   }
 }
 
+class SearchResultsVellumValue extends AstNode {
+  private value: SearchResult[];
+
+  public constructor(value: SearchResult[]) {
+    super();
+    this.value = value;
+  }
```
noanflaherty, line 265: A few things here that need to be addressed:

1. Whenever we create a new class that extends `AstNode`, we need to inheritReferences on any sub-nodes it creates within.
2. I don't think it's safe to assume that `document` is instantiated.

PR #301, opened 2024-12-04
Add support for function call

kirawoods:


ee/codegen/src/generators/vellum-variable-value.ts
```diff
@@ -217,6 +218,56 @@ class ImageVellumValue extends AstNode {
...
+    }
+
+    if (!isNil(this.value.state)) {
+      arguments_.push(
+        python.methodArgument({
+          name: "state",
+          value: python.TypeInstantiation.str(this.value.state),
+        })
+      );
+    }
```
dvargasfuertes: [nit] we should remove this - `state` is something we foresee removing from out function call types in the future and should always resolve to FULFILLED in the backend

dvargas92495 (approved):

ee/codegen/src/generators/vellum-variable-value.ts
```diff
@@ -256,6 +257,47 @@ class AudioVellumValue extends AstNode {
   }
 }
 
+class FunctionCallVellumValue extends AstNode {
+  private value: FunctionCall;
+
+  public constructor(value: FunctionCall) {
+    super();
+    this.value = value;
```
noanflaherty, line 284: [blocking] We need to inherit references of the ast nodes that get generated.

kirawoods: 👍

kirawoods: @noanflaherty Does this look right/is this what you meant? https://github.com/vellum-ai/vellum-python-sdks/pull/301/files#diff-669580c6ba7a67bf8abb45c1cf2dfeed737dcda7d72fb6eca8bc8bedded16742R289-R299

dvargas92495 (approved):

ee/codegen/src/generators/vellum-variable-value.ts
NgoHarrison: We already inherit references here for the node that gets constructed at the end in vellum value:
https://github.com/vellum-ai/vellum-python-sdks/pull/301/files#diff-669580c6ba7a67bf8abb45c1cf2dfeed737dcda7d72fb6eca8bc8bedded16742R416

kirawoods: @NgoHarrison Does this mean we don't need to inherit references on the more specific VellumValue types?

NgoHarrison: Not at that level, if you look at the other ones as examples they just construct the "specific" types. In this case we construct the FunctionCall type which is an ast node. In your case match, you should just assign this.astNode to your new class

PR #302, opened 2024-12-04
Add support for Array vellum values

kirawoods:


ee/codegen/src/__test__/__snapshots__/vellum-variable-value.test.ts.snap
```diff
@@ -1,5 +1,10 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`VellumValue > ARRAY > should write a ARRAY value correctly 1`] = `
+"["Hello, World!", 42]
```
dvargasfuertes: [blocking] this should actually be `StringVellumValue(...), NumberVellumValue(...)`

kirawoods: Looking at the snapshots for NumberVellumValue and StringVellumValue, it looks like those types just come back as strings and numbers. I added an AudioVellumValue to confirm we're getting the more complex types back. Is the current behavior of NumberVellumValue and StringVellumValue what we expect or should I update those to return the vellum types there?

dvargasfuertes: yea it is - ARRAY is a weird type in that it's _specifically_ an array of vellum values. This is a requirement that we are likely to try to smooth out in the future, but for now, need to keep

dvargasfuertes: Actually you know what... maybe our chance to smooth this out could be now. I'd be cool with leaving this as is for now and see what breaks down the road

kirawoods: I think it's working okay maybe? It's just that StringVellumValue and NumberVellumValue are set up to pass values through as primitives. It seems to pass the non-primitve VellumValue types through as expected (see snapshot to see how AudioVellumValue is passed through).

dvargasfuertes: It's more about downstream when we start running push and pull in tandem with each other and with existing workflows where I'd be curious to see how this performs. In any case, let's roll with what we have now

As an aside, `VellumAudio` _is_ considered the primitive and a distinct concept from `AudioVellumValue`. Which is to say we are generating the consistent type here with the primitive string and primitive number

dvargas92495 (approved):

ee/codegen/src/__test__/__snapshots__/vellum-variable-value.test.ts.snap
kirawoods: I see, I can definitely circle back to this later if it looks like this is causing issues.

PR #303, opened 2024-12-04
Kirawoods/sc 5661/codegen additional vellum value types search results

kirawoods:


ee/codegen/src/generators/vellum-variable-value.ts
```diff
@@ -283,32 +283,52 @@ class SearchResultsVellumValue extends AstNode {
         }),
         python.methodArgument({
           name: "document",
-          value: python.reference({
-            name: "document", // Assuming document is already instantiated
-            modulePath: VELLUM_CLIENT_MODULE_PATH,
+          value: python.instantiateClass({
```
kirawoods: @noanflaherty Could you take another look here to make sure this looks right for inheritance and document instatiation?

NgoHarrison (approved):
✅

PR #305, opened 2024-12-04
Add support for ChatMessage image content type

kirawoods:


dvargas92495 (approved):

ee/codegen/src/generators/chat-message-content.ts
```diff
@@ -42,6 +43,8 @@ export class ChatMessageContent extends AstNode {
     } else if (this.chatMessageContent.type === "FUNCTION_CALL") {
       this.addReference(this.getFunctionCallChatMessageContentRef());
       this.addReference(this.getFunctionCallChatMessageContentValueRef());
+    } else if (this.chatMessageContent.type === "IMAGE") {
+      this.addReference(this.getImageChatMessageContentRef());
```
NgoHarrison: @noanflaherty This is the old pattern and I believe we prefer using this.inheritReferences now correct?

dvargas92495 (approved):

NgoHarrison (approved):
✅

ee/codegen/src/generators/chat-message-content.ts
noanflaherty: That's correct – Kira and I also chatted about this offline

PR #306, opened 2024-12-04
[SC-5640] Flesh out API node body

NgoHarrison:


ee/codegen/src/generators/nodes/api-node.ts
```diff
@@ -18,9 +25,151 @@ export class ApiNode extends BaseSingleFileNode<ApiNodeType, ApiNodeContext> {
...
+      python.field({
+        name: "method",
+        initializer: this.convertMethodValueToEnum(),
+      })
+    );
+
+    const body = this.nodeInputsByKey.get("body");
+
+    if (body) {
+      const bodyType = this.isTypeStringOrJson(body.nodeInputData.value);
```
noanflaherty: [blocking] The need for this helper kinda spooks me. Especially because that helper only seems to handle constant values, whereas in practice, API Node bodies are often driven by the output of an upstream node.

I think for now we should only support JSON and just assume that that's the case. It's not possible to create string bodies in API Nodes in the UI today.

noanflaherty (approved):

kirawoods (approved):
+1 to what Noa said, but otherwise looks good!

ee/codegen/src/generators/nodes/api-node.ts
NgoHarrison: Ahh makes sense. Just to clarify, the changes should be to just handle all types of inputs and then assume json is the only one we support?

noanflaherty: Yeah that's correct! We could write a test to see what happens if we send an input that's not json and that may change my mind.

PR #307, opened 2024-12-05
Graph Generation - Port to Node

dvargas92495:
Satisfying the case where we have a conditional node pointing to nodes, with tests

### DIFF
--- a/ee/codegen/src/generators/workflow.ts
+++ b/ee/codegen/src/generators/workflow.ts
@@ -722,12 +722,28 @@ export class Workflow {
           reference: targetNode,
         };
       } else if (graphMutableAst.type === "node_reference") {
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/workflow.ts
```diff
@@ -758,6 +774,28 @@ export class Workflow {
             ],
           };
         }
+      } else if (graphMutableAst.type === "right_shift") {
+        if (graphMutableAst.lhs.type == "port_reference") {
```
dvargasfuertes, line 778: I definitely anticipate fancy recursion in a future PR, but am trying to build up to it by being explicit with test cases before doing it all at once

kirawoods (commented):
✅

PR #310, opened 2024-12-05
Graph Gen - Solving case with a set of an edge and a node

dvargas92495:
`graph = {A >> B, C}`

This test case finally pushed us over the edge into recursion. I would review by commit:
- first commit lays out the test case we want to solve
- The second commit is a logicless refactor to format our existing graph manipulation logic into a callable function
- The third commit uses recursion to solve the new test case

### DIFF
--- a/ee/codegen/src/__test__/workflow.test.ts
+++ b/ee/codegen/src/__test__/workflow.test.ts
@@ -559,6 +559,86 @@ describe("Workflow", () => {
           expect(await writer.toStringFormatted()).toMatchSnapshot();
         }
       );
+
+      it("should be correct for set of a branch and a node", async () => {
+        const inputs = codegen.inputs({ workflowContext });
+
+        const templatingNodeData1 = templatingNodeFactory();
+        const templatingNodeContext1 = await createNodeContext({
+          workflowContext,
+          nodeData: templatingNodeData1,
+        });
+        workflowContext.addNodeContext(templatingNodeContext1);
+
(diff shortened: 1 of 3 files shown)


ee/codegen/src/__test__/workflow.test.ts
```diff
@@ -559,6 +559,86 @@ describe("Workflow", () => {
...
+          workflowContext,
+          inputs,
+          nodes: [
+            templatingNodeData1,
+            templatingNodeData2,
+            templatingNodeData3,
+          ],
+          // This test case fails with the legacy graph generation algorithm
+          // so we don't parameterize over it.
+          breadthFirstGraphGeneration: true,
```
dvargasfuertes, line 636: third test case we've found so far that the old algo doesn't cover

ee/codegen/src/generators/workflow.ts
```diff
@@ -716,96 +716,111 @@ export class Workflow {
...
-            edge.sourceHandleId
-          );
-          if (sourcePortContext) {
-            graphMutableAst = {
-              type: "set",
-              values: [
-                graphMutableAst,
-                {
+        } else if (mutableAst.type === "node_reference") {
+          if (sourceNode && mutableAst.reference === sourceNode) {
```
dvargasfuertes: renaming the function scoped arg from `graphMutableAst` to `mutableAst` just to deconflict with the variable in the while loop scope that we are updating with each edge

kirawoods (commented):
Very cool solution to this!

PR #311, opened 2024-12-05
Graph Gen - Set up and implement test case for an edge between a node and a set

dvargas92495:
`graph = A >> {B, C}`

### DIFF
--- a/ee/codegen/src/__test__/workflow.test.ts
+++ b/ee/codegen/src/__test__/workflow.test.ts
@@ -639,6 +639,87 @@ describe("Workflow", () => {
         workflow.getWorkflowFile().write(writer);
         expect(await writer.toStringFormatted()).toMatchSnapshot();
       });
+
+      it.each([[true], [false]])(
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/workflow.test.ts
```diff
@@ -639,6 +639,87 @@ describe("Workflow", () => {
         workflow.getWorkflowFile().write(writer);
         expect(await writer.toStringFormatted()).toMatchSnapshot();
       });
+
+      it.each([[true], [false]])(
+        "should be correct for set a node to a set",
```
kirawoods, line 644: I'm a little confused at what this line is supposed to be saying

kirawoods (commented):
Added one comment to get clarity on the test description

ee/codegen/src/__test__/workflow.test.ts
dvargasfuertes: yea that first `set` shouldn't be there, good catch

PR #312, opened 2024-12-05
POC for discriminated unions in codegen

dvargas92495:


ee/codegen/src/__test__/serializers/vellum.test.ts
```diff
@@ -203,7 +203,7 @@ describe("vellum", () => {
         definition: null,
       };
 
-      const parsedData = PromptNodeSerializer.parse(data);
+      const parsedData = WorkflowNodeSerializer.parse(data);
```
dvargasfuertes, line 206: switching this to `WorkflowNodeSerializer` since the child no longer has the type field

ee/codegen/src/serializers/vellum.ts
```diff
@@ -14,6 +14,7 @@ import {
   Schema,
   record as recordSchema,
   unknown as unknownSchema,
+  union,
```
dvargasfuertes, line 17: all other changes stay confined to this file, without polluting the types we care about

ee/codegen/src/serializers/vellum.ts
```diff
@@ -585,10 +585,9 @@ export declare namespace BaseDisplayableWorkflowNodeSerializer {
 
 export const EntrypointNodeSerializer: ObjectSchema<
```
dvargasfuertes, line 586: I'd also argue we don't need these `ObjectSchema` annotations here at all, except for maybe in the case of recursion

noanflaherty (approved):

ee/codegen/src/serializers/vellum.ts
noanflaherty: They're def needed for recursion – I haven't figured out at what level of the recursion they're minimally needed, so got into the habit of just added everywhere.

PR #313, opened 2024-12-05
[Codegen] Use Discriminated Unions

noanflaherty:
We previously were using undiscriminated unions in our serializers, making serialization errors verbose and hard to debug. The hope is that by using discriminated unions, we get more direct error messages.

### DIFF
--- a/ee/codegen/src/serializers/vellum.ts
+++ b/ee/codegen/src/serializers/vellum.ts
@@ -611,7 +611,7 @@ export declare namespace EntrypointNodeSerializer {
 
 export const InlineSubworkflowNodeDataSerializer: ObjectSchema<
   InlineSubworkflowNodeDataSerializer.Raw,
-  InlineSubworkflowNodeData
+  Omit<InlineSubworkflowNodeData, "variant">
 > = objectSchema({
   workflowRawData: propertySchema(
     "workflow_raw_data",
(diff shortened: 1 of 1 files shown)


ee/codegen/src/serializers/vellum.ts
```diff
@@ -142,15 +142,21 @@ export const ChatMessagePromptTemplateBlockSerializer: ObjectSchema<
     ),
     blocks: propertySchema(
       "blocks",
-      listSchema(lazy(() => PromptTemplateBlockSerializer))
+      listSchema(
+        lazySchema(() => {
+          return PromptTemplateBlockSerializer as unknown as Schema<
+            PromptTemplateBlockSerializer.Raw,
+            PromptTemplateBlock
+          >;
```
noanflaherty, line 150: You'll notice I have to do type casting in a couple places.

This is because:
1. We have recursive schemas
2. A schema containing `lazy` requires you to explicitly declare the type via `ObjectSchema` if it's to be used by parent schemas
3. Our serializer `parsed` types now `Omit` their discriminators
4. Which means they don't conform to our internal types

I think the only way to get rid of the type casts would be to make duplicates of our internal types that don't include their discriminators, which doesn't seem worth it.

dvargasfuertes: [oos] We def should report this to fern as a bug. maybe if we upgrade our node generator they have already fixed it.

fwiw tho, I would expect `PromptTemplateBlockSerializer` to keep the block_type field even with how it exists today, obviating the need for the cast

dvargas92495 (approved):

ee/codegen/src/serializers/vellum.ts
noanflaherty: Yeah it's the fact that the interior blocks within no longer have their own block_type. Like `ChatMessagePromptTemplateBlock` and `RichTextPromptTemplateBlock` both are recursive types that omit their inner children's blocks.

PR #314, opened 2024-12-05
Remove new graph gen feature flag

dvargas92495:
Removes the Graph gen feature flag to cut over to the new algo.

Note that we are not yet guaranteeing that we could generate _all_ graphs. just that we are now at least a superset of the old algo. Will leave this one up for review while I continue bashing more test cases and productionizing in parallel


kirawoods (approved):
✅

PR #316, opened 2024-12-05
Skip Outputs class generation if there are no outputs defined

dvargas92495:
We don't need to generate Workflow Outputs class if none are defined. Has the additional benefit of slimming down future graph generation tests.

### DIFF
--- a/ee/codegen/src/__test__/workflow.test.ts
+++ b/ee/codegen/src/__test__/workflow.test.ts
@@ -721,7 +721,7 @@ describe("Workflow", () => {
         expect(await writer.toStringFormatted()).toMatchSnapshot();
       });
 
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/workflow.test.ts
```diff
@@ -721,7 +721,7 @@ describe("Workflow", () => {
         expect(await writer.toStringFormatted()).toMatchSnapshot();
       });
 
-      it("should be correct for set a node to a set", async () => {
+      it("should be correct for a node to a set", async () => {
```
dvargasfuertes, line 724: feedback caught in a prev pr

kirawoods (approved):
✅

noanflaherty (commented):
✅

PR #317, opened 2024-12-05
Add support for Arrays in chat messages

kirawoods:


ee/codegen/src/generators/chat-message-content.ts
```diff
@@ -153,9 +153,15 @@ export class ChatMessageContent extends AstNode {
     }
 
     if (contentType === "ARRAY") {
-      // TODO: Implement array types and call this recursively
-      //    https://app.shortcut.com/vellum/story/4937/flesh-out-codegen-for-all-chat-message-content-types
-      throw new Error("Unhandled type: ARRAY");
+      const arrayValue = this.chatMessageContent.value as unknown[];
```
kirawoods, line 171: Q: What should the type be here (probably not unknown[])? 

Q: I'm assuming we don't need to inheritReferences for this type since we'd be inheriting references on the other types that we're calling when we create the array. Is that a correct assumption?

ee/codegen/src/generators/chat-message-content.ts
```diff
@@ -153,9 +153,15 @@ export class ChatMessageContent extends AstNode {
...
+      const arrayElements = arrayValue.map(
+        (element) =>
```
kirawoods, line 173: Create custom AstNode here

NgoHarrison (approved):

dvargas92495 (approved):

PR #318, opened 2024-12-05
Separate the graph attribute into its own AstNode file

dvargas92495:
This PR is a logicless refactor to split `GraphAttribute` as its own ASTNode. I tried to follow the patterns established elsewhere, but could use some feedback on the constructor + write methods.

This allows us to concentrate the graph generation to a single place and even move our tests over to a `graph-attribute.test.ts` if we choose to in the future

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -2,9 +2,9 @@
 
 exports[`Workflow > write > graph > should be correct for a basic conditional node case 1`] = `
 "from vellum.workflows import BaseWorkflow
-from .nodes.conditional_node import ConditionalNode
 from .nodes.templating_node import TemplatingNode
 from .nodes.templating_node_2 import TemplatingNode2
+from .nodes.conditional_node import ConditionalNode
 
 
 class TestWorkflow(BaseWorkflow):
@@ -92,9 +92,9 @@ class TestWorkflow(BaseWorkflow):
 
 exports[`Workflow > write > graph > should be correct for a node to a set 1`] = `
 "from vellum.workflows import BaseWorkflow
-from .nodes.templating_node import TemplatingNode
 from .nodes.templating_node_2 import TemplatingNode2
 from .nodes.templating_node_3 import TemplatingNode3
+from .nodes.templating_node import TemplatingNode
 
 
 class TestWorkflow(BaseWorkflow):
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -2,9 +2,9 @@
 
 exports[`Workflow > write > graph > should be correct for a basic conditional node case 1`] = `
 "from vellum.workflows import BaseWorkflow
+from .nodes.conditional_node import ConditionalNode
 from .nodes.templating_node import TemplatingNode
 from .nodes.templating_node_2 import TemplatingNode2
-from .nodes.conditional_node import ConditionalNode
 
 
 class TestWorkflow(BaseWorkflow):
@@ -92,9 +92,9 @@ class TestWorkflow(BaseWorkflow):
 
 exports[`Workflow > write > graph > should be correct for a node to a set 1`] = `
 "from vellum.workflows import BaseWorkflow
+from .nodes.templating_node import TemplatingNode
 from .nodes.templating_node_2 import TemplatingNode2
 from .nodes.templating_node_3 import TemplatingNode3
-from .nodes.templating_node import TemplatingNode
 
 
 class TestWorkflow(BaseWorkflow):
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -2,9 +2,9 @@
 
 exports[`Workflow > write > graph > should be correct for a basic conditional node case 1`] = `
 "from vellum.workflows import BaseWorkflow
-from .nodes.conditional_node import ConditionalNode
```
dvargasfuertes: I was hoping for no snapshot changes, but something about how we're now doing references changed the import order for these pre-isort imports

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -0,0 +1,409 @@
...
+  private readonly entrypointNodeEdges: WorkflowEdge[];
+  private readonly entrypointNodeId: string;
+  private readonly astNode: python.AstNode;
+
+  public constructor({
+    entrypointNodeEdges,
+    entrypointNodeId,
+    edgesByPortId,
+    workflowContext,
+  }: GraphAttribute.Args) {
```
dvargasfuertes, line 57: I have ideas on how to simplify this constructor as we add more test cases. the issue is that we still reuse some of these entrypoint node and edges stuff for display class generation

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -0,0 +1,409 @@
...
+      }
+
+      const isPlural = (mutableAst: GraphMutableAst): boolean => {
+        return (
+          mutableAst.type === "right_shift" ||
+          (mutableAst.type === "set" && mutableAst.values.every(isPlural))
+        );
+      };
+
+      const getAstSources = (
```
dvargasfuertes, line 105: Am planning to pull out these nested helpers in a followup PR. Wanted to first do a logicless migration

ee/codegen/src/generators/workflow.ts
```diff
@@ -47,7 +45,6 @@ export class Workflow {
   private readonly inputs: Inputs;
   private readonly nodes: WorkflowDataNode[];
   private readonly edgesByPortId: Map<string, WorkflowEdge[]>;
-  private readonly entrypointPortContexts: PortContext[];
```
dvargasfuertes, line 50: was unused from a previous pr

src/vellum/workflows/workflows/base.py
```diff
@@ -86,8 +86,11 @@ def __new__(mcs, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any
         return super().__new__(mcs, name, bases, dct)
 
 
+GraphAttribute = Union[Type[BaseNode], Graph, Set[Type[BaseNode]], Set[Graph]]
```
dvargasfuertes, line 89: This is ultimately why I called the generator `graph-attribute` instead of `graph`. `Graph` is a specific class in the SDK, and we are ultimately looking to generate this union

NgoHarrison:
The write method for the graph attribute looks good and the constructor for the graph attribute also looks good to me in terms of pattern and convention.

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -0,0 +1,409 @@
...
+  }: GraphAttribute.Args) {
+    super();
+    this.entrypointNodeEdges = entrypointNodeEdges;
+    this.entrypointNodeId = entrypointNodeId;
+    this.workflowContext = workflowContext;
+    this.edgesByPortId = edgesByPortId;
+
+    const graphMutableAst = this.generateGraphMutableAst();
+    this.astNode = this.getGraphAttributeAstNode(graphMutableAst);
+    this.inheritReferences(this.astNode);
```
noanflaherty: nit: The convention I've been trying to establish is that any helper that returns an instance of a custom AstNode:
1. Has a name prefix of `generate`
2. Calls `this.inheritReferences` internally at the very end, right before returning the node.

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -0,0 +1,409 @@
...
+        lhs,
+        rhs,
+      });
+    }
+
+    return python.TypeInstantiation.none();
+  }
+
+  public write(writer: Writer): void {
+    this.astNode.write(writer);
```
noanflaherty, line 410: This is great! `write` methods should never create new AstNode instances within them. Any AstNode instance that gets created should happen as part of the constructor, which this does 🎉

noanflaherty (approved):
Looks great!

ee/codegen/src/generators/graph-attribute.ts
noanflaherty: cc @NgoHarrison and @kirawoods so that we're all aligned.

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -0,0 +1,409 @@
...
+
+    const graphMutableAst = this.generateGraphMutableAst();
+    this.astNode = this.getGraphAttributeAstNode(graphMutableAst);
+    this.inheritReferences(this.astNode);
+  }
+
+  /**
+   * Generates a mutable graph AST.
+   *
+   * The algorithm we implement is a Breadth-First Search (BFS) that traverses through
```
kirawoods, line 70: Glad that all my Leetcode hours weren't for nothing 🤣

kirawoods (approved):
🎉

PR #323, opened 2024-12-06
Setup test & implementation for new workflow state snapshot events

dvargas92495:
This PR introduces `workflow.execution.snapshotted` events. They are responsible for emitting to the user what the state was at that moment in time. They will contain in its body the full state and are emitted whenever the state is updated. Some opinionated stances that were made and seeking feedback on:
- It will be the only `workflow.*` level event not emitted to the user by default
- adding the new status `snapshotted` as part of our `entity.action.status` spec.
- getting rid of the `meta.is_terminated` field within state. We first introduced this to make resume from node slightly easier, but have since found a workaround, and we are trying to keep all state updates to come from a node thread instead of a workflow thread.

### DIFF
--- a/tests/workflows/basic_emitter_workflow/tests/test_workflow.py
+++ b/tests/workflows/basic_emitter_workflow/tests/test_workflow.py
@@ -49,49 +49,22 @@ def test_run_workflow__happy_path(mock_uuid4_generator, mock_datetime_now):
     assert events[0].span_id == workflow_span_id
     assert events[0].timestamp == frozen_datetime
 
-    assert events[1].name == "workflow.execution.snapshotted"
-    assert default_serializer(events[1].state) == {
+    assert events[1].name == "node.execution.initiated"
+    assert events[1].node_definition == StartNode
+
+    assert events[2].name == "workflow.execution.snapshotted"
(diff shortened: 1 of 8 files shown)


tests/workflows/basic_emitter_workflow/tests/test_workflow.py
```diff
@@ -51,179 +52,109 @@ def test_run_workflow__happy_path(mock_uuid4_generator, mock_datetime_now):
...
+    assert events[7].outputs == {"final_value": "Score: 13"}
+
+    assert events[8].name == "workflow.execution.fulfilled"
+    assert events[8].outputs == {"final_value": "Score: 13"}
+
+    assert len(events) == 9, final_event
+
+    # AND the emitter should have emitted all of the expected state snapshots
+    state_snapshots = list(emitter.state_snapshots)
+    assert len(state_snapshots) == 3
```
dvargasfuertes, line 160: I'm also currently planning to remove the snapshot_state method from emitters now that state can be emitted through events

dvargas92495:
Merging to keep moving, still interested in feedback on decisions made here

PR #324, opened 2024-12-06
Simplify and rename workflow event queues

dvargas92495:
This PR focuses on:
- Deleting the internal-only `WorkItemEvent` so that both of our main workflow runner queues emit workflow events
- rename both queues to indicate designated purpose
- Use the inner queue for subworkflows to ensure better order consistency

This PR is also a precursor to emitting state events: https://github.com/vellum-ai/vellum-python-sdks/pull/323


src/vellum/workflows/runner/runner.py
```diff
@@ -73,9 +72,7 @@ def __init__(
         parent_context: Optional[ParentContext] = None,
     ):
         if state and external_inputs:
-            raise ValueError(
-                "Can only run a Workflow providing one of state or external inputs, not both"
-            )
+            raise ValueError("Can only run a Workflow providing one of state or external inputs, not both")
```
dvargasfuertes, line 75: @marlenc3 do you have any special width configured locally? I'm going to add black and flake to our precommits later tn so we don't keep ping ponging haha. But i think our default is 120

src/vellum/workflows/runner/types.py
```diff
@@ -1,16 +0,0 @@
-"""Only intenral types and enums for WorkflowRunner should be defined in this module."""
```
dvargasfuertes, line 1: Deleting this file was the DoD. Also `intenral`

src/vellum/workflows/runner/runner.py
marlenc3: I do not but I can add in the default 120! 🫡

marlenc3 (approved):

PR #325, opened 2024-12-06
Graph generation test case for node to set to node

dvargas92495:
`graph = A >> {B, C} >> D`

### DIFF
--- a/ee/codegen/src/generators/graph-attribute.ts
+++ b/ee/codegen/src/generators/graph-attribute.ts
@@ -329,6 +329,11 @@ export class GraphAttribute extends AstNode {
 
           const newRhs = addEdgeToGraph(mutableAst.rhs, lhsTerminal.reference);
           if (newRhs) {
+            if (lhsTerminals.length > 1 && newRhs.type === "set") {
+              throw new Error(
+                "Adding an edge between two sets is not supported"
+              );
+            }
             return {
               type: "right_shift",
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -316,14 +318,22 @@ export class GraphAttribute extends AstNode {
...
-                type: "right_shift",
-                lhs: mutableAst,
-                rhs: { type: "node_reference", reference: targetNode },
-              };
-            }
+            return;
+          }
+
+          const lhsTerminals = getAstTerminals(mutableAst.lhs);
+          const lhsTerminal = lhsTerminals[0];
```
dvargasfuertes, line 325: Even the sdk itself doesn't support `{A, B} >> {C, D}` yet. We'll need to make sdk-side changes first to do so

NgoHarrison (approved):
✅

ee/codegen/src/generators/graph-attribute.ts
noanflaherty: Is it easy to detect this case and throw an error if we do? My goal is to have code that self-documents this edge case and makes it easy for a future developer to know where to update if/when we do support this.

noanflaherty (approved):

kirawoods (commented):
✅

PR #326, opened 2024-12-06
Introduce black & flake to pre commit and CI

dvargas92495:
This PR adds the formatter (black) and linter (flake) we're used to in our other repos to our sdk. I would focus the review on just the first commit where we do the actual configuration. The second commit is the actual reformatting to these introduced standards.


pyproject.toml
```diff
@@ -115,3 +122,16 @@ known_vellum = ["vellum", "vellum_cli", "vellum_ee"]
...
 sections = ["FUTURE", "TESTING", "STDLIB", "TYPING", "THIRDPARTY", "VELLUM", "FIRSTPARTY", "LOCALFOLDER"]
 no_lines_before = ["STDLIB", "TYPING"]
+
+[tool.black]
+line-length = 120
+target-version = [
+    "py38",
+]
+include = ".*\\.py$"
+exclude = "(.git|venv|src/vellum/client|tests/client|scripts)"
```
dvargasfuertes, line 132: should we auto format/auto import sort the fern generated files in our post processing script so that we could remove these exclusions?

pyproject.toml
```diff
@@ -115,3 +122,16 @@ known_vellum = ["vellum", "vellum_cli", "vellum_ee"]
...
+
+[tool.flake8]
+max-line-length = 120
+exclude = ".git,.venv,ee/codegen/node_modules,ee/codegen_integration/fixtures,scripts,src/vellum/__init__.py,src/vellum/client,src/vellum/core,src/vellum/environment.py,src/vellum/errors,src/vellum/resources,src/vellum/types,tests/client"
```
dvargasfuertes, line 136: Am highlighting `ee/codegen_integration/fixtures`, meaning we codegen code that fails our linter today (mostly strings that are too long for the file width)

siddseethepalli (commented):
✅

PR #327, opened 2024-12-06
[SC-5655] Add required/default/extensions to input variables

NgoHarrison:


kirawoods (approved):
✅

ee/codegen_integration/fixtures/simple_search_node/display_data/simple_search_node.json
```diff
@@ -311,9 +311,12 @@
       "id": "a6ef8809-346e-469c-beed-2e5c4e9844c5",
       "key": "query",
       "type": "STRING",
-      "default": null,
+      "default": {
+        "type": "STRING",
+        "value": "Search node query default"
+      },
       "required": true,
```
noanflaherty, line 318: [q] I think this should probably be `"required": false` since there is a default value. Wdyt?

noanflaherty (commented):
This looks awesome – nice work. Just one open question from me.

ee/codegen_integration/fixtures/simple_search_node/display_data/simple_search_node.json
NgoHarrison: Oops totally agree, will make a fast follow to change this

PR #328, opened 2024-12-06
Start injecting custom errors in response to orphaned node output id

dvargas92495:
I first started by triaging this sentry error: https://vellum.sentry.io/issues/6114814120/?project=4508362703634432&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=7d&stream_index=0

I then realized I wanted to use this as an opportunity to start bubbling up custom errors that eventually results in a list of warnings for the user. 

This PR simply adds a test that replicates the situation posed by the sentry error, adding in some custom error classes. In a future PR, will then use this to parlay into codegen warnings so that we generate as much as possible

### DIFF
--- a/ee/codegen/src/__test__/nodes/base-node.test.ts
+++ b/ee/codegen/src/__test__/nodes/base-node.test.ts
@@ -1,0 +1,42 @@
+import { workflowContextFactory } from "src/__test__/helpers";
+import { templatingNodeFactory } from "src/__test__/helpers/node-data-factories";
+import { createNodeContext } from "src/context";
+import { TemplatingNodeContext } from "src/context/node-context/templating-node";
+import { NodeAttributeGenerationError } from "src/generators/errors";
+import { TemplatingNode } from "src/generators/nodes/templating-node";
+
+describe("BaseNode", () => {
+  describe("failures", () => {
+    it("should throw the expected error when a input references an invalid node", async () => {
+      const workflowContext = workflowContextFactory();
+
+      const templatingNodeData = templatingNodeFactory({
+        inputRules: [
+          {
+            type: "NODE_OUTPUT",
+            data: {
+              nodeId: "12345678-1234-5678-1234-567812345678",
+              outputId: "90abcdef-90ab-cdef-90ab-cdef90abcdef",
+            },
+          },
+        ],
+      });
+      const templatingNodeContext = (await createNodeContext({
+        workflowContext,
+        nodeData: templatingNodeData,
+      })) as TemplatingNodeContext;
+      workflowContext.addNodeContext(templatingNodeContext);
+
+      expect(() => {
+        new TemplatingNode({
+          workflowContext,
+          nodeContext: templatingNodeContext,
+        });
+      }).toThrow(
+        new NodeAttributeGenerationError(
+          "Failed to generate attribute 'TemplatingNode.inputs.text': Failed to find node with id '12345678-1234-5678-1234-567812345678'"
+        )
+      );
+    });
+  });
+});
--- a/ee/codegen/src/context/workflow-context/workflow-context.ts
+++ b/ee/codegen/src/context/workflow-context/workflow-context.ts
@@ -7,6 +7,7 @@ import { PortContext } from "src/context/port-context";
 import { generateSdkModulePaths } from "src/context/workflow-context/sdk-module-paths";
 import { SDK_MODULE_PATHS } from "src/context/workflow-context/types";
 import { WorkflowOutputContext } from "src/context/workflow-output-context";
+import { NodeNotFoundError } from "src/generators/errors";
 import { BaseNode } from "src/generators/nodes/bases";
 import {
   EntrypointNode,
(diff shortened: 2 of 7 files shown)


ee/codegen/src/__test__/nodes/base-node.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import { TemplatingNodeContext } from "src/context/node-context/templating-node";
+import { NodeAttributeGenerationError } from "src/generators/errors";
+import { TemplatingNode } from "src/generators/nodes/templating-node";
+
+describe("BaseNode", () => {
+  describe("failures", () => {
+    it("should throw the expected error when a input references an invalid node", async () => {
+      const workflowContext = workflowContextFactory();
+
+      const templatingNodeData = templatingNodeFactory({
```
dvargasfuertes, line 13: I previously looked into adding a `baseNodeFactory` before realizing that bc of the reliance of id vs name, it wouldn't reproduce the same error. So continuing to use Templating Node as my standard "Base Node" for legacy nodes for now

ee/codegen/src/__test__/nodes/base-node.test.ts
```diff
@@ -0,0 +1,42 @@
...
+      expect(() => {
+        new TemplatingNode({
+          workflowContext,
+          nodeContext: templatingNodeContext,
+        });
+      }).toThrow(
+        new NodeAttributeGenerationError(
+          "Failed to generate attribute 'TemplatingNode.inputs.text': Failed to find node with id '12345678-1234-5678-1234-567812345678'"
+        )
+      );
```
dvargasfuertes, line 39: In a future PR, instead of expecting this to throw, we change this to expect that the node has multiple warnings. thoughts on this API:
```typescript
const node = new TemplatingNode({
    workflowContext,
    nodeContext: templatingNodeContext,
});
const warnings = node.getWarnings();
expect(warnings).toBe([
    {
        "code": "NODE_ATTRIBUTE_GENERATION_ERROR",
        "message": "Failed to etc etc etc...",
    }
])
```

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -235,8 +236,9 @@ export class WorkflowContext {
   }
 
   public async getMLModelNameById(mlModelId: string): Promise<string> {
-    if (this.mlModelNamesById[mlModelId]) {
-      return this.mlModelNamesById[mlModelId];
+    const mlModelName = this.mlModelNamesById[mlModelId];
+    if (mlModelName) {
+      return mlModelName;
```
dvargasfuertes, line 241: random ts warning I was seeing locally

ee/codegen/src/generators/nodes/templating-node.ts
```diff
@@ -46,12 +46,9 @@ export class TemplatingNode extends BaseSingleFileNode<
       python.field({
         name: "inputs",
         initializer: python.TypeInstantiation.dict(
-          otherInputs.map((input) => ({
-            key: python.TypeInstantiation.str(input.key),
-            value: new NodeInputValuePointer({
-              workflowContext: this.workflowContext,
-              nodeInputValuePointerData: input.value,
-            }),
+          otherInputs.map((codeInput) => ({
```
dvargasfuertes, line 49: switch this to how the other nodes do this to have a single consumer of `NodeInputValuePointer`

ee/codegen/src/project.ts
```diff
@@ -65,10 +66,6 @@ import {
 import { getNodeId } from "src/utils/nodes";
 import { assertUnreachable } from "src/utils/typing";
 
-export class ProjectSerializationError extends Error {
-  isProjectSerializationError = true;
```
dvargasfuertes, line 69: moved to `errors.ts`. We'll need to update codegen service to check `instanceof ProjectSerializationError`

ee/codegen/src/generators/errors.ts
```diff
@@ -0,0 +1,30 @@
+export type CodegenErrorCode =
+  | "PROJECT_SERIALIZATION_ERROR"
+  | "NODE_ATTRIBUTE_GENERATION_ERROR"
+  | "NODE_NOT_FOUND_ERROR";
+
+export abstract class BaseCodegenError extends Error {
+  abstract code: CodegenErrorCode;
```
dvargasfuertes, line 7: I'll note that in most `catch` checks, we'll need to do `instanceof MyCustomError` bc `catch` types the error as `unknown`. The `code` here is just for serialization purposes

ee/codegen/src/__test__/nodes/base-node.test.ts
noanflaherty: I'm imagining that our CLI has a flag that's like `--strict=True`. If strict mode is enabled, we error instead of warn. If False, we do what we can. This applies to both codegen and serialization.

In this world, I'd want this to be called `getErrors()` because that's what they are, and then the flag determines what to do about those errors.

noanflaherty (approved):

PR #330, opened 2024-12-06
Introduce function compilation

dvargas92495:
This PR introduces the utility for function compilation into our Vellum-native `FunctionDefinition` class. We simply solve the base case, and will be rolling out each test case per PR. We eventually (the last PR) will use this utility within our inline prompt nodes to allow users to pass in arbitrary functions defined locally as arguments.

### DIFF
--- a/src/vellum/workflows/utils/functions.py
+++ b/src/vellum/workflows/utils/functions.py
@@ -1,0 +1,13 @@
+from typing import Callable
+
+from vellum.client.types.function_definition import FunctionDefinition
+
+
+def compile_function_definition(function: Callable) -> FunctionDefinition:
+    """
+    Converts a Python function into our Vellum-native FunctionDefinition type.
+    """
+
+    return FunctionDefinition(
+        name=function.__name__,
+    )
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/utils/functions.py
```diff
@@ -0,0 +1,13 @@
+from typing import Callable
+
+from vellum.client.types.function_definition import FunctionDefinition
+
+
+def compile_function_definition(function: Callable) -> FunctionDefinition:
```
dvargasfuertes, line 6: I debated whether I should start the `src/vellum/prompts` module and put this in `src/vellum/prompts/utils/functions.py` instead. Thoughts?

src/vellum/workflows/utils/tests/test_functions.py
```diff
@@ -0,0 +1,16 @@
+from vellum.client.types.function_definition import FunctionDefinition
+from vellum.workflows.utils.functions import compile_function_definition
+
+
+def test_compile_function_definition__just_name():
```
dvargasfuertes, line 5: More tests coming in future PRs

marlenc3:
✅

src/vellum/workflows/utils/functions.py
noanflaherty: I'm game!

noanflaherty (commented):
I love where this is headed. I have some spicy ideas I wanna run by you

kirawoods (commented):
✅

PR #331, opened 2024-12-06
Setup test for and Implement function compilation for all parameter types

dvargas92495:
With just these two tests, we now support everything that [openai swarm does](https://github.com/openai/swarm/blob/main/swarm/util.py#L31-L87)

However we can't stop here. We still need to do unions, parameterized lists & dicts, dataclasses, optionals, etc etc

### DIFF
--- a/src/vellum/workflows/utils/tests/test_functions.py
+++ b/src/vellum/workflows/utils/tests/test_functions.py
@@ -13,4 +13,31 @@ def test_compile_function_definition__just_name():
     # THEN it should return the compiled function definition
     assert compiled_function == FunctionDefinition(
         name="my_function",
+        parameters={"type": "object", "properties": {}, "required": []},
+    )
+
+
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/utils/tests/test_functions.py
```diff
@@ -13,4 +13,31 @@ def my_function():
     # THEN it should return the compiled function definition
     assert compiled_function == FunctionDefinition(
         name="my_function",
+        parameters={"type": "object", "properties": {}, "required": []},
+    )
+
+
+def test_compile_function_definition__all_args():
```
dvargasfuertes, line 20: I considered `pytest.mark.parameterize`, but it was awkward to define the functions as a parameter. these tests are simple enough to do as individual ones anyway

noanflaherty (commented):
✅

kirawoods (commented):
✅

PR #332, opened 2024-12-06
Compile function union args

dvargas92495:


src/vellum/workflows/utils/tests/test_functions.py
```diff
@@ -41,3 +43,24 @@ def my_function(a: str, b: int, c: float, d: bool, e: list, f: dict):
...
+    # WHEN compiling the function
+    compiled_function = compile_function_definition(my_function)
+
+    # THEN it should return the compiled function definition
+    assert compiled_function == FunctionDefinition(
+        name="my_function",
+        parameters={
+            "type": "object",
+            "properties": {
+                "a": {"anyOf": [{"type": "string"}, {"type": "integer"}]},
```
dvargasfuertes, line 62: I hate that json schema allows `"type"` to be an array and prefer to avoid falling into that temptation indefinitely. This also plays better with our UI

noanflaherty (commented):
✅

kirawoods (commented):
✅

PR #336, opened 2024-12-06
Setup test and impl for compiling dataclasses

dvargas92495:
Took an opinionated stance here to say that _all_ dataclasses will be pushed to a `$defs` schema. I contemplated only doing this on the second reference of a dataclass, but that's slightly more complex and this is at least more consistent.


kirawoods (commented):
✅

PR #329, opened 2024-12-06
Creating a `workflows` subcommand for vellum push

dvargas92495:
This PR introduces support for:
- `vellum push workflows [module]` - push a specific workflow, as defined in config
- `vellum push [module]` - push a specific resource, as defined in config

In the near future, we plan to support:
- `vellum push` - push all of the resources
- `vellum push workflows` - push all of the workflows

PR #337, opened 2024-12-06
Use click's test runner to during testing

dvargas92495:
We are going to be making some changes to the CLI interface (sneak peak [here](https://github.com/vellum-ai/vellum-python-sdks/pull/329)), that will require us to override some of Click's default grouping behavior. Given this, I wanted to make our cli tests more higher level to account for the logic we will soon introduce here. Luckily, click exports this sweet `CliRunner` that makes this testing easy and performant.

So this PR simply cuts over our current cli tests to use the cli runner. We left one test in `test_main` to still test from `subprocess` to just ensure that it installs correctly.


m-abboud (approved):
What is a "click"?

dvargas92495:
it's the library powering our CLI. highly recommend!

PR #338, opened 2024-12-07
Create subcommand `vellum pull workflows`

dvargas92495:
This PR aims to create a subcommand for vellum pull for pulling specific entities.

Here's the list of commands now supported:

- vellum pull workflows [module] - pull a specific workflow, as defined in config
- vellum pull [module] - pull a specific module, as defined in config

In the future, we're looking to support the following as well:
- vellum pull - pull all of the things
- vellum pull workflows - pull all of the workflows

### DIFF
--- a/ee/vellum_cli/__init__.py
+++ b/ee/vellum_cli/__init__.py
@@ -40,7 +40,46 @@ def push(
     )
 
 
-@main.command()
+class PullGroup(ClickAliasedGroup):
+    def get_command(self, ctx, cmd_name):
+        # First try to get the command normally
+        cmd = super().get_command(ctx, cmd_name)
+        if cmd is not None:
+            return cmd
+
+        # If no command found, call our catch-all
+        return pull_module
+
+
+@main.group(invoke_without_command=True, cls=PullGroup)
+@click.pass_context
+@click.option(
+    "--include-json",
+    is_flag=True,
+    help="""Include the JSON representation of each Resource in the pull response. \
+Should only be used for debugging purposes.""",
+)
+@click.option(
+    "--exclude-code",
+    is_flag=True,
+    help="""Exclude the code definition of each Resource from the pull response. \
+Should only be used for debugging purposes.""",
+)
+def pull(
+    ctx: click.Context,
+    include_json: Optional[bool],
+    exclude_code: Optional[bool],
+) -> None:
+    """Pull the first configured Resource from Vellum"""
+
+    if ctx.invoked_subcommand is None:
+        pull_command(
+            include_json=include_json,
+            exclude_code=exclude_code,
+        )
+
+
+@pull.command(name="workflows")
 @click.argument("module", required=False)
 @click.option(
     "--include-json",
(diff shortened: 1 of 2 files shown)


ee/vellum_cli/__init__.py
```diff
@@ -40,7 +40,46 @@ def push(
...
+@main.group(invoke_without_command=True, cls=PullGroup)
+@click.pass_context
+@click.option(
+    "--include-json",
+    is_flag=True,
+    help="""Include the JSON representation of each Resource in the pull response. \
+Should only be used for debugging purposes.""",
+)
+@click.option(
+    "--exclude-code",
```
dvargasfuertes, line 63: I was thinking that these two arguments were useful for all three commands

ee/vellum_cli/__init__.py
```diff
@@ -40,7 +40,46 @@ def push(
...
+        # If no command found, call our catch-all
+        return pull_module
+
+
+@main.group(invoke_without_command=True, cls=PullGroup)
+@click.pass_context
+@click.option(
+    "--include-json",
+    is_flag=True,
+    help="""Include the JSON representation of each Resource in the pull response. \
```
dvargasfuertes, line 59: We need a name for the general "thing" that users can push and pull. `Resource`? `Entity`? `Executable`? `Executable Entity`? The last one is the one we use internally

ee/vellum_cli/__init__.py
```diff
@@ -40,7 +40,46 @@ def push(
...
+    is_flag=True,
+    help="""Exclude the code definition of each Resource from the pull response. \
+Should only be used for debugging purposes.""",
+)
+def pull(
+    ctx: click.Context,
+    include_json: Optional[bool],
+    exclude_code: Optional[bool],
+) -> None:
+    """Pull the first configured Resource from Vellum"""
```
dvargasfuertes, line 73: In the future, this will be all instead of just first

ee/vellum_cli/__init__.py
```diff
@@ -55,13 +94,17 @@ def push(
...
+def pull_workflows(
     module: Optional[str],
     include_json: Optional[bool],
     workflow_sandbox_id: Optional[str],
     exclude_code: Optional[bool],
 ) -> None:
-    """Pull Workflow from Vellum"""
+    """
+    Pull Workflows from Vellum. If a module is provided, only the Workflow for that module will be pulled.
+    If no module is provided, the first configured Workflow will be pulled.
```
dvargasfuertes, line 105: In the future, this will be all instead of just first

ee/vellum_cli/tests/test_pull.py
```diff
@@ -94,7 +127,7 @@ def test_pull__sandbox_id_with_no_config(vellum_client):
 
     # WHEN the user runs the pull command with the workflow sandbox id and no module
     runner = CliRunner()
-    result = runner.invoke(cli_main, ["pull", "--workflow-sandbox-id", workflow_sandbox_id])
+    result = runner.invoke(cli_main, ["pull", "workflows", "--workflow-sandbox-id", workflow_sandbox_id])
```
dvargasfuertes, line 130: Main goal of the PR. Note that removing it from `vellum pull` is a breaking change and we'll need to update our UI

dvargas92495:
Gonna keep moving but am keen on feedback here

ee/vellum_cli/__init__.py
noanflaherty: Let's go with Resource, this is consistent with other things like Terraform https://developer.hashicorp.com/terraform/language/resources

ee/vellum_cli/__init__.py
```diff
@@ -70,6 +113,35 @@ def pull(
     )
 
 
+@pull.command(name="*")
+@click.pass_context
+@click.option(
+    "--include-json",
```
noanflaherty, line 119: nit: It'd be cool to DRY out some of these options/help text

noanflaherty (commented):
✅

ee/vellum_cli/__init__.py
dvargasfuertes: I started to do so, but the wording was slightly different depending on Resource or Workflow. may still try to do so in a future pr

PR #339, opened 2024-12-07
Add tests for our CLI's `push` command

dvargas92495:
In a future PR, we are going to be adding the `workflows` subcommand, similar to how we recently added it to push. This PR simply adds tests to what we have so far so that we could be explicit about the new commands and breaking changes being introduced on the next one

### DIFF
--- a/ee/vellum_cli/tests/conftest.py
+++ b/ee/vellum_cli/tests/conftest.py
@@ -1,0 +1,40 @@
+import pytest
+import os
+import shutil
+import tempfile
+from uuid import uuid4
+from typing import Any, Callable, Dict, Generator, Tuple
+
+import tomli_w
+
+
+@pytest.fixture
+def mock_module() -> Generator[Tuple[str, str, Callable[[Dict[str, Any]], None]], None, None]:
+    current_dir = os.getcwd()
+    temp_dir = tempfile.mkdtemp()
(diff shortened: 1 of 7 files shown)


src/vellum/workflows/workflows/base.py
```diff
@@ -84,7 +83,7 @@
 class _BaseWorkflowMeta(type):
     def __new__(mcs, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any:
         if "graph" not in dct:
-            dct["graph"] = []
+            dct["graph"] = set()
```
dvargasfuertes, line 86: fixes a bug with serializing a base workflow with no graph attribute defined, noticed during the happy path push test. Added a serializer test to repro

ee/vellum_cli/tests/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+import shutil
+import tempfile
+from uuid import uuid4
+from typing import Any, Callable, Dict, Generator, Tuple
+
+import tomli_w
+
+
+@pytest.fixture
+def mock_module() -> Generator[Tuple[str, str, Callable[[Dict[str, Any]], None]], None, None]:
```
dvargasfuertes, line 12: moved from `test_pull.py`

PR #340, opened 2024-12-07
WIP - Context Wrapper Handler 

marlenc3:
we're allllmost there
there's some weird stuff happening with ports I'm digging into, but would love some overall vibe check -- 
* is this a pattern that looks promising? 
* are there any scary things that we'd like to avoid? 
* anything to chat about? 

- I'mma switch over to getting parent context into nodes in the meanwhile so we're not held up by this :sob:

### DIFF
--- a/src/vellum/workflows/context.py
+++ b/src/vellum/workflows/context.py
@@ -7,6 +7,7 @@ from typing import (
     Any,
     Callable,
     Dict,
+    Generator,
     Iterator,
     List,
     Optional,
@@ -24,6 +25,8 @@ ExecutionContext = Dict[str, Any]
 _CONTEXT_KEY = "_execution_context"
 
 local = threading.local()
+
+_context_lock = threading.Lock()
 
 F = TypeVar("F", bound=Callable[..., Any])
 T = TypeVar("T")
@@ -47,16 +50,18 @@ def get_parent_context() -> ParentContext:
 @contextmanager
 def execution_context(**kwargs: Any) -> Iterator[None]:
     """Context manager for handling execution context."""
-    prev_context = get_execution_context()
+    with _context_lock:
+        prev_context = get_execution_context()
+        new_context = {**prev_context, **kwargs}
+        set_execution_context(new_context)
     try:
-        set_execution_context({**prev_context, **kwargs})
         yield
     finally:
         set_execution_context(prev_context)
 
 
 # Defined set of keywords that should be extracted for the wrapper
-EXECUTION_CONTEXT_ONLY_KEYWORDS: List[str] = ["context", "parent_context", "execution_id"]
+EXECUTION_CONTEXT_ONLY_KEYWORDS: List[str] = ["parent_context"]
 EXECUTION_PASSTHROUGH_KEYWORDS: List[str] = []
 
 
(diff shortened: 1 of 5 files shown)


src/vellum/workflows/runner/runner.py
```diff
@@ -363,7 +353,11 @@ def _run_node_if_ready(
 
             worker_thread = Thread(
                 target=self._run_work_item,
-                kwargs={"node": node, "span_id": node_span_id},
+                kwargs={
+                    "node": node,
+                    "span_id": node_span_id,
+                    "parent_context": current_parent_context,
```
marlenc3, line 359: to pass context to threads, we can rely on kwargs on setting context for us

tests/workflows/basic_parent_context/tests/test_basic_wrapper.py
```diff
@@ -0,0 +1,71 @@
...
+
+def test_basic_wrapper():
+    a_instance = A(uuid4())
+    a_instance.inner_test()
+
+    assert a_instance._context is None
+
+    a_instance.test()
+
+    assert a_instance._context == NodeParentContext(
```
marlenc3, line 46: changed some of the logic here --- need to update

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,113 @@
...
+    try:
+        yield
+    finally:
+        set_execution_context(prev_context)
+
+
+# Defined set of keywords that should be extracted for the wrapper
+EXECUTION_CONTEXT_ONLY_KEYWORDS: List[str] = ["parent_context"]
+EXECUTION_PASSTHROUGH_KEYWORDS: List[str] = []
+
```
marlenc3, line 66: base wrapper --- 
- takes in current wanted state
- does not handle parent nesting for us

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,113 @@
...
+    Type,
+    TypeVar,
+    Union,
+    cast,
+    runtime_checkable,
+)
+
+from vellum.workflows.events.types import ParentContext
+
+ExecutionContext = Dict[str, Any]
```
dvargasfuertes, line 24: [q] this PR essentially replaces https://github.com/vellum-ai/vellum-python-sdks/pull/252 correct?

[nit] If so, let's followup on @siddseethepalli's feedback from there, in particular, turning this into a typed dataclass/pydantic model

src/vellum/workflows/runner/runner.py
```diff
@@ -600,7 +603,8 @@ def _is_terminal_event(self, event: WorkflowEvent) -> bool:
             return event.workflow_definition == self.workflow.__class__
         return False
 
-    def stream(self) -> WorkflowEventStream:
+    def stream(self, **kwargs):
```
dvargasfuertes, line 606: [blocking] do we still need `kwargs` here? looks like we also lost return type annotation here

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,113 @@
...
+    Args:
+        wrapper_args: Positional arguments for the wrapper context
+        wrapper_kwargs: Keyword arguments for the wrapper context
+    """
+
+    def wrapper(fn: F, *_: Any, **__: Any) -> Callable[..., Any]:
+        @wraps(fn)
+        def wrap(*args: Any, **kwds: Any) -> Any:
+            # Extract all wrapper-specific arguments from kwds
+            with _context_lock:
```
dvargasfuertes, line 80: [q] does this lock prevent multiple nodes from executing in parallel?

src/vellum/workflows/runner/runner.py
```diff
@@ -600,7 +603,8 @@ def _is_terminal_event(self, event: WorkflowEvent) -> bool:
...
+        parent = get_parent_context()
```
dvargasfuertes, line 607: [q] This line, plus the fact that we are passing it into the `Thread's` kwargs is what is starting to undermine the decorator for me (and the risk associated with overwriting the method signature).

dvargas92495 (commented):
Would you be open to a pair program sesh tomorrow for this?

I think there could be something that works quite well here, but just want to make sure we work through all the kinks and potential future footguns

EDIT: Just saw that we split out a good chunk of this here which I love: https://github.com/vellum-ai/vellum-python-sdks/pull/344#pullrequestreview-2488370532. Continuing to split out possibly less contentious portions is another alternative to pairing

PR #342, opened 2024-12-08
[SC-5083] Add stable ids for api nodes

NgoHarrison:
This PR adds stable ids. In a previous commit, I forgot to remove a include-fixtures from that set so we weren't running any integration tests. When running customer workflows, I encountered some issues with api node. This PR and the output display one just sets up the nodes to have the right data. In a follow up PR, I will address the whole serialization and test failures that come from that

### DIFF
--- a/ee/codegen/src/generators/nodes/api-node.ts
+++ b/ee/codegen/src/generators/nodes/api-node.ts
@@ -7,7 +7,6 @@ import { ApiNodeContext } from "src/context/node-context/api-node";
 import { NodeInput } from "src/generators";
 import { BaseSingleFileNode } from "src/generators/nodes/bases/single-file-base";
 import { ApiNode as ApiNodeType, ConstantValuePointer } from "src/types/vellum";
-import {OUTPUTS_CLASS_NAME} from "src/constants";
 
 export class ApiNode extends BaseSingleFileNode<ApiNodeType, ApiNodeContext> {
   baseNodeClassName = "APINode";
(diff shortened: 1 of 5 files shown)


ee/codegen/src/generators/nodes/api-node.ts
```diff
@@ -267,17 +268,21 @@ export class ApiNode extends BaseSingleFileNode<ApiNodeType, ApiNodeContext> {
...
                 );
               }
+              const key = new NodeInput({
+                workflowContext: this.workflowContext,
+                nodeInputData: nodeInput,
+              });
 
               return {
-                key: python.TypeInstantiation.str(nodeInput.key),
+                key: key,
```
noanflaherty:
```suggestion
                key,
```

ee/codegen/src/generators/nodes/api-node.ts
```diff
@@ -293,18 +298,22 @@ export class ApiNode extends BaseSingleFileNode<ApiNodeType, ApiNodeContext> {
...
               }
+              const key = new NodeInput({
+                workflowContext: this.workflowContext,
+                nodeInputData: nodeInput,
+              });
 
               return {
-                key: python.TypeInstantiation.str(nodeInput.key),
-                value: python.TypeInstantiation.uuid(nodeInput.id),
+                key: key,
```
noanflaherty:
```suggestion
                key,
```

noanflaherty (approved):
🚀

PR #343, opened 2024-12-08
WIP: fix mypy issues with codegen'd execution nodes

NgoHarrison:
This PR attempts to fix two issues like this below from codegened scripts from code execution nodes:
example/woflow/nodes/map_node/nodes/keywords_search/script.py:16: error: Need type annotation for "unigram_keywords" (hint: "unigram_keywords: dict[<type>, <type>] = ...")  [var-annotated]
example/woflow/nodes/map_node/nodes/keywords_search/script.py:17: error: Need type annotation for "multigram_keywords" (hint: "multigram_keywords: dict[<type>, <type>] = ...")  [var-annotated]

e.g: `unigram_keywords = {}`

AND:

Incompatible types in assignment (expression has type "OutputReference[BaseDescriptor[_OutputType?]]", base class "MapNode" defined the type as "Union[list[Any], BaseDescriptor[list[Any]]]")  [assignment]
`code_inputs = {"arr": IterateEachPage.Outputs.final_output}`

### DIFF
--- a/src/vellum/plugins/vellum_mypy.py
+++ b/src/vellum/plugins/vellum_mypy.py
@@ -4,7 +4,9 @@ from typing import Callable, Dict, List, Optional, Set, Type
 from mypy.nodes import (
     AssignmentStmt,
     CallExpr,
+    Context,
     Decorator,
+    DictExpr,
     MemberExpr,
     NameExpr,
     OverloadedFuncDef,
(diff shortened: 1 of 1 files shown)


NgoHarrison:
@dvargas92495 @noanflaherty 

There are currently two concerns/issues I have:

1. 
CONCERN: This problem can scale really fast as the code I have in place attempts to dynamically have mypy generate typings for dict assignments during the analysis phase (at least I think this what this hook is telling mypy to do). However users can potentially do these kinds of assignments for other things that mypy will not be happy with. Not sure if this is maintainable long term to keep having specific hooks on these errors.

SOLUTION: I am thinking it is easier/a better approach to just have a mypy ignore-errors comment on top of every codegened script file for code execution nodes

2. 
ISSUE: Having said all of that above, if we still want to address these kinds of issues with the plugin, I could use feedback/help debugging it. It currently doesn't work still and I am not sure how to proceed as the mypy debug outputs is not helpful as to why things are failing/not working 😭

src/vellum/plugins/vellum_mypy.py
```diff
@@ -644,6 +657,24 @@ def _resolve_descriptors_in_outputs(self, type_info: SymbolTableNode) -> SymbolT
 
         return new_type_info
 
+    def get_type_analyze_hook(self, fullname: str) -> Optional[Callable[[AttributeContext], MypyType]]:
+        if fullname == "builtins.dict":
+            return self._analyze_mapping_type
+        return None
```
dvargasfuertes, line 663: this could be dangerous bc we are amending _all_ dicts, instead of just ones defined on Nodes

dvargas92495 (commented):
@NgoHarrison can you post a link to a PR in the prototype repo showing where this mypy error is throwing? after that, I'd be able to give better guidance on whether this is a codegen error or a mypy error

dvargas92495:
We should be able to close this now right?

PR #344, opened 2024-12-09
Improvements to Parent Context

marlenc3:
small improvements to parent context -- allowing for parent context to be sent to threads and not effect the global parent context

This one is pretty much ready to go!

### DIFF
--- a/src/vellum/workflows/context.py
+++ b/src/vellum/workflows/context.py
@@ -1,0 +1,42 @@
+from contextlib import contextmanager
+import threading
+from typing import Any, Callable, Dict, Iterator, TypeVar, cast
+
+from vellum.workflows.events.types import ParentContext
+
+ExecutionLocalContext = Dict[str, Any]
+_CONTEXT_KEY = "_execution_context"
+
+local = threading.local()
+
+_context_lock = threading.Lock()
+
+F = TypeVar("F", bound=Callable[..., Any])
+T = TypeVar("T")
+P = TypeVar("P", bound=ParentContext)
+
+
+def get_execution_context() -> ExecutionLocalContext:
+    """Retrieve the current execution context."""
+    return getattr(local, _CONTEXT_KEY, {})
+
+
+def set_execution_context(context: ExecutionLocalContext) -> None:
+    """Set the current execution context."""
+    setattr(local, _CONTEXT_KEY, context)
+
+
+def get_parent_context() -> ParentContext:
+    return cast(ParentContext, get_execution_context().get("parent_context"))
+
+
+@contextmanager
+def execution_context(**kwargs: Any) -> Iterator[None]:
+    """Context manager for handling execution context."""
+    prev_context = get_execution_context()
+    new_context = {**prev_context, **kwargs}
+    try:
+        set_execution_context(new_context)
+        yield
+    finally:
+        set_execution_context(prev_context)
--- a/src/vellum/workflows/context.py
+++ b/src/vellum/workflows/context.py
@@ -1,10 +1,15 @@
 from contextlib import contextmanager
 import threading
-from typing import Any, Callable, Dict, Iterator, TypeVar, cast
+from typing import Any, Callable, Iterator, Optional, TypeVar, cast
 
+from vellum.client.core import UniversalBaseModel
 from vellum.workflows.events.types import ParentContext
 
-ExecutionLocalContext = Dict[str, Any]
+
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/workflows/event_filters.py
```diff
@@ -46,7 +46,10 @@ def root_workflow_event_filter(workflow_definition: Type["BaseWorkflow"], event:
     if event.parent.type != "WORKFLOW":
         return False
 
-    return event.parent.workflow_definition == CodeResourceDefinition.encode(workflow_definition)
```
dvargasfuertes, line 49: [nit] let's override the `__eq__` method in `CodeResourceDefinition` to ensure that `.name`s are equal and `.module`s are equal

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,42 @@
+from contextlib import contextmanager
+import threading
+from typing import Any, Callable, Dict, Iterator, TypeVar, cast
+
+from vellum.workflows.events.types import ParentContext
+
+ExecutionLocalContext = Dict[str, Any]
```
dvargasfuertes: [nit] can we dataclass/pydantic model this?

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,42 @@
...
+_CONTEXT_KEY = "_execution_context"
+
+local = threading.local()
+
+_context_lock = threading.Lock()
+
+F = TypeVar("F", bound=Callable[..., Any])
+T = TypeVar("T")
+P = TypeVar("P", bound=ParentContext)
```
dvargasfuertes: Are these used?

src/vellum/workflows/context.py
```diff
@@ -0,0 +1,42 @@
...
+
+def get_parent_context() -> ParentContext:
+    return cast(ParentContext, get_execution_context().get("parent_context"))
+
+
+@contextmanager
+def execution_context(**kwargs: Any) -> Iterator[None]:
+    """Context manager for handling execution context."""
+    prev_context = get_execution_context()
+    new_context = {**prev_context, **kwargs}
```
dvargasfuertes: [nit] can we make the `kwargs` types explicit? ie, `parent_context: ParentContext`.

[nit] It's still odd to me that `new_context` has the same args as `prev_context`. I would've expected the new context to be constructed and point to prev context, but maybe I'm missing something

EDIT: Let's dismiss the second nit, I understand now that `new_context` and `prev_context` are part of the `ExecutionContext`, _not_ `ParentContext`

src/vellum/workflows/runner/runner.py
```diff
@@ -504,9 +496,16 @@ def _stream(self) -> None:
...
+            span_id=self._initial_state.meta.span_id,
+            workflow_definition=self.workflow.__class__,
+            parent=self._parent_context,
+            type="WORKFLOW",
+        )
         for node_cls in self._entrypoints:
             try:
-                self._run_node_if_ready(self._initial_state, node_cls)
+                with execution_context(parent_context=current_parent):
+                    self._run_node_if_ready(self._initial_state, node_cls)
```
dvargasfuertes, line 508: [nit] could see us doing the same strategy as `_context_run_work_item` to avoid all of the repeated `with .. as` statements

src/vellum/workflows/runner/runner.py
```diff
@@ -125,7 +126,7 @@ def __init__(
 
         self._active_nodes_by_execution_id: Dict[UUID, BaseNode[StateType]] = {}
         self._cancel_signal = cancel_signal
-        self._parent_context = parent_context
+        self._parent_context = get_parent_context() or parent_context
```
dvargasfuertes, line 129: [q] could we remove this statement?

dvargas92495 (approved):
[q] do we already have a test showing that these changes preserve preexisting behavior?

marlenc3:
> [q] do we already have a test showing that these changes preserve preexisting behavior?

@dvargas92495  we do! there's a test for parent context inside of the basic_parent_context directory that has pre-existing behavior covered

src/vellum/workflows/context.py
marlenc3: I'll see about adding some typing to make it easier on the eyes!

kirawoods (commented):
✅

dvargas92495 (approved):

PR #345, opened 2024-12-09
Subworkflow Parent Context

marlenc3:
similar issue wth prompts --- the sdk isn't reflecting the API (thus the failures) 

but this should pass forward the subworkflow context! 
this is pretty much "ready" just need to make sure to test all the monitoring PRs end to end


marlenc3:
there is a parallelism test that's failing 🤔  but not locally? going to see what that's about in the morning

src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
```diff
@@ -24,13 +25,14 @@ class InlineSubworkflowNode(BaseSubworkflowNode[StateType], Generic[StateType, W
...
-            parent_state=self.state,
-            context=self._context,
-        )
-        subworkflow_stream = subworkflow.stream(
-            inputs=self._compile_subworkflow_inputs(),
-        )
+        with execution_context(parent_context=get_parent_context() or self._context.parent_context):
+            subworkflow = self.subworkflow(
+                parent_state=self.state,
+                context=self._context,
```
dvargasfuertes: [blocking] same feedback as here: https://github.com/vellum-ai/vellum-python-sdks/pull/346#discussion_r1875762073

src/vellum/workflows/nodes/displayable/subworkflow_deployment_node/node.py
```diff
@@ -98,7 +101,9 @@ def run(self) -> Iterator[BaseOutput]:
             event_types=["WORKFLOW"],
             metadata=self.metadata,
             request_options=self.request_options,
+            execution_context={"parent_context": parent_context},
         )
+        # for some reason execution context isn't showing as an option? ^ failing mypy
```
dvargasfuertes, line 110: [blocking] this is expected since we purposefully omitted it from showing up in our api clients.

instead, we'll need to fill in `request_options.additional_body_parameters` to include the extra body parameter

marlenc3: bless 💯

src/vellum/workflows/nodes/displayable/subworkflow_deployment_node/node.py
```diff
@@ -105,41 +125,112 @@ def run(self) -> Iterator[BaseOutput]:
...
+                    WorkflowExecutionInitiatedEvent(
+                        span_id=event.execution_id,
+                        trace_id=event.run_id,
+                        name="workflow.execution.initiated",
+                        body=WorkflowExecutionInitiatedBody(
+                            inputs=BaseInputs(
+                                **{value.name: cast(value.type, value.value) for value in event.data.inputs}
+                            ),
+                            # We need the workflow definition, or we make this an optional field
+                            workflow_definition=WorkflowType.__class__,
```
marlenc3: @dvargas92495 last piece here, besides test, is what to do about workflow_definition and parent fields

I think map should be fine - but linking this to a WIP PR here: https://github.com/vellum-ai/vellum/pull/6683

other option is asking vembda what the resource is -- or make a generic workflowbase that is a subclass of the base class

dvargasfuertes: this won't really be tenable here. we actually don't need the deployment events to be emitted back out here bc it will already have its own emitter through vellum's execute_workflow api.

We only need to emit nested events for executions that are executed fully within vembda, like inline subworkflow and map nodes

dvargas92495 (approved):

src/vellum/client/core/client_wrapper.py
```diff
@@ -1,10 +1,11 @@
 # This file was auto-generated by Fern from our API Definition.
 
-from ..environment import VellumEnvironment
```
dvargasfuertes, line 3: [nit] in the future, let's try to revert any changes made to files in `src/vellum/client/...`

dvargas92495 (commented):
✅

PR #346, opened 2024-12-09
Map Node Context

marlenc3:
All that's needed for map context -- we pass forward the parent context to the workflow runner.
now, the inner events will have the updated context, but doesnt seem like we stream events outside of map once they're done, just the outputs? but 🤷  

will pass it forward


src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -53,7 +55,15 @@ def run(self) -> Outputs:
         fulfilled_iterations: List[bool] = []
         for index, item in enumerate(self.items):
             fulfilled_iterations.append(False)
-            thread = Thread(target=self._run_subworkflow, kwargs={"item": item, "index": index})
+            parent_context = get_parent_context() or self._context.parent_context
```
dvargasfuertes, line 60: [oos] If this is `get_execution_context().parent_context`, we could start to move the other items in `self._context` into the execution context and remove the `_context` param from the base node init

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -88,6 +98,13 @@ def run(self) -> Outputs:
...
 
+    def _context_run_subworkflow(
+        self, *, item: MapNodeItemType, index: int, parent_context: Optional[ParentContext] = None
+    ) -> None:
+        parent_context = parent_context or self._context.parent_context
+        with execution_context(parent_context=parent_context):
+            self._run_subworkflow(item=item, index=index)
+
     def _run_subworkflow(self, *, item: MapNodeItemType, index: int) -> None:
         subworkflow = self.subworkflow(parent_state=self.state, context=self._context)
```
dvargasfuertes: [blocking] we'll need to do a change similar to [this file](https://github.com/vellum-ai/vellum-python-sdks/pull/294/files#diff-b1ecb01085162e6bb71295fa2b7372eb706f77bb97f45cae52809f3faec67f6e) in order for the inner subworkflow events to pipe through the api. I suggest we do so that we can make assertions about those events' parent context in the test

marlenc3: I was wondering how to get those events up, will add !

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -85,12 +101,27 @@ def run(self) -> Outputs:
...
+        self, *, item: MapNodeItemType, index: int, parent_context: Optional[ParentContext] = None
+    ) -> None:
+        parent_context = parent_context or self._context.parent_context
+        with execution_context(parent_context=parent_context):
+            self._run_subworkflow(item=item, index=index)
+
     def _run_subworkflow(self, *, item: MapNodeItemType, index: int) -> None:
-        subworkflow = self.subworkflow(parent_state=self.state, context=self._context)
-        events = subworkflow.stream(inputs=self.SubworkflowInputs(index=index, item=item, all_items=self.items))
+        context = WorkflowContext(
```
marlenc3: 🤔  going to come back --- want to get the parent emitter into execution context as well cuz this was fun to debug

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -85,12 +101,27 @@ def run(self) -> Outputs:
...
+            _vellum_client=self._context._vellum_client, _parent_context=self._context.parent_context
+        )
+        if self._context._event_queue:
+            context._register_event_queue(self._context._event_queue)
```
dvargasfuertes: I wouldn't expect to need this, since `self.subworkflow` will have its own event queueing.

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -85,12 +101,27 @@ def run(self) -> Outputs:
...
+    ) -> None:
+        parent_context = parent_context or self._context.parent_context
+        with execution_context(parent_context=parent_context):
+            self._run_subworkflow(item=item, index=index)
+
     def _run_subworkflow(self, *, item: MapNodeItemType, index: int) -> None:
-        subworkflow = self.subworkflow(parent_state=self.state, context=self._context)
-        events = subworkflow.stream(inputs=self.SubworkflowInputs(index=index, item=item, all_items=self.items))
+        context = WorkflowContext(
+            _vellum_client=self._context._vellum_client, _parent_context=self._context.parent_context
```
dvargasfuertes: `_parent_context=self._context.parent_context` is not needed bc of `execution_context(parent_context=parent_context)` right?

dvargas92495 (commented):
✅

PR #347, opened 2024-12-09
Prompt Parent Context

marlenc3:
stab at passing parent context for prompts --- not including adhoc
- for adhoc, there's a stab at implementation, but wasn't sure how useful passing parent context would be, considering we don't really save parent execution ids atm, but can be added in! 

want a sanity check - but this should be all that's needed 
second question : the API isn't updating to use the new execution_context that is available - anything I can do to make it reflect it?


tests/workflows/basic_prompt_deployment/tests/test_workflow.py
```diff
@@ -156,3 +169,11 @@ def generate_prompt_events(*args: Any, **kwargs: Any) -> Iterator[ExecutePromptE
         "results": expected_outputs,
         "text": "It was hot",
     }
+
+    assert node_events[0].name == "node.execution.initiated"
+    assert (
+        cast(WorkflowParentContext, node_events[0].parent).workflow_definition.model_dump()
```
dvargasfuertes, line 180: [nit] if we `assert node_events[0].parent` and `node_events[0].parent.type == "WORKFLOW"`, we don't need the cast

tests/workflows/basic_prompt_deployment/tests/test_workflow.py
```diff
@@ -156,3 +169,11 @@ def generate_prompt_events(*args: Any, **kwargs: Any) -> Iterator[ExecutePromptE
...
+        == WorkflowParentContext(
```
dvargasfuertes, line 181: [nit] instead of model_dump's, I'm still bullish on overriding the `__eq__` method on CodeResourceDefinition

dvargas92495 (approved):
> the API isn't updating to use the new execution_context that is available - anything I can do to make it reflect it?

Yup, now that the sdk will be sending it, we can make sure that both execute prompt APIs are using it to set parent execution id's approproately.

Then, we'll work this week to stand up APIs that will query clickhouse by parent execution id

dvargas92495 (approved):

PR #348, opened 2024-12-09
Setup test and fix for pulling by id error.

dvargas92495:
`vellum pull` had a bug where pulling by id while there was a _different_ workflow configured in the `pyproject.toml` would prefer the latter. We should always prefer cli args over config files


noanflaherty (commented):
✅

PR #355, opened 2024-12-09
[SC-5766] Fix api node serialization and tests

NgoHarrison:


ee/vellum_ee/workflows/display/nodes/vellum/utils.py
NgoHarrison, line 1: I had to make these changes because the json for workspace secret pointers can have this format:
```
{
                  "type": "WORKSPACE_SECRET",
                  "data": {
                    "type": "STRING",
                    "workspace_secret_id": null
                  }
                }
```

Before we just made constant value pointers for this if the value was `None`.

noanflaherty (approved):
Good stuff!

PR #357, opened 2024-12-10
[SC-5780] Fix duplicate variable assignments for workflows that have the same name for terminal nodes

NgoHarrison:
This was discovered during customer workflows pulling. It would pull and codegen workflows like the one below:
```
class Workflow(BaseWorkflow[Inputs, BaseState]):
    graph = CampaignStrategy >> {Summary, CampaignTactics}

        final_output_11 = FinalOutput11.Outputs.value
        final_output_9 = FinalOutput9.Outputs.value
        final_output = FinalOutput.Outputs.value
        final_output_10 = Tactics.Outputs.value
        final_output = Summary.Outputs.value
```

        
`final_output` has two assignments which technically isn't incorrect in python standards but when looking at the json, it has the following:

```
{
        "id": "05531472-903d-4c03-bafa-08510a80ff27",
        "type": "TERMINAL",
        "data": {
          "label": "Summary",
          "name": "final-output",
          "target_handle_id": "8016e648-7434-424d-b7f6-d324f88e23ad",
          "output_id": "7ef20c0c-6b68-4109-a990-604c218b8340",
          "output_type": "STRING",
          "node_input_id": "db4d7102-6ba1-49fe-8925-965d95351b56"
        }
```
        
        
and
        
        
```
        {
        "id": "31aa4dce-6c17-4eef-b059-50af89ca3d38",
        "type": "TERMINAL",
        "data": {
          "label": "Final Output",
          "name": "final-output",
          "target_handle_id": "6c489c7c-ca64-474b-86ac-16bab42f52d9",
          "output_id": "d3716282-7fc7-4364-a38b-2141bbe851c4",
          "output_type": "STRING",
          "node_input_id": "17934e2e-8174-4852-9872-8d93591cb983"
        }
```
        
        
Therefore, the workflow is trying to convey two terminal node outputs for this but the name is the one we key on when codegening this field

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -171,6 +171,22 @@ class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
 "
 `;
 
+exports[`Workflow > write > should generate correct code with multiple terminal nodes with the same name 1`] = `
+"from vellum.workflows import BaseWorkflow
+from .inputs import Inputs
+from vellum.workflows.state import BaseState
+from .nodes.final_output import FinalOutput
+
+
+class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
+    graph = None
+
+    class Outputs(BaseWorkflow.Outputs):
+        query = FinalOutput.Outputs.value
+        query_1 = FinalOutput.Outputs.value
+"
+`;
+
 exports[`Workflow > write > should handle edges pointing to non-existent nodes 1`] = `
 "from vellum.workflows import BaseWorkflow
 from .inputs import Inputs
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -87,7 +87,7 @@ class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
     graph = SearchNode
 
     class Outputs(BaseWorkflow.Outputs):
-        query = FinalOutput.Outputs.value
+        query_1 = FinalOutput.Outputs.value
 "
 `;
 
@@ -167,7 +167,7 @@ from .nodes.final_output import FinalOutput
 
 class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
     class Outputs(BaseWorkflow.Outputs):
-        query = FinalOutput.Outputs.value
+        query_1 = FinalOutput.Outputs.value
 "
 `;
 
@@ -182,7 +182,7 @@ class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
     graph = None
 
     class Outputs(BaseWorkflow.Outputs):
-        query = FinalOutput.Outputs.value
+        query_1 = FinalOutput.Outputs.value
         query_2 = FinalOutput.Outputs.value
 "
 `;
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -183,7 +183,7 @@ class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
 
     class Outputs(BaseWorkflow.Outputs):
         query = FinalOutput.Outputs.value
-        query_1 = FinalOutput.Outputs.value
+        query_2 = FinalOutput.Outputs.value
 "
 `;
 
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -28,9 +28,9 @@ export function entrypointNodeDataFactory(): EntrypointNode {
   };
 }
 
-export function terminalNodeDataFactory(): FinalOutputNode {
+export function terminalNodeDataFactory(id?: string): FinalOutputNode {
   return {
-    id: "terminal-node-1",
+    id: id ? id : "terminal-node-1",
     type: WorkflowNodeType.TERMINAL,
     inputs: [
       {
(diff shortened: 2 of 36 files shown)


ee/codegen/src/context/workflow-output-context.ts
```diff
@@ -23,6 +30,27 @@ export class WorkflowOutputContext {
...
+      const originalName = this.terminalNodeData.data.name;
+      let uniqueName = originalName;
+
+      while (names.has(uniqueName)) {
+        uniqueName = `${originalName}-${counter}`;
+        counter++;
+      }
+      this.workflowContext.addOutputName(this.terminalNodeData.id, uniqueName);
+      return uniqueName;
+    }
```
NgoHarrison: This is where the changes happen. Would love feedback on this approach as this is kind of hacky and doesn't really really hit the elegant approach

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -76,6 +76,8 @@ export class WorkflowContext {
 
   public readonly workflowRawEdges: WorkflowEdge[];
 
+  public readonly outputNamesById: Map<string, string>;
+
```
NgoHarrison, line 82: Just left this global tracker here for the output names we use for workflow outputs. Don't know if there is a better spot to put it

ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -28,9 +28,9 @@ export function entrypointNodeDataFactory(): EntrypointNode {
   };
 }
 
-export function terminalNodeDataFactory(): FinalOutputNode {
+export function terminalNodeDataFactory(id?: string): FinalOutputNode {
```
noanflaherty: nit: let's make this an optional set of kwargs from the get-go. Note: My syntax below might be wrong.

```suggestion
export function terminalNodeDataFactory({id}: {id?: string} = {}): FinalOutputNode {
```

ee/codegen/src/context/workflow-context/workflow-context.ts
noanflaherty: This makes sense to me! A comment above could help indicate what it's meant to be used for.

ee/codegen/src/generators/workflow-output.ts
```diff
@@ -12,19 +12,22 @@ export declare namespace WorkflowOutput {
...
     workflowOutputContext: WorkflowOutputContext;
+    outputNamesById: Map<string, string>;
   }
 }
 
 export class WorkflowOutput extends AstNode {
   private workflowContext: WorkflowContext;
   private workflowOutputContext: WorkflowOutputContext;
   private workflowOutput: Field;
+  private outputNamesById: Map<string, string>;
```
noanflaherty: We shouldn't need to pass/duplicate this here, right? We have the `workflowContext` above and can access this same info through that?

ee/codegen/src/context/workflow-output-context.ts
```diff
@@ -23,6 +30,27 @@ export class WorkflowOutputContext {
...
+      const names = new Set(this.workflowContext.outputNamesById.values());
+
+      const originalName = this.terminalNodeData.data.name;
+      let uniqueName = originalName;
+
+      while (names.has(uniqueName)) {
+        uniqueName = `${originalName}-${counter}`;
+        counter++;
+      }
+      this.workflowContext.addOutputName(this.terminalNodeData.id, uniqueName);
```
noanflaherty: [3/5] I wouldn't expect us to mutate `workflowContext` as part of this context's constructor. aka, We shouldn't call `addOutputName` here.

Instead, we can do this as part of `addWorkflowOutputContext` on the Workflow Context. Basically, as part of registering an output context with a workflow context, we add its output name to the registry. Wdyt?

Instead, it'd be

ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -171,6 +171,22 @@ class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
...
+from vellum.workflows.state import BaseState
+from .nodes.final_output import FinalOutput
+
+
+class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
+    graph = None
+
+    class Outputs(BaseWorkflow.Outputs):
+        query = FinalOutput.Outputs.value
+        query_1 = FinalOutput.Outputs.value
```
noanflaherty: nit: I'd expect this to be `query_1` and `query_2`.

noanflaherty (approved):
Overall approach seems sound! It makes sense to track this on the WorkflowContext. A few suggestions on how we might clean up the code.

ee/codegen/src/context/workflow-output-context.ts
NgoHarrison: Just updated it

ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -183,7 +183,7 @@ class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
 
     class Outputs(BaseWorkflow.Outputs):
         query = FinalOutput.Outputs.value
-        query_1 = FinalOutput.Outputs.value
+        query_2 = FinalOutput.Outputs.value
```
noanflaherty, line 186: The first one should be `query_1`, yeah?

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -263,4 +266,28 @@ export class WorkflowContext {
   public addOutputName(terminalNodeId: string, outputName: string): void {
     this.outputNamesById.set(terminalNodeId, outputName);
   }
+
+  public generateUniqueFinalOutputName(
```
noanflaherty: nit: this can be made private

noanflaherty (approved):

ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
NgoHarrison: Yes updated

PR #359, opened 2024-12-10
Setup Test Case and implementation fix for adding an edge to a non-terminal node within a set

dvargas92495:
```
graph = {
    A >> B >> C,
    D
} >> E >> F
```

The tricky thing about this edge case is that near the end, we have this graph:

```
graph = {
    A >> B >> C,
    D >> E >> F
}
```

When this graph then comes across the `C >> E` edge, it fails to locate `E` on the bottom to simplify, so the graph stays as two edges instead of simplifying into the merge node `E`. This PR tackles this problem, while starting to take some of our helper methods out of the main function and out into the GraphAttribute class


noanflaherty (approved):

kirawoods (commented):
✅

PR #361, opened 2024-12-10
[SC-5790] Handle null workflow input variable keys

NgoHarrison:


ee/codegen/src/__test__/__snapshots__/inputs.test.ts.snap
```diff
@@ -45,6 +45,26 @@ class Inputs(BaseInputs):
 "
 `;
 
+exports[`Inputs > write > should generate correct code when Inputs has variables with duplicate keys 1`] = `
+"from vellum.workflows.inputs import BaseInputs
+
+
+class Inputs(BaseInputs):
+    key_1: str
+    key_1_01: str
```
NgoHarrison: Some notes here on how this differs from how we handled output variable naming (largely influenced by my personal bias):

1. I decided to keep the original naming of the variable because I think a user would be a little more confused if we slightly changed the name of the input variable since this is more defined upfront from the user and only start incrementing after the original. For example having:

original:
```
var = foo
var = foo
```
current PR renaming impl:
```
var = foo
var_01 = foo
```
instead of:
```
var_1 = foo
var_2 = foo 
```
2. I also decided to appened the strings with `0n` where n is the count since I personally dislike seeing variables named `var_1_1` or `var_1_2`

I am totally okay with changing this to match final output naming pattern

noanflaherty: A few thoughts:
1. Thanks for championing your personal preference! Given that we're building for developers, hearing the opinions of ourselves as developers is awesome
2. I'm down for something like this, but think that we should be consistent between inputs/outputs one way or another. If we introduce a new convention here (which I'm cool with), we should update the convention for outputs to match. My rationale is that Final Output Nodes do in fact have you manually enter a `name` field so I argue it is also `more defined upfront from the user`
3. FWIW I think it's pretty rare for a customer to define a legit variable named like `var_1`, `var_2`, so the case of `var_1_1` seems even more rare. Given this, I'd counter and advocate for `_n` suffix instead of `_0n` suffix, but am 2/5 and down to roll with this if you feel more strongly.

noanflaherty (approved):

ee/codegen/src/__test__/__snapshots__/inputs.test.ts.snap
NgoHarrison: Because of point 3, I will move back to the original convention of _n

PR #363, opened 2024-12-10
Include sdk version in vellum push

m-abboud:
I thought about putting this in workflow_display but in the actual push call seemed more appropriate since I think push will also grab the container image from the toml in the future

### DIFF
--- a/ee/vellum_cli/push.py
+++ b/ee/vellum_cli/push.py
@@ -3,6 +3,7 @@ import json
 import os
 import sys
 import tarfile
+from importlib import metadata
 from uuid import UUID
 from typing import List, Optional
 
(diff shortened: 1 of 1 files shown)


ee/vellum_cli/push.py
```diff
@@ -50,6 +51,9 @@ def push_command(
     workflow = BaseWorkflow.load_from_module(workflow_config.module)
     workflow_display = get_workflow_display(base_display_class=VellumWorkflowDisplay, workflow_class=workflow)
     exec_config = workflow_display.serialize()
+    exec_config["runner_config"] = {
+        "sdk_version": metadata.version("vellum-ai"),
```
dvargasfuertes, line 55: [blocking] could we add the image params from the default runner config for now? as it stands the runner config will save without it: https://github.com/vellum-ai/vellum/blob/main/django/app/api/workflows/utils.py#L72-L75

m-abboud: I put the defaults in the RunnerConfigSerializer. And. I figured the defaults should be set server side for figuring out which sdk version to match to which image in the future.

dvargas92495 (approved):

PR #365, opened 2024-12-11
[Codegen] De-Duplicate Node Module Names

noanflaherty:
Today, if two modules have the same label, we'll generate duplicative module paths. This is bad because we'll overwrite one-another's files.

This PR fixes it such that we generate unique file names for every node.

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -195,6 +195,7 @@ exports[`Workflow > write > should generate correct code with multiple terminal 
 from .inputs import Inputs
 from vellum.workflows.state import BaseState
 from .nodes.final_output import FinalOutput
+from .nodes.final_output_1 import FinalOutput
 
 
 class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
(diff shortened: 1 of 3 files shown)


ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -195,6 +195,7 @@ exports[`Workflow > write > should generate correct code with multiple terminal
 from .inputs import Inputs
 from vellum.workflows.state import BaseState
 from .nodes.final_output import FinalOutput
+from .nodes.final_output_1 import FinalOutput
```
noanflaherty, line 198: Here you can see we generate unique module names, which is the goal of this PR.

We do still however generate duplicative class names without import aliases. This is a separate bug and will be fixed after this fern PR is merged and published: https://github.com/fern-api/fern/pull/5377

noanflaherty: I'll also add that this is an opinionated naming convention and I'm open to feedback/alternative ideas.

Basically the first time we encounter a name, we use it. Thereafter, we append with `_(idx+1)`.

dvargasfuertes: [super oos] do we anticipate allowing the user to specify class names for a node in the UI? could be an advanced feature way of addressing this issue long term.

Doesn't answer for the fact that even in that world, we need _some_ default behavior for if multiple nodes have the same class name specified anyway in its definition, which will be common early on

dvargas92495 (approved):

NgoHarrison (approved):

kirawoods (approved):

PR #366, opened 2024-12-11
[SC-5655] Fast follow - fix logic for required and default values

NgoHarrison:
This fixes the input variable changes I did with required and false.

More detail in the comment below but this is one of a couple of fast follows to address the changes I made before. TLDR: the failures in main were not caught by tests because they were hidden. There were 4 failing tests in main: search node, api node, map node, and subworkflow node. This one fixes search node. Api node works fine as is and map node/subworkflow node are the same error that relate to my duplicate output name handling PR

### DIFF
--- a/ee/codegen_integration/conftest.py
+++ b/ee/codegen_integration/conftest.py
@@ -29,8 +29,14 @@ def _get_fixture_paths(root: str) -> Tuple[str, str]:
 _fixture_paths = _get_fixtures(
     # TODO: Remove exclusions on all of these fixtures
     # https://app.shortcut.com/vellum/story/4649/remove-fixture-exclusions-for-serialization
-    exclude_fixtures={"simple_merge_node"},
-    include_fixtures={"simple_error_node"},
+    exclude_fixtures={
+        "simple_merge_node",
+        "faa_q_and_a_bot",
+        # TODO: Remove the bottom three in fast follows
+        "simple_inline_subworkflow_node",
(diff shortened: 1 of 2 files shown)


ee/codegen_integration/conftest.py
```diff
@@ -29,8 +29,14 @@ def _get_fixture_paths(root: str) -> Tuple[str, str]:
...
     # https://app.shortcut.com/vellum/story/4649/remove-fixture-exclusions-for-serialization
-    exclude_fixtures={"simple_merge_node"},
-    include_fixtures={"simple_error_node"},
+    exclude_fixtures={
+        "simple_merge_node",
+        "faa_q_and_a_bot",
+        # TODO: Remove the bottom three in fast follows
+        "simple_inline_subworkflow_node",
+        "simple_map_node",
+        "simple_api_node",
```
NgoHarrison, line 38: I will be removing these exclusions in fast follows. The reason main was failing was because in the api node PR here:
https://github.com/vellum-ai/vellum-python-sdks/pull/355

I removed the line `include_fixture="simple_error_node"`. This worked fine since that PR at the time did not have the other changes. However in the subsequent PRs of input variable using required/default and the PRs of handling duplicate output names, those did not have that include fixture line removed. Therefore, test failures were being masked. However once those got merged to main, then the tests were no longer being masked

noanflaherty (approved):
Thanks for getting to the bottom of it!

PR #370, opened 2024-12-12
[Codegen] Prepare for Returning Mapping Between Node ID & Node Definition

noanflaherty:
Codegen service will need to return not only the generated files, but also a mapping between Node IDs and Node Definitions.

This PR adds everything needed for codegen service to be able to do this. Specifically:
1. Make `workflowContext` publicly accessible
2. Move info about node base classes to node context and out of the ast node
3. Create a helper on node context to get the definition

I'll get a PR up in codegen service next that makes use of a new release of the package.

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/templating-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/templating-node.test.ts.snap
@@ -1,5 +1,27 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`TemplatingNode > basic > getNodeDefinition 1`] = `
+{
+  "bases": [
+    {
+      "module": [
+        "vellum",
+        "workflows",
+        "nodes",
+        "displayable",
+      ],
+      "name": "TemplatingNode",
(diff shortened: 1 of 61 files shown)


NgoHarrison (commented):
✅

ee/codegen/src/__test__/nodes/__snapshots__/templating-node.test.ts.snap
```diff
@@ -1,5 +1,27 @@
...
+{
+  "bases": [
+    {
+      "module": [
+        "vellum",
+        "workflows",
+        "nodes",
+        "displayable",
+      ],
+      "name": "TemplatingNode",
```
dvargasfuertes, line 13: [blocking] a few of these nodes, including TemplatingNode, are actually from the `core` module. It's important that we use the exact module where it's defined bc during event serialization, we're just going to be going off the `Node.__class__` to get name and module

dvargas92495 (commented):
✅ pending `core` feedback, which I think we'll need to iterate on before we could integrate

kirawoods (commented):
✅

PR #372, opened 2024-12-12
Use core module definitions for core nodes

dvargas92495:
[These](https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/nodes/displayable/__init__.py#L1-L4) four nodes will actually emit `module` paths in their events with `core` instead of `displayable`, bc that's where they actually live. This means there's a difference between the _import module_ and the _definition module_. This PR attempts to address this disparity

### DIFF
--- a/ee/codegen/src/context/node-context/base.ts
+++ b/ee/codegen/src/context/node-context/base.ts
@@ -29,6 +29,7 @@ export abstract class BaseNodeContext<T extends WorkflowDataNode> {
   public abstract readonly baseNodeClassName: string;
   public abstract readonly baseNodeDisplayClassName: string;
 
+  protected isCore = false;
   public readonly baseNodeClassModulePath: readonly string[];
   public readonly baseNodeDisplayClassModulePath: readonly string[];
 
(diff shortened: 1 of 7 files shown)


ee/codegen/src/context/node-context/base.ts
```diff
@@ -96,7 +97,12 @@ export abstract class BaseNodeContext<T extends WorkflowDataNode> {
       bases: [
         {
           name: this.baseNodeClassName,
-          module: [...this.baseNodeClassModulePath],
+          module: [
+            ...(this.isCore
+              ? this.workflowContext.sdkModulePathNames.CORE_NODES_MODULE_PATH
+              : this.workflowContext.sdkModulePathNames
+                  .DISPLAYABLE_NODES_MODULE_PATH),
+          ],
```
dvargasfuertes, line 105: `this.baseNodeClassModulePath` is used elsewhere for imports of the base class. I imagine we still want _those_ to say displayable, but for node definitions, we need the exact path of where it's defined. Open to other ideas of organizing this

noanflaherty: Seems reasonable. If there's reason to change later, we can

noanflaherty (approved):

kirawoods (commented):
✅

PR #377, opened 2024-12-12
Fix mypy errors when setting TemplatingNode.inputs[key] to a descriptor of an int

dvargas92495:
The first commit setups up a test that shows the mypy error we were seeing in some pulled workflows. The second commit is the one line fix. For more information on covariants and contravariants, see this explanation from cursor:
Screenshot 2024-12-12 at 4 24 09 PM

Based on what I read above, Descriptors are always meant to be a container type that should "go with", which is why I landed on covariant.


marlenc3:
👍

PR #378, opened 2024-12-12
move required and color to WorkflowInputsVellumDisplayOverrides instead of parent WorkflowInputsDisplayOverrides

NgoHarrison:
Stemmed from this:
https://vellum-ai.slack.com/archives/C06P13ABK0W/p1733938703919229


noanflaherty (approved):
Does codegen need to be updated or are we good there?

dvargas92495 (commented):
✅

I believe that this is what codegen was _already_ doing, from the original report

PR #380, opened 2024-12-12
[Codegen] Clean Up & Fix Input Variable Naming & De-Duplication

noanflaherty:
Fixes input name generation and de-duplication, following patterns establish by node name de-duping.

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/inputs.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/inputs.test.ts.snap
@@ -5,7 +5,7 @@ exports[`Inputs > write > should convert input variable names into valid python 
 
 
 class Inputs(BaseInputs):
-    my_input_1: str
+    my_input: str
     my_input_1: float
 "
 `;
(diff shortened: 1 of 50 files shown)


ee/codegen/src/__test__/__snapshots__/inputs.test.ts.snap
```diff
@@ -60,8 +60,8 @@ exports[`Inputs > write > should generate correct code when Inputs has variables
 
 
 class Inputs(BaseInputs):
-    var_1: str
-    var_2: str
+    input_: str
```
NgoHarrison, line 63: [q] is this intentional naming?

noanflaherty: Yeah – I wanted to make the default name "input", but that's a reserved keyword in python, so a common approach is to add a trailing "_"

Maybe `Inputs.input_` is bad though. Open to alternative naming suggestion, but I had a slight preference for this over `var`

NgoHarrison: Ah I see. Yup I like input_ over the other one as well

NgoHarrison:
✅

NgoHarrison (commented):
✅

PR #381, opened 2024-12-12
[Codegen] De-Dupe Output Names & Clean Up Input Variables

noanflaherty:
This repeats the pattern introduced for Input Variables in https://github.com/vellum-ai/vellum-python-sdks/pull/380, but applies it to Output Variables.

It also cleans up some input-variable related code along the way.


NgoHarrison (commented):
✅

PR #388, opened 2024-12-13
Preserve lock file deployments on future vellum pulls

dvargas92495:
If there was a deployment in the vellum lock file, pulls of _other_ workflows would clear it, causing diffs like this: https://github.com/vellum-ai/workflows-as-code-runner-prototype/pull/523/files#r1876411802


noanflaherty (commented):
✅

PR #389, opened 2024-12-13
Setup and solve test cases surrounding use of three nodes

dvargas92495:
Solves both of these test cases, both relevant for deepscribe's graph [generation](https://github.com/vellum-ai/workflows-as-code-runner-prototype/pull/529/files#diff-190954e59737e8c60702ee2c36c5a383d80611a212b71fcaa99f7e8c755335deR19)

`graph = {A, B, C}`

`graph = {A, B, C} >> D`

### DIFF
--- a/ee/codegen/package.json
+++ b/ee/codegen/package.json
@@ -24,6 +24,7 @@
     "test": "vitest --run",
     "prepublishOnly": "npm run build && npm run gar-login",
     "test:update": "vitest --run -u",
+    "test:debug": "vitest --run --inspect-brk --no-file-parallelism",
     "types": "tsc --noEmit",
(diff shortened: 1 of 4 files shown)


ee/codegen/package.json
```diff
@@ -24,6 +24,7 @@
     "test": "vitest --run",
     "prepublishOnly": "npm run build && npm run gar-login",
     "test:update": "vitest --run -u",
+    "test:debug": "vitest --run --inspect-brk --no-file-parallelism",
```
dvargasfuertes, line 27: Helps to debug codegen tests. Will flesh out a README section when approach is stable

noanflaherty (commented):
✅

kirawoods (commented):
✅

PR #390, opened 2024-12-13
[Codegen][SC-5803] Handle Numeric Values at Start of Python Statements

noanflaherty:
If a user uses a number at the start of a node label, input name, or output name, we currently include that number at the start of the generated python statement, resulting in invalid python.

This PR fixes this by cleaning the names that we generate.

### DIFF
--- a/ee/codegen/src/context/node-context/base.ts
+++ b/ee/codegen/src/context/node-context/base.ts
@@ -1,6 +1,7 @@
 import { WorkflowContext } from "src/context";
 import { PortContext } from "src/context/port-context";
 import { WorkflowDataNode, WorkflowNodeDefinition } from "src/types/vellum";
+import { toPythonSafeSnakeCase } from "src/utils/casing";
 import { getNodeId, getNodeLabel } from "src/utils/nodes";
 import {
   getGeneratedNodeDisplayModulePath,
@@ -120,6 +121,6 @@ export abstract class BaseNodeContext<T extends WorkflowDataNode> {
       throw new Error(`Node output name not found for output ID: ${outputId}`);
     }
 
-    return nodeOutputName;
+    return toPythonSafeSnakeCase(nodeOutputName, "output");
   }
 }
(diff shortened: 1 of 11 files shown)


ee/codegen/src/context/node-context/base.ts
```diff
@@ -120,6 +121,6 @@ export abstract class BaseNodeContext<T extends WorkflowDataNode> {
       throw new Error(`Node output name not found for output ID: ${outputId}`);
     }
 
-    return nodeOutputName;
+    return toPythonSafeSnakeCase(nodeOutputName, "output");
```
dvargasfuertes, line 124: oh cool I like this approach, a different prefix per module type

ee/codegen/src/utils/casing.ts
```diff
@@ -64,11 +64,26 @@ export function createPythonClassName(input: string): string {
...
     .replace(/([a-z])([A-Z])/g, "$1_$2") // Insert underscore between lower and upper case
     .replace(/[^a-zA-Z0-9]+/g, "_") // Replace any non-alphanumeric characters with underscore
     .replace(/^_+|_+$/g, "") // Remove any leading/trailing underscores
     .toLowerCase(); // Convert to lowercase
+
+  // Add underscore prefix if cleaned string started with unsafe chars
+  const cleanedSafetyPrefix =
+    safetyPrefix === "_"
+      ? "_"
+      : `${safetyPrefix}${safetyPrefix.endsWith("_") ? "" : "_"}`;
```
dvargasfuertes, line 87: [q] don't we control the safety prefix? do we need to sanitize it in that case?

dvargas92495 (approved):

ee/codegen/src/utils/casing.ts
noanflaherty: Yeah but I do it just to be extra safe, doesn't hurt.

PR #393, opened 2024-12-15
Allow ports to be valid edge targets

dvargas92495:
Follows up on the graph discovered here: https://github.com/vellum-ai/workflows-as-code-runner-prototype/pull/530

Allows Ports to be used as graph targets and set elements

### DIFF
--- a/src/vellum/workflows/graph/graph.py
+++ b/src/vellum/workflows/graph/graph.py
@@ -12,11 +12,13 @@ if TYPE_CHECKING:
 GraphTargetOfSets = Union[
     Set[NodeType],
     Set["Graph"],
-    Set[Union[Type["BaseNode"], "Graph"]],
+    Set["Port"],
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/graph/graph.py
```diff
@@ -53,9 +55,13 @@ def from_set(targets: GraphTargetOfSets) -> "Graph":
                 entrypoints.update(target._entrypoints)
                 edges.update(target._edges)
                 terminals.update(target._terminals)
-            else:
+            elif hasattr(target, "Ports"):
```
dvargasfuertes, line 58: we can't use `isinstance(Port)` or `issubclass(BaseNode)` bc they would introduce circular imports so they remain in dynamic TYPE_CHECKING imports above

marlenc3:
✅

noanflaherty (commented):
✅

PR #395, opened 2024-12-15
Harrison/sc 5150/search node metadata filters

NgoHarrison:


ee/vellum_ee/workflows/display/nodes/vellum/search_node.py
```diff
@@ -150,38 +151,42 @@ def _serialize_logical_expression(
...
+                else uuid4_from_hash(f"{self.node_id}|{hash(path)}")
             )
-            rhs_variable_id = (
-                variable_id_map.rhs.id
-                if variable_id_map and variable_id_map.rhs and variable_id_map.rhs.id
-                else uuid4_from_hash(f"{self.node_id}|{hash(tuple(path))}|rhs")
+            rhs_query_input_id = (
+                self.input_variable_ids_by_logical_id[rhs_variable_id]
+                if self.input_variable_ids_by_logical_id
+                else uuid4_from_hash(f"{self.node_id}|{hash(path)}")
```
NgoHarrison, line 162: Could use some context/feedback here: I am not sure why we needed the `variable_id_map` here? For conditional nodes we needed it cause every rule condition group has an id and then potentially a lhs/rhs to keep track of stable ids. But for logical condition groups, there is no id for it, just ids for the lhs/rhs of the individual logical condition. Therefore, I thought the `lhs_variable_id` and `rhs_variable_id` could be derived from just the lhs/rhs variables that are on the expression.

ee/codegen/src/generators/MetadataFilters.ts
```diff
@@ -0,0 +1,148 @@
...
+  VellumLogicalExpression as VellumLogicalExpressionType,
+} from "src/types/vellum";
+
+export declare namespace MetadataFilters {
+  export interface Args {
+    metadata: VellumLogicalExpressionType;
+  }
+}
+
+export class MetadataFilters extends AstNode {
```
noanflaherty: We should name this to be more specific and ideally, even co-locate with the search node ast node.

```suggestion
export class SearchNodeMetadataFilters extends AstNode {
```

ee/vellum_ee/workflows/display/nodes/vellum/search_node.py
noanflaherty: Maybe we don't need it? From what I can tell on main, `variable_id_map` isn't ever actually non-null.

ee/vellum_ee/workflows/display/nodes/vellum/utils.py
```diff
@@ -93,4 +95,7 @@ def create_pointer(
             )
 
     vellum_variable_value = primitive_to_vellum_value(value)
-    return ConstantValuePointer(type="CONSTANT_VALUE", data=vellum_variable_value)
+    if pointer_type is InputVariablePointer:
+        return InputVariablePointer(type="INPUT_VARIABLE", data=InputVariableData(input_variable_id=value))
+    else:
```
noanflaherty, line 100: [q] Are we going to need to add the other types here?

noanflaherty (approved):
Looks great!

dvargas92495 (approved):

ee/vellum_ee/workflows/display/nodes/vellum/utils.py
NgoHarrison: We definitely can to be safe but I have been slowly adding these types as these edge cases come up. Idk if there are going to be any use for them but I can add them preemptively anyway if you would like in a fast follow

PR #398, opened 2024-12-16
[SC-5827] Update code execution util to find correct script file path

NgoHarrison:
Should fix the issue where deepscribe soap workflow had empty code inputs

### DIFF
--- a/ee/codegen_integration/fixtures/simple_code_execution_node/code/nodes/code_execution_node/__init__.py
+++ b/ee/codegen_integration/fixtures/simple_code_execution_node/code/nodes/code_execution_node/__init__.py
@@ -6,7 +6,7 @@ from ...inputs import Inputs
 
 
 class CodeExecutionNode(BaseCodeExecutionNode[BaseState, str]):
-    filepath = "code_execution_node/script.py"
+    filepath = "./script.py"
     code_inputs = {"arg": Inputs.input}
     output_type = "STRING"
     runtime = "PYTHON_3_11_6"
--- a/ee/codegen_integration/fixtures/simple_code_execution_node/code/nodes/code_execution_node/__init__.py
+++ b/ee/codegen_integration/fixtures/simple_code_execution_node/code/nodes/code_execution_node/__init__.py
@@ -6,7 +6,7 @@ from ...inputs import Inputs
 
 
 class CodeExecutionNode(BaseCodeExecutionNode[BaseState, str]):
-    filepath = "./script.py"
+    filepath = "code_execution_node/script.py"
     code_inputs = {"arg": Inputs.input}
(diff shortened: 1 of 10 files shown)


ee/codegen_integration/fixtures/simple_code_execution_node/code/nodes/code_execution_node/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 
 class CodeExecutionNode(BaseCodeExecutionNode[BaseState, str]):
-    filepath = "./script.py"
+    filepath = "code_execution_node/script.py"
```
dvargasfuertes: hm, as a user, I would expect the former to work. Are we sure that this is a codegen problem and not a serialization problem?

NgoHarrison: It was a serialization problem but I wanted an easy way to know which directory the script lived in. If a user had multiple code execution nodes, I needed to know which node module it lived under since there can be multiple files named script. I could be overcomplicating this thinking tho. I modified codegen to pass in a little bit more info that I could use for serialization which ultimately was the motivation for my codegen change

dvargasfuertes: I suppose I find this line concerning: https://github.com/vellum-ai/vellum-python-sdks/pull/398/files#diff-c5094b531ceb9467a34a39269ca6a891fcf07152776c7f4a7c622849f2dd81ddL67

the script file should almost always be codegened relative to the current node file. push/serialization could make the same assumption. This tells me that we have something busted in the translation from local files to UI data where we lost the script we're trying to push to the UI.

Happy to talk through this more tn/tm morning!

tests/workflows/basic_code_execution_node/try_workflow.py
```diff
@@ -8,7 +8,7 @@
 
 @TryNode.wrap()
 class SimpleCodeExecutionNode(CodeExecutionNode[BaseState, int]):
-    filepath = "/".join(base_module + ["tests", "code.py"])
+    filepath = "tests/code.py"
```
dvargasfuertes:
```suggestion
    filepath = "./tests/code.py"
```

Can we quickly confirm that ^ works too (here and below)

dvargas92495 (approved):

tests/workflows/basic_code_execution_node/try_workflow.py
NgoHarrison: Yeah that works, I will just update the PR anyway with those paths to be more inline with our `./script.py` convention

PR #400, opened 2024-12-16
add more pointer types to display utils helper

NgoHarrison:
Fast follow from this comment:
https://github.com/vellum-ai/vellum-python-sdks/pull/395#discussion_r1887215934

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/utils.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/utils.py
@@ -10,6 +10,8 @@ from vellum_ee.workflows.display.utils.uuids import uuid4_from_hash
 from vellum_ee.workflows.display.utils.vellum import create_node_input_value_pointer_rule, primitive_to_vellum_value
 from vellum_ee.workflows.display.vellum import (
     ConstantValuePointer,
+    ExecutionCounterData,
+    ExecutionCounterPointer,
     InputVariableData,
     InputVariablePointer,
     NodeInput,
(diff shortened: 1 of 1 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/utils.py
```diff
@@ -97,5 +99,11 @@ def create_pointer(
     vellum_variable_value = primitive_to_vellum_value(value)
     if pointer_type is InputVariablePointer:
         return InputVariablePointer(type="INPUT_VARIABLE", data=InputVariableData(input_variable_id=value))
+    elif pointer_type is WorkspaceSecretPointer:
+        return WorkspaceSecretPointer(
+            type="WORKSPACE_SECRET", data=WorkspaceSecretData(type="STRING", workspace_secret_id=value)
+        )
+    elif pointer_type is ExecutionCounterPointer:
+        return ExecutionCounterPointer(type="EXECUTION_COUNTER", data=ExecutionCounterData(node_id=value))
     else:
```
noanflaherty: nit: I suggest looking for usages of `assertUnreachable`. It'd be cool to ensure we're covering all our bases here and throwing exceptions if we come across data we shouldn't see

noanflaherty (approved):

ee/vellum_ee/workflows/display/nodes/vellum/utils.py
NgoHarrison: Is there a python equivalent? I am trying to look for one but don't see it unless you meant for me to create one for our python utils or just do a generic raise error on else

noanflaherty: Oh woops no. But we should probs still raise a `ValueError` in our `else` block if we come across something we don't expect.

PR #401, opened 2024-12-17
[SC-5854] Remove source handle id from error node

NgoHarrison:
First error of headspace workflow pull (this seems like the only issue preventing a pull). Will look at the python files now to see if the codegen looks good

ERROR preventing pull:
error nodes do not have source handle ids


dvargas92495 (approved):
This will be an interesting one to cutover to Base Node - maybe we should update the sdk to say that this node should have no ports

noanflaherty (approved):

PR #402, opened 2024-12-17
Support ports within trynode

dvargas92495:
Ports specified on nodes were not previously carried over to adornments. This would result in runtime errors whenever a user tried to create an edge from an adornment port to another node.

This PR fixes this behavior to unblock the prompt fallback use case

### DIFF
--- a/src/vellum/workflows/nodes/core/try_node/node.py
+++ b/src/vellum/workflows/nodes/core/try_node/node.py
@@ -160,6 +160,7 @@ Message: {event.error.message}""",
                     "__module__": dynamic_module,
                     "on_error_code": _on_error_code,
                     "subworkflow": Subworkflow,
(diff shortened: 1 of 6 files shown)


src/vellum/workflows/nodes/core/try_node/node.py
```diff
@@ -160,6 +160,7 @@ class Outputs(inner_cls.Outputs):  # type: ignore[name-defined]
                     "__module__": dynamic_module,
                     "on_error_code": _on_error_code,
                     "subworkflow": Subworkflow,
+                    "Ports": type("Ports", (TryNode.Ports,), {port.name: port.copy() for port in inner_cls.Ports}),
```
dvargasfuertes, line 163: This is the key line of the PR - we copy over all of the ports from the wrapped node to the adornment

noanflaherty (approved):

PR #403, opened 2024-12-17
Ignore non py files in push

m-abboud:
~~We can't really use the .pyc files since their compiled based on the python version and think the python versions hafta be compatible. Also they were causing the encoding error I ran into. Im going also patch this on django by checking the encoding of a file somehow or something but this raises a question in my head of what if they want to include some non utf-8 data? How would vembda even handle that should we just.~~

I think it might be safe here just to exclude all non .py files instead of excluding specific things but not 100% sure?

Also why arn't we just serializing and running this through the codegen service for consistency?

### DIFF
--- a/ee/vellum_cli/push.py
+++ b/ee/vellum_cli/push.py
@@ -1,8 +1,8 @@
-import re
 from importlib import metadata
 import io
 import json
 import os
+import re
 import sys
 import tarfile
 from uuid import UUID
--- a/ee/vellum_cli/push.py
+++ b/ee/vellum_cli/push.py
@@ -1,3 +1,4 @@
+import re
 from importlib import metadata
 import io
 import json
@@ -19,6 +20,13 @@ from vellum_cli.logger import load_cli_logger
 from vellum_ee.workflows.display.workflows.get_vellum_workflow_display_class import get_workflow_display
 from vellum_ee.workflows.display.workflows.vellum_workflow_display import VellumWorkflowDisplay
 
+_IGNORE_DIRECTORIES = [
+    r".*__pycache__"
+]
+
+_IGNORE_FILES = [
+    r"\.DS_Store"
+]
 
 def push_command(
     module: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


ee/vellum_cli/push.py
```diff
@@ -80,6 +80,9 @@ def push_command(
         module_dir = workflow_config.module.replace(".", os.path.sep)
         for root, _, files in os.walk(module_dir):
             for filename in files:
+                if not filename.endswith(".py"):
+                    continue
```
dvargasfuertes, line 84: [blocking] there may be `.json` or other data files that we would want to let the user include. I think instead, we should _exclude_ the files that were causing problems, which appear to be namely the `.pyc` files.

Also note that in `pull.py` we use a `.ignore` field from the workflow config that we'll eventually want to use here as well

dvargas92495 (commented):
> Also why arn't we just serializing and running this through the codegen service for consistency?

This is something I'm happy to discuss further, possibly in wsdk sesh today. Serialization + codegen will always introduce some amount of risk, so for users that don't intend to use our UI and therefore have no need for serialization, we'd like to be able to support just deploying their artifacts directly and not risk any discrepancies through the serialization / codegen process

ee/vellum_cli/push.py
m-abboud: .json and other data files would currently be unavailable because only python modules are loaded at runtime which is why I decided to just switch it to py only since it seemed to make more sense just to exclude them if they can't currently be used properly and unlikely on future

dvargasfuertes: would the user not be able to do `open("local_data.json")`?

dvargasfuertes: Code execution nodes with typescript scripts is actually another example of ^

m-abboud: Pretty sure that would be going to local disk not the virtual file system for modules

dvargasfuertes: Are these files not available on the local disk? I suspect code exec nodes are busted in vembda currently if that's the case

m-abboud: Uh no anything going into the route is just kept in memory or a virtual file module loader, any files would have to be loaded from the docker image currently. My initial gut reaction to implementing this is a 5🥴 out of 5🥴 since it might involve weird monkeypatching since it needs to be able to access the virtual files and files on the docker image

PR #409, opened 2024-12-17
Setup test for complex nested conditional node test

dvargas92495:
This test case is quite involved and relevant for a customer we are looking to onboard this week. Basically, graph generation breaks down when there's a Conditional Node _within another_ Conditional node, as long as the former is in a set. This PR fixes this case

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -41,6 +41,19 @@ class TestWorkflow(BaseWorkflow):
 "
 `;
 
+exports[`Workflow > write > graph > should be correct for a basic merge node and an additional edge 2`] = `
+"from vellum.workflows import BaseWorkflow
+from .nodes.templating_node import TemplatingNode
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -41,10 +41,12 @@ repos:
...
         types: [python]
 
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.13.2
+  - repo: local
     hooks:
       - id: isort
+        name: isort
+        entry: bash -c "poetry run isort ."
+        language: system
```
dvargasfuertes: If you rebase, this would catch issues from the fern generated files, presumably bc it missed our config

ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -153,6 +175,22 @@ class TestWorkflow(BaseWorkflow):
 "
 `;
 
+exports[`Workflow > write > graph > should be correct for port within set to a set 1`] = `
```
dvargasfuertes, line 178: this was just moved from below, since I forgot to nest this under `graph` in a previous PR

PR #413, opened 2024-12-17
Fix hash of search node query inputs

NgoHarrison:
The path variable is a list so it isn't hashable


dvargas92495 (approved):

dvargas92495:
I would've expected this to fail a test, similar to whatever triggered this in the prototype repo. Could we fast follow with one?

PR #414, opened 2024-12-17
Setup test and implement solution for port within a set pointing to a set

dvargas92495:
```python
graph = {
    A.ports.if >> B,
    A.ports.else >> {C, D}
}
```

What made this case tricky is that when adding an edge to a set (eg `A.ports.else >> D`), our algorithm iterates over all members of the set to see which one to add it to. In this case, both members shared the node `A` so both were valid. We want our algorithm to only add it to one, so we implement a max priority score on which one to choose.

Eventually building up to: https://github.com/vellum-ai/vellum-python-sdks/pull/409/files

### DIFF
--- a/ee/codegen/src/generators/graph-attribute.ts
+++ b/ee/codegen/src/generators/graph-attribute.ts
@@ -236,19 +236,43 @@ export class GraphAttribute extends AstNode {
             };
           }
         } else if (mutableAst.type === "set") {
+          // debugger;
           const newSet = mutableAst.values.map((subAst) => {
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -236,19 +236,43 @@ export class GraphAttribute extends AstNode {
             };
           }
         } else if (mutableAst.type === "set") {
+          // debugger;
```
dvargasfuertes, line 239: Agh. I'll clean this up on the next one.

PR #418, opened 2024-12-17
add tests for search node

NgoHarrison:
We didn't have search node tests before. This adds some tests with an example search node workflow that has search node metadata filters


dvargas92495 (approved):
I need to do another rev on graph gen - so let's hold off on publishing 0.11.8 until I get that in

PR #427, opened 2024-12-18
Setup test and implement solution for branches merging from sets

dvargas92495:
See this slack thread for full context: https://vellum-ai.slack.com/archives/C06P13ABK0W/p1734477299960769

Codegen here implements Method 4 from that thread, for a few reasons over the alternatives:
- Intermediate variables, while more readable for the end developer, will be hard for codegen to reason about. This may be a space to be explored during workflow file merging.
- Our graph generation algorithm assumes adding a single node to the graph at a time, which is why we don't currently generate adding the edge to all the possible valuable sources. It also has the benefit that the chain of 100 nodes is only defined once

### DIFF
--- a/ee/codegen/src/generators/graph-attribute.ts
+++ b/ee/codegen/src/generators/graph-attribute.ts
@@ -400,9 +400,6 @@ export class GraphAttribute extends AstNode {
         return;
       };
 
-      if (edge.id === "edge-5") {
-        debugger;
-      }
       const newMutableAst = addEdgeToGraph(graphMutableAst, null);
       processedEdges.add(edge);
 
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -690,4 +726,30 @@ export class GraphAttribute extends AstNode {
   public write(writer: Writer): void {
     this.astNode.write(writer);
   }
+
+  private debug(mutableAst: GraphMutableAst): string {
```
dvargasfuertes, line 730: Found this useful while debugging intermediate states

dvargas92495:
Gonna continue to merge to unblock 0.11.8, but still interested on feedback here

noanflaherty (commented):
✅

PR #430, opened 2024-12-18
Add code execution tests and throw node attribute generation error

dvargas92495:
Adds our new error for the following [sentry error](https://vellum.sentry.io/issues/6155986060/?referrer=slack&notification_uuid=c2030f02-2c7e-454e-9738-8f3f28076f18&alert_rule_id=15639644&alert_type=issue)

Screenshot 2024-12-18 at 8 30 06 AM

In a future PR, we will be catching NodeAttributeGenerationError's and raising 4xx instead (from codegen).

Also used this as an opportunity to start the code execution node generation test in general

### DIFF
--- a/ee/codegen/src/generators/nodes/code-execution-node.ts
+++ b/ee/codegen/src/generators/nodes/code-execution-node.ts
@@ -4,6 +4,8 @@ import * as path from "path";
 import { python } from "@fern-api/python-ast";
 import { AstNode } from "@fern-api/python-ast/core/AstNode";
 
+import { BaseNode } from "./bases/base";
+
 import { OUTPUTS_CLASS_NAME } from "src/constants";
 import { CodeExecutionContext } from "src/context/node-context/code-execution-node";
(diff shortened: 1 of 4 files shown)


ee/codegen/src/generators/nodes/code-execution-node.ts
```diff
@@ -17,6 +19,15 @@ export class CodeExecutionNode extends BaseSingleFileNode<
...
 > {
   public declare readonly nodeContext: CodeExecutionContext;
+  private readonly scriptFileContents: string;
+
+  constructor({
+    workflowContext,
+    nodeContext,
+  }: BaseNode.Args<CodeExecutionNodeType, CodeExecutionContext>) {
+    super({ workflowContext, nodeContext });
+    this.scriptFileContents = this.generateScriptFileContents();
```
dvargasfuertes, line 29: Added this here so that the error could be thrown on node construction instead of midway through `persist()`

PR #432, opened 2024-12-18
Support workflows pull by deployment identifiers cli

dvargas92495:
Users could now pull workflows from vellum using the `--workflow-deployment` CLI arg


marlenc3 (approved):

noanflaherty (commented):
[q] Will this set the workflow sandbox id associated with the workflow deployment in the vellum.lock file?

dvargas92495:
Not yet - we need Fern (or our post processing script) to have a way for us to access response headers for multipart download APIs, which is where we would need to store this info

PR #433, opened 2024-12-18
add error output id to inline prompt node

NgoHarrison:


dvargas92495 (approved):
Can we fast follow with a test protecting against this regression?

PR #438, opened 2024-12-18
Fix circular import issue breaking workflows with merge node

dvargas92495:
Importing a workflows with merge node display was crashing - this should resolve and introduces a test to protect against any fast follows

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/base_node_vellum_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_vellum_display.py
@@ -30,6 +30,9 @@ class BaseNodeVellumDisplay(BaseNodeDisplay[NodeType]):
     def get_target_handle_id(self) -> UUID:
         return self._get_node_display_uuid("target_handle_id")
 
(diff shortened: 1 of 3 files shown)


ee/vellum_ee/workflows/display/nodes/base_node_vellum_display.py
```diff
@@ -30,6 +30,9 @@ def get_display_data(self) -> NodeDisplayData:
     def get_target_handle_id(self) -> UUID:
         return self._get_node_display_uuid("target_handle_id")
 
+    def get_target_handle_id_by_source_node_id(self, source_node_id: UUID) -> UUID:
+        return self.get_target_handle_id()
```
noanflaherty, line 34: Can we add a code comment like:

In the vast majority of cases, nodes will only have a single target handle and can be retrieved independently of the source node. However, in rare cases (such as legacy Merge nodes), this method can be overridden to account for the case of retrieving one amongst multiple target handles on a node.

noanflaherty (approved):

PR #442, opened 2024-12-18
add release cutting action and script

dvargas92495:
steps:
- add PyGithub dep
- change trigger to be based on PR label
- change github token auth


dvargas92495:
Gonna close this in favor of storing it in one place in our sdk-generator so that it's published to all of the sdks in the future

PR #444, opened 2024-12-18
fix logic so expressions with no rhs can be generated

NgoHarrison:
Came up from workflows not supporting not null operator for conditional nodes in UI


ee/codegen/src/generators/conditional-node-port.ts
```diff
@@ -163,8 +162,8 @@ export class ConditionalNodePort extends AstNode {
       doesNotContain: "does_not_contain",
       doesNotBeginWith: "does_not_begin_with",
       doesNotEndWith: "does_not_end_with",
-      null: "is_null",
-      notNull: "is_not_null",
+      null: "is_none",
+      notNull: "is_not_none",
```
NgoHarrison: I named these by mistake before, should be these I believe:
https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/descriptors/base.py#L266

dvargasfuertes: [blocking] let's actually keep `is_null` and `is_not_null` as that's more JSON compliant. I'll rename the descriptors now so we could rebase here

dvargasfuertes: Just merged https://github.com/vellum-ai/vellum-python-sdks/pull/447, let's rebase and rename!

PR #445, opened 2024-12-18
Add unit test in codegen for conditional nodes referencing prompt node as an input

NgoHarrison:
Follow up from this comment:
https://github.com/vellum-ai/vellum-python-sdks/pull/433#pullrequestreview-2512352592


src/vellum/client/core/client_wrapper.py
dvargasfuertes, line 1: [nit] can we revert changes to this file? I think it happens due to the rebase

dvargas92495 (approved):

PR #448, opened 2024-12-19
Rename Sdk's VellumError to a new WorkflowError

dvargas92495:
The client autogenerates a `VellumError` from our API. To be able to distinguish between the two, and fix what gets autogenerated to error node by default, we rename the SDK's main error class to `WorkflowError`.

Some opinionated takes on `src/vellum/workflows/errors/types.py` so will block this PR on review


src/vellum/workflows/errors/types.py
```diff
@@ -15,6 +21,42 @@ class VellumErrorCode(Enum):
...
+
+
+def vellum_error_to_workflow_error(error: VellumError) -> WorkflowError:
+    return WorkflowError(
+        message=error.message,
+        code=_VELLUM_ERROR_CODE_TO_WORKFLOW_ERROR_CODE.get(error.code, WorkflowErrorCode.INTERNAL_ERROR),
+    )
+
+
+_WORKFLOW_EVENT_ERROR_CODE_TO_WORKFLOW_ERROR_CODE: Dict[WorkflowExecutionEventErrorCode, WorkflowErrorCode] = {
```
dvargasfuertes, line 47: Special 👀 on this one.

`WorkflowEventError` was our first error object that tied to workflows, so consequently, it's emitted today by the API. I think `WorkflowError` should likely be a strict superset? what do we think?

src/vellum/workflows/errors/types.py
```diff
@@ -15,6 +21,42 @@ class VellumErrorCode(Enum):
...
+    "WORKFLOW_INITIALIZATION": WorkflowErrorCode.INVALID_WORKFLOW,
+    "WORKFLOW_CANCELLED": WorkflowErrorCode.WORKFLOW_CANCELLED,
+    "NODE_EXECUTION_COUNT_LIMIT_REACHED": WorkflowErrorCode.INVALID_STATE,
+    "INTERNAL_SERVER_ERROR": WorkflowErrorCode.INTERNAL_ERROR,
+    "NODE_EXECUTION": WorkflowErrorCode.PROVIDER_ERROR,
```
noanflaherty: This mapping here surprises me. Are all node execution errors due to provider errors? Seems unlikely

noanflaherty (approved):

src/vellum/workflows/errors/types.py
dvargasfuertes: They're not, but was having trouble which to map it to, or to introduce a new one

PR #449, opened 2024-12-19
Always infer NodeDisplay._node from generic parameter

dvargas92495:
Node display classes were previously always accepting a node arg in the __init__ method. This is strictly speaking not necessary - the generic class is sufficient to drive this behavior. 

This PR removes the init arg and infers the underlying `_node` attribute using the generic. This will set us up to have a future PR where display classes will be able to annotate the id directly onto the node class, so that we don't need to import logic from `ee` into our sdk

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -1,10 +1,9 @@
 from functools import cached_property
 import inspect
 from uuid import UUID
-from typing import TYPE_CHECKING, Any, Dict, Generic, Optional, Type, TypeVar, cast, get_args, get_origin
+from typing import TYPE_CHECKING, Any, Dict, ForwardRef, Generic, Optional, Type, TypeVar, cast, get_args, get_origin
 
 from vellum.workflows.nodes.bases.base import BaseNode
-from vellum.workflows.nodes.utils import get_wrapped_node, has_wrapped_node
 from vellum.workflows.ports import Port
 from vellum.workflows.references import OutputReference
 from vellum.workflows.types.core import JsonObject
@@ -27,9 +26,6 @@ class BaseNodeDisplay(Generic[NodeType]):
 
     # Used to store the mapping between node types and their display classes
     _node_display_registry: Dict[Type[NodeType], Type["BaseNodeDisplay"]] = {}
-
-    def __init__(self, node: Type[NodeType]):
-        self._node = node
 
     def serialize(self, display_context: "WorkflowDisplayContext", **kwargs: Any) -> JsonObject:
         raise NotImplementedError(f"Serialization for nodes of type {self._node.__name__} is not supported.")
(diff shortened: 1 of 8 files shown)


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -28,9 +27,6 @@ class BaseNodeDisplay(Generic[NodeType]):
     # Used to store the mapping between node types and their display classes
     _node_display_registry: Dict[Type[NodeType], Type["BaseNodeDisplay"]] = {}
 
-    def __init__(self, node: Type[NodeType]):
-        self._node = node
```
dvargasfuertes, line 32: Main goal of the PR

ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -124,17 +141,13 @@ def _get_explicit_node_display_attr(
     def __init_subclass__(cls, **kwargs: Any) -> None:
         super().__init_subclass__(**kwargs)
 
-        original_base = get_original_base(cls)
-        node_class = get_args(original_base)[0]
-        if isinstance(node_class, TypeVar):
-            bounded_class = node_class.__bound__
-            if inspect.isclass(bounded_class) and issubclass(bounded_class, BaseNode):
-                cls._node_display_registry[bounded_class] = cls
```
dvargasfuertes, line 132: This half was moved up to `infer_node_class`

ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -124,17 +141,5 @@ def _get_explicit_node_display_attr(
...
-        elif issubclass(node_class, BaseNode):
-            if has_wrapped_node(node_class):
-                wrapped_node = get_wrapped_node(node_class)
-                if wrapped_node._is_wrapped_node:
-                    cls._node_display_registry[wrapped_node] = cls
-                    return
```
dvargasfuertes, line 138: we ended up not needing this logic. I know this had something to do with mapping TryNode's adorned nodes to the try node's display, but that always felt risky. I think some other changes below relieve this. Ideally we could keep this simple - the registry is simply a map between the parameterized type and the display class

ee/vellum_ee/workflows/display/nodes/get_node_display_class.py
```diff
@@ -11,9 +12,17 @@ def get_node_display_class(
...
                 base_class, node_class.__bases__[0], node_class if root_node_class is None else root_node_class
             )
+
+            # `base_node_display_class` is always a Generic class, so it's safe to index into it
+            NodeDisplayBaseClass = base_node_display_class[node_class]  # type: ignore[index]
+            NodeDisplayClass = types.new_class(
+                f"{node_class.__name__}Display",
+                bases=(NodeDisplayBaseClass,),
+            )
+            return NodeDisplayClass
```
dvargasfuertes, line 25: For any node classes where the user did not specify a display class, we dynamically generate one that subclasses our out of box (or previously registered) display classes. This maintains the invariant we said above

src/vellum/workflows/nodes/utils.py
```diff
@@ -2,28 +2,24 @@
...
-        if isinstance(subworkflow, NodeReference) and subworkflow.instance:
-            graph = subworkflow.instance.graph
-            if issubclass(graph, BaseNode):
-                return graph
+    wrapped_node = getattr(node, "__wrapped_node__", None)
+    if wrapped_node is None:
+        raise AttributeError("Wrapped node not found")
 
-    raise TypeError("Wrapped subworkflow contains more than one node")
+    return wrapped_node
```
dvargasfuertes, line 16: in a separate pass, I think we could replace all uses of `get_wrapped_node`/`has_wrapped_node` with just this attribute

ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -71,6 +67,23 @@ def get_node_port_display(self, port: Port) -> PortDisplay:
     def get_from_node_display_registry(cls, node_class: Type[NodeType]) -> Type["BaseNodeDisplay"]:
         return cls._node_display_registry[node_class]
 
+    @classmethod
+    def infer_node_class(cls) -> Type[NodeType]:
```
noanflaherty, line 71: [q] Does this work as expected for adorned nodes?

ee/vellum_ee/workflows/display/nodes/get_node_display_class.py
noanflaherty: Neat

noanflaherty (approved):
Love it!

ee/vellum_ee/workflows/display/nodes/base_node_display.py
dvargasfuertes: What would be your expectation?

noanflaherty: Not sure tbh. I suppose that it'd return `TryNode`

dvargasfuertes: yea right now it's meant to simply return the node class that is parameterized

marlenc3 (approved):
🚀

PR #456, opened 2024-12-19
Setup test case for strict param during codegen

dvargas92495:
The plan is to solve this ticket in two stages:
- Stage 1 - pull `--strict`
    - Codegen will now pull as much as possible. Any errors it encounters will be emitted to the console.
    - users could run `vellum workflows pull --strict` to instead fail codegen upon seeing errors immediately instead of trying to complete.
- Stage 2 - push `--check`
    - Serialization by default will fail early if it sees any errors.
    - users could run `vellum workflows push --check` to instead run the full serialization and print any issues to the console
        - We likely should report back any _diffs_ as well, but this will be ticketed separately.

This PR is only the first part of Stage 1. Looking to use this for feedback on the api and overall approach first before continuing

### DIFF
--- a/ee/codegen/src/__test__/project.test.ts
+++ b/ee/codegen/src/__test__/project.test.ts
@@ -13,6 +13,7 @@ import { makeTempDir } from "./helpers/temp-dir";
 
 import { SpyMocks } from "src/__test__/utils/SpyMocks";
 import { WorkflowProjectGenerator } from "src/project";
+import { NodeAttributeGenerationError } from "src/generators/errors";
 
 describe("WorkflowProjectGenerator", () => {
   let tempDir: string;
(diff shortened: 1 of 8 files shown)


ee/codegen/src/__test__/project.test.ts
```diff
@@ -111,5 +112,53 @@ describe("WorkflowProjectGenerator", () => {
...
+        absolutePathToOutputDirectory: tempDir,
+        workflowVersionExecConfigData: displayData,
+        moduleName: "code",
+        vellumApiKey: "<TEST_API_KEY>",
+        strict: true,
+      });
+
+      expect(project.generateCode()).rejects.toThrow(
+        new NodeAttributeGenerationError("test")
+      );
```
noanflaherty: Rather than throwing the very first error that we come across, I'd be more partial to doing what DRF serializers do and accumulating all the errors that would arise and raise an exception at the end with details about all of them.

The reason being I claim it's a better devexp. Rather than running, getting an error, fixing, re-running, running into a new error, rinse and repeat – you can run once, get back info about everything you need to fix, fix them in one go, then re-run.

noanflaherty: Maybe that's what's happening here already?

dvargasfuertes: couple thoughts:
- That's more true for serialization (push) then it is for codegen (pull). Ideally whatever is stored in vellum could always be pulled no matter what. The devx that the user controls and iterates on will be on `push --check`, bc it's their code -> vellum
- Users still have the option of seeing the full list of errors without the strict flag, as shown in the above test case. It just won't fail and instead write `error.log` to the console. So I was originally thinking the strict option was the fail fast alternative

ultimately, only 2/5 though

ee/codegen/src/project.ts
```diff
@@ -349,168 +354,170 @@ ${errors.slice(0, 3).map((err) => {
...
-            case "DEPLOYMENT":
-              node = new SubworkflowDeploymentNode({
-                workflowContext: this.workflowContext,
-                nodeContext: nodeContext as SubworkflowDeploymentNodeContext,
-              });
-              break;
-            default: {
-              assertUnreachable(variant);
+    await Promise.all(
+      nodeIds.map(async (nodeId) => {
```
dvargasfuertes, line 366: This forEach instead of Promise.all(array.map()) was causing all sorts of problems trying to catch errors

dvargas92495:
@noanflaherty @NgoHarrison This PR is ready for review!

NgoHarrison (approved):
✅

PR #461, opened 2024-12-19
Updating Code Definitions w/ Ids

marlenc3:
majority of this update is updating tests -- but should be passing up all ids now


src/vellum/workflows/events/types.py
```diff
@@ -28,6 +28,15 @@ def serialize_type_encoder(obj: type) -> Dict[str, Any]:
     }
```
marlenc3, line 28: Main change is this file 🔥

src/vellum/workflows/events/node.py
```diff
@@ -21,7 +21,7 @@ class _BaseNodeExecutionBody(UniversalBaseModel):
```
marlenc3, line 21: event bodies will also carry their ids -- let me know if this is not something we want

src/vellum/workflows/events/workflow.py
```diff
@@ -27,7 +27,7 @@ class _BaseWorkflowExecutionBody(UniversalBaseModel):
 
     @field_serializer("workflow_definition")
     def serialize_workflow_definition(self, workflow_definition: Type, _info: Any) -> Dict[str, Any]:
-        return serialize_type_encoder(workflow_definition)
+        return serialize_type_encoder_with_id(workflow_definition)
```
marlenc3, line 32: workflow bodies to carry their ids

src/vellum/workflows/events/node.py
dvargasfuertes: this is something we need to replace the translator logic, so def keep

src/vellum/workflows/events/tests/test_event.py
```diff
@@ -73,6 +74,7 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
                 "name": "workflow.execution.initiated",
                 "body": {
                     "workflow_definition": {
+                        "id": str(uuid4_from_hash(MockWorkflow.__qualname__)),
```
dvargasfuertes: [nit] slight pref for setting these two as hardcoded files in the top of the file and using in tests

src/vellum/workflows/events/types.py
```diff
@@ -46,9 +55,17 @@ def encode(obj: type) -> "CodeResourceDefinition":
         return CodeResourceDefinition(**serialize_type_encoder(obj))
 
 
+class VellumCodeResource(CodeResourceDefinition):
+    id: UUID
```
dvargasfuertes: [nit] ah I think _here_, we're safe to add directly to `CodeResourceDefinition`. It's on the vellum side that we'll want to split out into a separate class

dvargas92495 (approved):

src/vellum/workflows/events/types.py
marlenc3: oh bet! that'll be a fast change

PR #467, opened 2024-12-20
Support various other json inputs in our prompt nodes

dvargas92495:
This was a poor devx issue that @awlevin ran into while spinning up extending base node docs. We can now accept any json serializable as a JSON input for a prompt by default.

PR is blocked on a vellum schema change we need to make, and a mypy issue

### DIFF
--- a/src/vellum/workflows/types/core.py
+++ b/src/vellum/workflows/types/core.py
@@ -27,10 +27,7 @@ class VellumSecret:
         self.name = name
 
 
-EntityInputsInterface = Dict[
-    str,
-    Any,
-]
+EntityInputsInterface = Dict[str, Any]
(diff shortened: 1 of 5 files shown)


src/vellum/workflows/types/core.py
```diff
@@ -42,42 +27,7 @@ def __init__(self, name: str):
         self.name = name
 
 
-VellumValuePrimitive = Union[
```
dvargasfuertes, line 45: https://www.loom.com/share/115cf5b5c4da49a183e21c9cb5c9e994

marlenc3 (approved):

dvargas92495:
Merging to include in next release

PR #472, opened 2024-12-20
Fix mypy issues with code execution nodes not inferring outputs results type

NgoHarrison:
Mypy does not like assignments like these:
`code_inputs = {"text", SomeNodeThatExtendsCodeExecutionNode.Outputs.Result}`

### DIFF
--- a/src/vellum/plugins/vellum_mypy.py
+++ b/src/vellum/plugins/vellum_mypy.py
@@ -157,7 +157,6 @@ class VellumMypyPlugin(Plugin):
         elif _is_subclass(
             ctx.cls.info, "vellum.workflows.nodes.displayable.code_execution_node.node.CodeExecutionNode"
         ):
-            self._code_execution_node_class_hook(ctx)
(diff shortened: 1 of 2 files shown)


src/vellum/plugins/vellum_mypy.py
```diff
@@ -172,41 +172,39 @@ def _base_class_hook(self, ctx: ClassDefContext) -> None:
...
-        base_templating_args = templating_node_bases[0].args
-        base_templating_node = templating_node_bases[0].type
-        if not _is_subclass(base_templating_node, "vellum.workflows.nodes.core.templating_node.node.TemplatingNode"):
-            return
+        base_args = node_bases[0].args
+        base_node = node_bases[0].type
 
-        if len(base_templating_args) != 2:
+        if len(base_args) != 2:
             return
```
dvargasfuertes, line 190: oh didn't even realize we were trying to inject code exec node above already!

dvargas92495 (approved):
Nice

PR #473, opened 2024-12-20
is_nil code generation for node outputs referenced in conditional node ports

dvargas92495:
This PR conditionally codegen's is_nil depending on whether it's referencing a node output within legacy conditional nodes.

### DIFF
--- a/ee/codegen/src/generators/conditional-node-port.ts
+++ b/ee/codegen/src/generators/conditional-node-port.ts
@@ -5,6 +5,8 @@ import { AstNode } from "@fern-api/python-ast/core/AstNode";
 import { Writer } from "@fern-api/python-ast/core/Writer";
 import { isNil } from "lodash";
 
+import { NodePortGenerationError } from "./errors";
+
 import { PortContext } from "src/context/port-context";
 import { Expression } from "src/generators/expression";
 import { NodeInput } from "src/generators/node-inputs";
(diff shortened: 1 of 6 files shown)


ee/codegen/src/generators/conditional-node-port.ts
```diff
@@ -172,8 +183,8 @@ export class ConditionalNodePort extends AstNode {
       doesNotContain: "does_not_contain",
       doesNotBeginWith: "does_not_begin_with",
       doesNotEndWith: "does_not_end_with",
-      null: "is_null()",
-      notNull: "is_not_null()",
+      null: isNodeOutput ? "is_nil()" : "is_null()",
+      notNull: isNodeOutput ? "is_not_nil()" : "is_not_null()",
```
dvargasfuertes, line 189: @NgoHarrison I find it odd that these look like methods here. We should probably codegen method expression calls in this case instead of saying that the parens are part of the operator name

ee/vellum_ee/workflows/display/nodes/vellum/conditional_node.py
```diff
@@ -242,9 +246,9 @@ def _convert_descriptor_to_operator(self, descriptor: BaseDescriptor) -> str:
             return "doesNotBeginWith"
         elif isinstance(descriptor, DoesNotEndWithExpression):
             return "doesNotEndWith"
-        elif isinstance(descriptor, IsNullExpression):
+        elif isinstance(descriptor, (IsNullExpression, IsNilExpression, IsUndefinedExpression)):
             return "null"
-        elif isinstance(descriptor, IsNotNullExpression):
+        elif isinstance(descriptor, (IsNotNullExpression, IsNotNilExpression, IsNotUndefinedExpression)):
```
dvargasfuertes, line 251: We likely should do the same special casing here - but am deferring that for now to prioritize getting pull + run going for relevant nodes with `null` operators in Vellum

ee/codegen/src/generators/conditional-node-port.ts
NgoHarrison: Agreed and good reminder for myself. Context: this was a rush from my part because I was trying to fix the original issue in a hurry and because we currently do something hacky to compose the whole port expression due to non-existent fern functionality, I didn't know if/how difficult it would be. But looking back at it now, it probably seems like an easy fix

PR #478, opened 2024-12-20
Support Flag for Code Execution Node Code Repr Override

noanflaherty:
With this, we can override codegen behavior such that Code Execution Nodes are forced to generate their code as either an inlined attribute or a standalone script file.


ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -88,6 +89,11 @@ export class WorkflowContext {
 
   public readonly workflowRawEdges: WorkflowEdge[];
 
+  public readonly codeExecutionNodeCodeRepresentationOverride:
+    | "STANDALONE"
+    | "INLINE"
+    | undefined;
```
dvargasfuertes, line 98: [q] why support `undefined` instead of defaulting to one of the two options?

dvargasfuertes: Ah I see to fallback to `.filepath` attribute

dvargas92495 (approved):

PR #479, opened 2024-12-20
Move conditional node is null function call generation to ast node write

NgoHarrison:
Change based on this feedback:
https://github.com/vellum-ai/vellum-python-sdks/pull/473#discussion_r1894198255


ee/codegen/src/generators/expression.ts
```diff
@@ -28,7 +28,7 @@ export class Expression extends AstNode {
     // TODO: We should ideally perform this using native fern functionality, but it requires being able to create
     //  a Reference object from an existing AstNode, which in turn requires all AstNode's to internally track their
     //  name and modulePath.
-    const rhsExpression = rhs ? `(${rhs.toString()})` : "";
+    const rhsExpression = rhs ? `(${rhs.toString()})` : "()";
     const rawExpression = `${lhs.toString()}.${expression}${rhsExpression}`;
 
     return python.codeBlock(rawExpression);
```
dvargasfuertes, line 34: Slightly better, but I think end state, we should look to remove this `python.codeblock` call to use actual asts. low pri though compared to most other things we have going on

dvargas92495 (approved):

PR #484, opened 2024-12-21
Introduce utilities for easy Sandbox running locally

dvargas92495:
This PR introduces some utilities that make it easier to run your workflow locally. It takes inspiration from our Sandbox UIs to create equivalent primitives locally.

If aligned on this approach, the next move is to introduce a `--include-sandbox` option that will codegen a file that looks somewhat like [this](https://github.com/vellum-ai/workflows-as-code-runner-prototype/blob/main/examples/soap_factory_default/sandbox.py), but using these new utilities instead of it inlined, pulling down the scenarios defined in the workflow sandbox


marlenc3 (approved):

marlenc3:
Love!! 🔥

PR #486, opened 2024-12-22
Implement initial codegen of the pull --include-sandbox param

dvargas92495:
This parameter will pull an initial `sandbox.py` file, that will make it easier for customers and contributors to start running and playing around with their workflow by running `python -m examples.my_workflow_sandbox`. There is no support for passing in an index yet (future PR)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -0,0 +1,17 @@
...
+"from vellum.workflows.sandbox.runner import SandboxRunner
+
+from .inputs import Inputs
+from .workflow import Workflow
+
+if __name__ != "__main__":
+    raise Exception("This file is not meant to be imported")
+
+
+runner = SandboxRunner(workflow=Workflow, inputs=[Inputs(input="foo"), Inputs(input="bar"), Inputs(input="hello")])
```
dvargasfuertes, line 13: Is there a way to get codegen to have each of these `Inputs` references to gen on a new line?

ee/codegen/src/generators/sandbox-file.ts
```diff
@@ -0,0 +1,113 @@
...
+    this.sandboxInputs = sandboxInputs;
+  }
+
+  protected getModulePath(): string[] {
+    return [this.workflowContext.moduleName, "sandbox"];
+  }
+
+  protected getFileStatements(): AstNode[] {
+    const sandboxRunnerField = python.field({
+      name: "runner",
```
dvargasfuertes: I'm using `python.field` as sort of a hack around the lack of a `python.assignment()`

ee/codegen/src/generators/sandbox-file.ts
```diff
@@ -0,0 +1,113 @@
...
+      arguments_: Object.entries(input).map(([key, value]) =>
+        python.methodArgument({
+          name: key,
+          value: this.getWorkflowInputValue(value),
+        })
+      ),
+    });
+  }
+
+  private getWorkflowInputValue(value: unknown): python.TypeInstantiation {
```
dvargasfuertes: deferring special classes (ChatMessage, FunctionCall, etc) to a future PR

src/vellum/workflows/sandbox/runner.py
```diff
@@ -10,12 +10,12 @@
 
 
 class SandboxRunner(Generic[WorkflowType]):
-    def __init__(self, workflow: Type[WorkflowType], dataset: Sequence[BaseInputs]):
-        if not dataset:
-            raise ValueError("Dataset is required to have at least one defined inputs")
+    def __init__(self, workflow: Type[WorkflowType], inputs: Sequence[BaseInputs]):
+        if not inputs:
+            raise ValueError("Inputs are required to have at least one defined inputs")
```
dvargasfuertes, line 15: PR feedback I forgot to address previously

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
noanflaherty: Not today.

ee/codegen/src/generators/sandbox-file.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export declare namespace SandboxFile {
+  interface Args extends BasePersistedFile.Args {
+    workflowFile: WorkflowFile;
+    inputsFile: Inputs;
+    sandboxInputs: Record<string, unknown>[];
+  }
+}
+
+export class SandboxFile extends BasePersistedFile {
```
noanflaherty: [blocking] Can we call this `WorkflowSandbox` everywhere that we currently say `Sandbox`? I want to disambiguate from a future of having prompt sandboxes, too.

ee/codegen/src/generators/sandbox-file.ts
```diff
@@ -0,0 +1,113 @@
...
+  private readonly workflowFile: WorkflowFile;
+  private readonly inputsFile: Inputs;
+  private readonly sandboxInputs: Record<string, unknown>[];
+
+  public constructor({
+    workflowContext,
+    workflowFile,
+    inputsFile,
```
noanflaherty: [blcoking] I'd really really like to avoid passing file ast nodes into another ast node. This makes it impossible to generate files in parallel and starts introducing dependencies between them. 

We should use/extend contexts as needed to achieve the same thing.

ee/codegen/src/generators/sandbox-file.ts
noanflaherty: This is the proper usage of `python.field`.

ee/codegen/src/generators/sandbox-file.ts
```diff
@@ -0,0 +1,113 @@
...
+            name: "workflow",
+            value: python.reference({
+              name: "Workflow",
+              modulePath: this.workflowFile.getModulePath(),
+            }),
+          }),
+          python.methodArgument({
+            name: "inputs",
+            value: python.TypeInstantiation.list(
+              this.sandboxInputs.map((input) => this.getWorkflowInput(input))
```
noanflaherty: We should use `InputVariableContext` for this.

ee/codegen/src/generators/sandbox-file.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+    this.inheritReferences(sandboxRunnerField);
+
+    return [
+      python.codeBlock(`\
+if __name__ != "__main__":
+    raise Exception("This file is not meant to be imported")
+        `),
+      sandboxRunnerField,
+      python.codeBlock("runner.run()"),
```
noanflaherty: You should be able to avoid a codeblock for this by using `python.invokeMethod`.

ee/codegen/src/generators/sandbox-file.ts
```diff
@@ -0,0 +1,113 @@
...
+        ],
+      }),
+    });
+    this.inheritReferences(sandboxRunnerField);
+
+    return [
+      python.codeBlock(`\
+if __name__ != "__main__":
+    raise Exception("This file is not meant to be imported")
+        `),
```
noanflaherty: This is a valid use of a code block (at least today) 👍

ee/codegen/src/generators/sandbox-file.ts
```diff
@@ -0,0 +1,113 @@
...
+    ];
+  }
+
+  private getWorkflowInput(
+    input: Record<string, unknown>
+  ): python.ClassInstantiation {
+    return python.instantiateClass({
+      classReference: python.reference({
+        name: "Inputs",
+        modulePath: this.inputsFile.getModulePath(),
```
noanflaherty: Use context here

ee/codegen/src/generators/sandbox-file.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+    if (Array.isArray(value)) {
+      return python.TypeInstantiation.list(
+        value.map(this.getWorkflowInputValue)
+      );
+    }
+    if (value === null) {
+      return python.TypeInstantiation.none();
+    }
+    return python.TypeInstantiation.dict(JSON.parse(JSON.stringify(value)));
```
noanflaherty: What is this case here?

I feel like we already have helpers/ ast nodes for generating vellum value primitives if that's what you're trying to do here.

ee/codegen/src/project.ts
```diff
@@ -186,14 +192,22 @@ ${errors.slice(0, 3).map((err) => {
...
     // error.log
     await this.generateErrorLogFile().persist();
 
+    // sandbox.py
+    if (this.sandboxInputs) {
+      await this.generateSandboxFile({
+        workflowFile,
+        inputsFile: inputs,
+      }).persist();
+    }
```
noanflaherty: [blocking] We should be generating both of these in parallel in the `Promise.all` above to minimize latency of codegen.

noanflaherty (changes requested):
Gonna request changes on this one given how many things I'd like to see iterated upon prior to merging.

ee/codegen/src/project.ts
dvargasfuertes: Error log I think needs to be after bc it collects the errors of all the above

ee/codegen/src/generators/sandbox-file.ts
dvargasfuertes: Was originally treating `value` as an `any`, but maybe should be treated as a VellumValue and reuse that logic

ee/codegen/src/generators/sandbox-file.ts
dvargasfuertes: I'm not sure I follow - these are input values, not input variables

ee/codegen/src/project.ts
```diff
@@ -86,6 +90,7 @@ export declare namespace WorkflowProjectGenerator {
     workflowsSdkModulePath?: readonly string[];
     workflowVersionExecConfigData: unknown;
     vellumApiKey?: string;
+    sandboxInputs?: SandboxInputs[];
```
noanflaherty: Rename to `WorkflowSandboxInputs`

ee/codegen/src/context/workflow-context/sdk-module-paths.ts
```diff
@@ -17,6 +17,11 @@ export function generateSdkModulePaths(
     CORE_NODES_MODULE_PATH: [...nodesModulePath, "core"] as const,
     DISPLAYABLE_NODES_MODULE_PATH: [...nodesModulePath, "displayable"] as const,
     INPUTS_MODULE_PATH: [...workflowsSdkModulePath, "inputs"] as const,
+    SANDBOX_RUNNER_MODULE_PATH: [
+      ...workflowsSdkModulePath,
+      "sandbox",
+      "runner",
```
noanflaherty: Remove one level of nesting and just create a file called `sandbox.py`

ee/codegen/src/generators/workflow-sandbox-file.ts
```diff
@@ -0,0 +1,102 @@
...
+
+type SandboxInput = StringInput | JsonInput | ChatHistoryInput | NumberInput;
+export type SandboxInputs = SandboxInput[];
+
+export declare namespace WorkflowSandboxFile {
+  interface Args extends BasePersistedFile.Args {
+    // Intentionally a nested array. Each `SandboxInputs` is a list of inputs for a single
+    // scenario / datapoint / test case. The outer array is a list of all the scenarios,
+    // dataset, test cases.
+    sandboxInputs: SandboxInputs[];
```
noanflaherty:
```suggestion
    sandboxInputs: SandboxInputs[];
```

ee/codegen/src/generators/workflow-sandbox-file.ts
```diff
@@ -0,0 +1,102 @@
...
+  NumberInput,
+  StringInput,
+} from "vellum-ai/api";
+
+import { vellumValue } from "src/codegen";
+import { BasePersistedFile } from "src/generators/base-persisted-file";
+import { getGeneratedInputsModulePath } from "src/utils/paths";
+
+type SandboxInput = StringInput | JsonInput | ChatHistoryInput | NumberInput;
+export type SandboxInputs = SandboxInput[];
```
noanflaherty: If this type is being imported into other files, let's move out of `generators/**` and somewhere in to `types/**`

ee/codegen/src/generators/workflow-sandbox-file.ts
noanflaherty: Remove vocab to avoid confusion

noanflaherty (approved):

PR #487, opened 2024-12-24
Dataclass's as default values

dvargas92495:
This PR supports using dataclasses as default values during function compilation. After this PR, we just have to do the same for pydantic and we could move on to integrating this util into prompt nodes


noanflaherty (approved):
Looking slick!

marlenc3 (commented):
🔥

PR #488, opened 2024-12-24
[Codegen] Fix Most Search Metadata Filter Issues

noanflaherty:
This fixes most but not all issues relating to `metadata_filters` in Search Nodes. There is one remaining issue that caused me to need to exclude the search node e2e test, but I claim what I have in here is an objective improvement over what's in main. Once we hash out the convo [here](https://vellum-ai.slack.com/archives/C06P13ABK0W/p1735079711823799), it'll inform what changes are needed from here on to un-exclude that fixture.

### DIFF
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -87,29 +87,6 @@ export function mergeNodeDataFactory(): MergeNode {
   };
 }
 
-const generateLogicalExpression = (): VellumLogicalConditionGroup => {
-  return {
-    type: "LOGICAL_CONDITION_GROUP",
-    negated: false,
-    combinator: "AND",
-    conditions: [
-      {
-        type: "LOGICAL_CONDITION_GROUP",
-        negated: false,
-        combinator: "AND",
-        conditions: [
-          {
-            type: "LOGICAL_CONDITION",
-            operator: "=",
-            lhsVariableId: "a6322ca2-8b65-4d26-b3a1-f926dcada0fa",
-            rhsVariableId: "c539a2e2-0873-43b0-ae21-81790bb1c4cb",
-          },
-        ],
-      },
-    ],
-  };
-};
-
 export function searchNodeDataFactory(args?: {
   metadataFiltersNodeInputId?: string;
   metadataFilters?: VellumLogicalConditionGroup;
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -1,5 +1,6 @@
 import { VellumVariableType } from "vellum-ai/api";
 
+import { VellumValueLogicalExpressionSerializer } from "src/serializers/vellum";
 import {
   EntrypointNode,
   CodeExecutionNode,
@@ -21,6 +22,7 @@ import {
   VellumLogicalConditionGroup,
   ConditionalNodeConditionData,
   NodeOutputData,
+  NodeInput,
 } from "src/types/vellum";
 
 export function entrypointNodeDataFactory(): EntrypointNode {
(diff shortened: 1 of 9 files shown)


ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -85,32 +87,69 @@ export function mergeNodeDataFactory(): MergeNode {
   };
 }
 
-const generateLogicalExpression: VellumLogicalConditionGroup = {
-  type: "LOGICAL_CONDITION_GROUP",
-  negated: false,
-  combinator: "AND",
-  conditions: [
+export function searchNodeDataFactory(args?: {
```
noanflaherty, line 90: Makes more about this factory parametrized.

ee/codegen/src/__test__/nodes/__snapshots__/search-node.test.ts.snap
```diff
@@ -37,6 +37,14 @@ class SearchNodeDisplay(BaseSearchNodeDisplay[SearchNode]):
...
     node_id = UUID("search")
     target_handle_id = UUID("370d712d-3369-424e-bcf7-f4da1aef3928")
+    metadata_filter_input_id_by_operand_id = {
+        UUID("a6322ca2-8b65-4d26-b3a1-f926dcada0fa"): UUID(
+            "c539a2e2-0873-43b0-ae21-81790bb1c4cb"
+        ),
+        UUID("c539a2e2-0873-43b0-ae21-81790bb1c4cb"): UUID(
+            "c539a2e2-0873-43b0-ae21-81790bb1c4cb"
+        ),
+    }
```
noanflaherty, line 47: We were hitting a failure case previously that caused us to not codegen these ids when we should have been

ee/codegen/src/__test__/nodes/__snapshots__/search-node.test.ts.snap
```diff
@@ -81,6 +89,133 @@ from vellum import (
...
+                    VellumValueLogicalConditionGroupRequest(
+                        type="LOGICAL_CONDITION_GROUP",
+                        combinator="AND",
+                        negated=False,
+                        conditions=[
+                            VellumValueLogicalConditionRequest(
+                                type="LOGICAL_CONDITION",
+                                lhs_variable=Inputs.var1,
+                                operator="=",
+                                rhs_variable=Inputs.var1,
```
noanflaherty, line 121: Notice that we now codegen these descriptors nested within the pydantic object. This is what's causing the new failure. Leaving like this so that we get the basic fixes in first and then we can solve this separately next.

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -31,6 +31,7 @@ export abstract class BaseNode<
   public readonly nodeContext: V;
 
   protected readonly nodeInputsByKey: Map<string, NodeInput>;
+  protected readonly nodeInputsById: Map<string, NodeInput>;
```
noanflaherty, line 34: Make it easy for downstream consumers to perform node input lookups not just by key, but also by id.

ee/codegen/src/generators/nodes/search-node.ts
```diff
@@ -192,19 +192,27 @@ export class SearchNode extends BaseSingleFileNode<
...
-        return parsedData.value;
-      }
+    if (!parsedData.ok) {
+      throw new Error(
+        `Failed to parse metadata filter JSON: ${JSON.stringify(
+          parsedData.errors
+        )}`
+      );
     }
-    return undefined;
```
noanflaherty, line 207: We were previously silently swallowing JSON parsing errors when we should instead fail loudly and early.

ee/codegen/src/generators/nodes/search-node.ts
```diff
@@ -259,39 +272,32 @@ export class SearchNode extends BaseSingleFileNode<
...
-        )?.nodeInputData?.value.rules[0] as InputVariablePointer;
-        const rhsQueryInput = this.nodeInputsByKey.get(
-          `${prefix}${logicalExpression.rhsVariableId}`
-        )?.nodeInputData?.value.rules[0] as InputVariablePointer;
+        const lhsQueryInput = this.nodeInputsById.get(
+          logicalExpression.lhsVariableId
+        )?.nodeInputData?.id;
+        const rhsQueryInput = this.nodeInputsById.get(
+          logicalExpression.rhsVariableId
+        )?.nodeInputData?.id;
```
noanflaherty, line 287: Perform lookups by id rather than by key (which required deriving the key the same way the main repo does and duplicating logic across the two)

ee/codegen/src/generators/nodes/search-node.ts
```diff
@@ -460,45 +477,15 @@ export class SearchNodeMetadataFilters extends AstNode {
...
-                name: "type",
-                value: python.TypeInstantiation.str("STRING"),
-              }),
-              python.methodArgument({
-                name: "value",
-                value: python.TypeInstantiation.str(lhsId),
-              }),
-            ],
-          }),
+          value: lhs,
```
noanflaherty, line 480: Here's where we codegen the descriptor instead of the id of the node input that represents the descriptor.

ee/codegen_integration/conftest.py
```diff
@@ -34,6 +34,7 @@ def _get_fixture_paths(root: str) -> Tuple[str, str]:
     # https://app.shortcut.com/vellum/story/4649/remove-fixture-exclusions-for-serialization
     exclude_fixtures={
         "faa_q_and_a_bot",
+        "simple_search_node",
```
noanflaherty, line 37: 😢

NgoHarrison (approved):
Nice stuff!

PR #489, opened 2024-12-24
[SC-5883] Accomodate strings in conditional node rule ids

NgoHarrison:


ee/vellum_ee/workflows/display/nodes/vellum/conditional_node.py
NgoHarrison, line 1: I changed all the UUID's in the dataclasses to be strings. I am not sure the extend of the bug in the main repo but it maniftested itself as invalid uuid's for field node inputs and value node inputs.

ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -809,9 +811,11 @@ export function conditionalNodeFactory({
         {
           id: "ad6bcb67-f21b-4af9-8d4b-ac8d3ba297cc",
           rules: [],
-          fieldNodeInputId: "2cb6582e-c329-4952-8598-097830b766c7",
+          fieldNodeInputId:
+            invalidUUID ?? "2cb6582e-c329-4952-8598-097830b766c7",
```
noanflaherty: nit: I notice that the hard-coded ids are not the same as each other, but that if I provide a value for `invalidUUID`, then that same value is used in multiple places. This makes it seem that these ids are supposed to match up, when in reality (I dont think?) they aren't meant to.

It might be better for tests that care about this case to not use this factory, or to set something higher level, like the `conditions` themselves.

ee/codegen/src/utils/uuid.ts
```diff
@@ -0,0 +1,8 @@
+import { python } from "@fern-api/python-ast";
+import { validate as uuidValidate } from "uuid";
+
+export function generateId(id: string): python.TypeInstantiation {
```
noanflaherty: nit: Given that this returns an AstNode, it seems reasonable for this to be a class called something like `UuidOrString` and extend `AstNode` directly. This would better follow the pattern's we've established so far.

ee/vellum_ee/workflows/display/nodes/vellum/conditional_node.py
```diff
@@ -40,17 +40,17 @@
...
 @dataclass
 class RuleIdMap:
-    id: UUID
+    id: Union[UUID, str]
     lhs: Optional["RuleIdMap"]
     rhs: Optional["RuleIdMap"]
-    field_node_input_id: Optional[UUID]
-    value_node_input_id: Optional[UUID]
+    field_node_input_id: Union[Optional[UUID], Optional[str]]
+    value_node_input_id: Union[Optional[UUID], Optional[str]]
```
noanflaherty: nit: I could see us using only strings here in `RuleIdMap` instead of a union between uuid and string. The main thing I wanted to avoid was having to change `node_input_ids_by_name: ClassVar[Dict[str, UUID]] = {}` to `node_input_ids_by_name: ClassVar[Dict[str, str]] = {}`, which you've achieved above with `BaseNodeVellumDisplay`. If you only ever use strings in `RuleIdMap`, it may simplify some codegen code too.

noanflaherty (approved):
I think there's some opportunity to clean some code up a tiny bit, but functionality looks solid 👍

PR #493, opened 2024-12-26
Use new function definition compilation in base inline prompt node

dvargas92495:
All of the new function compilation logic comes to a head as we incorporate it into base prompt node.

In the next PR we will use a new node that will spin off these function definitions into new nodes, resolve the values, and feed back into the model

### DIFF
--- a/src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
+++ b/src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
@@ -1,6 +1,6 @@
 import json
 from uuid import uuid4
-from typing import ClassVar, Generic, Iterator, List, Optional, Tuple, cast
+from typing import Callable, ClassVar, Generic, Iterator, List, Optional, Tuple, Union, cast
 
 from vellum import (
     AdHocExecutePromptEvent,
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
```diff
@@ -45,7 +46,7 @@ class BaseInlinePromptNode(BasePromptNode, Generic[StateType]):
     blocks: ClassVar[List[PromptBlock]]
 
     # The functions/tools that a Prompt has access to
-    functions: Optional[List[FunctionDefinition]] = OMIT
+    functions: Optional[List[Union[FunctionDefinition, Callable]]] = None
```
dvargasfuertes, line 49: I know we stole the `OMIT` idea from Fern, but this value is not falsy so it fails the `if self.functions` check on line 68 below.

Since we're already using optional here, I'd rather default to `None` as we did here, or to our undefined repr `UNDEF`, which _is_ falsy.

marlenc3 (commented):
✅

src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
noanflaherty: I'm generally a fan of replacing our `OMIT`s with `UNDEF`

noanflaherty (commented):
🔥

PR #496, opened 2024-12-26
[Codegen] Handle Missing Input Variables

noanflaherty:
It's possible for users to delete an Input Variable in the UI and not update Node Inputs afterwards. It doesn't appear that we go through and clean up these references and so instead, codegen must handle this edge case.


NgoHarrison (commented):
Lgtm!

dvargas92495 (commented):
✅

PR #497, opened 2024-12-26
[Codegen] Generate Standalone Script Files by Default

noanflaherty:
We were erroneously using the presence of a `filepath` to determine whether we should generate the code that backs a Code Execution Node as a standalone file or inlined attribute. Instead, we should default to standalone unless there's an override saying otherwise.


NgoHarrison (commented):
✅

dvargas92495 (commented):
✅

PR #504, opened 2024-12-27
Fix 500 errors associated with including invalid chat messages

dvargas92495:
WIP, Two issues to solve still

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -1,7 +1,8 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`WorkflowProjectGenerator > inlude sandbox > should include a sandbox.py file when passed sandboxInputs 1`] = `
-"from vellum.workflows.sandbox import SandboxRunner
+"from vellum import ARRAYRequest, StringChatMessageContentRequest, STRINGRequest
+from vellum.workflows.sandbox import SandboxRunner
 
 from .inputs import Inputs
 from .workflow import Workflow
@@ -10,7 +11,7 @@ if __name__ != "__main__":
     raise Exception("This file is not meant to be imported")
 
 
-runner = SandboxRunner(workflow=Workflow, inputs=[Inputs(input="foo"), Inputs(input="bar"), Inputs(input="hello")])
+runner = SandboxRunner(workflow=Workflow, inputs=[Inputs(input="foo", chat=[ChatMessageRequest(role="USER", text="foo")]), Inputs(input="bar", chat=[ChatMessageRequest(role="USER", content=StringChatMessageContentRequest(value="bar"))]), Inputs(input="hello", chat=[ChatMessageRequest(role="USER", content=[StringChatMessageContentRequest(value="hello")])]),])
 
 runner.run()
 "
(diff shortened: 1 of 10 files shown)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1,7 +1,8 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`WorkflowProjectGenerator > inlude sandbox > should include a sandbox.py file when passed sandboxInputs 1`] = `
-"from vellum.workflows.sandbox import SandboxRunner
+"from vellum import ARRAYRequest, StringChatMessageContentRequest, STRINGRequest
```
dvargasfuertes: Where is `ARRAYRequest` and `STRINGRequest` coming from? Why is `ChatMessageRequest` missing?

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -10,7 +11,7 @@ if __name__ != "__main__":
     raise Exception("This file is not meant to be imported")
 
 
-runner = SandboxRunner(workflow=Workflow, inputs=[Inputs(input="foo"), Inputs(input="bar"), Inputs(input="hello")])
+runner = SandboxRunner(workflow=Workflow, inputs=[Inputs(input="foo", chat=[ChatMessageRequest(role="USER", text="foo")]), Inputs(input="bar", chat=[ChatMessageRequest(role="USER", content=StringChatMessageContentRequest(value="bar"))]), Inputs(input="hello", chat=[ChatMessageRequest(role="USER", content=[StringChatMessageContentRequest(value="hello")])]),])
```
dvargasfuertes: why is this not newlining after introducing the comma?

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
noanflaherty: This looks like a bug in the logic here: https://github.com/vellum-ai/vellum-python-sdks/blob/175711d9ef4bc8579afff035aa485150b30573f5/ee/codegen/src/generators/chat-message-content.ts#L52

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
noanflaherty: My guess is you're hitting this case: https://github.com/vellum-ai/vellum-python-sdks/blob/b0c07f84f6694cf6a344bd0b7e261a4e616e9270/ee/codegen/src/generators/base-persisted-file.ts#L87

ee/codegen/src/generators/workflow-sandbox-file.ts
```diff
@@ -40,35 +40,31 @@ export class WorkflowSandboxFile extends BasePersistedFile {
...
     return [
       python.codeBlock(`\
 if __name__ != "__main__":
     raise Exception("This file is not meant to be imported")
 `),
       sandboxRunnerField,
-      runMethodInvocation,
+      // Using code block instead of method invocation since the latter tries to import `runner.run` after
+      // specifying as a reference, even though it's a locally defined variable.
+      python.codeBlock(`runner.run()`),
```
dvargasfuertes, line 67: cc: @noanflaherty, curious on your thoughts here

noanflaherty: Huh that's interesting. This seems like a bug that I'll have to fix on the fern side. Let's roll with this as is for now. I've created this to track: https://app.shortcut.com/vellum/story/5890/fix-imports-of-locally-defined-methods-classes-in-fern

noanflaherty (approved):

kirawoods (commented):
✅

PR #505, opened 2024-12-27
Move Utility Functions to Prepare for Prompt Block Compilation

noanflaherty:
This PR performs a refactor to centralize some utility functions that we'll need for compiling prompt blocks.

Notably, it introduces a new top-level `utils/` directory within `vellum/src`. Ideally I'd call this `core`, but that name is taken by fern for now. I think this is fine and we can rearrange things later if need be.

In my next PR, I'll introduce a new top-level `prompts/` directory.


dvargas92495 (approved):

src/vellum/utils/__init__.py
dvargasfuertes, line 1: [oos] there are some uuid utilities in both dirs I could see centralizing here too

PR #506, opened 2024-12-27
Introduce Prompt Block Compilation Code

noanflaherty:
This open-sources some of our prompt block compilation code – responsible for performing all variable substitutions and jinja templating.

### DIFF
--- a/src/vellum/prompts/blocks/compilation.py
+++ b/src/vellum/prompts/blocks/compilation.py
@@ -1,0 +1,190 @@
+import json
+from typing import Optional, cast
+
+from vellum import (
+    ChatMessage,
+    JsonVellumValue,
+    PromptBlock,
+    PromptRequestInput,
+    RichTextPromptBlock,
+    StringVellumValue,
+    VellumVariable,
+)
+from vellum.prompts.blocks.exceptions import PromptCompilationError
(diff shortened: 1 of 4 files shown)


src/vellum/prompts/blocks/compilation.py
```diff
@@ -0,0 +1,190 @@
...
+                compiled_blocks.extend(chat_message_blocks)
+            else:
+                raise ValueError(f"Invalid input type for variable block: {compiled_input.type}")
+
+        elif block.block_type == "RICH_TEXT":
+            value_block = _compile_rich_text_block_as_value_block(block=block, inputs=sanitized_inputs)
+            compiled_blocks.append(value_block)
+
+        elif block.block_type == "FUNCTION_DEFINITION":
+            raise ValueError("Function definitions shouldn't go through compilation process")
```
dvargasfuertes, line 114: [nit] why not raise `PromptCompilationError` in place of these `ValueError`s?

dvargas92495 (commented):
✅

src/vellum/prompts/blocks/compilation.py
noanflaherty: It probably should – I'll update.

PR #507, opened 2024-12-28
Introduce New Experimental `OpenAIChatCompletionNode`

noanflaherty:
This introduces the first of a few upcoming experimental nodes. This one – `OpenAIChatCompletionNode` – allows for using the standard Vellum prompt interface, but for making API calls directly to OpenAI rather than through Vellum. It's meant to serve as a drop-in replacement for `InlinePromptNode`.

For a demo, see here:


src/vellum/workflows/nodes/experimental/README.md
```diff
@@ -0,0 +1,6 @@
+# 🧪 Experimental
+
+This section is a proofing ground for new ideas and concepts. It's not meant to be used in production and is
+subject to breaking changes at any time.
+
+If a concept within is sufficiently interesting and validated, then introduced as a first-class node.
```
siddseethepalli: nit: `then introduced` -> `then it may be introduced`?

pyproject.toml
```diff
@@ -40,6 +40,7 @@ cdktf = "^0.20.5"
 click = "8.1.7"
 docker = "7.1.0"
 httpx = ">=0.21.2"
+openai = "1.58.1"
```
siddseethepalli: Does this mean we'll have a specific version of the OpenAI SDK as a dependency? That doesn't feel great since it might conflict with customer versions. Can we at least mark this as an extra?

siddseethepalli (commented):
Huge

pyproject.toml
dvargasfuertes: I'm not too worried about sub-dep conflict here since in theory they are using vellum and not the openai sdk.

however, the way around this is to specify a range like we do with the `httpx` dep above. So pick the minimum version that is supported by the code (eg let's say it's 1.0.0) and declare `openai = ">= 1.0.0"`

PR #508, opened 2024-12-29
[Codegen] Fix Reference Tracking for Chat Messages

noanflaherty:
While pulling down the `sandbox.py` file for our Basic RAG Workflow, I noticed that some imports were missing.

The root cause of this is that we were creating AST Nodes in various `write` methods instead of in our constructors. This PR performs some refactoring to make sure that all reference tracking works as expected and also modualrizes the code a bit further.

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/chat-message-content.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/chat-message-content.test.ts.snap
@@ -1,11 +1,11 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`ChatMessageContent > ARRAY > should write an array of content correctly 1`] = `
-"ArrayChatMessageContentRequest(
+"ArrayChatMessageContent(
     value=[
-        StringChatMessageContentRequest(value="First message"),
-        FunctionCallChatMessageContentRequest(
-            value=FunctionCallChatMessageContentValueRequest(
+        StringChatMessageContent(value="First message"),
+        FunctionCallChatMessageContent(
+            value=FunctionCallChatMessageContentValue(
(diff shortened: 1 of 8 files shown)


ee/codegen/src/__test__/__snapshots__/chat-message-content.test.ts.snap
```diff
@@ -1,11 +1,11 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`ChatMessageContent > ARRAY > should write an array of content correctly 1`] = `
-"ArrayChatMessageContentRequest(
+"ArrayChatMessageContent(
```
noanflaherty, line 4: I'm changing the default behavior to not include the `Request` suffix.

ee/codegen/src/__test__/chat-message-content.test.ts
```diff
@@ -16,6 +16,7 @@ describe("ChatMessageContent", () => {
       });
       chatMessageContent.write(writer);
       expect(await writer.toString()).toMatchSnapshot();
+      expect(chatMessageContent.getReferences()).toHaveLength(1);
```
noanflaherty, line 19: Adds test coverage to make sure we have references as expected

NgoHarrison (commented):
✅

ee/codegen/src/__test__/__snapshots__/chat-message-content.test.ts.snap
dvargasfuertes: [oos] We may need to make sdk changes to support this, will test today

dvargas92495 (commented):
✅

kirawoods (commented):
✅

PR #510, opened 2024-12-29
Exclude `prompts/` and `utils/` from Fern

noanflaherty:
Updates fern ignore so that these new directories don't get wiped during codegen.


NgoHarrison (commented):
✅

dvargas92495 (commented):
✅

PR #518, opened 2024-12-30
Improve Interfaces for `SandboxRunner` and `WorkflowContext`

noanflaherty:
# Rationale

When trying to run and debug Workflows locally, @NgoHarrison and I found ourselves wanting to have more control over the `WorkflowContext` that was being used in a `SandboxRunner`. Our goal was to drive the urls of the Vellum CLI, which I later discovered was possible via env vars and handled as part of `create_vellum_client`. Despite this, I think it's very reasonable to give users greater control by passing instances of things they care about, rather than forcing them to use env vars and know about them.

# Changes
1. Improve names of kwargs in `WorkflowContext.__init__`
2. Pass Workflow instances instead of their class when instantiating `SandboxRunner`
3. Rename from `SandboxRunner` to `WorkflowSandboxRunner`
4. Update codegen to generate the newly expected code.


NgoHarrison (approved):
Nice!

dvargas92495 (approved):

PR #519, opened 2024-12-30
Inline Subworkflow node supports inputs class's

dvargas92495:
Context: https://vellum-ai.slack.com/archives/C06P13ABK0W/p1735584326763649?thread_ts=1735309195.646639&cid=C06P13ABK0W


src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
```diff
@@ -24,6 +26,7 @@ class InlineSubworkflowNode(BaseSubworkflowNode[StateType], Generic[StateType, W
     """
 
     subworkflow: Type["BaseWorkflow[WorkflowInputsType, InnerStateType]"]
+    subworkflow_inputs: ClassVar[Union[EntityInputsInterface, BaseInputs]]
```
dvargasfuertes: Ideally this is parameterized with `WorkflowInputsType` instead of `BaseInputs`. In order to do so, would require some custom mypy-ing bc they don't have great support for typevar's used in this way. Not a huge issue though since if you're already instantiating an `Inputs` class, that class' init will drive most of the type checking

dvargasfuertes: https://app.shortcut.com/vellum/story/5986

src/vellum/workflows/nodes/bases/base_subworkflow_node/node.py
```diff
@@ -1,10 +0,0 @@
-from typing import ClassVar, Generic
-
-from vellum.workflows.nodes.bases import BaseNode
-from vellum.workflows.types.core import EntityInputsInterface
-from vellum.workflows.types.generics import StateType
-
-
-class BaseSubworkflowNode(BaseNode[StateType], Generic[StateType]):
-    # Inputs that are passed to the Subworkflow
-    subworkflow_inputs: ClassVar[EntityInputsInterface] = {}
```
dvargasfuertes, line 10: The only thing this class reused between workflow nodes was subworkflow_inputs. With this new feature, the type now differs between the two, meaning there's no reason now to have a base class between the two nodes

src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
```diff
@@ -66,4 +69,9 @@ def run(self) -> Iterator[BaseOutput]:
 
     def _compile_subworkflow_inputs(self) -> WorkflowInputsType:
         inputs_class = self.subworkflow.get_inputs_class()
-        return inputs_class(**self.subworkflow_inputs)
+        if isinstance(self.subworkflow_inputs, dict):
```
noanflaherty, line 72: Are we sure we want to support two forms here? If so, we will need to pick one that we want codegen to use. My 1/5 preference is to have one representation in general and my 3/5 preference is to codegen the class representation even if we support the dict repr.

noanflaherty (approved):

src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
dvargasfuertes: My 2/5 preference with user-facing interfaces in general is as flexible as reasonably possible on the input, as restrictive as reasonably possible in the output. In this case, `subworkflow_inputs` pointing to a JSON output of an upstream node seems like a reasonable use case to support

Agree with codegenning Inputs classes

PR #520, opened 2024-12-30
Add support for casting to json types for templating nodes

NgoHarrison:
This came up from a customer workflow with this class:
`class CastToJSON(TemplatingNode[BaseState, Any]):`

We don't currently support Any casting. This PR tries to cast to json if it is Any since codegen is generating `Any` if we have output type of `Json`


noanflaherty (approved):
I think this makes sense for now. I'd love for us to ticket and explore later how/whether we could use our special `Json` type from [here](https://github.com/vellum-ai/vellum-python-sdks/blob/fca0a47e882c38c27ea23ded017081a1ff9cad63/src/vellum/workflows/types/core.py#L14) instead of `Any`. Users may legitimately not want to do any special casting on the output of a custom type and here we're forcing it into JSON without giving them that optionality.

PR #521, opened 2024-12-30
[SC-5996] Use custom json type instead of Any when type checking templating node output type

NgoHarrison:
Spawned from this comment here:
https://github.com/vellum-ai/vellum-python-sdks/pull/520#pullrequestreview-2526027106

### DIFF
--- a/src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
+++ b/src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
@@ -1,9 +1,9 @@
 import json
-from typing import Any
 
 from vellum.workflows.nodes.bases.base import BaseNode
 from vellum.workflows.nodes.core.templating_node.node import TemplatingNode
 from vellum.workflows.state import BaseState
+from vellum.workflows.types.core import Json
 
 
 def test_templating_node__dict_output():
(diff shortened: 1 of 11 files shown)


src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
```diff
@@ -74,7 +74,7 @@ class BoolTemplateNode(TemplatingNode[BaseState, bool]):
 
 def test_templating_node__json_output():
     # GIVEN a templating node that outputs JSON
-    class JSONTemplateNode(TemplatingNode[BaseState, Any]):
+    class JSONTemplateNode(TemplatingNode[BaseState, Json]):
```
noanflaherty, line 77: [blocking] Can we add a test to confirm that we're able to produce the expected JSON representation when we serialize a node that looks like this?

noanflaherty (commented):
Looks promising! I suspect we'll need to update serialization to handle this case though

noanflaherty:
Also, note that any patterns we establish for JSON with `TemplatingNode`, we likely should also propagate over to `CodeExecutionNode`. This could be ticketed and tackled separately

noanflaherty (approved):

PR #522, opened 2024-12-31
Search node use is_valid_uuid instead of is instance

m-abboud:
The UI sets ids instead of names for document_index and deserialize of a union[str|uuid] isn't working and isn't really possible if you think about it unless you ban uuids for names. I'm also planning on switching the UI to save these as names instead of the ids so the vellum pulls are prettier (could also do it in deserialize step I guess but that breaks users that just wanna use ids?) and maybe change the document search route to do a double lookup to handle the case of a document_name being a uuid string.


src/vellum/workflows/nodes/displayable/bases/search_node.py
```diff
@@ -12,6 +12,7 @@
     SearchWeightsRequest,
 )
 from vellum.core import ApiError, RequestOptions
+from vellum.evaluations.utils.uuid import is_valid_uuid
```
noanflaherty: [blocking] Can we move this `uuid.py` file into the new top-level `vellum.utils` directory? We should avoid importing from `evaluations/` until there's good reason to.

noanflaherty (approved):
> I'm also planning on switching the UI to save these as names instead of the ids so the vellum pulls are prettier (could also do it in deserialize step I guess but that breaks users that just wanna use ids?

(1/5) I'd probably do this in the codegen step personally and just be opinionated that we always use the document index name upon pulling. (4/5) big fan of doing one or both of these proposals though! Let's ticket if not tackling soon so we don't forget and add to one of the epics in here: https://app.shortcut.com/vellum/objective/5089/wac-marketing-release


> and maybe change the document search route to do a double lookup to handle the case of a document_name being a uuid string.

I like this idea and think it's worth doing in any case. [Here's an example](https://github.com/vellum-ai/vellum/blob/e8b0814b36fc17a962688171915c0586990a1ea0/django/app/api/deployments/views.py#L198) in case helpful.

m-abboud:
> > I'm also planning on switching the UI to save these as names instead of the ids so the vellum pulls are prettier (could also do it in deserialize step I guess but that breaks users that just wanna use ids?
> 
> (1/5) I'd probably do this in the codegen step personally and just be opinionated that we always use the document index name upon pulling. (4/5) big fan of doing one or both of these proposals though! Let's ticket if not tackling soon so we don't forget and add to one of the epics in here: https://app.shortcut.com/vellum/objective/5089/wac-marketing-release
> 
> > and maybe change the document search route to do a double lookup to handle the case of a document_name being a uuid string.
> 
> I like this idea and think it's worth doing in any case. [Here's an example](https://github.com/vellum-ai/vellum/blob/e8b0814b36fc17a962688171915c0586990a1ea0/django/app/api/deployments/views.py#L198) in case helpful.

@noanflaherty  wouldn't this mean codegen would have to do a document index lookup to django since it would only have  the id, making codegen slower? If I did I add to codegen though is there somewhere where were already prefetching stuff?

noanflaherty:
@m-abboud that's correct. We do similar in a number of other instances too. For example, [here](https://github.com/vellum-ai/vellum-python-sdks/blob/abd2d8b85275e6eccc1a4508cd028285adc22367/ee/codegen/src/context/node-context/create-node-context.ts#L57) and [here](https://github.com/vellum-ai/vellum-python-sdks/blob/abd2d8b85275e6eccc1a4508cd028285adc22367/ee/codegen/src/context/node-context/create-node-context.ts#L94).

Given the API call is between two services in our own infra, I wouldn't expect the slowdown to be _too_ bad? IMO adding this logic here to codegen for now then finding ways to remove the need for it later is a fine approach that we've been taking for other use-cases so far.

src/vellum/workflows/nodes/displayable/bases/search_node.py
```diff
@@ -73,11 +74,13 @@ class Outputs(BaseOutputs):
...
     def _perform_search(self) -> SearchResponse:
+        index_is_uuid = True if isinstance(self.document_index, UUID) else is_valid_uuid(self.document_index)
+
         try:
             return self._context.vellum_client.search(
                 query=self.query,
-                index_id=str(self.document_index) if isinstance(self.document_index, UUID) else None,
-                index_name=self.document_index if isinstance(self.document_index, str) else None,
+                index_id=str(self.document_index) if index_is_uuid else None,
+                index_name=str(self.document_index) if not index_is_uuid else None,
```
dvargasfuertes, line 83: [oos] this is where the `document_index` param could be useful as we'd be able to just do `document_index=str(self.document_index)` without the isinstance/is_valid_uuid logic above

dvargas92495 (commented):
✅

PR #529, opened 2024-12-31
Codegen search node document_index as name instead of ID

m-abboud:


ee/codegen/src/context/node-context/create-node-context.ts
```diff
@@ -35,8 +37,26 @@ export async function createNodeContext(
...
+      // with name instead of ID. We can only do this if the document index input is a constant value.
+      let documentIndex: DocumentIndexRead | null = null;
+      const inputValue = searchNodeData.inputs.find(
+        (input) => input.id === searchNodeData.data.documentIndexNodeInputId
+      )?.value;
+
+      const rule = inputValue?.rules?.[0];
+      if (rule?.type === "CONSTANT_VALUE") {
+        if (rule.data.value?.toString()) {
+          documentIndex = await new DocumentIndexesClient({
```
m-abboud, line 51: If the id/name is invalid this will fail and throw an error that might end up being obfuscated behind an internal server error, do we have/plan on having a pre-validation step anywhere? Should this have a try catch to throw a specific error type that the user will actually see?

noanflaherty: Yeah we should generally catch errors and use this guy: https://github.com/vellum-ai/vellum-python-sdks/blob/89b079213c3c80a3ea851be1429b2cb82bdf0d01/ee/codegen/src/context/workflow-context/workflow-context.ts#L320

ee/codegen/src/context/node-context/create-node-context.ts
```diff
@@ -35,8 +37,26 @@ export async function createNodeContext(
...
+      // Grab the associated document index from api if available to always populate document_index field
+      // with name instead of ID. We can only do this if the document index input is a constant value.
+      let documentIndex: DocumentIndexRead | null = null;
+      const inputValue = searchNodeData.inputs.find(
+        (input) => input.id === searchNodeData.data.documentIndexNodeInputId
+      )?.value;
+
+      const rule = inputValue?.rules?.[0];
+      if (rule?.type === "CONSTANT_VALUE") {
+        if (rule.data.value?.toString()) {
```
noanflaherty, line 50: nit: we could set a temp var for `rule.data.value?.toString()` since it's used twice here.

noanflaherty (approved):
Looks good!

dvargas92495 (commented):
✅

PR #530, opened 2024-12-31
WIP - Static Node IDs in new map.py file

marlenc3:
quick loom over new file: https://www.loom.com/share/d8bcd0a884f54954bfef4860f5f2f85e

the gist is that the events are using hashes of the node/workflow modules instead of the node_ids that we pull from sandboxes

this is the first step to mapping ids -- we'll check this file for our path on init, and if we have a UUID, we'll use it

looking for quick thoughts, one other thought I had was to just save the __id__ on each file for us, but I can't imagine that would be a great experience for developers

cc: @dvargas92495 PR w/ Issue found: https://github.com/vellum-ai/vellum/pull/7056

### DIFF
--- a/ee/codegen/src/codegen.ts
+++ b/ee/codegen/src/codegen.ts
@@ -8,6 +8,7 @@ import {
   Workflow,
 } from "./generators";
 import { BasePersistedFile } from "./generators/base-persisted-file";
+import { IdentifierFile } from "./generators/identifier-file";
 import { WorkflowSandboxFile } from "./generators/workflow-sandbox-file";
 
 export function vellumVariable(args: VellumVariable.Args): VellumVariable {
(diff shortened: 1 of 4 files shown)


dvargas92495:
We eventually want to do some sort of state file, but in the meantime, we'll want to use the node id that we're _already generating_, for example: https://github.com/vellum-ai/workflows-as-code-runner-prototype/blob/main/examples/basic_rag_workflow/display/nodes/prompt_node.py#L13

Once this display class is declared, it should already set the `__id__` properly here: https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/vellum_ee/workflows/display/nodes/base_node_display.py#L43

So my hunch is that when Vembda loads the workflow here: https://github.com/vellum-ai/vembda-service/blob/90248d19bc755e4a448f3ffba3a4826ca36a8155/python-workflow-server/src/workflow_server/executor.py#L388

We are for some reason not loading the display classes here: https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/workflows/base.py#L418

Now I thought that [these init files](https://github.com/vellum-ai/workflows-as-code-runner-prototype/blob/main/examples/basic_rag_workflow/__init__.py#L3) would guarantee that the display classes would load but it doesn't seem like the case. So I'm thinking that in the `load_from_module` method linked above we manually load this init file

PR #534, opened 2025-01-02
Prevent Displayable Prompt Nodes from failing on just function call responses

dvargas92495:
Prompt Nodes that will be used in the UI currently fail if the prompt returns _just_ a function call. Our behavior here should model what we do with legacy runner as much as possible, which is to always return a text output using [this helper](https://github.com/vellum-ai/vellum/blob/main/django/app/predict/utils.py#L68-L82).

This PR updates the run method of our `InlinePromptNode` to use an updated/adapted version of the above logic.

### DIFF
--- a/src/vellum/workflows/nodes/displayable/inline_prompt_node/node.py
+++ b/src/vellum/workflows/nodes/displayable/inline_prompt_node/node.py
@@ -1,3 +1,4 @@
+import json
 from typing import Iterator
 
 from vellum.workflows.errors import WorkflowErrorCode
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/displayable/inline_prompt_node/node.py
```diff
@@ -37,13 +38,19 @@ def run(self) -> Iterator[BaseOutput]:
...
+        for output in outputs:
+            if output.value is None:
+                continue
+
+            if output.type == "STRING":
+                string_outputs.append(output.value)
+            elif output.type == "JSON":
+                string_outputs.append(json.dumps(output.value, indent=4))
+            elif output.type == "FUNCTION_CALL":
+                string_outputs.append(output.value.model_dump_json(indent=4))
```
dvargasfuertes, line 51: In legacy prompt runner, we return the openai representation of sunction calls stringified here. this is for legacy reasons of when function calling was originally introduced and is part of the tech debt that we want to clean up there anyway. we swap to using our representation here.

src/vellum/workflows/nodes/displayable/inline_prompt_node/tests/test_node.py
```diff
@@ -117,3 +116,15 @@ def generate_prompt_events(*args: Any, **kwargs: Any) -> Iterator[ExecutePromptE
...
+    assert (
+        outputs[-1].value
+        == """{
+    "arguments": {
+        "foo": "hello",
+        "bar": 1
+    },
+    "id": null,
+    "name": "my_function",
+    "state": null
```
dvargasfuertes, line 128: we should remove this `state` field in a followup

noanflaherty (approved):

marlenc3 (approved):

PR #535, opened 2025-01-02
Setup tests and flesh out implementation for node mocking

dvargas92495:
This PR introduces node mocking to the SDK.

In the monorepo, we have data model support for conditional mocks. However, we haven't seen user requests for it yet and haven't prioritized it yet. So I decided to opt for a simpler API to start but would be amenable to iterating it.

### DIFF
--- a/src/vellum/workflows/runner/runner.py
+++ b/src/vellum/workflows/runner/runner.py
@@ -4,7 +4,7 @@ import logging
 from queue import Empty, Queue
 from threading import Event as ThreadingEvent, Thread
 from uuid import UUID
-from typing import TYPE_CHECKING, Any, Dict, Generic, Iterable, Iterator, Optional, Sequence, Set, Type, Union
+from typing import TYPE_CHECKING, Any, Dict, Generic, Iterable, Iterator, List, Optional, Sequence, Set, Type, Union
 
 from vellum.workflows.constants import UNDEF
 from vellum.workflows.context import execution_context, get_parent_context
@@ -72,6 +72,7 @@ class WorkflowRunner(Generic[StateType]):
         entrypoint_nodes: Optional[RunFromNodeArg] = None,
         external_inputs: Optional[ExternalInputsArg] = None,
         cancel_signal: Optional[ThreadingEvent] = None,
+        mocks: Optional[List[BaseOutputs]] = None,
         parent_context: Optional[ParentContext] = None,
     ):
         if state and external_inputs:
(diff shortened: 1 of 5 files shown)


src/vellum/workflows/runner/runner.py
```diff
@@ -72,6 +72,7 @@ def __init__(
         entrypoint_nodes: Optional[RunFromNodeArg] = None,
         external_inputs: Optional[ExternalInputsArg] = None,
         cancel_signal: Optional[ThreadingEvent] = None,
+        mocks: Optional[List[BaseOutputs]] = None,
```
noanflaherty: nit: I suggest we call this `node_mocks` or `node_output_mocks` to be more explicit about what it is that we're mocking.

src/vellum/workflows/workflows/base.py
```diff
@@ -194,6 +194,7 @@ def run(
         entrypoint_nodes: Optional[RunFromNodeArg] = None,
         external_inputs: Optional[ExternalInputsArg] = None,
         cancel_signal: Optional[ThreadingEvent] = None,
+        mocks: Optional[List[BaseOutputs]] = None,
```
noanflaherty: If we rename above, let's rename here, too.

noanflaherty (approved):
Implementation looks good!

I think before merging we should internally align on how/whether this plays a first-class role when writing tests for a Workflow. I think it could be really cool to think about this within that context and make sure they play nicely together.

dvargas92495:
I was actually writing tests for [vargas jr](https://github.com/dvargas92495/vargasjr-dev/blob/main/agent/tests/test_runner.py#L108-L109) when I decided it was time to add mocking support. So yea, I was envisioning this in the context of first class testing support.

Happy to push another commit to help show this - originally thought the [first test](https://github.com/vellum-ai/vellum-python-sdks/pull/535/files#diff-3640fb8a2ccecd8ebbcc5d190eed944b9932c16a5e8ec43fa0d08828b9ac94ceR8-R13) would. Should I migrate one of our existing tests in this repo that is using prompt nodes to show case it? Or did you have something else in mind?

noanflaherty:
> Should I migrate one of our existing tests in this repo that is using prompt nodes to show case it

I think this would help a lot. If you have confidence the API plays nice with this use-case, we can do as a fast-follow

dvargas92495:
@noanflaherty all tests in _this_ repo should actually continue to mock the ad_hoc client since they are all inline prompt node specific tests.

However, here's a preview of how this starts to look like for end users in the prototype repo: https://github.com/vellum-ai/workflows-as-code-runner-prototype/pull/564

noanflaherty:
@dvargas92495 given that mocking happens at the node-level, but are provided at the node-output level, what happens if a user doesn't provide mock outputs for all of a node's outputs?

noanflaherty:
One idea is to have the user provide an instance of the node's outputs class so that all outputs are provided and we treat this as a node-level mock.

noanflaherty (approved):
Okay yeah this is looking awesome!

PR #536, opened 2025-01-02
Update sandbox codegen to use sanitized names for input args

NgoHarrison:
This was coming up from a customer workflow. They had the input variable names for the workflow sandbox as camel case (i.e `someFoo`) This was causing a compile issue since the `Inputs.py` file is being codegen'd with the sanitized name (includes converting it to python snake case). This PR is to fix that

### DIFF
--- a/ee/codegen/src/context/input-variable-context.ts
+++ b/ee/codegen/src/context/input-variable-context.ts
@@ -38,6 +38,10 @@ export class InputVariableContext {
     return this.inputVariableData;
   }
 
+  public getRawName(): string {
+    return this.inputVariableData.key;
+  }
+
   private generateSanitizedInputVariableName(): string {
     const defaultName = "input_";
     const rawInputVariableName = this.inputVariableData.key;
--- a/ee/codegen/src/context/workflow-context/workflow-context.ts
+++ b/ee/codegen/src/context/workflow-context/workflow-context.ts
@@ -232,6 +232,31 @@ export class WorkflowContext {
     return inputVariableContext;
   }
 
+  public findInputVariableContextByRawName(
+    rawName: string
+  ): InputVariableContext | undefined {
+    const inputVariableContext = Array.from(
+      this.inputVariableContextsById.values()
+    ).find((inputContext) => inputContext.getRawName() === rawName);
+
+    return inputVariableContext;
+  }
+
+  public getInputVariableContextByRawName(
+    rawName: string
+  ): InputVariableContext {
+    const inputVariableContext =
+      this.findInputVariableContextByRawName(rawName);
+
+    if (!inputVariableContext) {
+      throw new Error(
+        `Input variable context not found for raw name: ${rawName}`
+      );
+    }
+
+    return inputVariableContext;
+  }
+
   public isOutputVariableNameUsed(outputVariableName: string): boolean {
     return this.outputVariableNames.has(outputVariableName);
   }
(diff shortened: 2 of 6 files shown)


ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -54,7 +54,7 @@ export class WorkflowContext {
 
   // Track what input variables names are used within this workflow so that we can ensure name uniqueness when adding
   // new input variables.
-  private readonly inputVariableNames: Set<string> = new Set();
+  public readonly inputVariableNames: Set<string> = new Set();
```
noanflaherty: Can we try making use of `InputVariableContext`, which already handles name sanitization, rather than making this public and duplicating sanitization logic below?

noanflaherty (changes requested):

ee/codegen/src/context/workflow-context/workflow-context.ts
NgoHarrison: Updated!

ee/codegen/src/context/input-variable-context.ts
```diff
@@ -38,6 +38,10 @@ export class InputVariableContext {
     return this.inputVariableData;
   }
 
+  public getRawName(): string {
+    return this.inputVariableData.key;
+  }
```
noanflaherty, line 43: Introduced this helper to centralize where/how we access the raw name.

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -232,6 +232,31 @@ export class WorkflowContext {
     return inputVariableContext;
   }
 
+  public findInputVariableContextByRawName(
```
noanflaherty, line 235: Introduced these helpers to follow similar patterns that we already have. It has the benefit of centralizing helpers to retrieve an input variable context by an identifier onto the `WorkflowContext` so that other consumers can use this down the road if need be.

ee/codegen/src/generators/workflow-sandbox-file.ts
```diff
@@ -79,14 +79,17 @@ if __name__ != "__main__":
...
+
+        return python.methodArgument({
+          name: inputVariableContext.name,
           value: vellumValue({
             vellumValue: input,
           }),
-        })
-      ),
+        });
+      }),
```
noanflaherty, line 92: Removed the `getName` helper from this class. If we were to keep it, we should have named it something like `getInputVariableName` to be more explicit.

ee/codegen/src/generators/workflow-sandbox-file.ts
```diff
@@ -79,14 +79,17 @@ if __name__ != "__main__":
         name: "Inputs",
         modulePath: getGeneratedInputsModulePath(this.workflowContext),
       }),
-      arguments_: inputs.map((input) =>
-        python.methodArgument({
-          name: input.name,
+      arguments_: inputs.map((input) => {
+        const inputVariableContext =
+          this.workflowContext.getInputVariableContextByRawName(input.name);
```
noanflaherty, line 84: Here we use the new helpers from above.

noanflaherty (approved):
@NgoHarrison I pushed a commit that does some code cleanup and included some comments with why I made the changes I did. Lmk what you think and if you're cool with them.

NgoHarrison:
> @NgoHarrison I pushed a commit that does some code cleanup and included some comments with why I made the changes I did. Lmk what you think and if you're cool with them.

Yup totally cool with it and it makes sense. Thanks for the quick cleanup!

dvargas92495 (commented):
✅

PR #537, opened 2025-01-02
Use addError for entity 404 errors in create node context

m-abboud:
So based off what was said about the addError in my last codegen PR, it seems like catching the 404 and adding an error would be advantageous and just move onto the other nodes so we can get all of the errors for every node with the workflow and they don't get hidden as internal server errors? Will add test if my understanding of our error handling is correct here

### DIFF
--- a/ee/codegen/src/__test__/helpers/workflow-context-factory.ts
+++ b/ee/codegen/src/__test__/helpers/workflow-context-factory.ts
@@ -7,6 +7,7 @@ export function workflowContextFactory(
     workflowClassName,
     workflowRawEdges,
     codeExecutionNodeCodeRepresentationOverride,
+    strict = true,
   }: Partial<WorkflowContext.Args> = {
     codeExecutionNodeCodeRepresentationOverride: "STANDALONE",
   }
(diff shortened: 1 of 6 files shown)


dvargas92495 (approved):
>  it seems like catching the 404 and adding an error would be advantageous and just move onto the other nodes so we can get all of the errors for every node with the workflow and they don't get hidden as internal server errors?

Correct! We want to remove all raw `throw new Error` and move to using custom errors and `addError()` handling

m-abboud:
> > it seems like catching the 404 and adding an error would be advantageous and just move onto the other nodes so we can get all of the errors for every node with the workflow and they don't get hidden as internal server errors?
> 
> Correct! We want to remove all raw `throw new Error` and move to using custom errors and `addError()` handling



> > it seems like catching the 404 and adding an error would be advantageous and just move onto the other nodes so we can get all of the errors for every node with the workflow and they don't get hidden as internal server errors?
> 
> Correct! We want to remove all raw `throw new Error` and move to using custom errors and `addError()` handling

@dvargas92495 Cool so should the `throw e` I did here when it wasn't a 404 also be handled and added to like an UnknownError thing? Or maybe we should have a try catch around the whole thing per node that adds the error to the context so it can continue code genning?

PR #538, opened 2025-01-02
Serialize queue to list in state encoder

m-abboud:


dvargas92495 (approved):
Can we add a test similar to https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/state/tests/test_state.py#L104 covering this case?

PR #539, opened 2025-01-03
Display Node Imports on Load from Module

marlenc3:
aight SECOND go

stuff is complicated due to the virtual file loading -- there's no real directory and we can't really exec the import by itself (we need to load up the modules in order to run the import)

anway loom: https://www.loom.com/share/81739a276f4647b09ebdd539f3279317?sid=ac41ebbb-4b6c-4f2d-8517-0f51ada4ac9c

im up for cleaning some of this up as well -- or if there's better suggestions

### DIFF
--- a/src/vellum/workflows/workflows/base.py
+++ b/src/vellum/workflows/workflows/base.py
@@ -435,37 +435,38 @@ class BaseWorkflow(Generic[WorkflowInputsType, StateType], metaclass=_BaseWorkfl
         elif len(workflows) > 1:
             raise ValueError(f"Multiple workflows found in {module_path}")
         try:
-            # Import the nodes package
-            nodes_package = importlib.import_module(f"{module_path}.display.nodes")
-
-            # Use the loader to get the code
-            if hasattr(nodes_package, "__spec__") and nodes_package.__spec__ and nodes_package.__spec__.loader:
-                loader = nodes_package.__spec__.loader
-
-                # Check if the loader has a code attribute
-                if hasattr(loader, "code"):
-                    code = loader.code
-
-                    # Parse the code to find import statements
-                    import_lines = [line.strip() for line in code.splitlines() if line.startswith("from ")]
-
-                    # Import each module specified in the code
-                    for line in import_lines:
-                        try:
-                            # Extract module name from the import line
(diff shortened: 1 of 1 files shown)


src/vellum/workflows/workflows/base.py
```diff
@@ -435,6 +434,36 @@ def load_from_module(module_path: str) -> Type["BaseWorkflow"]:
             raise ValueError(f"No workflows found in {module_path}")
         elif len(workflows) > 1:
             raise ValueError(f"Multiple workflows found in {module_path}")
+        try:
+            # Import the nodes package
+            nodes_package = importlib.import_module(f"{module_path}.display.nodes")
```
dvargasfuertes: [q] hm, I would've expected `importlib.import_module(f"{module_path}.display")` to be sufficient, as that usually imports our `.nodes` module, which in turn imports each node file. Was that not the case in testing? Or is it bc of how our virtual file loader works?

dvargas92495 (approved):
Would love to add a test for this! If there's a difference between hoe `import_module` works for normal python and with the virtual file loader, then maybe we consider moving the virtual file loader to this package so we could create a test for this case (curious on @m-abboud's thoughts)

src/vellum/workflows/workflows/base.py
marlenc3: yeah its due to the file loader! we need to load in the files with the classes themselves to pull in the class instance. Since when we import, it's a "ModuleType" -- and will only have it's __init__.py code inside of the spec loader, despite being a directory. I also couldn't figure out a way to crawl through the virtual file directory from that Module, so this is where I landed. I'd love to simplify it up 🙏  so I can do some more digging later. I also need to double check that running from the CLI either picks up the __init__.py for us, or needs some additional work.

marlenc3:
this works locally- will be fast following with tests, otherwise moving forward to unblock

m-abboud:
> Would love to add a test for this! If there's a difference between hoe `import_module` works for normal python and with the virtual file loader, then maybe we consider moving the virtual file loader to this package so we could create a test for this case (curious on @m-abboud's thoughts)

Ya I can move the virtual file loader here if it helps

PR #542, opened 2025-01-03
add basic test for base node serialization

peicodes:


poetry.lock
peicodes, line 1: Not sure why this file is being updated

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_generic_node_serialization.py
```diff
@@ -0,0 +1,174 @@
...
+            ],
+            "bases": [],
+        },
+    }
+
+    api_node = workflow_raw_data["nodes"][1]
+    assert not DeepDiff(
+        {
+            "id": "c2ed23f7-f6cb-4a56-a91c-2e5f9d8fda7f",
+            "type": "GENERIC",
```
peicodes, line 85: The serialization of the content of generic nodes will be in a subsequent PR

ee/vellum_ee/workflows/display/nodes/vellum/base_node.py
```diff
@@ -0,0 +1,21 @@
...
+from vellum.workflows.types.core import JsonObject
+from vellum_ee.workflows.display.nodes.base_node_vellum_display import BaseNodeVellumDisplay
+from vellum_ee.workflows.display.types import WorkflowDisplayContext
+
+_BaseNodeType = TypeVar("_BaseNodeType", bound=BaseNode)
+
+
+class BaseNodeDisplay(BaseNodeVellumDisplay[_BaseNodeType], Generic[_BaseNodeType]):
+    def serialize(
+        self, display_context: WorkflowDisplayContext, error_output_id: Optional[UUID] = None, **kwargs: Any
```
noanflaherty: Generic nodes won't have the `Reject on Error` toggle, hence we shouldn't allow it as a kwarg

```suggestion
        self, display_context: WorkflowDisplayContext, **kwargs: Any
```

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_generic_node_serialization.py
```diff
@@ -0,0 +1,174 @@
+from deepdiff import DeepDiff
+
+from vellum_ee.workflows.display.workflows import VellumWorkflowDisplay
+from vellum_ee.workflows.display.workflows.get_vellum_workflow_display_class import get_workflow_display
+
+from tests.workflows.basic_generic_node.workflow import BasicGenericNodeWorkflow
+
+
+def test_serialize_workflow(vellum_client):
```
noanflaherty, line 9: We might want to think of a testing pattern that let's us just unit test the node's serialization outside of the context of a full workflow. I suspect there's go na be lots of permutations of generic node functionality that we'll want to unit test and it'll be a pain to create separate workflows for each permutation.

I'd also encourage that we reduce the assertions in here to only assert generic-node-related functionality, rather than workflow functionality like entrypoint nodes, etc.

poetry.lock
noanflaherty: Let's exclude this diff from the pr prior to merge

noanflaherty (approved):

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_generic_node_serialization.py
peicodes: I think if we're doing other permutations then yeah, let's find a way to just test serialization of the node. But I do want at least one full workflow test where we are testing entrypoint nodes and final output nodes

NgoHarrison (approved):
✅

dvargas92495 (approved):

PR #546, opened 2025-01-03
[SC-6050] Add output id and array output id to prompt deployment node

NgoHarrison:


ee/codegen/src/__test__/nodes/multiple-nodes.test.ts
```diff
@@ -74,3 +76,63 @@ describe("InlinePromptNode with Conditional Node", () => {
...
+        inputVariableData: {
+          id: "90c6afd3-06cc-430d-aed1-35937c062531",
+          key: "text",
+          type: "STRING",
+        },
+        workflowContext,
+      })
+    );
+
+    const outputId = "fa015382-7e5b-404e-b073-1c5f01832169";
```
dvargasfuertes: [nit] would be nice for tests to include (either in this one or a new one) references to the `results` output and `error` output

dvargas92495 (approved):

PR #549, opened 2025-01-03
Move virtual File Loader here

m-abboud:


src/vellum/utils/virtual_file_loader.py
```diff
@@ -0,0 +1,42 @@
+import importlib
+
+
+class VirtualFileLoader(importlib.abc.Loader):
```
dvargasfuertes: [nit] Could we leave docstrings in each class specifying that these are only meant for use by vellum? In fact, maybe we put it in the `ee` directory for now...

dvargas92495 (approved):
[oos] as a followup, would love to create a test and fix for the issue referenced here: https://github.com/vellum-ai/vellum-python-sdks/pull/539#discussion_r1901917696

eg, it seems that virtually loading an __init__.py file is not virtually loading its imports

src/vellum/utils/virtual_file_loader.py
m-abboud: Moved to ee/workflows/server. "vellum-use-only" doc string is overkill though, like most we already have tons of public stuff nobody would ever really use by themselves

PR #550, opened 2025-01-03
Abstract TryNode's decorator logic to a separate utility

dvargas92495:
There are nodes in the prototype repo that use retry node, which currently fail node output mocking because retry node and map node's adornment logic is lagging behind Try Node. This PR starts down the path of consolidating the three:
- This PR: Separate out into its own utility method
- Next PR: Cutover RetryNode
- Last PR: Cutover MapNode

There should be little to no logic changes in this PR. Simply moving TryNode's `wrap` logic to its own utility


noanflaherty (commented):
✅

PR #551, opened 2025-01-03
Cutover Retry node to create_adornment utility

dvargas92495:
Part 2 of plan referenced here: https://github.com/vellum-ai/vellum-python-sdks/pull/550

PR #552, opened 2025-01-03
Consolidate Adornable class logic into new BaseAdornableNode

dvargas92495:
Quick detour referenced here: https://github.com/vellum-ai/vellum-python-sdks/pull/551#discussion_r1902219687

With this PR, users in the future will be able to define _their own_ adornable nodes. There should be no logic changes in this PR - simply a consolidation of common logic.

Will await this PR on naming feedback before merging

### DIFF
--- a/src/vellum/workflows/nodes/bases/base_adornable_node.py
+++ b/src/vellum/workflows/nodes/bases/base_adornable_node.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
     from vellum.workflows import BaseWorkflow
 
 
-class BaseAdornableNodeMeta(BaseNodeMeta):
+class _BaseAdornmentNodeMeta(BaseNodeMeta):
     def __new__(cls, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any:
         node_class = super().__new__(cls, name, bases, dct)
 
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/nodes/bases/base_adornable_node.py
```diff
@@ -0,0 +1,78 @@
...
+
+class BaseAdornableNode(
+    BaseNode[StateType],
+    Generic[StateType],
+    metaclass=BaseAdornableNodeMeta,
+):
+    """
+    A base node that makes the node "Adornable" - meaning it can wrap another node. The
+    wrapped node is stored in the `__wrapped_node__` attribute and is redefined as a single
+    node subworkflow.
```
noanflaherty:
```suggestion
    wrapped node is stored in the `__wrapped_node__` attribute and is redefined as a
    single-node subworkflow.
```

noanflaherty (approved):
Neat 👍 

I think the naming you're looking to align on is `BaseAdornableNode`? If so, 🚢

noanflaherty:
Oh wait – Adornable...




I think this means that a node that you can add an adornment to would be considered "adornable" – I think this means we'd want to call this `BaseAdornmentNode`?

PR #553, opened 2025-01-04
Cutover Map node to new BaseAdornmentNode

dvargas92495:
This PR completes our adornment consolidation series. All feature work that we did for Try Nodes should now apply for map and retry node adornments once we start working on them, and in the future, users will be able to create their own adornment nodes


src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -98,7 +97,12 @@ def run(self) -> Outputs:
                     )
         except Empty:
             pass
-        return self.Outputs(**mapped_items)
```
dvargasfuertes, line 101: mypy gave me issues with this syntax. The other two adornment nodes use `setattr` for now

noanflaherty: nit: To me using `return self.Outputs(**mapped_items)  # type: ignore` is less gross than using `setattr`

noanflaherty (commented):
✅

PR #554, opened 2025-01-04
Support multiple mocks for the same node

dvargas92495:
In trying to cutover the prototype repo to using mocks, I noticed that some of the workflows would require mocking the same node multiple times in the context of a loop. This PR introduces a `CycleMap` that enables this functionality by cycling through the mocks defined for a particular node.

### DIFF
--- a/tests/workflows/basic_looping/tests/test_workflow.py
+++ b/tests/workflows/basic_looping/tests/test_workflow.py
@@ -1,4 +1,4 @@
-from tests.workflows.basic_looping.workflow import BasicLoopingWorkflow
+from tests.workflows.basic_looping.workflow import BasicLoopingWorkflow, StartNode
 
 
 def test_workflow__happy_path():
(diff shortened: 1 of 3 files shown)


tests/workflows/basic_looping/tests/test_workflow.py
```diff
@@ -12,3 +12,21 @@ def test_workflow__happy_path():
...
+    # GIVEN a workflow that defines a loop
+    workflow = BasicLoopingWorkflow()
+
+    # WHEN we run the workflow
+    terminal_event = workflow.run(
+        node_output_mocks=[
+            StartNode.Outputs(final_value=5),
+            StartNode.Outputs(final_value=15),
+        ]
+    )
```
noanflaherty, line 27: It'd be cool to think about whether there's some way to make this feel more like pytest's `.return_value` vs `.side_effect`.

noanflaherty (commented):
✅

tests/workflows/basic_looping/tests/test_workflow.py
dvargasfuertes: https://app.shortcut.com/vellum/story/6061

PR #556, opened 2025-01-04
Introduce a constant value reference

dvargas92495:
We've talked a bit about the need of a constant reference. This is a wrapper of constants, compatible with our descriptor language and could serialize/codegen from existing `CONSTANT_VALUE` pointers to and from Vellum.

Looking to use it in some of the mocks work


noanflaherty (commented):
✅

PR #557, opened 2025-01-04
Support node output mocks for nested nodes

dvargas92495:
This PR introduces mocks for nested nodes. There was a bit tricky bc Map nodes execute iterations in parallel, so if you want to support multiple mocks for a node, you'll need:
- something that is race condition resistant
- Allows users to specify the condition by which a mock can be used.

Both of these led this PR to introducing `when_condition`s from the [monorepo](https://github.com/vellum-ai/vellum/blob/8597ab54940b690bd274ab2356b9b9ed7c69b172/django/app/domain_models/types/workflows.py#L1700). This will eventually make it easier for mocks to serialize anyway when it comes time to support mocks from the Sandbox

Gonna keep this PR in review for feedback on user-facing API and naming.


src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
```diff
@@ -36,6 +37,8 @@ def run(self) -> Iterator[BaseOutput]:
             )
             subworkflow_stream = subworkflow.stream(
                 inputs=self._compile_subworkflow_inputs(),
+                event_filter=all_workflow_event_filter,
```
dvargasfuertes, line 80: Some of the changes in this PR were unrelated and were just context related features I noticed while implementing. Will split these out to merge through in the meantime to keep this PR more focused

src/vellum/workflows/runner/runner.py
```diff
@@ -125,7 +126,6 @@ def __init__(
 
         self._dependencies: Dict[Type[BaseNode], Set[Type[BaseNode]]] = defaultdict(set)
         self._state_forks: Set[StateType] = {self._initial_state}
-        self._mocks_by_node_outputs_class = CycleMap(items=node_output_mocks or [], key_by=lambda mock: mock.__class__)
```
dvargasfuertes, line 147: This `CycleMap` class ended up being pretty short lived, due to the read/write race it exhibited

tests/workflows/basic_looping/tests/test_workflow.py
```diff
@@ -21,8 +23,14 @@ def test_workflow__happy_path_with_mocks():
     # WHEN we run the workflow
     terminal_event = workflow.run(
         node_output_mocks=[
```
dvargasfuertes, line 25: important API to 👀 below

tests/workflows/nested_node_mocks/tests/test_workflow.py
```diff
@@ -0,0 +1,32 @@
...
+from tests.workflows.nested_node_mocks.workflow import MapFruitNode, NestedNodeMockWorkflow, NodeToMock
+
+
+def test_run_workflow__happy_path():
+    # GIVEN a workflow that contains a MapNode that contains the actual node we want to mock
+    workflow = NestedNodeMockWorkflow()
+
+    # WHEN the workflow is run with nested mocks
+    terminal_event = workflow.run(
+        node_output_mocks=[
```
dvargasfuertes, line 12: Same here 👀

tests/workflows/basic_looping/tests/test_workflow.py
```diff
@@ -21,8 +23,14 @@ def test_workflow__happy_path_with_mocks():
...
-            StartNode.Outputs(final_value=5),
-            StartNode.Outputs(final_value=15),
+            MockNodeExecution(
+                when_condition=StartNode.Execution.count.equals(0),
+                then_outputs=StartNode.Outputs(final_value=5),
+            ),
+            MockNodeExecution(
+                when_condition=StartNode.Execution.count.equals(1),
+                then_outputs=StartNode.Outputs(final_value=15),
+            ),
```
noanflaherty, line 33: This is where it feels like there's opportunity to lean on patterns established by pytest. In pytest and mock, you have `.return_value = x`, `.side_effect = [x, y]` and `.side_effect=lambda x: x` – three solid APIs for the basic return case, the multiple-execution case, and the arbitrary logic case.

I don't have immediate suggestions for a better API that's closer to this, but sharing in case it inspires something.

dvargas92495:
Planning to merge this through for the following reasons:
- We are not locked out from the DevX discussed [here](https://github.com/vellum-ai/vellum-python-sdks/pull/603) in the future. Those could still serve as "pytest-like" wrappers that compile down to this json
- The current state on main has bugs that this PR will resolve
- This is the version most compatible with what's in the UI rn, which is my primary concern over the next week
- Looking to clear the slate of "High" priority tickets from shortcut so that we could use as a tool this week

PR #558, opened 2025-01-04
Pipe Inline Subworkflow node events back through parent

dvargas92495:
Merging through the context related changes from https://github.com/vellum-ai/vellum-python-sdks/pull/557/files


noanflaherty (commented):
✅

PR #559, opened 2025-01-05
Setup test and implementation for solving pydantic validation errors for fern models

dvargas92495:
We need our pydantic plugin to be registered and loaded before any of the fern generated models are defined.

I think taking this to the extreme, we will need to figure out how to register our plugin upon initial pydantic import, but plugins are still pretty new to pydantic, and this will unblock our most immediate needs.

I will need to follow up this PR with an immediate release cut that includes a postprocessing script change since we are changing logic in fern generated files


noanflaherty (commented):
✅

PR #562, opened 2025-01-06
Fix Graph Generation issue related to multiple branches from the same node

dvargas92495:
```
graph = A >> { B >> C, C }
```

Normally we want to "optimize common targets" here but having each element in the set pop the last node. In this case, popping one of the branches produces a `None`


noanflaherty (commented):
✅

PR #563, opened 2025-01-06
Full implementation for supporting `--dry-run` option on vellum push

dvargas92495:
This PR is meant to align on the CLI experience. Once we have alignment, I then plan to implement the vellum-side and then the client generator updates needed to flesh out the rest of this CLI PR


ee/vellum_cli/tests/test_push.py
```diff
@@ -171,3 +171,75 @@ class ExampleWorkflow(BaseWorkflow):
...
+    assert (
+        result.output
+        == """\x1b[38;20mLoading workflow from examples.mock\x1b[0m
+\x1b[38;20mRan check on Mock\x1b[0m
+"""
+    )
+
+
+# TODO: Implement in Vellum push `check` and output results
+# https://app.shortcut.com/vellum/story/5416
```
dvargasfuertes: This is what we would log to the CLI upon running `--check`. Alternatively, we could write to a `.log` file and output the path to that file here

ee/vellum_cli/push.py
```diff
@@ -50,7 +51,7 @@ def push_command(
     # https://app.shortcut.com/vellum/story/5585
     workflow = BaseWorkflow.load_from_module(workflow_config.module)
     workflow_display = get_workflow_display(base_display_class=VellumWorkflowDisplay, workflow_class=workflow)
-    exec_config = workflow_display.serialize()
+    exec_config = workflow_display.serialize(raise_errors=not check)
```
dvargasfuertes: didn't know we had this already, sweet

ee/vellum_cli/tests/test_push.py
```diff
@@ -171,3 +171,75 @@ class ExampleWorkflow(BaseWorkflow):
...
+#         == """# Workflow Push Report
+
+# ## Errors
+# - `NotSupportedNode`
+#     - Serialization is not supported.
+
+# ## Diffs
```
dvargasfuertes: First half is generated from running `.serialize()` locally. Second half is based on an `expected_diffs: List[ExpectedDiff]` array we get back from the API

ee/vellum_cli/tests/test_push.py
```diff
@@ -171,3 +171,75 @@ class ExampleWorkflow(BaseWorkflow):
...
+# + `AddedNode`
+# ~ `ModifiedNode`
+#     + `ModifiedNode.added_attribute`
+#     ~ `ModifiedNode.modified_attribute`
+#         ~ "Hello, World" -> "Hello, World!"
+#     - `ModifiedNode.removed_attribute`
```
dvargasfuertes: This syntax is very "python"-based. We may need to switch to a more JSON diffing output since the vellum endpoint only has access to past and expected JSONs

ee/vellum_cli/__init__.py
```diff
@@ -49,13 +49,17 @@ def workflows():
 @click.option("--deployment-name", type=str, help="Unique name for the Deployment")
 @click.option("--deployment-description", type=str, help="Description for the Deployment")
 @click.option("--release-tag", type=list, help="Release Tag for the Deployment", multiple=True)
+@click.option(
+    "--check", is_flag=True, help="Check the Workflow for errors and expected changes, without updating Vellum."
```
noanflaherty:
```suggestion
    "--check", is_flag=True, help="Check the Workflow for errors and expected changes, without updating its state in Vellum."
```

ee/vellum_cli/__init__.py
```diff
@@ -79,13 +83,17 @@ def workflows_push(
...
```
noanflaherty:
```suggestion
    "--check", is_flag=True, help="Check the Workflow for errors and expected changes, without updating its state in Vellum."
```

ee/vellum_cli/tests/test_push.py
noanflaherty: I'm generally a fan of returning data in a response and then (oos) using a separate flag to allow the user to opt in to writing logs to a file if they wish

ee/vellum_cli/tests/test_push.py
```diff
@@ -171,3 +171,66 @@ class ExampleWorkflow(BaseWorkflow):
...
+    assert result.exit_code == 0
+
+    # AND we should have called the push API with the dry-run option
+    vellum_client.workflows.push.assert_called_once()
+    call_args = vellum_client.workflows.push.call_args.kwargs
+    assert call_args["dry_run"] is True
+
+    # AND the report should be in the output
+    assert "## Errors" in result.output
+    # assert "Serialization is not supported." in result.output
```
dvargasfuertes, line 234: @NgoHarrison / @noanflaherty I could use help on this assertion.

It works in isolation, but fails in running all tests. Something tells me it has to do with order of display class registration

PR #572, opened 2025-01-08
Support Subworkflow Nodes with inputs as Attributes

dvargas92495:
This PR implements support for the ideal case as specified here: https://vellum-ai.slack.com/archives/C06P13ABK0W/p1735310036893499?thread_ts=1735309195.646639&cid=C06P13ABK0W

Left OOS from this PR is type hint support for these attributes. While this is something worth interrogating and supporting, the full solution would require pyright/mypy-as-ide-plugin support, both of which we are deferring to post marketing release.

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/inline_subworkflow_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/inline_subworkflow_node.py
@@ -2,6 +2,7 @@ from uuid import UUID
 from typing import ClassVar, Dict, Generic, List, Optional, Tuple, Type, TypeVar, cast
 
 from vellum import VellumVariable
+from vellum.workflows.inputs.base import BaseInputs
 from vellum.workflows.nodes import InlineSubworkflowNode
 from vellum.workflows.types.core import JsonObject
 from vellum_ee.workflows.display.nodes.base_node_vellum_display import BaseNodeVellumDisplay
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/nodes/core/inline_subworkflow_node/tests/test_node.py
```diff
@@ -39,3 +39,19 @@ class MyNode(InlineSubworkflowNode):
     assert events == [
         BaseOutput(name="out", value="bar"),
     ]
+
+
+def test_inline_subworkflow_node__support_inputs_as_attributes():
+    # GIVEN a node setup with subworkflow inputs
+    class MyNode(InlineSubworkflowNode):
+        subworkflow = MySubworkflow
+        foo = "bar"
```
dvargasfuertes, line 48: the current implementation supports attributes both on `MyNode` and on `SomeOtherNode(MyNode)`

ee/vellum_ee/workflows/display/nodes/vellum/inline_subworkflow_node.py
```diff
@@ -60,12 +61,25 @@ def _generate_node_and_workflow_inputs(
...
+        subworkflow = raise_if_descriptor(node.subworkflow)
+        subworkflow_inputs_class = subworkflow.get_inputs_class()
         subworkflow_inputs = raise_if_descriptor(node.subworkflow_inputs)
-        subworkflow_entries = (
-            [(variable_name, variable_value) for variable_name, variable_value in subworkflow_inputs.items()]
-            if isinstance(subworkflow_inputs, dict)
-            else [(variable_ref.name, variable_value) for variable_ref, variable_value in subworkflow_inputs]
-        )
+
+        if isinstance(subworkflow_inputs, BaseInputs):
```
dvargasfuertes, line 68: spent some cycles trying to consolidate this with `_compile_subworkflow_inputs`, but one operates on the instance and one on the class, and it wasn't worth converting the former to the latter because of nested node resolution.

marlenc3:
🔥

PR #575, opened 2025-01-08
Support RetryNode.retry_on_condition

dvargas92495:
Allows users to declaratively express any condition to invoke a retry loop

### DIFF
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -324,6 +324,10 @@ class BaseNode(Generic[StateType], metaclass=BaseNodeMeta):
             if not descriptor.instance:
                 continue
 
+            if any(isinstance(t, type) and issubclass(t, BaseDescriptor) for t in descriptor.types):
+                # We don't want to resolve attributes that are _meant_ to be descriptors
+                continue
+
             resolved_value = resolve_value(descriptor.instance, self.state, path=descriptor.name, memo=inputs)
             setattr(self, descriptor.name, resolved_value)
 
--- a/src/vellum/workflows/nodes/core/retry_node/node.py
+++ b/src/vellum/workflows/nodes/core/retry_node/node.py
@@ -1,5 +1,6 @@
 from typing import Callable, Generic, Optional, Type
 
+from vellum.workflows.descriptors.base import BaseDescriptor
 from vellum.workflows.errors.types import WorkflowErrorCode
 from vellum.workflows.exceptions import NodeException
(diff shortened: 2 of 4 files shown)


src/vellum/workflows/nodes/bases/base.py
```diff
@@ -324,6 +324,10 @@ def __init__(
             if not descriptor.instance:
                 continue
 
+            if any(isinstance(t, type) and issubclass(t, BaseDescriptor) for t in descriptor.types):
+                # We don't want to resolve attributes that are _meant_ to be descriptors
+                continue
```
dvargasfuertes, line 329: This is the trickiest part of this PR, but it feels correct

src/vellum/workflows/nodes/displayable/api_node/node.py
```diff
@@ -24,8 +24,8 @@ class APINode(BaseAPINode):
 
     authorization_type: Optional[AuthorizationType] = None
     api_key_header_key: Optional[str] = None
-    api_key_header_value: Optional[Union[str, VellumSecretReference]] = None
-    bearer_token_value: Optional[Union[str, VellumSecretReference]] = None
+    api_key_header_value: Optional[Union[str, VellumSecret]] = None
+    bearer_token_value: Optional[Union[str, VellumSecret]] = None
```
dvargasfuertes, line 28: We can't (shouldn't have ever) use descriptors as a type for a Node Attribute, unless we intend for it to not be resolved by the node

src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -55,18 +58,36 @@ def run(self) -> BaseNode.Outputs:
...
+        retry_on_error_code: Optional[WorkflowErrorCode] = None,
+        retry_on_condition: Optional[BaseDescriptor] = None,
     ) -> Callable[..., Type["RetryNode"]]:
         return create_adornment(
-            cls, attributes={"max_attempts": max_attempts, "retry_on_error_code": retry_on_error_code}
+            cls,
+            attributes={
+                "max_attempts": max_attempts,
+                "retry_on_error_code": retry_on_error_code,
+                "retry_on_condition": retry_on_condition,
```
noanflaherty, line 91: nit: I think we could nix the `retry_` prefix for both of these since it's implied by the fact that it's a `RetryNode`.

noanflaherty (approved):

awlevin:
@VishnuK1947 this will unblock the polling PR you were working on! For context, this approach will be favorable over the infinite while loop because the execution logs in the UI will show the repeated invocations of this node until success

PR #580, opened 2025-01-09
[SC-5606] Generate array output id for inline prompt node context

NgoHarrison:
This issue came from the pdf to image workflow: https://app.vellum.ai/workflow-sandboxes/8dd47593-ce0d-4ddc-a131-005096d18414

The templating node in this workflow was referencing the array output id of the prompt node. However, we currently do not expose that array output id in the prompt node context. This PR adds it to the context so that it can be referenced

### DIFF
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -642,6 +642,8 @@ export function templatingNodeFactory({
   errorOutputId,
   inputRules,
   outputType = VellumVariableType.String,
+  inputReferenceId,
+  inputReferenceNodeId,
 }: {
   id?: string;
   label?: string;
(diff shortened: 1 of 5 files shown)


noanflaherty:
Could you share some context in the PR description for what the goal is here? (maybe I'm just out of the loop)

ee/codegen/src/__test__/nodes/__snapshots__/templating-node.test.ts.snap
```diff
@@ -178,7 +191,9 @@ from vellum.workflows.nodes.core import TryNode
 
 @TryNode.wrap()
 class TemplatingNode(BaseTemplatingNode[BaseState, str]):
-    template = """Hello, world!"""
-    inputs = {}
+    template = """{{ output[0].type }}"""
+    inputs = {
+        "text": "Hello, world!",
+    }
```
noanflaherty: nit: this looks like an invalid template given these inputs.

noanflaherty (approved):

ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -650,30 +652,59 @@ export function templatingNodeFactory({
...
+        id: "7b8af68b-cf60-4fca-9c57-868042b5b616",
+        key: "template",
+        value: {
+          rules: [
             {
               type: "CONSTANT_VALUE",
               data: {
                 type: "STRING",
-                value: "Hello, world!",
+                value: "{{ output[0].type }}",
```
dvargasfuertes: [blocking] It seems like there there a bunch of unrelated tests that changed bc of this change.

I would prefer if we reverted this and for the new test to pass in an optional `template` option - in general, I would only expect one test to be updated/change with a feature/bug fix like this

NgoHarrison: Updated the factory to take a template option. The other tests are being updated because I changed a lot of the ids that the old factory had. The old factory had id values for templating node and inline prompt node (I think I copied pasted these values back then). It worked by themselves but now that I am using both of them together, it caused some errors

ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -650,33 +653,63 @@ export function templatingNodeFactory({
...
+    id: id ?? "46e221ab-a749-41a2-9242-b1f5bf31f3a5",
     type: WorkflowNodeType.TEMPLATING,
     data: {
       label: label ?? "Templating Node",
       outputId: "2d4f1826-de75-499a-8f84-0a690c8136ad",
       errorOutputId,
-      sourceHandleId: sourceHandleId ?? "dd8397b1-5a41-4fa0-8c24-e5dffee4fb98",
-      targetHandleId: targetHandleId ?? "3feb7e71-ec63-4d58-82ba-c3df829a2948",
+      sourceHandleId: sourceHandleId ?? "6ee2c814-d0a5-4ec9-83b6-45156e2f22c4",
+      targetHandleId: targetHandleId ?? "3960c8e1-9baa-4b9c-991d-e399d16a45aa",
```
dvargasfuertes, line 676: [nit] would prefer if we reverted these too so that we minimize blast radius on other tests as aggressively as possible

dvargas92495 (approved):

PR #581, opened 2025-01-09
Virtual File Testing - Displays pt 2

marlenc3:
A few things I found + fixed, but we should be importing all files needed as expected now!

### DIFF
--- a/ee/vellum_ee/workflows/server/virtual_file_loader.py
+++ b/ee/vellum_ee/workflows/server/virtual_file_loader.py
@@ -26,13 +26,13 @@ class VirtualFileFinder(importlib.abc.MetaPathFinder, importlib.abc.Loader):
         key_name = "__init__" if fullname == self.namespace else fullname.replace(f"{self.namespace}.", "")
 
         files_key = f"{key_name.replace('.', '/')}.py"
-        if not self.files.get(files_key):
+        if self.files.get(files_key) is None:
             files_key = f"{key_name.replace('.', '/')}/__init__.py"
 
         file = self.files.get(files_key)
         is_package = "__init__" in files_key
 
-        if file:
+        if file is not None:
             return importlib.machinery.ModuleSpec(
                 prefixed_name,
                 VirtualFileLoader(file, is_package),
--- a/ee/vellum_ee/workflows/server/virtual_file_loader.py
+++ b/ee/vellum_ee/workflows/server/virtual_file_loader.py
@@ -10,7 +10,7 @@ class VirtualFileLoader(importlib.abc.Loader):
         return None  # use default module creation
 
     def exec_module(self, module):
-        if not self.is_package:
+        if not self.is_package or self.is_package and self.code:
             exec(self.code, module.__dict__)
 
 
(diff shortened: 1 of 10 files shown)


tests/server/tests/test_virtual_files.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Track sys.modules keys before import
+    existing_modules = set(sys.modules.keys())
+
+    # Import the base module (should implicitly load its dependencies)
+    base_class_module = importlib.import_module(full_module_name)
+    assert base_class_module, f"Failed to import module: {full_module_name}"
+
+    baseclass = None
+    for name in dir(base_class_module):
```
marlenc3, line 48: same logic we use in load_from_module

ee/vellum_ee/workflows/server/virtual_file_loader.py
```diff
@@ -10,7 +10,7 @@ def create_module(self, spec):
         return None  # use default module creation
 
     def exec_module(self, module):
-        if not self.is_package:
+        if not self.is_package or self.is_package and self.code:
```
marlenc3, line 13: unsure why we don't exec __init__ files - since we do serialize imports there but let me know if there was a concern here for it!

marlenc3: @m-abboud lemme know if there are concerns with this change!

ee/vellum_ee/workflows/server/virtual_file_loader.py
```diff
@@ -26,13 +26,13 @@ def find_spec(self, fullname, path, target=None):
         key_name = "__init__" if fullname == self.namespace else fullname.replace(f"{self.namespace}.", "")
 
         files_key = f"{key_name.replace('.', '/')}.py"
-        if not self.files.get(files_key):
+        if self.files.get(files_key) is None:
```
marlenc3, line 29: if an init file that is empty is found - not sure if we want to also load it in, but added in case, will iterate and see if really needed

ee/vellum_ee/workflows/server/virtual_file_loader.py
dvargasfuertes: [nit] the `or self.is_package` is redundant

ee/vellum_ee/workflows/server/virtual_file_loader.py
dvargasfuertes: I'm on board with this change

src/vellum/workflows/workflows/base.py
```diff
@@ -437,40 +437,8 @@ def load_from_module(module_path: str) -> Type["BaseWorkflow"]:
...
             raise ValueError(f"Multiple workflows found in {module_path}")
-        try:
-            BaseWorkflow.import_node_display(module_path)
-        except ModuleNotFoundError:
-            pass
-
         return workflows[0]
 
-    @staticmethod
-    def import_node_display(module_path):
```
dvargasfuertes, line 448: [q] did we confirm with an example that this still worked

dvargas92495 (approved):
Great work!

src/vellum/workflows/workflows/base.py
marlenc3: locally! working with this branch to test adornment nodes

PR #582, opened 2025-01-09
Improved Jinja Template Serialization of pydantic objects

dvargas92495:
@awlevin ran into a bug yesterday building a workflow that will also affect our SDK. The first two commits are the test and implementations respectively to solve that bug, which I also plan to apply to Vellum.

The third commit was a mypy issue that I also investigated yesterday, that came up bc of the new test, but ended up working around it in the prototype repo. Basically Templating Node Outputs are not being inferred with the correct Outputs class (since it's auto amended in the metaclass).

I made good headway in making the Outputs class correct, but attributes are still typed incorrectly, so am deferring further investigation.


noanflaherty (commented):
✅

marlenc3:
✅

PR #588, opened 2025-01-10
Update Serialization & Codegen to Remove WorkflowNodeDefinition

noanflaherty:
Sibling PR: https://github.com/vellum-ai/vellum/pull/7235

This PR updates serialization and codegen to ensure that we're operating with the new data model. Note that this is likely to introduce breaking changes and will warrant a bump in our sdk versions.

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -57,14 +57,16 @@ class BaseNodeDisplay(Generic[NodeType], metaclass=BaseNodeDisplayMeta):
     def get_base(self) -> CodeResourceDefinition:
         node = self._node
 
-        bases = [
-            CodeResourceDefinition(
-                name=base.__name__,
-                module=base.__module__.split("."),
-            )
-            for base in node.__bases__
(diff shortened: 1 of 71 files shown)


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -54,18 +54,23 @@ class BaseNodeDisplay(Generic[NodeType], metaclass=BaseNodeDisplayMeta):
...
-        node_definition = NodeDefinition(
+
+        bases = [
+            CodeResourceDefinition(
+                name=base.__name__,
+                module=base.__module__.split("."),
+            )
+            for base in node.__bases__
+        ]
+        return bases[0]
```
noanflaherty: There might be a more robust check we want to make here rather than blindly grabbing the first. I can follow up in a separate PR if we feel we want this.

dvargasfuertes: yea something like `next((base for base in node.__bases__ if issubclass(base, BaseNode)))`

dvargas92495 (approved):

PR #603, opened 2025-01-11
Explore pytest-esque patterns for mocks

dvargas92495:
This PR was split off from [this one](https://github.com/vellum-ai/vellum-python-sdks/pull/557) to show an alternative approach to how we could implement node mocking, based on [this comment](https://github.com/vellum-ai/vellum-python-sdks/pull/557/files#r1903178062). If it's preferred, I can flesh out implementation, but going through it gave me some concerns which I'll specify inline


tests/workflows/basic_looping/tests/test_workflow.py
```diff
@@ -20,19 +18,13 @@ def test_workflow__happy_path_with_mocks():
     # GIVEN a workflow that defines a loop
     workflow = BasicLoopingWorkflow()
 
+    StartNode.Outputs._mock_side_effect = [
```
dvargasfuertes: multi-execution case. Besides the odd double `StartNode.Outputs` reference, this has the same read/write race condition as the previous solution, which will require us to carry mock information in `State`

tests/workflows/basic_node_mocking/tests/test_workflow.py
```diff
@@ -5,13 +5,14 @@ def test_workflow__happy_path():
...
 
-    # WHEN we run the workflow with a mock defined
-    final_event = workflow.run(
-        node_output_mocks=[
-            StartNode.Outputs(greeting="Hello"),
-        ]
+    # AND we mock the node's output
+    StartNode.Outputs._mock_return_value(
+        greeting="Hello",
     )
```
dvargasfuertes: The single case. Because we support mocks in sandboxes, implementing mocks off of the node like this (in all three cases) means we'll need to implement additional logic in Vembda instead of just passing the request data through to the `.run()` method.

tests/workflows/nested_node_mocks/tests/test_workflow.py
```diff
@@ -1,30 +1,19 @@
...
-            MockNodeExecution(
-                when_condition=MapFruitNode.SubworkflowInputs.index.equals(1),
-                then_outputs=NodeToMock.Outputs(result="eggplant"),
-            ),
-            MockNodeExecution(
-                when_condition=MapFruitNode.SubworkflowInputs.index.equals(2),
-                then_outputs=NodeToMock.Outputs(result="fig"),
-            ),
-        ]
+    NodeToMock.Outputs._mock_side_effect = lambda inputs: (
```
dvargasfuertes: The arbitrary logic case. Because this is a "custom method", a case that was previously serializable by the UI no longer is.

tests/workflows/basic_looping/tests/test_workflow.py
noanflaherty: FWIW I might have expected this to live on the node itself rather than its outputs class. So something like:

```python
StartNode._mock_side_effect = [...]
```

But in general, it feels like an anti-pattern to define mocks on the thing you're mocking directly. This is why python has the `mock` library and `patch`, whose return value you define `side_effect` and `return_value` attributes for.

The more pytest-ey way of doing things might be more like:

```python
with patch_node(StartNode) as patched_start_node:
    patched_start_node.side_effect = [
        StartNode.Outputs(final_value=5),
    ]
```

dvargas92495:
Iterated on the design based on the `with patch():` approach. I think all three of my comments tho still hold true - I'd like to come resolution during the checkin tn so we could move forward to UI integration

tests/workflows/nested_node_mocks/tests/test_workflow.py
noanflaherty: I'd be interested to see if there are any meaningful workflows out in the wild actually making use of this feature today. If not or negligible, then this wouldn't be a significant downside in my mind.

tests/workflows/basic_node_mocking/tests/test_workflow.py
noanflaherty: While true, this doesn't surprise me in the world of mocking and would be more consistent with how mocking in most languages/frameworks work, making this not a big concern in my mind unless the scope needed to do so is huge.

tests/workflows/basic_looping/tests/test_workflow.py
noanflaherty: This becomes a non-concern to me with the patch-based approach.

tests/workflows/nested_node_mocks/tests/test_workflow.py
dvargasfuertes: We never built UI support for it so the answer is a definitive no

tests/workflows/basic_looping/tests/test_workflow.py
dvargasfuertes: This was the concern:
> this has the same read/write race condition as the previous solution, which will require us to carry mock information in State

There may be a way around it though using execution counts

tests/workflows/basic_node_mocking/tests/test_workflow.py
dvargasfuertes: > unless the scope needed to do so is huge

That's the concern

PR #606, opened 2025-01-12
Support a concurrency parameter to define max number of threads in workflow runs

dvargas92495:
Workflows are currently hard to debug during runs that have parallelism, since multiple breakpoints will trigger at once. This PR provides a way to simplify this, as well as provide other rate-limiting benefits by supporting a `concurrency` arg. 

This argument controls the maximum number of node threads running at once. In terms of naming, it was meant to match the `MapNode.concurrency` attribute but am open to feedback on both.

### DIFF
--- a/src/vellum/workflows/workflows/base.py
+++ b/src/vellum/workflows/workflows/base.py
@@ -188,7 +188,7 @@ class BaseWorkflow(Generic[WorkflowInputsType, StateType], metaclass=_BaseWorkfl
         external_inputs: Optional[ExternalInputsArg] = None,
         cancel_signal: Optional[ThreadingEvent] = None,
         node_output_mocks: Optional[List[BaseOutputs]] = None,
-        concurrency: Optional[int] = None,
+        max_concurrency: Optional[int] = None,
     ) -> TerminalWorkflowEvent:
         """
         Invoke a Workflow, returning the last event emitted, which should be one of:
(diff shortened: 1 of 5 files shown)


src/vellum/workflows/workflows/base.py
```diff
@@ -215,6 +216,10 @@ def run(
 
         node_output_mocks: Optional[List[Outputs]] = None
             A list of Outputs to mock for Nodes during Workflow Execution.
+
+        concurrency: Optional[int] = None
+            The number of concurrent threads to run the Workflow with. If not provided, the Workflow will run without
```
noanflaherty:
```suggestion
            The max number of concurrent threads to run the Workflow with. If not provided, the Workflow will run without
```
```suggestion
            The number of concurrent threads to run the Workflow with. If not provided, the Workflow will run without
```

src/vellum/workflows/workflows/base.py
```diff
@@ -215,6 +216,10 @@ def run(
 
         node_output_mocks: Optional[List[Outputs]] = None
             A list of Outputs to mock for Nodes during Workflow Execution.
+
+        concurrency: Optional[int] = None
```
noanflaherty:
```suggestion
        max_concurrency: Optional[int] = None
```

noanflaherty (approved):
I’d be a fan of renaming this to max_concurrency throughout to make clear that this is the maximum amount and not the definitive amount.

it would also be nice to update help text to make clear whether this is across a single level of the workflow stack, or across all sub workflows too (I.e. do two map nodes share the same max concurrency of the parent? Ideally our doc strings make this clear one way or another)

dvargas92495:
cool will rename the map node param in a fast followup since there are codegen/serialization implications. Will rename the param to workflow.run here

PR #607, opened 2025-01-12
Add concurrency to adornment parameters

dvargas92495:
Noticed that we were missing the existing Map Node concurrency attribute from the adornment - this PR quickly adds support for that parameter


noanflaherty (commented):
As mentioned in the other PR, I’d be a fan of renaming to max_concurrency.

PR #608, opened 2025-01-12
Fix infer types in future versions of python

dvargas92495:
I was using a `list[str]` output in a vargas jr workflow (which is on python 3.13) and was getting a failure to run bc of `infer_types()` method. This PR is the fix


noanflaherty (commented):
✅

NgoHarrison (commented):
✅

PR #612, opened 2025-01-13
Replace all instances of throw new Error in codegen with a BaseCodegenError

dvargas92495:
This PR makes it so that all errors that we are expected now go through our error logging system and are handled appropriately. All new errors that we raise going forward should follow suit


NgoHarrison (approved):
✅

PR #613, opened 2025-01-13
Operator precendence tests

peicodes:
This PR proves that we don't need grouped expressions to handle operator and parenthesis precendence while we are using just `&` and `|` operators.


dvargas92495 (approved):

noanflaherty (commented):
Nice!

PR #616, opened 2025-01-14
[Codegen][Part 1] Add outputs to generic node serializer

NgoHarrison:
Fast follow to previous generic node serializer. This PR adds outputs to serializer

### DIFF
--- a/ee/codegen/src/types/vellum.ts
+++ b/ee/codegen/src/types/vellum.ts
@@ -656,7 +656,7 @@ export interface NodeOutput {
   id: string;
   name: string;
   type: VellumVariableType;
-  value: WorkflowValueDescriptor;
+  value?: WorkflowValueDescriptor;
 }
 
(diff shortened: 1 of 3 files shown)


ee/codegen/src/types/vellum.ts
```diff
@@ -652,6 +652,13 @@ export interface GenericNodeDisplayData {
   position?: { x: number; y: number };
 }
 
+export interface NodeOutput {
+  id: string;
+  name: string;
+  type: VellumVariableType;
+  value: WorkflowValueDescriptor;
```
noanflaherty: Value should be optional and nullable here

noanflaherty (approved):
Looks good pending one change

ee/codegen/src/types/vellum.ts
NgoHarrison: Ah good catch

dvargas92495 (commented):
✅

PR #617, opened 2025-01-14
Setup test and implementation for streaming events from Map Nodes

dvargas92495:
This PR aims to replicate these streaming events from Map nodes today: https://github.com/vellum-ai/vellum/blob/main/django/app/core/workflows/runner/nodes/map.py#L281-L320

These streaming events help enable our UIs that allow paginating across iterations and opening up the execution details for a given execution

### DIFF
--- a/tests/workflows/basic_map_node/tests/test_workflow.py
+++ b/tests/workflows/basic_map_node/tests/test_workflow.py
@@ -65,7 +65,7 @@ def test_map_node_streaming_events():
     assert len(node_fulfilled) == 3
 
     # Check first iteration
-    first_event = node_fulfilled[0]
+    first_event = next(e for e in node_fulfilled if e.span_id == node_initiated[1].span_id)
     assert isinstance(first_event, NodeExecutionFulfilledEvent)
     assert first_event.outputs.count == len("apple")  # 5
     assert first_event.parent is not None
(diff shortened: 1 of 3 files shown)


tests/workflows/basic_map_node/tests/test_workflow.py
```diff
@@ -118,3 +113,37 @@ def test_map_node_streaming_events():
...
+    assert node_streaming[3].output.is_streaming
+    assert node_streaming[3].output.name == "count"
+    assert node_streaming[4].output.is_streaming
+    assert node_streaming[4].output.name == "count"
+    assert {
+        node_streaming[3].output.delta,
+        node_streaming[4].output.delta,
+    } == {
+        (5, 0, "FULFILLED"),
+        (7, 1, "FULFILLED"),
```
dvargasfuertes, line 144: Most important part of the PR to 👀 , along with lines 131-132.

Legacy workflows emit three pieces of info with each map node streaming event - the output, the index, and whether the iteration just started or finished.

I'm hopeful that with subworkflows events now just streaming up too, that these node streaming events are actually redundant with their corresponding subworkflow's initiated and fulfilled events, but am including for now to ease the cutover for now.

noanflaherty: Do we want to change this to `BaseMapNode` and introduce a new `MapNode` that extends it in `displayable/`? This is how we've handled cases where we need to introduce functionality purely for backwards compatibility's sake and not because "it's how we actually want it to work" sake for other node types. Defer to you ultimately.

noanflaherty (approved):

tests/workflows/basic_map_node/tests/test_workflow.py
dvargasfuertes: "how we actually want it to work" it's not yet clear to me - as a consumer, it's useful to get index back with each streaming chunk. So am opting to focus on simplifying QA backwards-compat for now, and a basemapnode could always be split out later if there's some signal of user demand for it

noanflaherty: Sounds good 👍

marlenc3:
🔥

PR #623, opened 2025-01-14
Implement `--strict` on vellum push CLI

dvargas92495:
Finishes end to end support for the `--strict` parameter on `vellum push`, which will detect diffs in our serialization and codegen, reporting back any differences.

### DIFF
--- a/ee/vellum_cli/push.py
+++ b/ee/vellum_cli/push.py
@@ -9,6 +9,7 @@ from typing import List, Optional
 
 from dotenv import load_dotenv
 
+from vellum.client.core.api_error import ApiError
 from vellum.resources.workflows.client import OMIT
 from vellum.types import WorkflowPushDeploymentConfigRequest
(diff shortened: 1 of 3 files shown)


ee/vellum_cli/push.py
```diff
@@ -109,18 +111,66 @@ def push_command(
...
+            artifact=artifact,
+            # We should check with fern if we could auto-serialize typed object fields for us
+            # https://app.shortcut.com/vellum/story/5568
+            deployment_config=deployment_config_serialized,  # type: ignore[arg-type]
+            dry_run=dry_run,
+            strict=strict,
+        )
+    except ApiError as e:
+        if e.status_code == 400 and isinstance(e.body, dict) and "diffs" in e.body:
+            diffs: dict = e.body["diffs"]
```
noanflaherty: nit we could unnest a lot of code if we checked for the inverse case and raise

noanflaherty (approved):
🚀

PR #624, opened 2025-01-14
Serialize input variable name with override

NgoHarrison:
This PR is to address the bug when we have a capitalized input variable (e.g `Foo`). We currently codegen it to be lowercased and we then serialize it using the lowercased naming. This PR makes it so that if there is an override (which is going to be the raw name), we use that during serialization


dvargas92495:
going to sync on this one offline first

dvargas92495 (approved):
While this doesn't fix the bug at hand, it fixed another related bug in our serialization!

PR #625, opened 2025-01-15
[Codegen][Part 2] Codegen triggers of generic nodes

NgoHarrison:
This PR aims to codegen for triggers of generic nodes. Subsequent PRs will have the other attributes


ee/codegen/src/generators/node-trigger.ts
```diff
@@ -0,0 +1,54 @@
+import { python } from "@fern-api/python-ast";
+import { AstNode } from "@fern-api/python-ast/core/AstNode";
+import { Writer } from "@fern-api/python-ast/core/Writer";
+
+import { NodeTrigger as NodeTriggerType } from "src/types/vellum";
+
+export declare namespace NodeTrigger {
+  export interface Args {
+    nodeTrigger: NodeTriggerType;
+    baseClassRef: python.Reference;
```
noanflaherty: Why do we need to pass this ref as an arg? Any time we pass an AstNode as an arg to another AstNode I get spooked because it's probably a sign that we should be using contexts instead.

ee/codegen/src/generators/node-trigger.ts
```diff
@@ -0,0 +1,54 @@
...
+  private nodeTrigger: NodeTriggerType;
+  private baseClassRef: python.Reference;
+  private astNode: AstNode;
+
+  public constructor(args: NodeTrigger.Args) {
+    super();
+
+    this.nodeTrigger = args.nodeTrigger;
+    this.baseClassRef = args.baseClassRef;
+    this.astNode = this.constructNodeTrigger();
```
noanflaherty: nit: Rather than saving these to `this`, we could just pass as args to `constructNodeTrigger` since nothing else will depend on them.

ee/codegen/src/generators/node-trigger.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  private constructNodeTrigger(): AstNode {
+    const clazz = python.class_({
+      name: "NodeTrigger",
+      extends_: [
+        python.reference({
+          name: this.baseClassRef.name,
+          modulePath: this.baseClassRef.modulePath,
+          alias: this.baseClassRef.alias,
```
noanflaherty: [Blocking] This looks a lot like we should be using Node Context to get this info, not the reference itself.

ee/codegen/src/generators/node-trigger.ts
```diff
@@ -0,0 +1,54 @@
...
+          attribute: ["Trigger"],
+        }),
+      ],
+    });
+
+    clazz.add(
```
noanflaherty, line 47: nit: `python.class_` now supports a `statements` arg in its constructor, so you could pass right away without having to call `.add` afterwards.

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -11,11 +12,18 @@ export class GenericNode extends BaseSingleFileNode<
...
+        nodeTrigger: this.nodeData.trigger,
+        baseClassRef: this.getNodeBaseClass(),
+      })
+    );
     return statements;
   }
 
   getNodeDisplayClassBodyStatements(): AstNode[] {
     const statements: AstNode[] = [];
+    statements.push();
```
noanflaherty: Remove this?

ee/codegen/src/generators/node-trigger.ts
```diff
@@ -0,0 +1,54 @@
...
+import { NodeTrigger as NodeTriggerType } from "src/types/vellum";
+
+export declare namespace NodeTrigger {
+  export interface Args {
+    nodeTrigger: NodeTriggerType;
+    baseClassRef: python.Reference;
+  }
+}
+
+export class NodeTrigger extends AstNode {
```
noanflaherty, line 15: Can we add a unit test for this?

noanflaherty (changes requested):
I'd like to give another pass once we're using contexts and have a unit test

ee/codegen/src/generators/node-trigger.ts
NgoHarrison: Good point, used node context for this

ee/codegen/src/generators/node-trigger.ts
NgoHarrison: I don't see this still? I tried npm installing but still have this as the fern constructor:
`constructor({ docs, name, extends_, decorators }: Class.Args);`

ee/codegen/src/generators/node-trigger.ts
NgoHarrison: Added

noanflaherty (approved):
Love it!

ee/codegen/src/__test__/nodes/__snapshots__/generic-node.test.ts.snap
```diff
@@ -19,6 +19,7 @@ exports[`GenericNode > basic > getNodeFile 1`] = `
 
 
 class MyCustomNode(BaseNode):
-    pass
+    class NodeTrigger(BaseNode.Trigger):
+        merge_behavior = "AWAIT_ALL"
```
dvargasfuertes, line 23: [oos] `MergeBehavior` is currently an enum: https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/nodes/bases/base.py#L241

We likely should update it on the SDK side to be a literal of three string options

dvargas92495 (approved):

PR #626, opened 2025-01-15
Unblock pushes of generic nodes

dvargas92495:
Was working on pushing this workflow to the UI: https://github.com/vellum-ai/workflows-as-code-runner-prototype/pull/577

All that was missing was port.name and outputs array, so I edited the pip package directly to unblock that above. Copy pasting those changes upstream in the form of this PR


noanflaherty (commented):
✅

PR #627, opened 2025-01-15
Serialize node attributes

peicodes:


ee/vellum_ee/workflows/display/nodes/vellum/base_node.py
```diff
@@ -26,6 +26,19 @@ def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> J
         node = self._node
         node_id = self.node_id
 
+        attributes: JsonArray = []
+        for name in node.__dict__.keys():
+            if name not in BaseNode.__dict__:
```
dvargasfuertes: [nit] you should be able to do `for attribute in node:` and `attribute` will be a `NodeReference` (a descriptor that references an attribute)

dvargas92495 (approved):

ee/vellum_ee/workflows/display/nodes/vellum/base_node.py
peicodes: @dvargas92495 I did this but I had to add a cast because `attribute.instance` is not typed as a `BaseDescriptor`

noanflaherty (commented):
I appreciate the thorough tests!

It'd be cool to have one that ensure we throw the expected exception if trying to serialize an attribute with a constant value whose type is not one we know how to serialize (e.g. a custom dataclass type)

NgoHarrison (commented):
✅

ee/vellum_ee/workflows/display/nodes/vellum/base_node.py
dvargasfuertes: That's because it shouldn't be - it can be Any:
`foo = "string"` -> constant values
`foo = MyNode.Outputs.foo` -> node reference

PR #628, opened 2025-01-15
Codegen fails container serialization

dvargas92495:
runner configs can now be empty objects

### DIFF
--- a/ee/codegen/src/serializers/vellum.ts
+++ b/ee/codegen/src/serializers/vellum.ts
@@ -1995,9 +1995,12 @@ export const WorkflowVersionExecConfigSerializer: ObjectSchema<
     objectSchema({
       containerImageName: propertySchema(
         "container_image_name",
-        stringSchema()
+        stringSchema().optional()
       ),
-      containerImageTag: propertySchema("container_image_tag", stringSchema()),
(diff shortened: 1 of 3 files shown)


ee/codegen/src/serializers/vellum.ts
```diff
@@ -1995,9 +1995,12 @@ export const WorkflowVersionExecConfigSerializer: ObjectSchema<
     objectSchema({
```
dvargasfuertes, line 1995: can we actually remove this field? we don't use this anywhere in codegen right?

m-abboud: My vote would be yes since we decided all this lives in the toml and I didn't even realize this was here.

m-abboud (approved):

ee/codegen/src/serializers/vellum.ts
dvargasfuertes: to be clear, I was simply advocating removing it from codegen request handling and not from the sandbox yet. I'm pretty sure we need it in the sandbox for custom images

NgoHarrison (commented):
✅

NgoHarrison (commented):
✅

PR #629, opened 2025-01-15
CD for codegen service codegen package versions

m-abboud:
Local run example: https://github.com/vellum-ai/codegen-service/pull/58

### DIFF
--- a/ee/codegen/package.json
+++ b/ee/codegen/package.json
@@ -28,7 +28,7 @@
     "types": "tsc --noEmit",
     "start": "node --loader ts-node/esm --experimental-specifier-resolution=node src/index.ts",
     "version:codegen": "node scripts/version.js",
-    "upgrade-codegen-service": "ts-node-esm --project tsconfig.json scripts/upgrade-codegen-service.mts"
+    "upgrade-codegen-service": "node --loader ts-node/esm --experimental-specifier-resolution=node  scripts/upgrade-codegen-service.mts"
   },
   "dependencies": {
     "@fern-api/python-ast": "^0.0.20",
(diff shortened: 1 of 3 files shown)


NgoHarrison (approved):

ee/codegen/package.json
```diff
@@ -28,7 +28,7 @@
     "types": "tsc --noEmit",
     "start": "node --loader ts-node/esm --experimental-specifier-resolution=node src/index.ts",
     "version:codegen": "node scripts/version.js",
-    "upgrade-codegen-service": "ts-node-esm --project tsconfig.json scripts/upgrade-codegen-service.mts"
+    "upgrade-codegen-service": "node --loader ts-node/esm --experimental-specifier-resolution=node  scripts/upgrade-codegen-service.mts"
```
m-abboud, line 31: Works with the nvm version included with the package

ee/codegen/src/index.ts
```diff
@@ -1 +1,3 @@
 export * as codegen from "./codegen";
+export * as errors from "src/generators/errors";
```
m-abboud, line 2: Included additional exports here so it just needs to do 1 module lookup since we now need errors too

ee/codegen/package.json
dvargasfuertes: I hope all of this is solved by latest noe and we can just skip to that

ee/codegen/scripts/upgrade-codegen-service.mts
```diff
@@ -29,17 +30,63 @@ const main = async () => {
...
+    const githubToken = await getGithubToken();
+    const repoUrl = `https://x-access-token:${githubToken}@github.com/vellum-ai/codegen-service.git`;
+    execSync(`git clone ${repoUrl} ${targetDir}`, { stdio: "inherit" });
+  }
 
   process.chdir(targetDir);
   execSync(`npm run gar-login`, { stdio: "inherit" });
   execSync(`npm install @vellum-ai/vellum-codegen@${version} --save-exact`, {
     stdio: "inherit",
   });
```
dvargasfuertes, line 48: [q] are we keeping this long term or plan to remove it once all of the versioning stuff is all setup?

dvargas92495 (commented):
✅

PR #630, opened 2025-01-15
Serialize node outputs

peicodes:
Includes tests for annotated, workflow input, and node output variations

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/base_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/base_node.py
@@ -1,5 +1,6 @@
 from typing import Any, Generic, TypeVar, cast
 
+from vellum.workflows.constants import UNDEF
 from vellum.workflows.descriptors.base import BaseDescriptor
 from vellum.workflows.expressions.between import BetweenExpression
 from vellum.workflows.expressions.is_not_null import IsNotNullExpression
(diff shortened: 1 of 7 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/base_node.py
```diff
@@ -62,6 +64,24 @@ def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> J
                     }
                 )
 
+        outputs: JsonArray = []
+        for output in node.Outputs:
+            type = primitive_type_to_vellum_variable_type(output)
+            value = (
+                self.serialize_value(display_context, output.instance)
+                if output.instance is not None and output.instance != UNDEF
```
noanflaherty, line 72: I could see us introducing an `is_nil` helper to do this check as I suspect we'll be doing this a lot.

```suggestion
                if output.instance is not None and output.instance != UNDEF
```

noanflaherty (approved):

ee/vellum_ee/workflows/display/nodes/vellum/base_node.py
peicodes: I agree it's a good helper to have. Surprisingly though, this is only the second place in the code where we do a `==` or `!=` check against UNDEF :O

ee/vellum_ee/workflows/display/nodes/vellum/base_node.py
```diff
@@ -62,6 +64,24 @@ def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> J
...
+                else None
+            )
+
+            outputs.append(
+                {
+                    "id": str(uuid4_from_hash(f"{node_id}|{output.name}")),
```
dvargasfuertes, line 78: [q] does this not need to look up output displays for id's?

dvargas92495 (commented):
✅

PR #645, opened 2025-01-16
[Codegen][Part 3] Add basic node ports codegen without expression support

NgoHarrison:
This PR aims to add basic support for codegen'ing node ports on generic nodes. Will do a fast follow to add that expression support which will complete node ports functionality on codegen side.

### DIFF
--- a/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
@@ -23,6 +23,6 @@ class MyCustomNode(BaseNode):
         merge_behavior = "AWAIT_ALL"
 
     class Ports(BaseNode.Ports):
-        default_port = Port.on_else(None)
+        pass
 "
 `;
--- a/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
@@ -21,5 +21,8 @@ exports[`GenericNode > basic > getNodeFile 1`] = `
 class MyCustomNode(BaseNode):
     class NodeTrigger(BaseNode.Trigger):
         merge_behavior = "AWAIT_ALL"
+
+    class Ports(BaseNode.Ports):
+        default_port = Port.on_else(None)
 "
 `;
--- a/ee/codegen/src/__test__/nodes/generic-nodes/node-ports.test.ts
+++ b/ee/codegen/src/__test__/nodes/generic-nodes/node-ports.test.ts
@@ -22,11 +22,6 @@ describe("NodePorts", () => {
     beforeEach(async () => {
       const nodePortsData: NodePort[] = [
         {
-          type: "DEFAULT",
-          id: "port-1",
-          name: "default_port",
-        },
-        {
           type: "IF",
           id: "port-2",
           name: "if_port",
--- a/ee/codegen/src/__test__/nodes/generic-nodes/node-ports.test.ts
+++ b/ee/codegen/src/__test__/nodes/generic-nodes/node-ports.test.ts
@@ -1,0 +1,71 @@
+import { Writer } from "@fern-api/python-ast/core/Writer";
+import { beforeEach, describe, expect, it } from "vitest";
+
+import { workflowContextFactory } from "src/__test__/helpers";
+import { genericNodeFactory } from "src/__test__/helpers/node-data-factories";
+import { createNodeContext, WorkflowContext } from "src/context";
+import { GenericNodeContext } from "src/context/node-context/generic-node";
+import { NodePorts } from "src/generators/node-port";
+import { NodePort } from "src/types/vellum";
+
+describe("NodePorts", () => {
+  let workflowContext: WorkflowContext;
+  let writer: Writer;
+  let nodePorts: NodePorts;
+
+  beforeEach(() => {
+    workflowContext = workflowContextFactory();
+    writer = new Writer();
+  });
+
+  describe("basic", () => {
+    beforeEach(async () => {
+      const nodePortsData: NodePort[] = [
+        {
+          type: "DEFAULT",
+          id: "port-1",
+          name: "default_port",
+        },
+        {
+          type: "IF",
+          id: "port-2",
+          name: "if_port",
+          expression: {
+            type: "CONSTANT_VALUE",
+            data: {
+              type: "STRING",
+              value: "test",
+            },
+          },
+        },
+        {
+          type: "ELSE",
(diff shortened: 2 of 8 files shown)


ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
```diff
@@ -21,5 +21,8 @@ exports[`GenericNode > basic > getNodeFile 1`] = `
 class MyCustomNode(BaseNode):
     class NodeTrigger(BaseNode.Trigger):
         merge_behavior = "AWAIT_ALL"
+
+    class Ports(BaseNode.Ports):
+        default_port = Port.on_else(None)
```
NgoHarrison: Will fix

ee/codegen/src/__test__/nodes/generic-nodes/node-ports.test.ts
NgoHarrison, line 1: In the fast follow PR, will do more thorough testing of the different expressions

ee/codegen/src/generators/node-port.ts
```diff
@@ -0,0 +1,124 @@
...
+        case "IF":
+        case "ELIF":
+          return nodePort.expression;
+        case "ELSE":
+        case "DEFAULT":
+          return undefined;
+        default:
+          assertUnreachable(nodePort);
+      }
+    })();
```
noanflaherty: It's pretty funky to make this a function only to immediately call it. Let's just use a good ol' `let` and `switch`, not wrapped in a function.

ee/codegen/src/generators/node-port.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+
+    return clazz;
+  }
+
+  write(writer: Writer): void {
+    this.astNode.write(writer);
+  }
+
+  private getPortAttribute(type: string): string {
```
noanflaherty: I think we can/should strengthen the type safety here with something like this

```suggestion
  private getPortAttribute(type:NodePort["type"]): string {
```

Or if that doesn't work, then `"DEFAULT" | "IF" | ...`

ee/codegen/src/generators/node-port.ts
```diff
@@ -0,0 +1,124 @@
...
+export class NodePorts extends AstNode {
+  private nodeContext: GenericNodeContext;
+  private workflowContext: WorkflowContext;
+  private astNode: AstNode;
+
+  public constructor(args: NodePorts.Args) {
+    super();
+
+    this.nodeContext = args.nodeContext;
+    this.workflowContext = args.workflowContext;
```
noanflaherty, line 31: nit: Doesn't seem like you need to save these to `this` if you're just passing as an arg below

ee/codegen/src/__test__/nodes/generic-nodes/node-ports.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  describe("basic", () => {
+    beforeEach(async () => {
+      const nodePortsData: NodePort[] = [
+        {
+          type: "DEFAULT",
+          id: "port-1",
+          name: "default_port",
+        },
```
noanflaherty: We generally shouldn't be seeing a default port if there are explicitly defined if/elif/else ports present. This tells me either:
1. There's a bug that's causing us to generate default ports when we shouldn't be; or
2. Our test fixture isn't producing realistic data.

noanflaherty (approved):
Looks good pending feedback!

ee/codegen/src/__test__/nodes/generic-nodes/node-ports.test.ts
NgoHarrison: Ah gotcha, thanks for clarifying. It is the second one as I hard coded these into the fixture

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
```diff
@@ -21,5 +21,8 @@ exports[`GenericNode > basic > getNodeFile 1`] = `
 class MyCustomNode(BaseNode):
     class NodeTrigger(BaseNode.Trigger):
         merge_behavior = "AWAIT_ALL"
+
+    class Ports(BaseNode.Ports):
+        pass
```
dvargasfuertes, line 26: [nit] we should skip codegen of ports if it would result in a pass like this

ee/codegen/src/generators/node-port.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  private getPortAttribute(type: NodePortType["type"]): string | undefined {
+    switch (type) {
+      case "DEFAULT":
+        return undefined;
+      case "IF":
+        return "on_if";
+      case "ELIF":
+        return "on_elif";
```
dvargasfuertes, line 79: [oos] we'll need to handle the case where there are multiple ifs or elifs on the node

ee/codegen/src/serializers/vellum.ts
```diff
@@ -1883,7 +1883,7 @@ export declare namespace GenericNodeSerializer {
     } | null;
     definition?: CodeResourceDefinitionSerializer.Raw | null;
     trigger: NodeTriggerSerializer.Raw;
-    ports: NodePortSerializer.Raw;
+    ports: NodePortSerializer.Raw[];
     adornments?: AdornmentNodeSerializer.Raw | null;
     attributes: NodeAttributeSerializer.Raw;
     outputs: NodeOutputSerializer.Raw;
```
dvargasfuertes, line 1889: [nit] can we quickly change these to arrays? it's blocking ordinary codegen

dvargas92495 (commented):
✅

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
NgoHarrison: This is only for when we get default only. Would you still like to skip if this is the case?

ee/codegen/src/serializers/vellum.ts
NgoHarrison: Fast follow coming right up

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
noanflaherty: Yes – we should skip in that case

dvargasfuertes: Yea, any class that is just a `class Foo(SuperFoo): pass` should be skipped, since it's redundant for the user

PR #649, opened 2025-01-17
[Codegen][Part 4] Add codegen support of workflow value descriptors

NgoHarrison:
This PR introduces the base for codegen'ing workflow value descriptors. This can then be used later in node ports and node attributes which I will update in subsequent PRs

### DIFF
--- a/ee/codegen/src/generators/expression.ts
+++ b/ee/codegen/src/generators/expression.ts
@@ -2,24 +2,31 @@ import { python } from "@fern-api/python-ast";
 import { AstNode } from "@fern-api/python-ast/core/AstNode";
 import { Writer } from "@fern-api/python-ast/core/Writer";
 
+import { NodeAttributeGenerationError } from "src/generators/errors";
+
 export declare namespace Expression {
   interface Args {
     lhs: AstNode;
     expression: string;
     rhs?: AstNode | undefined;
+    base?: AstNode | undefined;
   }
 }
 
 export class Expression extends AstNode {
   private readonly astNode: AstNode;
 
-  constructor({ lhs, expression, rhs }: Expression.Args) {
+  constructor({ lhs, expression, rhs, base }: Expression.Args) {
     super();
-
-    this.astNode = this.generateAstNode({ lhs, expression, rhs });
+    this.astNode = this.generateAstNode({ lhs, expression, rhs, base });
   }
 
-  private generateAstNode({ lhs, expression, rhs }: Expression.Args): AstNode {
+  private generateAstNode({
+    lhs,
+    expression,
+    rhs,
+    base,
+  }: Expression.Args): AstNode {
     this.inheritReferences(lhs);
     if (rhs) {
       this.inheritReferences(rhs);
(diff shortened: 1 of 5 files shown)


NgoHarrison:
[q] I am making the assumption that in workflow value descriptors on ports, the outermost expression is going to be either a unary, binary, or ternary expression. Is this correct to make? If not I can add additional handling. I am asking because I have defined the shapes to be:
```
export type WorkflowValueDescriptor =
  | UnaryWorkflowExpression
  | BinaryWorkflowExpression
  | TernaryWorkflowExpression
  | NodeOutputPointer
  | InputVariablePointer
  | ConstantValuePointer
  | WorkspaceSecretPointer
  | ExecutionCounterPointer;
```
  
  and:
```
  export interface IfConditionNodePort {
  type: "IF";
  id: string;
  name: string;
  expression?: WorkflowValueDescriptor;
}

export interface ElifConditionNodePort {
  type: "ELIF";
  id: string;
  name: string;
  expression?: WorkflowValueDescriptor;
}

```

dvargas92495:
> the outermost expression is going to be either a unary, binary, or ternary expression. Is this correct to make?

[oos] The SDK supports the rest, so we should be able to serialize the rest. I believe the FE folks have some ux to make it "look like" a unary expression while being one of the remaining 5

ee/codegen/src/generators/expression.ts
```diff
@@ -28,12 +35,38 @@ export class Expression extends AstNode {
...
     //  a Reference object from an existing AstNode, which in turn requires all AstNode's to internally track their
     //  name and modulePath.
-    const rhsExpression = rhs ? `(${rhs.toString()})` : "()";
-    const rawExpression = `${lhs.toString()}.${expression}${rhsExpression}`;
+
+    const rawExpression = base
+      ? this.generateExpressionWithBase(base, expression, lhs, rhs)
+      : this.generateStandardExpression(lhs, expression, rhs);
 
     return python.codeBlock(rawExpression);
```
dvargasfuertes, line 43: [oos] do we have all of the fern ast primitives we need to make this an astnode instead of a codeblock?

ee/codegen/src/generators/workflow-value-descriptor.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  private isExpression(workflowValueDescriptor: WorkflowValueDescriptorType) {
+    return (
+      workflowValueDescriptor.type === "UNARY_EXPRESSION" ||
+      workflowValueDescriptor.type === "BINARY_EXPRESSION" ||
+      workflowValueDescriptor.type === "TERNARY_EXPRESSION"
+    );
+  }
+
+  private isPointer(workflowValueDescriptor: WorkflowValueDescriptorType) {
```
dvargasfuertes: [nit] we've been calling these "References"

dvargas92495 (approved):

ee/codegen/src/generators/expression.ts
noanflaherty: Could be trickier than expected. I remember looking into this with Harrison a while back and we landed on the TODO above.

ee/codegen/src/generators/workflow-value-descriptor.ts
```diff
@@ -0,0 +1,155 @@
...
+        "<=": "less_than_or_equal_to",
+        ">=": "greater_than_or_equal_to",
+        contains: "contains",
+        beginsWith: "begins_with",
+        endsWith: "ends_with",
+        doesNotContain: "does_not_contain",
+        doesNotBeginWith: "does_not_begin_with",
+        doesNotEndWith: "does_not_end_with",
+        null: isNodeOutput ? "is_nil" : "is_null",
+        notNull: isNodeOutput ? "is_not_nil" : "is_not_null",
```
noanflaherty: [q] Why do we perform this conditional here?

ee/codegen/src/generators/workflow-value-descriptor.ts
```diff
@@ -0,0 +1,155 @@
...
+        });
+      }
+      default:
+        assertUnreachable(workflowValueDescriptor);
+    }
+  }
+
+  private convertOperatorType(
+    workflowValueDescriptor: WorkflowValueDescriptorType
+  ): string {
```
noanflaherty: Do we have a stricter type we can use as the return value here?

noanflaherty (approved):

ee/codegen/src/generators/workflow-value-descriptor.ts
NgoHarrison: I copied over the logic we used for conditional nodes and this logic was originally there for legacy conditional nodes using is_nil to represent is_null. I can update this now to use is_null

ee/codegen/src/generators/workflow-value-descriptor.ts
NgoHarrison: Added

PR #652, opened 2025-01-17
Load Display Meta

marlenc3:
So I think this is the basic blob we need to have the name of the fields map to the proper names for input/output displays

WIP- in case we want to add more to this, or if there's a better/cleaner way to do this mapping 😩 

additionally, if we want to cut down on some of these values -- double down

also readding in some tests that fern deleted RIP

### DIFF
--- a/ee/vellum_ee/workflows/display/types.py
+++ b/ee/vellum_ee/workflows/display/types.py
@@ -75,7 +75,7 @@ class WorkflowDisplayContext(
             current_node = node_displays[node]
             input_display = {}
             if issubclass(current_node.__class__, BaseNodeVellumDisplay):
-                input_display = current_node.node_input_ids_by_name
+                input_display = current_node.node_input_ids_by_name  # type: ignore[attr-defined]
             node_display_meta = {
                 output.name: current_node.output_display[output].id for output in current_node.output_display
             }
--- a/ee/vellum_ee/workflows/display/types.py
+++ b/ee/vellum_ee/workflows/display/types.py
@@ -46,7 +46,7 @@ class NodeDisplay(UniversalBaseModel):
     port_display: Dict[str, PortDisplayOverrides]
 
 
-class WorkflowDisplayMeta(UniversalBaseModel):
+class WorkflowEventDisplayContext(UniversalBaseModel):
     node_displays: Dict[str, NodeDisplay]
     workflow_inputs: Dict[str, InputDisplay]
     workflow_outputs: Dict[str, OutputDisplay]
@@ -79,7 +79,7 @@ class WorkflowDisplayContext(
     edge_displays: Dict[Tuple[Port, Type[BaseNode]], EdgeDisplayType] = field(default_factory=dict)
     port_displays: Dict[Port, "PortDisplay"] = field(default_factory=dict)
 
-    def build_meta(self) -> WorkflowDisplayMeta:
+    def build_meta(self) -> WorkflowEventDisplayContext:
         # type ignores due to bound types mapping to values there
         workflow_outputs = {}
         for output in self.workflow_output_displays:
@@ -119,7 +119,7 @@ class WorkflowDisplayContext(
                 output_display=node_display_meta,
                 port_display=port_display_meta,
             )
-        display_meta = WorkflowDisplayMeta(
+        display_meta = WorkflowEventDisplayContext(
             workflow_outputs=workflow_outputs,
             workflow_inputs=workflow_inputs,
             node_displays=temp_node_displays,
--- a/ee/vellum_ee/workflows/display/types.py
+++ b/ee/vellum_ee/workflows/display/types.py
@@ -73,10 +73,10 @@ class WorkflowDisplayContext(
         temp_node_displays = {}
         for node in node_displays:
             current_node = node_displays[node]
-            outputs = current_node.output_display
-            node_display_meta = {output.name: outputs[output].id for output in outputs}
-            ports = current_node.port_displays
-            port_display_meta = {port.name: ports[port].id for port in ports}
+            node_display_meta = {
+                output.name: current_node.output_display[output].id for output in current_node.output_display
+            }
+            port_display_meta = {port.name: current_node.port_displays[port].id for port in current_node.port_displays}
 
             temp_node_displays[node] = NodeDisplay(
                 input_display=current_node.node_input_ids_by_name,  # type: ignore[attr-defined]
--- a/ee/vellum_ee/workflows/display/types.py
+++ b/ee/vellum_ee/workflows/display/types.py
@@ -65,14 +65,10 @@ class WorkflowDisplayContext(
 
     def build_meta(self) -> WorkflowEventDisplayContext:
         # type ignores due to bound types mapping to values there
-        workflow_outputs = {}
-        for output in self.workflow_output_displays:
-            current_output = self.workflow_output_displays[output]
-            workflow_outputs[output.name] = current_output.id
-        workflow_inputs = {}
-        for input in self.workflow_input_displays:
-            current_inputs = self.workflow_input_displays[input]
-            workflow_inputs[input.name] = current_inputs.id
+        workflow_outputs = {
+            output.name: self.workflow_output_displays[output].id for output in self.workflow_output_displays
+        }
+        workflow_inputs = {input.name: self.workflow_input_displays[input].id for input in self.workflow_input_displays}
         node_displays = {str(node.__id__): self.node_displays[node] for node in self.node_displays}
         temp_node_displays = {}
         for node in node_displays:
--- a/ee/vellum_ee/workflows/display/types.py
+++ b/ee/vellum_ee/workflows/display/types.py
@@ -15,7 +15,6 @@ from vellum_ee.workflows.display.base import (
     WorkflowOutputDisplayType,
 )
 from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplay, PortDisplayOverrides
-from vellum_ee.workflows.display.vellum import WorkflowInputsVellumDisplay, WorkflowOutputVellumDisplay
 
 if TYPE_CHECKING:
     from vellum_ee.workflows.display.nodes.base_node_display import BaseNodeDisplay
@@ -23,6 +22,22 @@ if TYPE_CHECKING:
 
 NodeDisplayType = TypeVar("NodeDisplayType", bound="BaseNodeDisplay")
 WorkflowDisplayType = TypeVar("WorkflowDisplayType", bound="BaseWorkflowDisplay")
+
+
+class OutputDisplay(UniversalBaseModel):
+    id: UUID
+    name: str
+    label: str
+    node_id: UUID
+    node_input_id: UUID
+    target_handle_id: UUID
+    edge_id: UUID
+
+
+class InputDisplay(UniversalBaseModel):
+    id: UUID
+    name: str
+    required: bool
 
 
 class NodeDisplay(UniversalBaseModel):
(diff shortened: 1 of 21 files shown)


src/vellum/workflows/workflows/base.py
```diff
@@ -429,6 +430,74 @@ def get_most_recent_state(self) -> StateType:
 
         return most_recent_state_snapshot
```
marlenc3: This is the main file to look at

ee/vellum_ee/workflows/tests/test_display_meta.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def test_base_class_dynamic_import(files):
+    namespace = str(uuid4())  # Create a unique namespace
+    sys.meta_path.append(VirtualFileFinder(files, namespace))
+
+    Workflow = BaseWorkflow.load_from_module(namespace)
+    WorkflowDisplay = BaseWorkflow.get_workflow_display(namespace)
+    display_meta = BaseWorkflow.gather_display_meta(WorkflowDisplay, Workflow) if WorkflowDisplay else {}
+
```
marlenc3, line 33: the output for the meta we'll be sending -- it's rather large, but down to be concise and only pull certain fields

src/vellum/workflows/workflows/base.py
marlenc3: the first option I had was to just dump all of the display_context- but the mapping for field name to actual input/output name would have been push to vellum, plus it would have been a much larger payload

ee/vellum_ee/workflows/tests/test_display_meta.py
NgoHarrison: If we know for sure these are always going to be the fields we are concerned with, the latter suggestion sounds good with me

src/vellum/workflows/workflows/base.py
```diff
@@ -429,6 +430,74 @@ def get_most_recent_state(self) -> StateType:
...
 
+    @staticmethod
+    def gather_display_meta(
+        display_class: Type[BaseWorkflowDisplay], workflow_class: Type["BaseWorkflow"]
+    ) -> Dict[Any, Any]:
+        display_context = display_class(workflow_class).display_context
+        display_meta = {}
```
marlenc3: tempted to put this as a display_context.build_meta() from the display_context dataclass

src/vellum/workflows/workflows/base.py
```diff
@@ -72,6 +72,7 @@
 from vellum.workflows.types.utils import get_original_base
 from vellum.workflows.utils.uuids import uuid4_from_hash
 from vellum.workflows.workflows.event_filters import workflow_event_filter
+from vellum_ee.workflows.display.workflows import BaseWorkflowDisplay
```
dvargasfuertes: [blocking] we cannot import from the `vellum_ee` into our the open source part of our library. Let's have this be a static method on the display class instead

ee/vellum_ee/workflows/tests/test_virtual_files.py
```diff
@@ -0,0 +1,82 @@
...
+            file_path = os.path.join(root, filename)
+            # Key will be the relative path inside `local_files`
+            relative_path = str(os.path.relpath(file_path, start=base_directory))
+            with open(file_path, encoding="utf-8") as f:
+                files[relative_path] = f.read()
+
+    return files
+
+
+def test_base_class_dynamic_import(files):
```
dvargasfuertes, line 26: [q] Are these the tests that fern deleted? can you share the commit where that happened?

everything in the `ee` directory should be ignored by fern

ee/vellum_ee/workflows/tests/test_display_meta.py
dvargasfuertes: For now, we only need info that we need for reverse mapping. so I think that means:
- node input/output/port names
- workflow input/output names

ee/vellum_ee/workflows/tests/test_display_meta.py
```diff
@@ -0,0 +1,80 @@
...
+    expected_result = {
+        "global_node_output_displays": {
+            "result": {"id": UUID("423bc529-1a1a-4f72-af4d-cbdb5f0a5929"), "name": "result"},
+            "value": {"id": UUID("5469b810-6ea6-4362-9e79-e360d44a1405"), "name": "value"},
+        },
+        "global_workflow_input_displays": {
```
dvargasfuertes: [nit] these two seem redundant with `node_displays` and `workflow_inputs`

src/vellum/workflows/workflows/base.py
```diff
@@ -429,6 +430,74 @@ def get_most_recent_state(self) -> StateType:
 
         return most_recent_state_snapshot
 
+    @staticmethod
+    def gather_display_meta(
+        display_class: Type[BaseWorkflowDisplay], workflow_class: Type["BaseWorkflow"]
+    ) -> Dict[Any, Any]:
```
dvargasfuertes: [blocking] let's return a typed object here, extending `UniversalBaseModel`

dvargas92495 (commented):
I like the direction! Think we could clean up a bit and type the object

ee/vellum_ee/workflows/tests/test_virtual_files.py
marlenc3: :') https://github.com/vellum-ai/vellum-python-sdks/pull/594/files

dvargasfuertes: Ah ok they were originally added in the `tests` directory. Anything within _that_ directory is managed by fern unless we explicitly exclude it like [this](https://github.com/vellum-ai/vellum-python-sdks/blob/main/.fernignore#L14)

ee/vellum_ee/workflows/display/types.py
```diff
@@ -48,3 +78,50 @@ class WorkflowDisplayContext(
...
+        for output in self.workflow_output_displays:
+            current_output = self.workflow_output_displays[output]
+            workflow_outputs[output.name] = OutputDisplay(
+                id=current_output.id,  # type: ignore[attr-defined]
+                name=current_output.name,  # type: ignore[attr-defined]
+                label=current_output.label,  # type: ignore[attr-defined]
+                node_id=current_output.node_id,  # type: ignore[attr-defined]
+                node_input_id=current_output.node_input_id,  # type: ignore[attr-defined]
+                target_handle_id=current_output.target_handle_id,  # type: ignore[attr-defined]
+                edge_id=current_output.edge_id,  # type: ignore[attr-defined]
```
marlenc3: because the values here of the bounded type -- seems to not like pulling these values w/ mypy

ee/vellum_ee/workflows/display/types.py
```diff
@@ -22,6 +24,34 @@
...
 
 
+class OutputDisplay(UniversalBaseModel):
+    id: UUID
+    name: str
+    label: str
+    node_id: UUID
+    node_input_id: UUID
+    target_handle_id: UUID
+    edge_id: UUID
```
dvargasfuertes: Let's remove these fields that we won't need

ee/vellum_ee/workflows/display/types.py
dvargasfuertes: not sure I follow - what's the reported type of `current_output`?

marlenc3: it's  `WorkflowOutputDisplayType` 
which comes from 
`WorkflowOutputDisplayType = TypeVar("WorkflowOutputDisplayType", bound=WorkflowOutputDisplay)`

similar problem happens below. but mypy isn't following that TypeVar is of WorkflowOutputDisplay

dvargasfuertes: ah I think these mypy errors are legit then, in the sense that the only field we _can_ guarantee is `id`. This ties into my other feedback where I think the only field we need to send back is the id and our django server will have the mapping back to the original names.

marlenc3: ahhhh, okok that makes sense. I simplified the model!

ee/vellum_ee/workflows/display/types.py
```diff
@@ -48,3 +62,30 @@ class WorkflowDisplayContext(
     workflow_output_displays: Dict[BaseDescriptor, WorkflowOutputDisplayType] = field(default_factory=dict)
     edge_displays: Dict[Tuple[Port, Type[BaseNode]], EdgeDisplayType] = field(default_factory=dict)
     port_displays: Dict[Port, "PortDisplay"] = field(default_factory=dict)
+
+    def build_meta(self) -> WorkflowEventDisplayContext:
+        # type ignores due to bound types mapping to values there
```
dvargasfuertes: [nit] could nuke this

ee/vellum_ee/workflows/display/types.py
```diff
@@ -48,3 +62,30 @@ class WorkflowDisplayContext(
     workflow_output_displays: Dict[BaseDescriptor, WorkflowOutputDisplayType] = field(default_factory=dict)
     edge_displays: Dict[Tuple[Port, Type[BaseNode]], EdgeDisplayType] = field(default_factory=dict)
     port_displays: Dict[Port, "PortDisplay"] = field(default_factory=dict)
+
+    def build_meta(self) -> WorkflowEventDisplayContext:
```
dvargasfuertes: [nit] I'd rename, eg `build_event_display_context(self) -> WorkflowEventDisplayContext `

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -333,3 +335,14 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
 
         workflow_class = get_args(cls.__orig_bases__[0])[0]  # type: ignore [attr-defined]
         cls._workflow_display_registry[workflow_class] = cls
+
+    @staticmethod
+    def gather_display_meta(
```
dvargasfuertes: [nit] same here, `gather_workflow_event_display_context`

ee/vellum_ee/workflows/display/types.py
```diff
@@ -48,3 +62,30 @@ class WorkflowDisplayContext(
...
+        temp_node_displays = {}
+        for node in node_displays:
+            current_node = node_displays[node]
+            node_display_meta = {
+                output.name: current_node.output_display[output].id for output in current_node.output_display
+            }
+            port_display_meta = {port.name: current_node.port_displays[port].id for port in current_node.port_displays}
+
+            temp_node_displays[node] = NodeDisplay(
+                input_display=current_node.node_input_ids_by_name,  # type: ignore[attr-defined]
```
dvargasfuertes: [nit] why the type ignore here?

dvargas92495 (approved):

ee/vellum_ee/workflows/display/types.py
marlenc3: oh - for some reason mypy doesn't like node_input_ids_by_name

it's also a bounded type, and I know BaseNodeDisplay has it, but for some reason -- the typing 
`NodeDisplayType = TypeVar("NodeDisplayType", bound="BaseNodeDisplay")`
isn't playing with mypy

marlenc3: specifically 
mypy .
```
ee/vellum_ee/workflows/display/types.py:81: error: "NodeDisplayType" has no attribute "node_input_ids_by_name"  [attr-defined]
Found 1 error in 1 file (checked 1995 source files)
```

and BaseNodeDisplay does have this attribute

dvargasfuertes: Ah it's only available in `BaseNodeVellumDisplay`. I would ensure that `issubclass(node, BaseNodeVellumDisplay)` before accessing to be safe

marlenc3: bet! thanks!

PR #653, opened 2025-01-18
[Codegen][Part 7] Codegen generic node outputs

NgoHarrison:
This PR handles codegen of generic node outputs

### DIFF
--- a/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
@@ -27,7 +27,7 @@ class MyCustomNode(BaseNode):
         merge_behavior = "AWAIT_ALL"
 
     class Outputs(BaseNode.Outputs):
-        output: str = "default-value"
+        output = "default-value"
 "
 `;
 
@@ -57,6 +57,6 @@ class MyCustomNode(BaseNode):
         merge_behavior = "AWAIT_ALL"
 
     class Outputs(BaseNode.Outputs):
-        output: str = "default-value"
+        output = "default-value"
 "
 `;
--- a/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
@@ -25,6 +25,9 @@ class MyCustomNode(BaseNode):
 
     class NodeTrigger(BaseNode.Trigger):
         merge_behavior = "AWAIT_ALL"
+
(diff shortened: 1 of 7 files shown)


noanflaherty (approved):
Looks good!

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
```diff
@@ -25,6 +25,9 @@ class MyCustomNode(BaseNode):
 
     class NodeTrigger(BaseNode.Trigger):
         merge_behavior = "AWAIT_ALL"
+
+    class Outputs(BaseNode.Outputs):
+        output: str = "default-value"
```
dvargasfuertes: [blocking] we should either generate either the type annotation or the initial value, not both

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/node-output.test.ts.snap
```diff
@@ -0,0 +1,19 @@
...
+
+exports[`NodeOutput > node output reference > generates correct output class with node output reference 1`] = `
+"class Outputs(BaseNode.Outputs):
+    output: str = TestNode.Outputs.my_output
+"
+`;
+
+exports[`NodeOutput > workflow input output > generates correct output class with workflow input reference 1`] = `
+"class Outputs(BaseNode.Outputs):
+    output: str = None
```
dvargasfuertes: [blocking] this doesn't look right - shouldnt it be referencing `Inputs.something`?

ee/codegen/src/generators/node-output.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export declare namespace NodeOutput {
+  export interface Args {
+    nodeOutputs: NodeOutputType[];
+    nodeContext: GenericNodeContext;
+    workflowContext: WorkflowContext;
+  }
+}
+
+export class NodeOutput extends AstNode {
```
dvargasfuertes: [nit] I'd make this class and file plural, since we are generating the full class

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/node-output.test.ts.snap
NgoHarrison: Yup, good catch. Updated

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
NgoHarrison: Updated

dvargas92495 (approved):

PR #654, opened 2025-01-19
Actual support for codegenning two nodes with the same label

dvargas92495:
Workflow Sandbox that repro's the issue: http://localhost:3000/workflow-sandboxes/b34abe22-6cc9-4c73-a704-9f7a53489bc7

We had a test for this logic at the `workflow.test.ts` level, but the logic actually lived in `project.ts`. So I added a test to `project.test.ts` to reproduce the bug and proposed a fix.

The most important implementation file to review is `ee/codegen/src/project.ts` and I'd start there. Everything else is logicless downstream implications of that change

### DIFF
--- a/ee/codegen/src/project.ts
+++ b/ee/codegen/src/project.ts
@@ -349,12 +349,10 @@ ${errors.slice(0, 3).map((err) => {
 
           nodesToGenerate.push(nodeData);
 
-          const nodeContext = createNodeContext({
+          await createNodeContext({
             workflowContext: this.workflowContext,
             nodeData,
           });
-          this.workflowContext.addNodeContext(nodeContext);
-          await nodeContext.buildProperties();
         }
       )
     );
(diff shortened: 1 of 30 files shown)


ee/codegen/src/project.ts
```diff
@@ -349,11 +349,12 @@ ${errors.slice(0, 3).map((err) => {
 
           nodesToGenerate.push(nodeData);
 
-          const nodeContext = await createNodeContext({
+          const nodeContext = createNodeContext({
             workflowContext: this.workflowContext,
             nodeData,
           });
           this.workflowContext.addNodeContext(nodeContext);
+          await nodeContext.buildProperties();
```
dvargasfuertes: Because `createNodeContext` was async, every node would enter that method first before the first node would enter `this.workflowContext.addNodeContext`, which is where the workflow context node-classname reuse logic resided.

I could see us adding the bottom two lines as part of `createNodeContext`, which would limit the blast radius in the test changes and actually allow us to clean up all of the `this.workflowContext.addNodeContext(nodeContext);` lines. but the key is moving the async work from before `addNodeContext` to after `addNodeContext`

noanflaherty (approved):
Okay yeah this makes sense 👍 

I'd be a fan of the cleanup mentioned

PR #658, opened 2025-01-21
[SC-6273] Handle null/notNull operators on search node metadata filters and improve error messages

NgoHarrison:
Context: This came up from QAing workflows and the data we get from main vellum is that we have a `rhsVariableId` when given a null/notNull operator but no corresponding vellum-query-builder input variable. This PR handles the bad data issue

### DIFF
--- a/ee/codegen/src/generators/nodes/search-node.ts
+++ b/ee/codegen/src/generators/nodes/search-node.ts
@@ -21,6 +21,7 @@ import {
   VellumLogicalExpression as VellumLogicalExpressionType,
   VellumLogicalExpression,
 } from "src/types/vellum";
+import { isUnaryOperator } from "src/utils/nodes";
 
 export class SearchNode extends BaseSingleFileNode<
   SearchNodeType,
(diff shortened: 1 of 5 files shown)


ee/codegen/src/utils/nodes.ts
```diff
@@ -28,3 +28,7 @@ export function getNodeLabel(nodeData: WorkflowNode): string {
       return nodeData.data.label;
   }
 }
+
+export function isUnaryOperator(operator: string): boolean {
+  return operator === "null" || operator === "notNull";
```
dvargasfuertes, line 33: [oos] we have a few more descriptors that will be considered unary (undefined is two). But since we don't have serialization or UI support for them either, let's be sure to track for now

ee/codegen/src/generators/nodes/search-node.ts
```diff
@@ -463,19 +467,41 @@ export class SearchNodeMetadataFilters extends AstNode {
...
+      })
+    );
+
     const rhsId = data.rhsVariableId;
     const rhs = this.nodeInputsById.get(rhsId);
-    if (!rhs) {
+    if (!isUnaryOperator(data.operator) && !rhs) {
       throw new NodeAttributeGenerationError(
-        `Could not find node input for id ${rhsId}`
+        `Could not find search node input for id ${rhsId}`
```
dvargasfuertes, line 495: [oos] it would be nice to eventually have the node label as part of this error message. Here's an example of how we do this and add it to our codegen error handling: https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/codegen/src/generators/nodes/bases/base.ts#L160-L163

dvargas92495 (approved):

ee/codegen/src/generators/nodes/search-node.ts
NgoHarrison: Agreed. Can do this as part of a larger refactor to all the error messages that look like this throughout codegen

PR #660, opened 2025-01-22
[Codegen][Part 8] Add generic node adornment support

NgoHarrison:
This PR adds support for codegen'ing node adornments on generic nodes

### DIFF
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -28,6 +28,7 @@ import {
   NodePort,
   NodeAttribute,
   NodeOutput,
+  AdornmentNode,
 } from "src/types/vellum";
 
 export function entrypointNodeDataFactory(): EntrypointNode {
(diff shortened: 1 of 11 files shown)


ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -1423,6 +1426,7 @@ export function genericNodeFactory(
         },
       },
     ],
+    adornments: adornments ?? undefined,
```
noanflaherty: nit: I think this is equivalent

```suggestion
    adornments,
```

ee/codegen/src/generators/vellum-variable-value.ts
```diff
@@ -496,3 +504,7 @@ export class VellumValue extends AstNode {
     this.astNode.write(writer);
   }
 }
+
+export interface IterableConfig {
+  endWithComma?: boolean;
+}
```
noanflaherty: In a dream world, I'd have exported this from fern such that you don't have to manually redefine. Given that I haven't, let's at least move this somewhere within `src/types/` rather than tightly coupling it to vellum variable values.

noanflaherty (approved):

dvargas92495 (approved):

PR #665, opened 2025-01-22
[SC-6284] Safe cast search node limit param to int if it is valid string

NgoHarrison:
This PR tries a safe cast on limit param if the value happens to be a string that represents an int. This came up from QA of trust center workflow

### DIFF
--- a/ee/codegen/src/__test__/nodes/search-node.test.ts
+++ b/ee/codegen/src/__test__/nodes/search-node.test.ts
@@ -291,12 +291,45 @@ describe("TextSearchNode", () => {
       });
     });
 
+    it("should throw ValueGenerationError when trying to parse non-integer string", async () => {
+      await expect(async () => {
+        node.getNodeFile().write(writer);
+        await writer.toStringFormatted();
+      }).rejects.toThrow(
+        'Failed to parse search node limit value "not-a-number" as an integer'
+      );
+    });
+  });
+
+  describe("limit param should throw exception if not a constant value of type number", () => {
+    beforeEach(async () => {
+      const nodeData = searchNodeDataFactory({
+        limitInput: {
+          type: "CONSTANT_VALUE",
+          data: {
+            type: "JSON",
+            value: "{}",
+          },
+        },
+      });
+
+      const nodeContext = (await createNodeContext({
+        workflowContext,
+        nodeData,
+      })) as TextSearchNodeContext;
+
+      node = new SearchNode({
+        workflowContext,
+        nodeContext,
+      });
+    });
+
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/nodes/search-node.test.ts
```diff
@@ -239,6 +239,73 @@ describe("TextSearchNode", () => {
...
+
+    it("getNodeFile", async () => {
+      node.getNodeFile().write(writer);
+      expect(await writer.toStringFormatted()).toMatchSnapshot();
+    });
+
+    it("getNodeDisplayFile", async () => {
+      node.getNodeDisplayFile().write(writer);
+      expect(await writer.toStringFormatted()).toMatchSnapshot();
+    });
```
dvargasfuertes: [nit] Unless there's something that changes in our display logic, we don't need to include display tests for cases like this

ee/codegen/src/__test__/nodes/search-node.test.ts
```diff
@@ -239,6 +239,73 @@ describe("TextSearchNode", () => {
...
+        nodeData,
+      })) as TextSearchNodeContext;
+
+      node = new SearchNode({
+        workflowContext,
+        nodeContext,
+      });
+    });
+
+    it("should throw NodeAttributeGenerationError when trying to parse non-integer string", async () => {
```
dvargasfuertes: Nice!

ee/codegen/src/__test__/nodes/search-node.test.ts
```diff
@@ -239,6 +239,73 @@ describe("TextSearchNode", () => {
...
+      await expect(async () => {
+        node.getNodeFile().write(writer);
+        await writer.toStringFormatted();
+      }).rejects.toThrow(
+        'Failed to parse search node limit value "not-a-number" to integer'
```
dvargasfuertes:
```suggestion
        'Failed to parse search node limit value "not-a-number" as an integer'
```

ee/codegen/src/generators/nodes/search-node.ts
```diff
@@ -50,12 +53,33 @@ export class SearchNode extends BaseSingleFileNode<
...
+          );
+        }
+      } else {
+        bodyStatements.push(
+          python.field({
+            name: "limit",
+            initializer: limitInput,
+          })
+        );
+      }
```
dvargasfuertes, line 89: [nit] We should throw an error if it's a constant value that isn't a number

dvargas92495 (approved):

PR #666, opened 2025-01-22
Await Attributes Default

marlenc3:


src/vellum/workflows/descriptors/utils.py
```diff
@@ -108,7 +108,7 @@ def is_unresolved(value: Any) -> bool:
     if isinstance(value, Mapping):
         return any(is_unresolved(item) for item in value.values())
 
-    if isinstance(value, Sequence):
+    if isinstance(value, Sequence) and not isinstance(value, str):
         return any(is_unresolved(item) for item in value)
```
marlenc3, line 113: string is sequence - so would run forever

dvargasfuertes: brutal - great catch

tests/workflows/missing_node_output/workflow.py
```diff
@@ -2,6 +2,7 @@
 
 from vellum.workflows.nodes.bases import BaseNode
 from vellum.workflows.outputs.base import BaseOutput
+from vellum.workflows.types import MergeBehavior
```
dvargasfuertes, line 5: [nit] I'd remove it from this test to keep the test simpler

dvargasfuertes: oh jk we need it bc the attribute is missing and that's what the test is focused on. In that case, I'd remove from StartNode and change NextNode to an AWAIT_ALL

dvargas92495 (approved):

PR #596, opened 2025-01-10
Solve metadata filter conditions on SearchNode with new descriptor friendly types

dvargas92495:
Still need to fix some tests and codegen, but wanted to send out for review now as I did

PR #667, opened 2025-01-22
[SC-6285] Update search node to use weights attribute if present and not default weights value

NgoHarrison:
This PR fixes search node to use weights attribute if present and not default weights value. This came up from QAing trust center workflow


dvargas92495 (approved):
nice

dvargas92495:
I was actually supposed to remove `options` entirely (context: https://github.com/vellum-ai/vellum-python-sdks/pull/596#discussion_r1911577413). @marlenc3 could you squeeze that into sprint when you get a chance? hoping it's straight forward

PR #669, opened 2025-01-23
[SC-6298] Allow codegen to go through even if node inputs are not mapping to anything

NgoHarrison:
Context: QAing a workflow showed node inputs that were node pointers that pointed to non existent node ids. This PR allows the codegen process to go through and does not block


noanflaherty (approved):
If easy, it'd be nice to include unit tests that cover these null cases

dvargas92495:
Going to merge this in now since it affected a potential lead, but please fast follow with some tests as Noa suggested

PR #672, opened 2025-01-23
Fix serialization error that throws when referencing `TryNode.Outputs.error`

dvargas92495:
This is blocking us from running the headspace workflow. Adornments continue to be the gift that keep on giving.

Thankfully, we have a ton of testing around it all that I feel motivated to do a few minor cleanups after this PR

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/utils.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/utils.py
@@ -28,8 +28,7 @@ from vellum.workflows.expressions.not_between import NotBetweenExpression
 from vellum.workflows.expressions.not_in import NotInExpression
 from vellum.workflows.expressions.or_ import OrExpression
 from vellum.workflows.nodes.displayable.bases.utils import primitive_to_vellum_value
-from vellum.workflows.nodes.utils import get_wrapped_node, has_wrapped_node
-from vellum.workflows.references import NodeReference, OutputReference
+from vellum.workflows.references import NodeReference
 from vellum.workflows.utils.uuids import uuid4_from_hash
 from vellum_ee.workflows.display.types import WorkflowDisplayContext
 from vellum_ee.workflows.display.utils.vellum import create_node_input_value_pointer_rule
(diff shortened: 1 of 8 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/utils.py
```diff
@@ -56,15 +55,6 @@ def create_node_input(
...
 ) -> NodeInput:
     input_id = str(input_id) if input_id else str(uuid4_from_hash(f"{node_id}|{input_name}"))
-    if (
-        isinstance(value, OutputReference)
-        and value.outputs_class._node_class
-        and has_wrapped_node(value.outputs_class._node_class)
-    ):
-        wrapped_node = get_wrapped_node(value.outputs_class._node_class)
-        if wrapped_node._is_wrapped_node:
-            value = getattr(wrapped_node.Outputs, value.name)
```
dvargasfuertes, line 66: This is where the error was throwing. `"error"` is not an output of the inner node.

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -150,7 +150,7 @@ def _enrich_global_node_output_displays(
 
             # TODO: Make sure this output ID matches the workflow output ID of the subworkflow node's workflow
             # https://app.shortcut.com/vellum/story/5660/fix-output-id-in-subworkflow-nodes
-            node_output_displays[node_output] = node, node_display.get_node_output_display(node_output)
+            node_output_displays[node_output] = node_display.get_node_output_display(node_output)
```
dvargasfuertes, line 153: moved this tuple inside of `get_node_output_display`, since adornments like try node want to always have the inner node be the one that's referenced by id, not itself

src/vellum/plugins/vellum_mypy.py
```diff
@@ -239,10 +240,13 @@ def _redefine_class_attributes_with_descriptors(self, ctx: ClassDefContext) -> N
         the original defined type, and the descriptor version of the type.
         """
 
-        for sym in ctx.cls.info.names.values():
+        for key, sym in ctx.cls.info.names.items():
             if not isinstance(sym.node, Var):
                 continue
 
+            if not re.match(NODE_ATTRIBUTE_REGEX, key):
+                continue
```
dvargasfuertes, line 248: our mypy plugin kept thinking that `__wrapped_node__` was a valid node attribute

noanflaherty (commented):
✅

PR #673, opened 2025-01-23
Remove the  `_is_wrapped_node` attribute from BaseNode

dvargas92495:
`BaseNode._is_wrapped_node` is redundant with `BaseAdornmentNode.__wrapped_node__`. Ideally, a child node doesn't care what context it's executing/serializing in, so we opt to have adornment state driven by the adornment node rather than the wrapped node, allowing us to remove `BaseNode._is_wrapped_node` entirely.

First in a series of cleanups after doing https://github.com/vellum-ai/vellum-python-sdks/pull/672


noanflaherty (commented):
✅

PR #674, opened 2025-01-23
Hoist check to enrich node displays outside of output/port loop

dvargas92495:
Continuing on the adornment related cleanups (previous: https://github.com/vellum-ai/vellum-python-sdks/pull/673)

These display class were redundantly re-enriching for each output and port in an adornment node class


noanflaherty (commented):
✅

PR #675, opened 2025-01-23
Cleanup has wrapped node util

dvargas92495:
Last of the adornment cleanup PRs I wanted to do for now (prev: https://github.com/vellum-ai/vellum-python-sdks/pull/674). Cleans up a util that didn't add much now that BaseAdornmentNode carries the relevant state

### DIFF
--- a/ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
+++ b/ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
@@ -11,7 +11,7 @@ from vellum.workflows.descriptors.base import BaseDescriptor
 from vellum.workflows.edges import Edge
 from vellum.workflows.expressions.coalesce_expression import CoalesceExpression
 from vellum.workflows.nodes.bases import BaseNode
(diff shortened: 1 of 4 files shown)


ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -214,8 +214,8 @@ def display_context(
             global_node_displays[node] = node_display
 
             # Nodes wrapped in a decorator need to be in our node display dictionary for later retrieval
```
dvargasfuertes, line 216: We don't recurse in this one, and we should eventually

src/vellum/workflows/nodes/utils.py
```diff
@@ -28,22 +29,11 @@ def get_unadorned_port(port: Port) -> Port:
     return getattr(unadorned_node.Ports, port.name)
 
 
-@cache
```
dvargasfuertes, line 31: this is a lookup now so no need for redundant cache

noanflaherty (commented):
✅

PR #681, opened 2025-01-23
Try node adornment serialization

peicodes:
Major changes:
- Renames `BaseNodeDisplay` to `GenericNodeDisplay`. There are two `BaseNodeDisplay` and it gets pretty confusing - one is the base class for all displays, the other actually holds the display logic for generic nodes. Maybe they should be merged instead but it was kinda hard the last time I tried.
- Adds an escape hatch on `TryNodeDisplay` to use `GenericNodeDisplay` when it's an adornment on a base node.

The rename likely has codegen implications. I'll need help on that front.

### DIFF
--- a/ee/vellum_cli/tests/test_push.py
+++ b/ee/vellum_cli/tests/test_push.py
@@ -194,12 +194,12 @@ def test_push__dry_run_option_returns_report(mock_module, vellum_client):
 from typing import Dict
 from vellum.workflows import BaseWorkflow
 from vellum.workflows.nodes import BaseNode
-from vellum_ee.workflows.display.nodes import BaseNodeDisplay
+from vellum_ee.workflows.display.nodes import GenericNodeDisplay
 
 class NotSupportedNode(BaseNode):
     pass
 
-class NotSupportedNodeDisplay(BaseNodeDisplay[NotSupportedNode]):
+class NotSupportedNodeDisplay(GenericNodeDisplay[NotSupportedNode]):
     def serialize(self, display_context, **kwargs) -> Dict:
         raise NotImplementedError(f"Serialization is not supported.")
 
--- a/ee/vellum_cli/tests/test_push.py
+++ b/ee/vellum_cli/tests/test_push.py
@@ -194,12 +194,12 @@ def test_push__dry_run_option_returns_report(mock_module, vellum_client):
 from typing import Dict
 from vellum.workflows import BaseWorkflow
 from vellum.workflows.nodes import BaseNode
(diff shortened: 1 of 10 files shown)


ee/vellum_cli/tests/test_push.py
```diff
@@ -194,12 +194,12 @@ def test_push__dry_run_option_returns_report(mock_module, vellum_client):
 from typing import Dict
 from vellum.workflows import BaseWorkflow
 from vellum.workflows.nodes import BaseNode
-from vellum_ee.workflows.display.nodes import BaseNodeDisplay
+from vellum_ee.workflows.display.nodes import GenericNodeDisplay
```
peicodes: We can do this rename later if we prefer.

dvargasfuertes: > Maybe they should be merged instead but it was kinda hard the last time I tried.

How open are we to merging them? Correct me if I'm wrong @noanflaherty, but I thought the original BaseNodeDisplay was meant to account for generic nodes?

If yes, then am happy to start the series of PRs to merge them together

noanflaherty: We have `BaseNodeDisplay` and then `BaseNodeVellumDisplay`.

The latter was intended to be used for existing node types and legacy concepts. `BaseNodeDisplay` was meant to be more general and might now be synonymous with "Generic Node Display"

It's possible the two could be merged together.

I (4/5) would like to avoid renames that result in codegen implications for 2 reasons:
1. I don't want to suddenly increase surface area for codegen bugs amidst our current QA efforts and make more work for those that are doing QA.
2. The word "Generic" is internal. It really should just be `BaseNodeDisplay` because someday all nodes will just be nodes that extend from Base Node. The word "Generic" is just an enum value used to differentiate new custom nodes from older legacy nodes.

peicodes: Bailing on the rename for now, I think we'll merge the two at a later date instead.

dvargas92495 (commented):
✅ 

Discussed PR offline

ee/vellum_cli/tests/test_push.py
dvargasfuertes: > We have BaseNodeDisplay and then BaseNodeVellumDisplay.

Incorrect, we currently have `vellum_ee.workflows.display.nodes.BaseNodeDisplay`, `vellum_ee.workflows.display.nodes.vellum.BaseNodeDisplay`, and `BaseNodeVellumDisplay`. I'm planning to merge the first two, confirmed by your latest message not realizing that there were two and that the original was intended for generic nodes

PR #683, opened 2025-01-24
fix: update Makefile to use poetry for pre-commit installation

vincent0426:
Running `pre-commit` in poetry environment


dvargas92495 (approved):
Thanks for the contribution!

PR #687, opened 2025-01-24
Fully consolidate the two base node displays

dvargas92495:
Part 2 of this PR: https://github.com/vellum-ai/vellum-python-sdks/commit/f97189e28fec4f86a87b3ad2df9b2ac11995946f

With this PR, we now only have 1 `BaseNodeDisplay`, deleting the one that was introduced for generic node serialization. Note that this is different from `BaseVellumNodeDisplay`, which handle all of our legacy nodes and will still exist.


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -107,6 +202,9 @@ def get_node_port_display(self, port: Port) -> PortDisplay:
 
         return PortDisplay(id=port_id, node_id=self.node_id)
 
+    def get_trigger_id(self) -> UUID:
+        return uuid4_from_hash(f"{self.node_id}|trigger")
```
dvargasfuertes, line 206: All other changes in this file were ported over from the deleted `BaseNodeDispay`. This is the only new lines

ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
```diff
@@ -13,7 +13,6 @@
 from vellum_ee.workflows.display.nodes.get_node_display_class import get_node_display_class
 from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay
 from vellum_ee.workflows.display.nodes.utils import raise_if_descriptor
-from vellum_ee.workflows.display.nodes.vellum.base_node import BaseNodeDisplay as GenericBaseNodeDisplay
```
dvargasfuertes, line 16: The import that got us started on this path haha

ee/vellum_ee/workflows/display/workflows/vellum_workflow_display.py
```diff
@@ -43,7 +44,7 @@ class VellumWorkflowDisplay(
         WorkflowMetaVellumDisplayOverrides,
         WorkflowInputsVellumDisplay,
         WorkflowInputsVellumDisplayOverrides,
-        BaseNodeVellumDisplay,
+        BaseNodeDisplay,
```
dvargasfuertes, line 47: This file is the one that requires the closest 👀 and was likely why we initially went down the path of a second `BaseNodeDisplay` that extended from `BaseNodeVellumDisplay`. This `VellumWorkflowDisplay` class is responsible for serializing to our UI, which now is both `BaseNodeVellumDisplay` and `BaseNodeDisplay`, the common denominator of which is ideally `BaseNodeDisplay`.

This makes me wonder if we should just consolidate `BaseWorkflowDisplay` and `VellumWorkflowDisplay` and just start flattening a lot of these generics/classes. At first we wanted to consider these edges and ids in general as "legacy", but we have evolved to keep a lot of that anyway for ease of migration. And I imagine in the short term it would help ease maintenance to have less similar sounding classes around.

I have a 1/5 on all of that second paragraph, but feel like pretty 4/5 strongly that we'll need to have `BaseNodeDisplay` here, or we'll need a separate Workflow Display class that has `BaseNodeDisplay` as the `node_display_base_class`, at which point this one would feel redundant.

ee/vellum_ee/workflows/display/workflows/vellum_workflow_display.py
```diff
@@ -170,15 +171,19 @@ def serialize(self) -> JsonObject:
...
                     synthetic_output_edges.append(
                         {
                             "id": str(workflow_output_display.edge_id),
                             "source_node_id": str(source_node_display.node_id),
-                            "source_handle_id": str(
-                                source_node_display.get_source_handle_id(
-                                    port_displays=self.display_context.port_displays
-                                )
-                            ),
+                            "source_handle_id": str(source_handle_id),
```
dvargasfuertes, line 187: note that this means even before this PR, we don't support synthetic output edges from non-default ports

ee/vellum_ee/workflows/display/workflows/vellum_workflow_display.py
```diff
@@ -293,8 +298,12 @@ def _generate_entrypoint_display(
         )
 
         entrypoint_target = get_unadorned_node(entrypoint)
-        target_node_id = node_displays[entrypoint_target].node_id
-        target_handle_id = node_displays[entrypoint_target].get_target_handle_id()
```
dvargasfuertes, line 297: note that this means even before this PR, we don't support an entrypoint feeding into a merge node. Which is dumb but still valid

ee/vellum_ee/workflows/display/workflows/vellum_workflow_display.py
noanflaherty: I'm down to collapse them and agree with the rationale. Only reason not to would be if it introduced more bugs in the middle of our QA efforts.

noanflaherty (approved):

PR #694, opened 2025-01-24
Remove error node input id value map from display

NgoHarrison:


dvargas92495 (approved):
Oh nice we didn't even need it. Even better!

PR #697, opened 2025-01-24
Setup test for pushing workflow to another workspace

dvargas92495:
Going to first request review on just the test before fleshing out the implementation

### DIFF
--- a/ee/vellum_cli/tests/test_push.py
+++ b/ee/vellum_cli/tests/test_push.py
@@ -320,6 +320,7 @@ Files that were different between the original project and the generated artifac
     )
 
 
+@pytest.mark.skip(reason="TODO: https://app.shortcut.com/vellum/story/6315")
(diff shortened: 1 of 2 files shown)


ee/vellum_cli/tests/test_push.py
```diff
@@ -318,3 +318,79 @@ class ExampleWorkflow(BaseWorkflow):
...
+        {
+            "workflows": [
+                {
+                    "module": module,
+                    "workflow_sandbox_id": workflow_sandbox_id,
+                }
+            ],
+            "workspaces": [
+                {
+                    "name": "my_other_workspace",
```
dvargasfuertes, line 342: Note that this `"name"` is sdk only. It does _not_ map to the Workspace's label or name on Vellum

NgoHarrison (approved):

m-abboud (commented):
✅

PR #703, opened 2025-01-25
fix: __eq__ of BaseDescriptor and External Outputs

vincent0426:
While testing with stop and resume workflows, I found that if there are two different `MiddleNode` classes with the same field name in `ExternalInput`, the behavior is incorrect. For example:

```python
class MiddleNode(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str

class MiddleNode2(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str
```

In this case, the two MiddleNode classes are treated as the same because they share the same field name message. As a result, the second external input will not receive its value, since the code assumes that the value already exists. 

As the following line

https://github.com/vellum-ai/vellum-python-sdks/blob/dce5b50d4fad9af830757fcafc05a98dba8ce63c/src/vellum/workflows/runner/runner.py#L396

Instead of being considered `UNDEF`, it will catch the previous value, causing the code to resume instead of paused

### DIFF
--- a/src/vellum/workflows/runner/runner.py
+++ b/src/vellum/workflows/runner/runner.py
@@ -97,30 +97,6 @@ class WorkflowRunner(Generic[StateType]):
 
         self.workflow = workflow
         self._is_resuming = False
-
-        # This queue is responsible for sending events from WorkflowRunner to the outside world
-        self._workflow_event_outer_queue: Queue[WorkflowEvent] = Queue()
-
-        # This queue is responsible for sending events from the inner worker threads to WorkflowRunner
-        self._workflow_event_inner_queue: Queue[WorkflowEvent] = Queue()
-
-        self._max_concurrency = max_concurrency
-        self._concurrency_queue: Queue[Tuple[StateType, Type[BaseNode], Optional[Edge]]] = Queue()
-
-        # This queue is responsible for sending events from WorkflowRunner to the background thread
-        # for user defined emitters
-        self._background_thread_queue: Queue[BackgroundThreadItem] = Queue()
-
-        self._dependencies: Dict[Type[BaseNode], Set[Type[BaseNode]]] = defaultdict(set)
-
-        self._mocks_by_node_outputs_class = CycleMap(items=node_output_mocks or [], key_by=lambda mock: mock.__class__)
-
-        self._active_nodes_by_execution_id: Dict[UUID, BaseNode[StateType]] = {}
-        self._cancel_signal = cancel_signal
-        self._parent_context = get_parent_context() or parent_context
-
-        self.workflow.context._register_event_queue(self._workflow_event_inner_queue)
-
         if entrypoint_nodes:
             if len(list(entrypoint_nodes)) > 1:
                 raise ValueError("Cannot resume from multiple nodes")
@@ -138,8 +114,6 @@ class WorkflowRunner(Generic[StateType]):
             for descriptor, value in external_inputs.items():
                 self._initial_state.meta.external_inputs[descriptor] = value
 
-            # Trigger a snapshot of the state
-            self._snapshot_state(self._initial_state)
             self._entrypoints = [
                 ei.inputs_class.__parent_class__
                 for ei in external_inputs
@@ -155,13 +129,33 @@ class WorkflowRunner(Generic[StateType]):
                 self._initial_state = self.workflow.get_default_state(normalized_inputs)
             self._entrypoints = self.workflow.get_entrypoints()
 
-        # Depends on self._initial_state
+        # This queue is responsible for sending events from WorkflowRunner to the outside world
+        self._workflow_event_outer_queue: Queue[WorkflowEvent] = Queue()
+
+        # This queue is responsible for sending events from the inner worker threads to WorkflowRunner
+        self._workflow_event_inner_queue: Queue[WorkflowEvent] = Queue()
+
+        self._max_concurrency = max_concurrency
+        self._concurrency_queue: Queue[Tuple[StateType, Type[BaseNode], Optional[Edge]]] = Queue()
+
+        # This queue is responsible for sending events from WorkflowRunner to the background thread
+        # for user defined emitters
+        self._background_thread_queue: Queue[BackgroundThreadItem] = Queue()
+
+        self._dependencies: Dict[Type[BaseNode], Set[Type[BaseNode]]] = defaultdict(set)
         self._state_forks: Set[StateType] = {self._initial_state}
+        self._mocks_by_node_outputs_class = CycleMap(items=node_output_mocks or [], key_by=lambda mock: mock.__class__)
+
+        self._active_nodes_by_execution_id: Dict[UUID, BaseNode[StateType]] = {}
+        self._cancel_signal = cancel_signal
+        self._parent_context = get_parent_context() or parent_context
+
         setattr(
             self._initial_state,
             "__snapshot_callback__",
             lambda s: self._snapshot_state(s),
         )
+        self.workflow.context._register_event_queue(self._workflow_event_inner_queue)
 
     def _snapshot_state(self, state: StateType) -> StateType:
         self._workflow_event_inner_queue.put(
@@ -401,6 +395,7 @@ class WorkflowRunner(Generic[StateType]):
 
                 if state.meta.external_inputs.get(descriptor, UNDEF) is UNDEF:
                     state.meta.external_inputs[descriptor] = UNDEF
+                    self._snapshot_state(state)
                     return
 
             all_deps = self._dependencies[node_class]
--- a/src/vellum/workflows/runner/runner.py
+++ b/src/vellum/workflows/runner/runner.py
@@ -97,6 +97,30 @@ class WorkflowRunner(Generic[StateType]):
 
         self.workflow = workflow
         self._is_resuming = False
+
+        # This queue is responsible for sending events from WorkflowRunner to the outside world
+        self._workflow_event_outer_queue: Queue[WorkflowEvent] = Queue()
+
+        # This queue is responsible for sending events from the inner worker threads to WorkflowRunner
+        self._workflow_event_inner_queue: Queue[WorkflowEvent] = Queue()
+
+        self._max_concurrency = max_concurrency
+        self._concurrency_queue: Queue[Tuple[StateType, Type[BaseNode], Optional[Edge]]] = Queue()
+
+        # This queue is responsible for sending events from WorkflowRunner to the background thread
+        # for user defined emitters
+        self._background_thread_queue: Queue[BackgroundThreadItem] = Queue()
+
+        self._dependencies: Dict[Type[BaseNode], Set[Type[BaseNode]]] = defaultdict(set)
+
+        self._mocks_by_node_outputs_class = CycleMap(items=node_output_mocks or [], key_by=lambda mock: mock.__class__)
+
+        self._active_nodes_by_execution_id: Dict[UUID, BaseNode[StateType]] = {}
+        self._cancel_signal = cancel_signal
+        self._parent_context = get_parent_context() or parent_context
+
+        self.workflow.context._register_event_queue(self._workflow_event_inner_queue)
+
         if entrypoint_nodes:
             if len(list(entrypoint_nodes)) > 1:
                 raise ValueError("Cannot resume from multiple nodes")
@@ -114,6 +138,8 @@ class WorkflowRunner(Generic[StateType]):
             for descriptor, value in external_inputs.items():
                 self._initial_state.meta.external_inputs[descriptor] = value
 
+            # Trigger a snapshot of the state
+            self._snapshot_state(self._initial_state)
             self._entrypoints = [
                 ei.inputs_class.__parent_class__
                 for ei in external_inputs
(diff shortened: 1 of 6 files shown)


tests/workflows/basic_input_node_multi_stop/tests/test_workflow.py
```diff
@@ -0,0 +1,104 @@
...
+    Inputs,
+    MiddleNode,
+    MiddleNode2,
+    State,
+)
+
+
+def test_workflow__happy_path_multi_stop():
+    """
+    Runs the non-streamed execution of a workflow with an Input Node.
```
dvargasfuertes:
```suggestion
    Runs the non-streamed execution of a workflow with a two Nodes defining similar ExternalInputs.
```

tests/workflows/basic_input_node_multi_stop/tests/test_workflow.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+
+    # GIVEN a workflow that uses an Input Node
+    workflow = BasicInputNodeWorkflow()
+
+    # WHEN we run the workflow with initial inputs and state
+    terminal_event = workflow.run(
+        inputs=Inputs(input_value="hello"),
+        state=State(state_value="world"),
```
dvargasfuertes: [nit] we already have a test using state/inputs in an external inputs example. we could remove reference to both in this test so we really minimally focus on the fact that there's two equivalent ExternalInputs

tests/workflows/basic_input_node_multi_stop/workflow.py
```diff
@@ -0,0 +1,67 @@
...
+    class Outputs(BaseNode.Outputs):
+        final_value: str
+
+    def run(self) -> Outputs:
+        return self.Outputs(
+            final_value=f"{self.start_input} {self.middle_message} {self.middle_message2} {self.start_state}"
+        )
+
+
+class BasicInputNodeWorkflow(BaseWorkflow[Inputs, State]):
```
dvargasfuertes:
```suggestion
class BasicInputNodeWorkflow(BaseWorkflow[Inputs, State]):
    """"
    This Workflow has two nodes that accept `ExternalInputs` of the same shape to ensure that they
    could each receive external data separately.
    """"
```

tests/workflows/basic_input_node_multi_stop/workflow.py
```diff
@@ -0,0 +1,67 @@
...
+    graph = StartNode >> MiddleNode >> TransitionNode >> MiddleNode2 >> EndNode
```
dvargasfuertes: [nit] I think we could simplify this workflow to just MiddleNode >> MiddleNode2 >> EndNode, to focus on the case we're testing. At that point, we could just call them `InputNode1` and `InputNode2`

dvargas92495 (approved):
Couple of nits, then we could ship! Also as a heads up, if you request my review, I'll get the notification sooner and give it a 👀 sooner

vincent0426:
> Couple of nits, then we could ship! Also as a heads up, if you request my review, I'll get the notification sooner and give it a 👀 sooner

@dvargas92495 I think I might need permission to request a reviewer? [github](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review#:~:text=Pull%20request%20authors%20can%27t%20request%20reviews%20unless%20they%20are%20either%20a%20repository%20owner%20or%20collaborator%20with%20write%20access%20to%20the%20repository.)
Screenshot 2025-01-26 at 6 43 40 PM

dvargas92495:
Ah Ok. I'll add a CODEOWNERS file soon to enable auto reviews

tests/workflows/basic_input_node_multi_stop/tests/test_workflow.py
vincent0426: @dvargas92495 I try to remove it

However, if we make StartNode pass, the test would somehow fail and final_value will not capture InputNode1.message -> `AssertionError: assert 'None rain' == 'sunny rain'`, not sure why in this case, it depends on StartNode

```py
class StartNode(BaseNode):
    pass


class InputNode(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class InputNode2(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class EndNode(BaseNode):
    middle_message = InputNode.ExternalInputs.message
    middle_message2 = InputNode2.ExternalInputs.message

        final_value: str

    def run(self) -> Outputs:
        return self.Outputs(
            final_value=f"{self.middle_message} {self.middle_message2}"
        )


class BasicInputNodeWorkflow(BaseWorkflow[Inputs, State]):
    """
    This Workflow has two nodes that accept `ExternalInputs` of the same shape to ensure that they
    could each receive external data separately.
    """

    graph = StartNode >> InputNode >> InputNode2 >> EndNode

        final_value = EndNode.Outputs.final_value
```

but this would pass

```py


class Inputs(BaseInputs):
    input_value: str


class State(BaseState):
    state_value: str


class StartNode(BaseNode):
        temp_value: str

    def run(self) -> Outputs:
        return self.Outputs(temp_value="temp")  # If change to aaa="temp", test will fail again with `AssertionError: assert 'None rain' == 'sunny rain'`


class InputNode(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class InputNode2(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class EndNode(BaseNode):
    middle_message = InputNode.ExternalInputs.message
    middle_message2 = InputNode2.ExternalInputs.message

        final_value: str

    def run(self) -> Outputs:
        return self.Outputs(
            final_value=f"{self.middle_message} {self.middle_message2}"
        )


class BasicInputNodeWorkflow(BaseWorkflow[Inputs, State]):
    """
    This Workflow has two nodes that accept `ExternalInputs` of the same shape to ensure that they
    could each receive external data separately.
    """

    graph = StartNode >> InputNode >> InputNode2 >> EndNode

        final_value = EndNode.Outputs.final_value

```

vincent0426: Also traced the code and seems like the following state in the fail case will not include sunny

https://github.com/vellum-ai/vellum-python-sdks/blob/18e6dbeeeceddfd56a4071d158a1fffa61edff53/src/vellum/workflows/references/external_input.py#L36

Success:
`{<vellum.workflows.references.external_input.ExternalInputReference object at 0x10884b220>: 'sunny', <vellum.workflows.references.external_input.ExternalInputReference object at 0x10910faf0>: 'rain'}`

Fail:
`{<vellum.workflows.references.external_input.ExternalInputReference object at 0x119e49dc0>: 'rain'}`

dvargasfuertes: Ah we might have a bug due to our treatment of initial nodes being different of n+1 nodes. This would be interesting as a standalone test case, A single node workflow where the one node accepts external inputs

src/vellum/workflows/runner/runner.py
```diff
@@ -114,6 +138,8 @@ def __init__(
             for descriptor, value in external_inputs.items():
                 self._initial_state.meta.external_inputs[descriptor] = value
 
+            # Trigger a snapshot of the state
+            self._snapshot_state(self._initial_state)
```
vincent0426: Hi @dvargas92495, this might be a workaround. Found that the snapshot isn’t triggered because `setattr` is called after this function. However, it requires `initial_state`, so we can’t simply move it above.

If this is the cause, we could refactor by pulling out the state initialization layer. The flow would then become: initialize state -> set attributes -> handle external input.

vincent0426:
force push rebase main

src/vellum/workflows/runner/runner.py
dvargasfuertes: [blocking] we do not want to emit state snapshots or workflow events in general, unless the workflow is already executing. Is this looking to resolve the External Inputs as first node issue? If so let's OOS that out of this PR for now and have this PR focused only on the duplicate issue with just the equality changes to external_input.py as the only impl change for now

vincent0426: this is to solve the case if the first start node simply pass or does not return output that will change state (not triggering), single external input node would pass even without this change.


```py
```

vincent0426: I see, this is to solve the case if there are > 1 external input with either
- incorrect prev node, such as pass, or return output that is not defined which would not change state
- serve as the first node and there exist other external which I provide both case below

single external input node would pass even without explicit snapshot

we could make the test case to pass by having a simple output in `StartNode` so it focuses on only handling two external input?

```py


class StartNode(BaseNode):
        message: str

    def run(self) -> Outputs:
        return self.Outputs(message="start")


class InputNode(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class InputNode2(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class EndNode(BaseNode):
    middle_message = InputNode.ExternalInputs.message
    middle_message2 = InputNode2.ExternalInputs.message

        final_value: str

    def run(self) -> Outputs:
        return self.Outputs(final_value=f"{self.middle_message} {self.middle_message2}")


class BasicInputNodeWorkflow(BaseWorkflow):
    """
    This Workflow has two nodes that accept `ExternalInputs` of the same shape to ensure that they
    could each receive external data separately.
    """

    graph = StartNode >> InputNode >> InputNode2 >> EndNode

        final_value = EndNode.Outputs.final_value

```

## Fail Case
where `middle_message = InputNode.ExternalInputs.message` is `None` in `EndNode`

### StartNode return pass only
```py


class StartNode(BaseNode):
    pass


class InputNode(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class InputNode2(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class EndNode(BaseNode):
    middle_message = InputNode.ExternalInputs.message
    middle_message2 = InputNode2.ExternalInputs.message

        final_value: str

    def run(self) -> Outputs:
        return self.Outputs(final_value=f"{self.middle_message} {self.middle_message2}")


class BasicInputNodeWorkflow(BaseWorkflow):
    """
    This Workflow has two nodes that accept `ExternalInputs` of the same shape to ensure that they
    could each receive external data separately.
    """

    graph = StartNode >> InputNode >> InputNode2 >> EndNode

        final_value = EndNode.Outputs.final_value

```

### InputNode as first node and have > 1 external input
```py


class InputNode(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class InputNode2(BaseNode):
    class ExternalInputs(BaseNode.ExternalInputs):
        message: str


class EndNode(BaseNode):
    middle_message = InputNode.ExternalInputs.message
    middle_message2 = InputNode2.ExternalInputs.message

        final_value: str

    def run(self) -> Outputs:
        return self.Outputs(final_value=f"{self.middle_message} {self.middle_message2}")


class BasicInputNodeWorkflow(BaseWorkflow):
    """
    This Workflow has two nodes that accept `ExternalInputs` of the same shape to ensure that they
    could each receive external data separately.
    """

    graph = InputNode >> InputNode2 >> EndNode

        final_value = EndNode.Outputs.final_value

```

src/vellum/workflows/runner/runner.py
```diff
@@ -395,6 +395,7 @@ def _run_node_if_ready(
 
                 if state.meta.external_inputs.get(descriptor, UNDEF) is UNDEF:
                     state.meta.external_inputs[descriptor] = UNDEF
+                    self._snapshot_state(state)
```
vincent0426, line 398: @dvargas92495 I look into the problem again, and think it was caused by this one which will not trigger snapshot as it's only adding into dict but not `setattr`?

https://github.com/vellum-ai/vellum-python-sdks/blob/0e23fb70a2a581937d873e401a299fb16caf5798/src/vellum/workflows/runner/runner.py#L396-L398

dvargasfuertes: ah I see - yea there's some magic we could do within `BaseState` to allow this to happen automatically there, but happy to [oos] that for now

tests/workflows/basic_external_input/tests/test_workflow.py
```diff
@@ -0,0 +1,31 @@
+from tests.workflows.basic_external_input.workflow import BasicInputNodeWorkflow, InputNode
+
+
+def test_workflow__happy_path_multi_stop():
+    """
+    Runs the non-streamed execution of a workflow with a two Nodes defining similar ExternalInputs.
```
dvargasfuertes:
```suggestion
    Runs the non-streamed execution of a Workflow with a single Nodes defining ExternalInputs.
```

tests/workflows/basic_external_input/tests/test_workflow.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+
+    # GIVEN a workflow that uses an Input Node
+    workflow = BasicInputNodeWorkflow()
+
+    # WHEN we run the workflow with initial inputs and state
```
dvargasfuertes:
```suggestion
    # WHEN we run the workflow
```

tests/workflows/basic_external_input/workflow.py
```diff
@@ -0,0 +1,19 @@
...
+
+class InputNode(BaseNode):
+    class ExternalInputs(BaseNode.ExternalInputs):
+        message: str
+
+
+class BasicInputNodeWorkflow(BaseWorkflow):
+    """
+    This Workflow has two nodes that accept `ExternalInputs` of the same shape to ensure that they
+    could each receive external data separately.
```
dvargasfuertes:
```suggestion
    This Workflow has a single node that defines an `ExternalInputs` to ensure that it
    could receive external data as the first node.
```

tests/workflows/basic_input_node_multi_stop/tests/test_workflow.py
```diff
@@ -0,0 +1,92 @@
...
+
+def test_workflow__happy_path_multi_stop():
+    """
+    Runs the non-streamed execution of a workflow with a two Nodes defining similar ExternalInputs.
+    """
+
+    # GIVEN a workflow that uses an Input Node
+    workflow = BasicInputNodeWorkflow()
+
+    # WHEN we run the workflow with initial inputs and state
```
dvargasfuertes:
```suggestion
    # WHEN we run the workflow
```

tests/workflows/basic_input_node_multi_stop/tests/test_workflow.py
```diff
@@ -0,0 +1,92 @@
+from tests.workflows.basic_input_node_multi_stop.workflow import BasicInputNodeWorkflow, InputNode, InputNode2
+
+
+def test_workflow__happy_path_multi_stop():
+    """
+    Runs the non-streamed execution of a workflow with a two Nodes defining similar ExternalInputs.
+    """
+
+    # GIVEN a workflow that uses an Input Node
```
dvargasfuertes:
```suggestion
    # GIVEN a workflow that defines two Input Nodes
```

dvargas92495 (approved):
Just some copy feedback and then I'll 🚢 !

vincent0426:
> Just some copy feedback and then I'll 🚢 !

Yes, I forgot to fix, thx!

PR #705, opened 2025-01-25
Add mock emit_event to test ExpensiveEmitter calls

vincent0426:


dvargas92495 (commented):
What was the issue you were running into that inspired this PR? PR descriptions and a context of the problem are especially helpful

vincent0426:
> What was the issue you were running into that inspired this PR? PR descriptions and a context of the problem are especially helpful

Hi @dvargas92495, no issues at all, trying to make the test robust if making some changes to emitter

tests/workflows/background_emitter/tests/test_workflow.py
```diff
@@ -1,16 +1,21 @@
...
-def test_workflow__happy_path():
+def test_workflow__happy_path(mocker):
     """
     Test that a Workflow with an expensive data emitter completes without waiting
     for the emitter to complete.
     """
 
+    # MOCK the emitter's emit_event
+    emitter = ExpensiveEmitter()
+    mock_emit_event = mocker.patch.object(emitter, "emit_event", side_effect=emitter.emit_event)
```
dvargasfuertes: [blocking] instead of reinstantiating the emitter and then mocking it with its own method, let's instead:
- call some other event within `ExpensiveEmitter` (e.g., self._emit())
- use mocker.patch to mock _only_ the new method
- Can just assert that new method is called

vincent0426: Fixed!

tests/workflows/background_emitter/workflow.py
```diff
@@ -10,9 +10,12 @@
 class ExpensiveEmitter(BaseWorkflowEmitter):
     delay = 0.02
 
-    def emit_event(self, event: WorkflowEvent) -> None:
+    def _emit(self, event: WorkflowEvent) -> None:
         time.sleep(self.delay)
 
+    def emit_event(self, event: WorkflowEvent) -> None:
+        self._emit(event)
```
dvargasfuertes: [nit] Let's keep the `time.sleep(self.delay)` here - the self._emit should be empty so that we could assert mock call in our test

dvargas92495 (approved):

PR #707, opened 2025-01-25
Add regex match expression

vincent0426:
Add support for regex match, could be helpful for 
- Email Validation
- Website URL Validation
- Phone Number Matching

e.g.,
```py
result = Inputs.input_string.matches_regex(Inputs.pattern)
```

### DIFF
--- a/src/vellum/workflows/descriptors/base.py
+++ b/src/vellum/workflows/descriptors/base.py
@@ -26,6 +26,7 @@ if TYPE_CHECKING:
     from vellum.workflows.expressions.is_undefined import IsUndefinedExpression
     from vellum.workflows.expressions.less_than import LessThanExpression
     from vellum.workflows.expressions.less_than_or_equal_to import LessThanOrEqualToExpression
+    from vellum.workflows.expressions.matches_regex import MatchesRegexExpression
     from vellum.workflows.expressions.not_between import NotBetweenExpression
     from vellum.workflows.expressions.not_in import NotInExpression
     from vellum.workflows.expressions.or_ import OrExpression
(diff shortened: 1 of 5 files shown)


src/vellum/workflows/descriptors/base.py
```diff
@@ -349,3 +350,8 @@ def is_not_blank(self) -> "IsNotBlankExpression":
         from vellum.workflows.expressions.is_not_blank import IsNotBlankExpression
 
         return IsNotBlankExpression(expression=self)
+
+    def matches_regex(self, regex: "Union[BaseDescriptor[str], str]") -> "MatchesRegexExpression":
+        from vellum.workflows.expressions.matches_regex import MatchesRegexExpression
+
+        return MatchesRegexExpression(expression=self, regex=regex)
```
dvargasfuertes, line 357: [blocking] let's just use the word `matches` and `MatchesExpression`

src/vellum/workflows/descriptors/tests/test_utils.py
```diff
@@ -77,6 +77,10 @@ class Outputs(BaseNode.Outputs):
         (FixtureState.zeta["foo"], "bar"),
         (ConstantValueReference(1), 1),
         (FixtureState.theta[0], "baz"),
+        (FixtureState.gamma.matches_regex(r"^h.*o$"), True),  # "hello" matches the regex
+        (FixtureState.gamma.matches_regex(r"^e.*o$"), False),  # "hello" does not match the regex
+        (FixtureState.delta.matches_regex(r"^el$"), True),  # "el" matches the regex exactly
+        (FixtureState.delta.matches_regex(r"^E.*"), False),  # Case-sensitive mismatch
```
dvargasfuertes, line 83: [nit] We should only need one of these four tests. They collectively are testing `re.match`, and are not testing different aspects of the SDK's implementation.

tests/workflows/basic_match_regex/tests/test_workflow.py
dvargasfuertes, line 1: [nit] we usually only add `tests/workflows` level tests when we have a fundamental change coming to BaseWorkflow or BaseNode, so we don't necessarily need to include one here

dvargas92495 (commented):
[blocking] We also need to ensure serialization raises the proper exception when it detects a MatchExpression

dvargas92495:
Let's actually pause on this one - there are some UI implications to adding _new_ descriptors that we need to resolve internally first

PR #710, opened 2025-01-26
Add default values for environment variable

vincent0426:
- Add test to check if workflow is rejected if environment variable not found
- Allow users to set default environment variable if it's not in env


vincent0426:
Or this could be integrated into the get, e.g., `get(“some env”, “default value”)` 

I think it would be better if need this feature?

src/vellum/workflows/references/environment_variable.py
```diff
@@ -8,13 +8,17 @@
 
 
 class EnvironmentVariableReference(BaseDescriptor[str]):
-    def __init__(self, *, name: str) -> None:
+    def __init__(self, *, name: str, default: Union[str, None] = None):
         super().__init__(name=name, types=(str,))
+        self.default = default
```
dvargasfuertes: [nit] private attributes should start with an underscore: `self._default`

dvargas92495 (approved):
Nice, I dig it. Will merge once we fix the private attr

src/vellum/workflows/environment/environment.py
```diff
@@ -1,7 +1,9 @@
+from typing import Union
+
 from vellum.workflows.references import EnvironmentVariableReference
 
 
 class Environment:
     @staticmethod
-    def get(name: str) -> EnvironmentVariableReference:
-        return EnvironmentVariableReference(name=name)
+    def get(name: str, default: Union[str, None] = None) -> EnvironmentVariableReference:
```
dvargasfuertes: [nit] Sorry, let's make this `Optional[str]` instead of `Union[str, None]`

dvargasfuertes: the ones below too

dvargas92495:
Also, I noticed that the CI checks are not running on your PRs. Is there a place for you to request that they run on your end?

vincent0426:
> Also, I noticed that the CI checks are not running on your PRs. Is there a place for you to request that they run on your end?

mm, only some tests (mypy, pytest, tsc, vitest..) when pushing from local, I could check when open next PR

PR #723

PR #711, opened 2025-01-26
Simplify the graph attribute interface to be just workflow context

dvargas92495:
Evaluating our set of customer workflows revealed some pretty gnarly graph generation cases, that has motivated me to want to separate out the set of graph gen cases and build out some more going forward. The first step of that was to make the interface simpler by requiring just the `workflowContext`. This will allow the graph attribute tests to not need to do the same work that `Workflow()` currently does for each case, so that the snapshots are just the graph attribute itself

This ended up being a bigger lift than expected, but doable, so going to start splitting out these PRs to merge them in


dvargas92495:
Closing this now - no further changes here needed

PR #712, opened 2025-01-26
Simplify workflowContextFactory usage

dvargas92495:

In this PR, we just simplify the workflowContextFactory usage such that the default was the same across the two test files that use it


noanflaherty (commented):
✅

PR #713, opened 2025-01-26
Remove entrypoint node edges from the Graph Attribute interface

dvargas92495:

In this PR, we just remove 1 of the three attributes we're looking to remove from the `GraphAttribute` class: `entrypointNodeEdges`. Instead, this can be inferred from the workflow context


ee/codegen/src/generators/workflow.ts
```diff
@@ -574,7 +566,6 @@ export class Workflow {
       name: "graph",
       initializer: new GraphAttribute({
         edgesByPortId: this.edgesByPortId,
-        entrypointNodeEdges: this.entrypointNodeEdges,
```
dvargasfuertes, line 577: This is the goal of this PR

noanflaherty (commented):
✅

PR #714, opened 2025-01-26
Remove the edgesByPortId constructor param from GraphAttribute

dvargas92495:

In this PR, we remove the `edgesByPortId` from the `GraphAttribute` class, by always inferring `edgesByPortId` from the `workflowContext`. We only read from this state in one spot outside of `GraphAttribute`, so perf impact should be negligible


noanflaherty (commented):
✅

PR #715, opened 2025-01-27
Remove `entrypointNodeId` from the GraphAttribute constructor interface

dvargas92495:

In this PR, we remove the last argument from the GraphAttribute constructor. After splitting out the bits I needed, I realized that I could now split out the tests into its own file and not necessarily implement the rest of the above linked PR


noanflaherty (commented):
✅

PR #716, opened 2025-01-27
Create a separate graph attribute test file

dvargas92495:
Evaluating our set of customer workflows revealed some pretty gnarly graph generation cases, that has motivated me to want to separate out the set of graph gen cases to make building out some more test cases easier going forward. The first step of that was to make the interface simpler by requiring just the workflowContext, which I explored and executed on through the series of PRs referenced [here](https://github.com/vellum-ai/vellum-python-sdks/pull/711).

This PR does the actual file move. There is no logical changes - simply a move of the `describe("graph", {})` block into its own file.

Once we have it in its own file, there are some more optimizations I have in mind to ship before tackling the actual graph gen case that was failing for customer workflows


marlenc3:
✅

PR #717, opened 2025-01-27
Remove the `nodes` param from codegen.workflow()

dvargas92495:
Workflow and GraphAttribute testing have some redundancy that I'm looking to cleanup to make adding new GraphAttribute test cases easier. The hope is that this continues to also simplify the codegen interfaces themselves.

In this PR, we remove the `nodes` parameter from the `codegen.workflow()` constructor. It was only used to determine whether or not to generate the GraphAttribute, which we could already infer from the `workflowContext`. This removes one unit of redundancy from the Graph Attribute tests


marlenc3:
✅

PR #719, opened 2025-01-27
Simplify the graph attribute tests to just snapshot the graph

dvargas92495:
With this PR, we complete our work around simplifying Graph Attribute tests to allow for new graph generation cases to be added. We simply remove workflow related fluff so that only the graph gets generated for each snapshot

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
@@ -1,324 +1,175 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`Workflow > graph > should be correct for a basic conditional node case 1`] = `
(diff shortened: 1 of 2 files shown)


ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
```diff
@@ -1,324 +1,175 @@
...
-from .nodes.search_node import SearchNode
-from .nodes.final_output import FinalOutput
-
-
-class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
-    graph = SearchNode
-
-    class Outputs(BaseWorkflow.Outputs):
-        query = FinalOutput.Outputs.value
+"TemplatingNode
```
dvargasfuertes, line 62: Switched this from Search -> Templating node to not need to mock document index retrieve in this file's test suite

noanflaherty (commented):
Nice simplification 👍

PR #720, opened 2025-01-27
Graph Generation case: node pointing to three nodes

dvargas92495:
`graph = A >> { B, C, D }`

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
@@ -173,3 +173,12 @@ exports[`Workflow > graph > should be correct for two branches merging from sets
 }
 "
 `;
+
+exports[`Workflow > graph > should define nested sets of nodes without compilation errors 1`] = `
+"TopNode >> {
+    OutputTopNode,
+    OutputMiddleNode,
+    OutputBottomNode,
+}
+"
+`;
--- a/ee/codegen/src/generators/graph-attribute.ts
+++ b/ee/codegen/src/generators/graph-attribute.ts
@@ -141,27 +141,6 @@ export class GraphAttribute extends AstNode {
         }
       };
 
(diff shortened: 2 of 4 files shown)


ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
```diff
@@ -173,3 +173,12 @@ exports[`Workflow > graph > should be correct for two branches merging from sets
 }
 "
 `;
+
+exports[`Workflow > graph > should define nested sets of nodes without compilation errors 1`] = `
+"TopNode >> {
+    OutputTopNode,
+    OutputMiddleNode,
+    OutputBottomNode,
+}
```
dvargasfuertes, line 182: Old graph was outputting:

```python
TopNode >> {
    {
        OutputTopNode,
        OutputMiddleNode,
    }
    OutputBottomNode,
}
```

Which fails to even compile since sets are unhashable and can't live in a set

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -141,27 +141,6 @@ export class GraphAttribute extends AstNode {
         }
       };
 
-      const popSources = (mutableAst: GraphMutableAst): GraphMutableAst => {
```
dvargasfuertes, line 144: Was just moved to a class method below instead of defined inline each iteration of this while loop. Over time, expect to do the same for the rest of these methods

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -681,6 +660,42 @@ export class GraphAttribute extends AstNode {
...
+   *
+   * Becomes:
+   *
+   * ```
+   * B >> C
+   * ```
+   */
+  private popSources = (mutableAst: GraphMutableAst): GraphMutableAst => {
+    if (mutableAst.type === "set") {
+      return this.flattenSet({
```
dvargasfuertes, line 680: This `this.flattenSet()` is the only actual change to the method to get the new test to pass

PR #725, opened 2025-01-27
Code Execution node optimization - running python code inline

dvargas92495:
Context: https://vellum-ai.slack.com/archives/C06P13ABK0W/p1738008892483549


src/vellum/workflows/nodes/displayable/code_execution_node/tests/test_code_execution_node.py
```diff
@@ -24,6 +26,12 @@ class State(BaseState):
     class ExampleCodeExecutionNode(CodeExecutionNode[State, int]):
         filepath = fixture
         runtime = "PYTHON_3_11_6"
+        packages = [
+            CodeExecutionPackage(
```
dvargasfuertes, line 31: Added packages to legacy tests to maintain reference to vellum client for now

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
```diff
@@ -10,3 +15,81 @@ def read_file_from_path(node_filepath: str, script_filepath: str) -> Union[str,
         with open(full_filepath) as file:
             return file.read()
     return None
+
+
+class ListWrapper(list):
```
dvargasfuertes, line 23: Copied most of this from: https://github.com/vellum-ai/code-executor-service/blob/main/server/sandboxes/python/runner.py

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
```diff
@@ -10,3 +15,81 @@ def read_file_from_path(node_filepath: str, script_filepath: str) -> Union[str,
...
+{code}
+
+import json
+
+out = json.dumps(main({", ".join(run_args)})).strip().splitlines()[-1][:{max_output_length}]
+print("{delimiter}")
+print(out)"""
+    local_namespace = {input_value.name: _clean_for_dict_wrapper(input_value.value) for input_value in input_values}
+
+    with redirect_stdout(log_buffer):
```
m-abboud: This will very likely break on vembda and also any threading that goes on in wsdk, and as of right now I do not see a way around it unless it magically works with the log proxy hack, without using a subprocess which is extremely undesirable because of the compute cost. So I would disable logs for now until I can noodle on a better solution.

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
```diff
@@ -10,3 +15,81 @@ def read_file_from_path(node_filepath: str, script_filepath: str) -> Union[str,
...
+    delimiter = "--vellum-output-start--"
+    max_output_length = "16_000_000"
+
+    run_args = [f"{input_value.name}={input_value.name}" for input_value in input_values]
+    execution_code = f"""\
+{code}
+
+import json
+
+out = json.dumps(main({", ".join(run_args)})).strip().splitlines()[-1][:{max_output_length}]
```
m-abboud: The way you would want to do this most efficiently is the way the older versions of python workflow server did by passing in a global context ex: https://github.com/vellum-ai/vembda-service/blob/f0d914f0206e28e01eaf0844fb556813ba9a4a5e/workflow-executor/src/workflow_executor/executor.py

Code exec just does the printing because it has to be called by a process.

Because of the problem with stdout in my other comment you will need to change this to using that way even though it is working.

m-abboud (commented):
Oops forgot to submit this earlier, blocking on log redirects

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
dvargasfuertes: yea I like this better

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
dvargasfuertes: I was under the impression that this would only capture the logs in this context. Is that not the case? will it affect threads as well?

dvargasfuertes: I think this feedback currently blocks this PR from merging since it would be a regression from the API experience.

I suspect that we may want Process's anyway in case there are other side effects that could affect Workflow Execution? We're replacing code exec node so from an overhead perspective I think we still net out positive

m-abboud (approved):
sweet

PR #735, opened 2025-01-28
Fix issue around pulling the same workflow twice creating redundant entries in the Vellum Lockfile

dvargas92495:
When a user pulls from the same workflow with the workflow-sandbox-id param, it would create duplicate issues. This PR fixes


NgoHarrison (commented):
✅

PR #723, opened 2025-01-27
Update CI workflow to trigger on pull requests in addition to pushes

vincent0426:
Add pull_request to ci option to see if it works

PR #736, opened 2025-01-28
Fix the GH Action triggers to allow for external collaborators

dvargas92495:
We now have external [contributors](https://github.com/vellum-ai/vellum-python-sdks/pull/723)! In order to support their development, we need the CI checks to fire on their PRs. These new triggers should enable that.


m-abboud (commented):
✅

PR #737, opened 2025-01-28
Fix snapshot state for external input

vincent0426:
Based on feedback from #703
- add state management for external input
- remove explicit snapshot


dvargas92495 (approved):
[nit] Can we add a test for this case to `src/vellum/workflows/state/tests/test_state.py`?

PR #738, opened 2025-01-28
Fix for mismatch of script files for codegen

marlenc3:
for now, since there's only two types we support, its one or the other
eventually we might want to introduce file extension matching to our runtimes

### DIFF
--- a/ee/codegen/src/__test__/nodes/code-execution-node.test.ts
+++ b/ee/codegen/src/__test__/nodes/code-execution-node.test.ts
@@ -1,5 +1,5 @@
 import { Writer } from "@fern-api/python-ast/core/Writer";
-import { beforeEach } from "vitest";
+import { beforeEach, describe } from "vitest";
 
 import { workflowContextFactory } from "src/__test__/helpers";
 import { codeExecutionNodeFactory } from "src/__test__/helpers/node-data-factories";
@@ -7,6 +7,7 @@ import { createNodeContext, WorkflowContext } from "src/context";
 import { CodeExecutionContext } from "src/context/node-context/code-execution-node";
 import { NodeAttributeGenerationError } from "src/generators/errors";
 import { CodeExecutionNode } from "src/generators/nodes/code-execution-node";
+import { NodeInputValuePointerRule } from "src/types/vellum";
 
 describe("CodeExecutionNode", () => {
   let workflowContext: WorkflowContext;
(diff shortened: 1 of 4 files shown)


ee/codegen/src/generators/nodes/code-execution-node.ts
```diff
@@ -242,7 +242,27 @@ export class CodeExecutionNode extends BaseSingleFileNode<
...
+        );
+      }
+      runtime = runtimeRule?.data.value ?? "";
+    } catch {
+      runtime = "";
+    }
+    const filepath =
+      this.nodeData.data.filepath ?? runtime?.includes("TYPESCRIPT")
+        ? "./script.ts"
+        : "./script.py";
```
marlenc3, line 266: once we start supporting more runtimes, we can split this into an object match for runtime prefix to file type

ee/codegen/src/generators/nodes/code-execution-node.ts
```diff
@@ -242,7 +242,27 @@ export class CodeExecutionNode extends BaseSingleFileNode<
...
+      if (
+        !runtimeRule ||
+        runtimeRule.type !== "CONSTANT_VALUE" ||
+        runtimeRule.data.type !== "STRING"
+      ) {
+        throw new NodeAttributeGenerationError(
+          "Expected to find runtime input with constant string value"
+        );
+      }
+      runtime = runtimeRule?.data.value ?? "";
```
marlenc3: also down to remove this catch, and break generator if we can't find runtime (I assume we'll always have it)

NgoHarrison: I think the throw will automatically stop it and if we don't want it to stop codegen, we can add the error the the workflowContext i.e `this.workflowContext.addError(new NodeAttributeGenerationError(
        "Expected to find runtime input with constant string value"
      );`

marlenc3: Ohhhhhh that's how we add errors?? Huge! Let me change it to that

ee/codegen/src/__test__/nodes/code-execution-node.test.ts
```diff
@@ -102,4 +103,43 @@ describe("CodeExecutionNode", () => {
...
+      });
+
+      const overrideInputValue: NodeInputValuePointerRule = {
+        type: "CONSTANT_VALUE",
+        data: {
+          type: "STRING",
+          value:
+            override == "TYPESCRIPT_5_3_3"
+              ? "async function main(inputs: {\n\n}): Promise<number> {\n\n  return;\n}"
+              : "def main(inputs):\n  return",
```
dvargasfuertes, line 122: [nit] I always lean towards a second test case arg over using conditionals in tests

dvargas92495 (approved):
sick 🎉

PR #739, opened 2025-01-28
Add Trigger workflow server CD

m-abboud:
This triggers a version bump in vembda-service whenever there is a new vellum-ai release.

### DIFF
--- a/.github/workflows/ci-workflow-server.yaml
+++ b/.github/workflows/ci-workflow-server.yaml
@@ -1,17 +1,16 @@
 name: ci-workflow-server
 
-on: [push]
-#on:
-#  push:
-#    branches:
-#      - main
+on:
+  push:
+    branches:
+      - main
 jobs:
   publish:
     permissions:
       contents: "read"
       id-token: "write"
 
-#    if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
+    if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout repo
--- a/.github/workflows/ci-workflow-server.yaml
+++ b/.github/workflows/ci-workflow-server.yaml
@@ -1,0 +1,45 @@
+name: ci-workflow-server
+
+on: [push]
+
+jobs:
+  publish:
+    permissions:
+      contents: "read"
+      id-token: "write"
+
+#    if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repo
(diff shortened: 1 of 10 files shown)


.github/workflows/ci-workflow-server.yaml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18.17.0'
+
+      - name: Install dependencies
+        run: npm install
+        working-directory: ee/codegen
```
dvargasfuertes: [blocking] we have an `ee/automation` directory for this exact use case. Can we add these scripts there instead? we'll already have the octokit/rest dependency and helper for getting github token there too

ee/codegen/scripts/upgrade-codegen-service.mts
```diff
@@ -1,30 +1,9 @@
...
-    privateKey,
-    installationId,
-  });
-
-  const { token } = await auth({ type: "installation" });
-  return token;
-};
+import { getGithubToken } from "./get-github-token.mjs";
 
 const main = async () => {
```
dvargasfuertes, line 8: [oos] I could see us moving this guy to `ee/automation` as well - can follow up to do so

ee/codegen/scripts/upgrade-workflow-server.mts
```diff
@@ -0,0 +1,30 @@
...
+ * so it can take advantage of the existing github app auth token.
+ */
+const main = async () => {
+  const version = packageJson.version;
+  console.log("Upgrading workflow server to version", version);
+  const octokit = new Octokit({ auth: await getGithubToken() });
+  await octokit.actions.createWorkflowDispatch({
+    owner: "vellum-ai",
+    repo: "vembda-service",
+    workflow_id: "version-python-workflow-server.yaml",
```
dvargasfuertes: [oos] This is cool - part of me sorta wished I designed the codegen server bump like this to use an action in _that_ repo.

.github/workflows/ci-workflow-server.yaml
m-abboud: Oh didn't notice that oops.

ee/codegen/package.json
```diff
@@ -40,6 +40,7 @@
   },
   "devDependencies": {
     "@octokit/auth-app": "^7.1.3",
+    "@octokit/rest": "^19.0.13",
```
dvargasfuertes: [blocking] can we now remove this dep and package-lock changes?

dvargas92495 (approved):
last blocking comment should be quick

PR #741, opened 2025-01-28
[SC-6295] Remove __all__ from codegen in __init__ files

NgoHarrison:
Context: Fern handles the logic for normalizing import alias naming. Therefore, this PR just removes __all__ from both init files since our imports aren't wildcards


ee/codegen/src/project.ts
NgoHarrison, line 1: All the other files are test updates. This is where the deletion occurs

ee/codegen/src/__test__/project.test.ts
```diff
@@ -503,119 +503,4 @@ Encountered 1 error(s) while generating code:
       expect(fs.existsSync(workflowPath)).toBe(true);
     });
   });
-
-  describe("multiple nodes with the same label", () => {
```
noanflaherty: Are we intentionally removing this test case?

noanflaherty (approved):

ee/codegen/src/__test__/project.test.ts
dvargasfuertes: Yea would prefer to keep this

dvargas92495 (approved):

ee/codegen/src/__test__/project.test.ts
NgoHarrison: Oops good catch, unintentional

PR #769, opened 2025-01-30
Introduce an edgesFactory to continue simplifying Graph attribute tests

dvargas92495:
I decided that each new graph attribute test case that comes up, going to pre-empt it with one change to simplify defining these graph generation tests to make each one easier to add. This PR focuses on defining edges, using a new `edgesFactory` to eliminate about ~700 lines of fluff

### DIFF
--- a/ee/codegen/src/serializers/vellum.ts
+++ b/ee/codegen/src/serializers/vellum.ts
@@ -1864,6 +1864,7 @@ export const GenericNodeSerializer: ObjectSchema<
   GenericNodeSerializer.Raw,
   Omit<GenericNode, "type">
 > = objectSchema({
+  id: stringSchema(),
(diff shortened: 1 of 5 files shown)


ee/codegen/src/serializers/vellum.ts
```diff
@@ -1864,6 +1864,7 @@ export const GenericNodeSerializer: ObjectSchema<
   GenericNodeSerializer.Raw,
   Omit<GenericNode, "type">
 > = objectSchema({
+  id: stringSchema(),
```
dvargasfuertes, line 1867: Generic node schema was missing its id. I plan to swap out all of the nodes in the graph attribute tests eventually with Generic nodes

NgoHarrison (approved):
Nice!

PR #771, opened 2025-01-30
[SC-6309] Handle duplicate node label for class name generation

NgoHarrison:
This PR handles duplicate name generation by using the same logic we have for input and output variables and renames the class when given the same label

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -46,11 +46,11 @@ class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
 exports[`Workflow > write > should handle the case of multiple nodes with the same label 1`] = `
 "from vellum.workflows import BaseWorkflow
 from .nodes.templating_node import TemplatingNode
-from .nodes.templating_node_1 import TemplatingNode as TemplatingNode1TemplatingNode
+from .nodes.templating_node_1 import TemplatingNode1
 
 
 class TestWorkflow(BaseWorkflow):
-    graph = TemplatingNode >> TemplatingNode1TemplatingNode
+    graph = TemplatingNode >> TemplatingNode1
 "
 `;
 
--- a/ee/codegen/src/context/workflow-context/workflow-context.ts
+++ b/ee/codegen/src/context/workflow-context/workflow-context.ts
@@ -21,6 +21,7 @@ import {
   WorkflowDataNode,
   WorkflowEdge,
 } from "src/types/vellum";
+import { createPythonClassName } from "src/utils/casing";
 
 type InputVariableContextsById = Map<string, InputVariableContext>;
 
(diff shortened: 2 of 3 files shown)


ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -46,10 +46,10 @@ class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
 exports[`Workflow > write > should handle the case of multiple nodes with the same label 1`] = `
 "from vellum.workflows import BaseWorkflow
 from .nodes.templating_node import TemplatingNode
-from .nodes.templating_node_1 import TemplatingNode as TemplatingNode1TemplatingNode
+from .nodes.templating_node_1 import TemplatingNode1
```
NgoHarrison, line 49: This essentially solved the other problem with `__all__` needing to know what fern generates as the alias on write time. Now, we just change the class name so it doesn't conflict and fern doesn't need to use it's alias reference conflict naming logic

dvargasfuertes: Yea we _may_ need to reintroduce `__all__` after all... https://vellum-ai.slack.com/archives/C06P13ABK0W/p1738211053520019

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -389,4 +393,31 @@ export class WorkflowContext {
...
+
+  public isClassNameUsed(className: string): boolean {
+    return this.classNames.has(className);
+  }
+
+  private addUsedClassName(className: string): void {
+    this.classNames.add(className);
+  }
+
+  public getUniqueClassName(baseName: string): string {
```
dvargasfuertes, line 405: > by using the same logic we have for input and output variables

can the logic be consolidated or more trouble than its worth?

ee/codegen/src/utils/paths.ts
```diff
@@ -54,25 +54,21 @@ export function getGeneratedNodeModuleInfo({
   let rawModuleName: string;
   let nodeClassName: string;
 
-  // In the case of adorned Nodes, we need to traverse the Adornment Node's definition to get
-  // info about the inner Node that it adorns.
-  // TODO: Handle case where there's multiple adornments on the same Node
-  //  https://app.shortcut.com/vellum/story/5699
```
dvargasfuertes: [nit] This is still open right? would like to keep the comment if so

dvargas92495 (approved):

ee/codegen/src/utils/paths.ts
NgoHarrison: Oops good catch, will re-add

ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
NgoHarrison: OOF, got it I will do a separate PR to revert it before merging this in

ee/codegen/src/context/workflow-context/workflow-context.ts
NgoHarrison: A little bit more work since the others use different state trackers on what names were used specifically and reside in their own contexts

ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
NgoHarrison: Merged in revert PR: https://github.com/vellum-ai/vellum-python-sdks/pull/772

PR #773, opened 2025-01-30
[SC-6397] Cast rhs value from string to number if given numeric operators

NgoHarrison:
This PR casts rhs values from string to number when given numeric operators in the conditional rule

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
@@ -1,59 +1,5 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`Conditional Node with numeric operator casts rhs to NUMBER > getNodeDisplayFile 1`] = `
-"from vellum_ee.workflows.display.nodes import BaseConditionalNodeDisplay
-from ...nodes.conditional_node import ConditionalNode
-from uuid import UUID
-from vellum_ee.workflows.display.nodes.vellum.conditional_node import (
-    RuleIdMap,
-    ConditionId,
-)
-from vellum_ee.workflows.display.vellum import NodeDisplayData, NodeDisplayPosition
-
-
-class ConditionalNodeDisplay(BaseConditionalNodeDisplay[ConditionalNode]):
-    label = "Conditional Node"
-    node_id = UUID("b81a4453-7b80-41ea-bd55-c62df8878fd3")
-    target_handle_id = UUID("842b9dda-7977-47ad-a322-eb15b4c7069d")
-    source_handle_ids = {0: UUID("63345ab5-1a4d-48a1-ad33-91bec41f92a5")}
-    rule_ids = [
-        RuleIdMap(
-            id="fa50fb0c-8d62-40e3-bd88-080b52efd4b2",
-            lhs=RuleIdMap(
-                id="ad6bcb67-f21b-4af9-8d4b-ac8d3ba297cc",
-                lhs=None,
-                rhs=None,
-                field_node_input_id=None,
-                value_node_input_id=None,
-            ),
-            rhs=None,
-            field_node_input_id=None,
-            value_node_input_id=None,
-        )
-    ]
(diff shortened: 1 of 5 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
```diff
@@ -1,5 +1,71 @@
...
+
+exports[`Conditional Node with numeric operator casts rhs to NUMBER > getNodeFile 1`] = `
+"from vellum.workflows.nodes.displayable import ConditionalNode as BaseConditionalNode
+from vellum.workflows.ports import Port
+from ..inputs import Inputs
+
+
+class ConditionalNode(BaseConditionalNode):
+    class Ports(BaseConditionalNode.Ports):
+        branch_1 = Port.on_if(Inputs.rhs.greater_than(0.5))
```
NgoHarrison, line 11: How it's supposed to look

ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
```diff
@@ -1,5 +1,71 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`Conditional Node with numeric operator casts rhs to NUMBER > getNodeDisplayFile 1`] = `
```
dvargasfuertes: [nit] Again, feel free to omit snapshots tests for display files if the additional thing we are testing a fix for does not have display implications. A notable tell that it doesn't is noticing that no display files were touched on the implementation half of this PR

ee/codegen/src/generators/expression.ts
```diff
@@ -16,14 +16,14 @@ export declare namespace Expression {
...
+  constructor({ lhs, operator, rhs, base }: Expression.Args) {
     super();
-    this.astNode = this.generateAstNode({ lhs, expression, rhs, base });
+    this.astNode = this.generateAstNode({ lhs, operator, rhs, base });
   }
 
   private generateAstNode({
     lhs,
-    expression,
+    operator,
```
dvargasfuertes, line 26: I dig the api change here

ee/codegen/src/generators/conditional-node-port.ts
```diff
@@ -204,6 +250,23 @@ export class ConditionalNodePort extends AstNode {
...
+    ]).has(operator);
+  }
+
+  private isConstantStringPointer(input: NodeInput): boolean {
+    return (
+      !isNilOrEmpty(input.nodeInputData.value.rules) &&
+      input.nodeInputData.value.rules[0]?.type === "CONSTANT_VALUE" &&
+      input.nodeInputData.value.rules[0]?.data.type === "STRING"
+    );
+  }
```
dvargasfuertes, line 268: [oos] it feels like we do this enough that it may be worth util-ing it soon

ee/codegen/src/generators/conditional-node-port.ts
```diff
@@ -149,23 +153,65 @@ export class ConditionalNodePort extends AstNode {
       rhsKey = this.valueInputKeysByRuleId.get(ruleId);
     }
     const lhs = this.nodeInputsByKey.get(lhsKey);
```
dvargasfuertes, line 155: [oos] I imagine we have the same with lhs? Is it possible in the UI?

dvargas92495 (approved):
niiice

ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
NgoHarrison: Oh yes sorry, will remove

ee/codegen/src/generators/conditional-node-port.ts
NgoHarrison: It IS possible but this is leaning more towards user error imo. Since lhs is always input variables, if a user adds an input variable and chooses string and then proceeds to put a number value, I don't think this would be the same edge case. I can add a follow up tho if we still want to account for this

PR #779, opened 2025-01-31
Align generic node data shape to be like other workflow nodes

NgoHarrison:


ee/codegen/src/types/vellum.ts
```diff
@@ -660,19 +660,24 @@ export interface NodeOutput {
...
   ports: NodePort[];
   adornments?: AdornmentNode[];
   attributes: NodeAttribute[];
   outputs: NodeOutput[];
 }
 
+export interface GenericNode extends BaseWorkflowNode {
+  id: string;
+  type: "GENERIC";
+  data: GenericNodeData;
```
NgoHarrison: The main thing I wanted out of this was having this type/data in the outer level like other nodes

NgoHarrison:
Closing because after looking at the json structure from network logs, I realized it doesn't have this structure 😬

ee/codegen/src/types/vellum.ts
```diff
@@ -660,17 +660,22 @@ export interface NodeOutput {
...
   type: "GENERIC";
+  data: GenericNodeData;
   displayData?: GenericNodeDisplayData;
   base: CodeResourceDefinition;
   definition?: CodeResourceDefinition;
-  trigger: NodeTrigger;
-  ports: NodePort[];
-  adornments?: AdornmentNode[];
-  attributes: NodeAttribute[];
-  outputs: NodeOutput[];
```
dvargasfuertes, line 673: to add, the long term goal is actually to go the opposite direction. Other Workflow Nodes -> Generic Nodes

PR #780, opened 2025-01-31
Dynamic Outputs

marlenc3:
I am going to start consolidating -- since I think we can use the same meta class for adornments, inline (deployments? unsure) 

anywho -- would love to know how to get the typing/warnings about attributes to go away

but this first stab does work, just need to iterate 

WIP mainly to show we can reuse this one


dvargas92495 (approved):
> I think we can use the same meta class for adornments

Yea! I think it could eventually extend from BaseAdornmentNode too since eventually a `@InlineSubworkflowNode.wrap` could be allowed and the logic is similar to try node/retry node

PR #781, opened 2025-01-31
Add interval parameter to RetryNode

### LINKED ISSUE #746: [Feat] Add intervals between retry for RetryNode
Allow users to add intervals between each retry? can be helpful for API requests or db operations

can sleep here:
https://github.com/vellum-ai/vellum-python-sdks/blob/bb09e11f5a02277174894bee554c770b3e43e6e3/src/vellum/workflows/nodes/core/retry_node/node.py#L70-L74

vincent0426:
Closes #746 

- uses float seconds to match with `time.sleep`

### DIFF
--- a/ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_adornments_serialization.py
+++ b/ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_adornments_serialization.py
@@ -20,7 +20,7 @@ class Inputs(BaseInputs):
     input: str
 
 
-@RetryNode.wrap(max_attempts=3)
+@RetryNode.wrap(max_attempts=3, interval=1)
 class InnerRetryGenericNode(BaseNode):
     input = Inputs.input
 
@@ -81,6 +81,11 @@ def test_serialize_node__retry(serialize_node):
                     },
                     "attributes": [
                         {
+                            "id": "6810e294-15c0-48a0-b77c-fa7318fd1c33",
+                            "name": "interval",
+                            "value": {"type": "CONSTANT_VALUE", "value": {"type": "NUMBER", "value": 1.0}},
+                        },
+                        {
                             "id": "f388e93b-8c68-4f54-8577-bbd0c9091557",
                             "name": "max_attempts",
                             "value": {"type": "CONSTANT_VALUE", "value": {"type": "NUMBER", "value": 3.0}},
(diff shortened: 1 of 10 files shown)


src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -17,11 +18,13 @@ class RetryNode(BaseAdornmentNode[StateType], Generic[StateType]):
     Used to retry a Subworkflow a specified number of times.
 
     max_attempts: int - The maximum number of attempts to retry the Subworkflow
+    interval: float = 0 - The number of seconds to wait between retries
     retry_on_error_code: Optional[VellumErrorCode] = None - The error code to retry on
     subworkflow: Type["BaseWorkflow[SubworkflowInputs, BaseState]"] - The Subworkflow to execute
     """
 
     max_attempts: int
+    interval: float = 0
```
dvargasfuertes: `@retry` module calls this `delay`, let's align with that

ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_adornments_serialization.py
```diff
@@ -20,7 +20,7 @@ class Inputs(BaseInputs):
     input: str
 
 
-@RetryNode.wrap(max_attempts=3)
+@RetryNode.wrap(max_attempts=3, interval=1)
```
dvargasfuertes: can we remove the changes in this file? Our licensing is such that we can't accept external contributions to the `ee` directory for now

src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -72,20 +75,23 @@ def run(self) -> BaseNode.Outputs:
                     terminal_event.error.message,
                     code=terminal_event.error.code,
                 )
+                time.sleep(self.interval)
```
dvargasfuertes: [blocking] let's make the parameter optional that defaults to None, then only run time.sleep if defined:
Screenshot 2025-01-30 at 11 19 23 PM

ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_adornments_serialization.py
vincent0426: yup, but `test_adornments_serialization.py` may fail as it will miss the `delay` field , I bypass pre-commit now

dvargas92495:
I'll fast follow with the test fix - in general, we should work to not have `ee` tests fail on non `ee` changes

PR #782, opened 2025-01-31
Fix the Graph.from_set import for graphs that need it

dvargas92495:
The `MethodInvocation` from fern imports as `from vellum import Graph.from_set` which is invalid. So we introduce a new AstNode that we need to support the correct import


ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
```diff
@@ -183,19 +183,6 @@ exports[`Workflow > graph > should define nested sets of nodes without compilati
 "
 `;
 
-exports[`Workflow > graph > should handle loops of conditionals 1`] = `
```
dvargasfuertes, line 186: we just moved this test up a level to account for the import

PR #786, opened 2025-01-31
Add codegen support for workflow output values and variables

NgoHarrison:


ee/codegen/src/generators/workflow.ts
```diff
@@ -107,6 +112,29 @@ export class Workflow {
     return outputsClass;
   }
 
+  private getTargetHandleId(nodeData: WorkflowDataNode): string {
+    let targetHandleId: string;
+    if (nodeData.type === "GENERIC") {
+      // TODO: Fill this out
+      targetHandleId = "";
```
NgoHarrison: How should I grab the target handle id when given a generic node? Should I be doing cross lookups from the generic node ports and the workflow edges?

ee/codegen/src/generators/workflow.ts
```diff
@@ -416,19 +444,59 @@ export class Workflow {
         initializer: python.TypeInstantiation.dict(
           this.workflowContext.workflowOutputContexts.map(
             (workflowOutputContext) => {
-              const terminalNodeData =
+              const finalOutput =
                 workflowOutputContext.getFinalOutputNodeData();
+              let outputNodeId: string;
+              let targetHandleId: string;
+              let outputId: string;
+              let nodeInputId: string | undefined;
```
NgoHarrison: How should I define this value when we have `output_values` defined? This is meant to represent the input on terminal nodes but if we have output values, should we leave this as undefined or populate it?

poetry.lock
dvargasfuertes, line 1: [blocking] why was this deleted? Can we restore?

ee/codegen_integration/fixtures/simple_workflow_node_with_output_values/code/nodes/templating_node.py
```diff
@@ -0,0 +1,11 @@
+from vellum.workflows.nodes.displayable import TemplatingNode as BaseTemplatingNode
+from vellum.workflows.state import BaseState
+
+from ..inputs import Inputs
+
+
+class TemplatingNode(BaseTemplatingNode[BaseState, str]):
```
dvargasfuertes, line 7: [nit] Would love to use `BaseNode` as our "default" node that we use for node agnostic functionality/testing going forward

ee/vellum_ee/workflows/display/vellum.py
```diff
@@ -113,7 +113,7 @@ class WorkflowOutputVellumDisplayOverrides(WorkflowOutputDisplay, WorkflowOutput
     name: str
     label: str
     node_id: UUID
-    node_input_id: UUID
+    node_input_id: Union[UUID, None]
```
dvargasfuertes: [blocking] why are we making this field nullable?

ee/codegen/src/types/vellum.ts
```diff
@@ -710,11 +717,17 @@ export interface WorkflowDisplayData {
   viewport: WorkflowDisplayDataViewport;
 }
 
+export interface WorkflowOutputValue {
+  outputVariableId: string;
+  value: WorkflowOutputPointer;
```
dvargasfuertes: [blocking] this should just use `WorkflowValueDescriptor` - https://github.com/vellum-ai/vellum/blob/main/django/app/domain_models/types/workflows.py#L521

ee/codegen/src/project.ts
```diff
@@ -333,19 +335,48 @@ ${errors.slice(0, 3).map((err) => {
...
+                  if (outputValue) {
+                    this.workflowContext.addWorkflowOutputContext(
+                      new WorkflowOutputContext({
+                        workflowContext: this.workflowContext,
+                        workflowOutputValue: outputValue,
+                      })
+                    );
+                  }
+                }
+              );
```
dvargasfuertes: [blocking] We'll want this generation to happen outside of the nodes loop. As stated, a workflow with two terminal nodes with two final outputs would generate four output attributes. Which now, I'd love a test in `project.test.ts` protecting against that bug 😁

ee/vellum_ee/workflows/display/vellum.py
dvargasfuertes: Oh I see. Will comment above in the other comment thread.

ee/codegen/src/generators/workflow.ts
dvargasfuertes: I suggest we split out a refactor from this PR that removes `node_input_id` from `WorkflowOutputVellumDisplayOverrides`. It arguably should've never been there since it's a node-level concern just for final output nodes rather than a global workflow output level concern

ee/codegen/src/generators/workflow.ts
```diff
@@ -416,19 +444,59 @@ export class Workflow {
...
+                targetHandleId = this.getTargetHandleId(
+                  referencedNode.nodeData
+                );
+                outputId = finalOutput.value.nodeOutputId;
+                name = referencedOutput.name;
+                label = this.getLabel(referencedNode.nodeData);
+                displayData = referencedNode.nodeData.displayData;
+              }
 
               const edge = this.getEdges().find((edge) => {
```
dvargasfuertes: [oos] We may also want to split out a refactor that removes `edge_id` from `WorkflowOutputVellumDisplayOverrides`. It's only used to eventually fill in this value: https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/vellum_ee/workflows/display/workflows/vellum_workflow_display.py#L185. But, by nature of the edge being synthetic, we are likely fine to depend only on synthetic ids for a synthetically made edge

ee/codegen/src/generators/workflow.ts
dvargasfuertes: `node.trigger.id` is the replacement for target_handle_id: https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/codegen/src/__test__/helpers/edge-data-factories.ts#L72

However, as I mentioned below, we may not need this logic if we remove the `edge_id` attribute from `WorkflowOutputVellumDisplayOverrides`

ee/codegen/src/generators/workflow.ts
NgoHarrison: Updated

poetry.lock
NgoHarrison: Restored

ee/vellum_ee/workflows/display/vellum.py
NgoHarrison: Updated

ee/codegen/src/types/vellum.ts
NgoHarrison: Updated

ee/codegen/src/project.ts
NgoHarrison: Updated

ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -1412,33 +1412,34 @@ export function nodePortsFactory(ports?: Partial<NodePort>[]): NodePort[] {
...
     id: id ?? uuidv4(),
+    label: label,
     type: WorkflowNodeType.GENERIC,
     base: {
       module: ["vellum", "workflows", "nodes", "bases", "base"],
       name: "BaseNode",
     },
     definition: {
-      name,
+      name: label,
```
dvargasfuertes, line 1442: [nit] we likely want to throw this into whatever class name normalization fcn we have

ee/codegen/src/generators/workflow.ts
```diff
@@ -107,6 +116,28 @@ export class Workflow {
...
+      targetHandleId = nodeData.trigger.id;
+    } else if ("targetHandleId" in nodeData.data) {
+      targetHandleId = nodeData.data.targetHandleId;
+    } else {
+      throw new WorkflowGenerationError(
+        `${nodeData.data.label} with id ${nodeData.id} is missing required targetHandleId`
+      );
+    }
+    return targetHandleId;
+  }
```
dvargasfuertes: [nit] We can now delete this and the `const edge = this.getEdges()` logic below

ee/codegen/src/project.ts
dvargasfuertes: looks like it's still in the `nodes` loop?

ee/codegen/src/generators/workflow.ts
NgoHarrison: We ultimately use this for the target handle id field on WorkflowOutputVellumDisplayOverrides. Is it correct to say we would still this helper but not the getEdges() logic

dvargasfuertes: Ah, then we likely could remove target handle id from `WorkflowOutputVellumDisplayOverrides` too!

After removing node_input_id and edge_id, I realized those were technically breaking changes that workflows on old versions and devs before installing could run into. Let's split out a pr making that field optional and has us stop filling it, so that we could eventually remove this here

We can then remove it outright along with any other breaking changes we want to group for a 0.14.0 release, to get into the habit of how we should manage breaking changes

NgoHarrison: Updated

dvargas92495 (approved):
Good to 🚢  once CI is resolved and `getTargetHandleId` is removed

PR #788, opened 2025-01-31
Fix the error output coming out of templating nodes

dvargas92495:
Context:

Screenshot 2025-01-31 at 11 24 12 AM

That output fails due to an attribute error


marlenc3 (approved):

NgoHarrison (commented):
✅

PR #796, opened 2025-02-01
[SC-6289] Correctly serialize subworkflow deployment inpud ids

NgoHarrison:
Before we were using `subworkflow_input_ids_by_name: ClassVar[Dict[str, UUID]] ` in subworkflow deployment serialization but this field was never codegen'd. Now we use `node_input_ids_by_name` field to look up the id for the input given name

### DIFF
--- a/ee/codegen_integration/conftest.py
+++ b/ee/codegen_integration/conftest.py
@@ -2,9 +2,10 @@ import pytest
 from datetime import datetime
 import glob
 import os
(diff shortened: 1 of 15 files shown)


marlenc3 (approved):
nice! 🔥

dvargas92495 (commented):
I've seen this in a few cases - would be nice to continue to standardize wherever we see discrepancies

✅

PR #824, opened 2025-02-01
Reproducing the test case where nested set of conditionals point to t…

dvargas92495:
…he wrong final output nodes

Discovered a variant of this graph on a customer workflow. I tried to MVP it as much as possible, but without going into solution mode, couldn't break it down further.

I believe the issue is once we create a set of nodes as an edge target, we need to break the set if we discover an edge pointing to a member of that set.

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
@@ -174,6 +174,25 @@ exports[`Workflow > graph > should be correct for two branches merging from sets
 "
 `;
 
+exports[`Workflow > graph > should be pointing to the correct terminal nodes from a nested set of conditionals 1`] = `
+"{
+    FirstCheckNode.Ports.if_port
+    >> {
+        FirstInnerCheckNode.Ports.if_port >> {
+            SecondInnerCheckNode.Ports.if_port >> FinalCheckNode,
+            SecondInnerCheckNode.Ports.else_port >> OuterOutputNode,
+        },
+        FirstInnerCheckNode.Ports.else_port
+        >> {
+            FinalCheckNode.Ports.if_port >> InnerTerminalNode,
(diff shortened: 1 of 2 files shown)


ee/codegen/src/__test__/graph-attribute.test.ts
```diff
@@ -686,5 +687,99 @@ describe("Workflow", () => {
...
+        [entrypointNode, firstCheckNode],
+        [[firstCheckNode, "if_port"], firstInnerCheckNode],
+        [[firstCheckNode, "else_port"], finalCheckNode],
+        [[finalCheckNode, "if_port"], innerTerminalNode],
+        [[finalCheckNode, "else_port"], outerOutputNode],
+        [[secondInnerCheckNode, "if_port"], finalCheckNode],
+        [[secondInnerCheckNode, "else_port"], outerOutputNode],
+        [[firstInnerCheckNode, "if_port"], secondInnerCheckNode],
+        [[firstInnerCheckNode, "else_port"], finalCheckNode],
+      ]);
```
dvargasfuertes, line 758: This graph is why I had all those PRs cleaning up the graph attribute tests. I anticipate way more complex edge cases to come

ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
```diff
@@ -174,6 +174,25 @@ exports[`Workflow > graph > should be correct for two branches merging from sets
 "
 `;
 
+exports[`Workflow > graph > should be pointing to the correct terminal nodes from a nested set of conditionals 1`] = `
```
dvargasfuertes, line 177: Manually edited this snapshot to match the intended target

dvargasfuertes: Note that there are actually three valid graphs possible. The set for `FinalCheckNode` could go in any of the three spots where its referenced

marlenc3 (approved):

PR #826, opened 2025-02-01
Fix test surrounding multiple nodes pointing to nested set of conditional ports

dvargas92495:
Finally resolves a pretty intricate test case of graph generation that we weren't covering. Had to do with improper splitting of a Graph set


ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
```diff
@@ -176,16 +176,14 @@ exports[`Workflow > graph > should be correct for two branches merging from sets
...
+    FirstCheckNode.Ports.if_port >> FirstInnerCheckNode >> SecondInnerCheckNode,
+    FirstCheckNode.Ports.else_port,
+} >> Graph.from_set(
+    {
         FinalCheckNode.Ports.if_port >> InnerTerminalNode,
         FinalCheckNode.Ports.else_port >> OuterOutputNode,
-    },
-}
+    }
+)
```
dvargasfuertes, line 186: Found an optimization too while I was at it, hence the snapshot edit

PR #829, opened 2025-02-02
Update code generator function to pass future builds

dvargas92495:
Context: https://github.com/vellum-ai/codegen-service/pull/66/files


m-abboud (approved):

m-abboud:
Can be improved a little by running the type check after installing the current package and then changing it to the version with the skip check comment.

dvargas92495:
> after installing the current package

At this point, each of the dynamic modules fail the type check bc they don't have type declaration files any more (they're now being removed after installation). Unless I'm misunderstanding?

PR #863, opened 2025-02-03
Update workflow value descriptors to use reference instead of old pointer types

NgoHarrison:
Context: This PR came up from this comment: https://github.com/vellum-ai/vellum-python-sdks/pull/786#discussion_r1937358227

Previously, I made workflow value descriptor use the old `NodeInputValuePointerRule` where it could a union of multiple pointers that essentially have the `data` key in it's shape. However, the new workflow reference pointer types do not have this `data` key and this PR updates it so that workflow value descriptors use these new generators. I was contemplating just make a soft wrapper on the old pointers and just constructing these internal objects which essentially just wrap the values with `data` but this seemed wrong and also would make it more work to uncouple in the future


ee/codegen/src/generators/workflow-value-descriptor-reference/BaseNodeInputWorkflowReferencePointer.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export declare namespace BaseNodeInputWorkflowReferencePointer {
+  export interface Args<T extends WorkflowValueDescriptorReferenceType> {
+    workflowContext: WorkflowContext;
+    nodeInputWorkflowReferencePointer: T;
+    iterableConfig?: IterableConfig;
+  }
+}
+
+export abstract class BaseNodeInputWorkflowReferencePointer<
```
dvargasfuertes: This is neat

dvargasfuertes: [nit] Reference and Pointer are a bit redundant with each other, Id go with the former

dvargas92495 (approved):
🚀

PR #867, opened 2025-02-03
Prompt Deployment was failing to support chat history from Vembda

dvargas92495:
Screenshot 2025-02-03 at 1 59 03 PM

PR #868, opened 2025-02-03
Test reproducing no support in subworkflow deployment nodes for chat …

dvargas92495:
…history

Same as https://github.com/vellum-ai/vellum-python-sdks/pull/867 but for subworkflows


NgoHarrison (commented):
✅

PR #869, opened 2025-02-03
Support any array as an input to subworkflow deployments

dvargas92495:
While working on https://github.com/vellum-ai/vellum-python-sdks/pull/868, I noticed that we don't support lists as inputs in our workflow deployment nodes


NgoHarrison (commented):
✅

PR #873, opened 2025-02-03
Fix circular dependency triggered when importing WorkflowEventDisplayContext

dvargas92495:
I was making some Workflow Server changes when I kept hitting this running tests locally:

Screenshot 2025-02-03 at 3 44 19 PM

Not sure why this is not affecting us in prod. But in any case, repro'd the import error with a test and solved by moving a method

### DIFF
--- a/ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
+++ b/ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
@@ -29,9 +29,15 @@ from vellum_ee.workflows.display.base import (
     WorkflowOutputDisplayOverridesType,
     WorkflowOutputDisplayType,
 )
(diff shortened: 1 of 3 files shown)


ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -343,4 +349,42 @@ def gather_event_display_context(
             display_class = importlib.import_module(workflow_display_module)
         except ModuleNotFoundError:
             return None
-        return display_class.WorkflowDisplay(workflow_class).display_context.build_event_display_context()
+
+        display_context = display_class.WorkflowDisplay(workflow_class).display_context
```
dvargasfuertes, line 353: no logical changes from above, just copy paste.

marlenc3 (approved):

PR #875, opened 2025-02-03
Disable formatting codegen option

m-abboud:
For when we are not pulling it down. Shaves off lots of latency not doing the isort in sandbox


dvargas92495 (approved):
Long term, I could see grouping this with the same process that is running our code merging (where we're also running isort/black I think) so that there's only one python subprocess invoked

m-abboud:
> Long term, I could see grouping this with the same process that is running our code merging (where we're also running isort/black I think) so that there's only one python subprocess invoked

@dvargas92495  Most optimal is to have this in django with code merge. If isort/black can be called programmatically without causing a new process that is

dvargas92495:
Oh that's a great idea.

Isort for sure can, I'd have to double check black

PR #877, opened 2025-02-03
[Part 1] Plumb Through Node Contexts

noanflaherty:
In order to support `LazyReference`, we need to know which node we're currently operating on. For this, we need `nodeContext`.

This PR simply passes `nodeContext` everywhere we need it, and simplify the call signature by removing `workflowContext` since that can be derived from `nodeContext`.

Next Steps:
- Make use of this plumbing to actually detect self-references and use `LazyReference`
- Do the same for Workflow Value Descriptors once everything is proven out for legacy Node Inputs

### DIFF
--- a/ee/codegen/src/__test__/node-inputs/node-input-value-pointer-rules/execution-counter-pointer.test.ts
+++ b/ee/codegen/src/__test__/node-inputs/node-input-value-pointer-rules/execution-counter-pointer.test.ts
@@ -29,10 +29,13 @@ describe("ExecutionCounterPointer", () => {
     const workflowContext = workflowContextFactory();
 
     const node = searchNodeDataFactory();
-    await nodeContextFactory({ workflowContext, nodeData: node });
+    const nodeContext = await nodeContextFactory({
+      workflowContext,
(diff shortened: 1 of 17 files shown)


NgoHarrison (approved):

ee/codegen/src/__test__/node-inputs/node-input-value-pointer-rules/execution-counter-pointer.test.ts
```diff
@@ -29,10 +29,13 @@ describe("ExecutionCounterPointer", () => {
...
     const node = searchNodeDataFactory();
-    await nodeContextFactory({ workflowContext, nodeData: node });
+    const nodeContext = await nodeContextFactory({
+      workflowContext,
+      nodeData: node,
+    });
 
     const executionCounterPointer = new ExecutionCounterPointerRule({
-      workflowContext: workflowContext,
+      nodeContext,
```
dvargasfuertes, line 38: this is a dope change

dvargas92495 (commented):
🔥 ✅

PR #878, opened 2025-02-03
Fix for inline code exec with json output

marlenc3:
still todo: issues w/ script files (running into some problems with finding file)


dvargas92495 (approved):
Let's please add a test reproducing the failure! either in this PR or in a fast follow

PR #879, opened 2025-02-04
[SC-6449] Add support for conditions beginning with constant value references on generic nodes

NgoHarrison:
This PR adds codegen support for conditions beginning with constant value references on generic nodes. The one key assumption is that the data we are getting is valid. Therefore, we don't do any guardrails / type checks.

### DIFF
--- a/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/workflow-value-descriptor.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/workflow-value-descriptor.test.ts.snap
@@ -20,48 +20,17 @@ exports[`WorkflowValueDescriptor > expressions > generates unary expression with
 "
 `;
 
-exports[`WorkflowValueDescriptor > expressions that begin with constant values > generates binary expression with in operator and string array 1`] = `
-""test" in [
-    "test",
-    "other",
-]
+exports[`WorkflowValueDescriptor > expressions that begin with constant values > generates binary expression beginning with constant value reference 1`] = `
+"ConstantValueReference("Hello, World!").equals(Inputs.count)
 "
 `;
 
-exports[`WorkflowValueDescriptor > expressions that begin with constant values > generates binary expression with lhs of number and rhs of number 1`] = `
-"5 == 500
+exports[`WorkflowValueDescriptor > expressions that begin with constant values > generates ternary expression beginning with constant value reference 1`] = `
+"ConstantValueReference(123).between(Inputs.count, Inputs.another_count)
 "
 `;
 
-exports[`WorkflowValueDescriptor > expressions that begin with constant values > generates binary expression with lhs of string and rhs of string 1`] = `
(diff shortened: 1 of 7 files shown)


ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/workflow-value-descriptor.test.ts.snap
```diff
@@ -19,3 +19,49 @@ exports[`WorkflowValueDescriptor > expressions > generates unary expression with
 "Inputs.count.is_null()
 "
 `;
+
+exports[`WorkflowValueDescriptor > expressions that begin with constant values > generates binary expression with in operator and string array 1`] = `
+""test" in [
+    "test",
+    "other",
+]
```
dvargasfuertes: [blocking] These are all going to resolve immediately, which will be incorrect if the RHS is a descriptor. For example, if I wanted to specify `"hello" == Inputs.world`, python will resolve this _immediately to False_, regardless of whether the use submits a workflow with `"hello"` as an input value or not.

This is why we have [this guy](https://github.com/vellum-ai/vellum-python-sdks/blob/37d94e6faad41cf7d78e7e2e534ca6a602cea435/src/vellum/workflows/references/constant.py#L11) that we need to use here. This allows us to delay resolution until the node comes online. So, using my example above, I would expect these all to compile to `ConstantValueReference("hello").equals(Inputs.world)`

This way, we likely don't need a test for each operator bc we can reuse our existing WorkflowValueDescriptor generation since `ConstantValueReference` is also a descriptor

lmk if ^ is clear, and happy to huddle this morning if not

NgoHarrison: Ahhh I see, thanks for clarifying! Updated this to wrap lhs with `ConstantValueReference`

ee/codegen/src/generators/expression.ts
```diff
@@ -49,22 +59,58 @@ export class Expression extends AstNode {
...
+
+  // We are assuming that the expression contains "good data". If the expression contains data
+  // where the generated expression is not correct, update the logic here with guardrails similar to the UI
+  private generateLhsAsConstantReference(lhs: AstNode): AstNode {
+    const constantValueReference = python.reference({
+      name: "ConstantValueReference",
+      modulePath: [
+        ...this.workflowContext.sdkModulePathNames.WORKFLOWS_MODULE_PATH,
+        "references",
+        "constant",
```
dvargasfuertes: [nit] instead of injecting the workflow context into each of these classes, let's just add this to our constants file.

I originally added `sdkModulePathNames` in order to navigate our migration from prototype repo -> sdk, but now that that's over, we likely could have them all as constants again

dvargas92495 (approved):
love it

ee/codegen/src/generators/expression.ts
NgoHarrison: Updated

PR #880, opened 2025-02-04
Add test for no graph workflow

vincent0426:
Ensure a workflow will still fulfilled without a graph with empty outputs


tests/workflows/no_graph/workflow.py
```diff
@@ -0,0 +1,7 @@
+from vellum.workflows import BaseWorkflow
+from vellum.workflows.inputs.base import BaseInputs
+from vellum.workflows.state.base import BaseState
```
dvargasfuertes: [nit] Let's remove these imports, showing that we don't need generic parameters in the base case

dvargas92495 (approved):

PR #885, opened 2025-02-04
Resolve more chat history model nonsense

dvargas92495:
I learned two things in this saga:
- We should prioritize getting all of these `*Request` equivalent dataclasses out of our SDK soon after marketing release
- Where there's a usage of `cast`, there will eventually be a bug


marlenc3 (commented):
✅

PR #887, opened 2025-02-04
[Codegen] Support for Lazy References of Nodes

noanflaherty:
This upgrades our the fern python generator and adds support for `LazyReference`. I've verified the changes manually and locally and will fast-follow with some tests. Getting this up asap to unblock customer workflows.

### DIFF
--- a/ee/codegen/src/generators/access-attribute.ts
+++ b/ee/codegen/src/generators/access-attribute.ts
@@ -1,0 +1,30 @@
+import { AstNode } from "@fern-api/python-ast/core/AstNode";
+import { Writer } from "@fern-api/python-ast/core/Writer";
+
+export declare namespace AccessAttribute {
+  interface Args {
+    lhs: AstNode;
+    rhs: AstNode;
+  }
+}
+
+export class AccessAttribute extends AstNode {
+  private readonly lhs: AstNode;
+  private readonly rhs: AstNode;
+
+  constructor({ lhs, rhs }: AccessAttribute.Args) {
+    super();
+
+    this.lhs = lhs;
+    this.inheritReferences(lhs);
+
+    this.rhs = rhs;
+    this.inheritReferences(rhs);
+  }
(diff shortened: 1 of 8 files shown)


ee/codegen/src/generators/access-attribute.ts
```diff
@@ -0,0 +1,30 @@
+import { AstNode } from "@fern-api/python-ast/core/AstNode";
+import { Writer } from "@fern-api/python-ast/core/Writer";
+
+export declare namespace AccessAttribute {
```
noanflaherty, line 4: Introduces a new low-level ast node that should probably live in the fern library.

ee/codegen/src/generators/node-inputs/node-input-value-pointer.ts
```diff
@@ -47,29 +56,74 @@ export class NodeInputValuePointer extends AstNode {
...
         break;
       }
 
-      writer.write(".coalesce(");
-      rule.write(writer);
-      writer.write(")");
+      const coalesceMethod = python.invokeMethod({
+        methodReference: python.reference({ name: "coalesce" }),
+        arguments_: [python.methodArgument({ value: rule })],
+      });
```
dvargasfuertes, line 83: oh neat - I wonder if I could've used this instead of `StaticMethodInvocation` the other day

dvargasfuertes: I suppose I still needed `AccessAttribute`

ee/codegen/src/generators/access-attribute.ts
dvargasfuertes: I think we should liberally continue to add more of these here to reduce fern roundtrips, until they add their own

dvargas92495 (approved):

NgoHarrison (commented):
✅

PR #890, opened 2025-02-05
[Fix] Fix Lazy Reference Imports

noanflaherty:
This fixes LazyReference such that we generate valid code.

## Before
```
from vellum.workflows.nodes.displayable import TemplatingNode as BaseTemplatingNode
from vellum.workflows.references import LazyReference
from  import coalesce
class TemplatingNode(BaseTemplatingNode[BaseState, str]):
    template = """Hello {{ var_1 }}"""
    inputs = {"var_1": LazyReference(lambda: TemplatingNode.Outputs.result.coalesce("World")),}
```

## After
```
from vellum.workflows.nodes.displayable import TemplatingNode as BaseTemplatingNode
from vellum.workflows.references import LazyReference


class TemplatingNode(BaseTemplatingNode[BaseState, str]):
    template = """Hello {{ var_1 }}"""
    inputs = {
        "var_1": LazyReference(lambda: TemplatingNode.Outputs.result.coalesce("World")),
    }
```

### DIFF
--- a/ee/codegen/src/generators/static-method-invocation.ts
+++ b/ee/codegen/src/generators/static-method-invocation.ts
@@ -4,7 +4,7 @@ import { Writer } from "@fern-api/python-ast/core/Writer";
 import type { python } from "@fern-api/python-ast";
 
 export class StaticMethodInvocation extends AstNode {
-  private reference: python.Reference;
+  private reference: python.AstNode;
   private methodName: string;
   private arguments: python.MethodArgument[];
   constructor({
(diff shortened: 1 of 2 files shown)


ee/codegen/src/generators/static-method-invocation.ts
```diff
@@ -4,22 +4,23 @@ import { Writer } from "@fern-api/python-ast/core/Writer";
 import type { python } from "@fern-api/python-ast";
 
 export class StaticMethodInvocation extends AstNode {
-  private reference: python.Reference;
+  private reference: python.AstNode;
```
dvargasfuertes, line 7: hm I guess. It _was_ originally meant for references lol

dvargas92495 (approved):

ee/codegen/src/generators/static-method-invocation.ts
noanflaherty: I've generally found that every time I try to use a specific type, I later find that there are valid use cases to just make it `AstNode`.

PR #899, opened 2025-02-05
Serialize String Lazy REferences

dvargas92495:
Serialization support for the new string Lazy Reference's

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -344,7 +344,25 @@ class BaseNodeDisplay(Generic[NodeType], metaclass=BaseNodeDisplayMeta):
 
         if isinstance(value, LazyReference):
             if isinstance(value._get, str):
-                raise NotImplementedError("LazyReference with a string is not implemented")
+                reference_parts = value._get.split(".")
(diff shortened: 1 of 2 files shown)


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -344,7 +344,25 @@ def serialize_value(self, display_context: "WorkflowDisplayContext", value: Base
...
+
+                output_name = reference_parts[-1]
+                nested_class_name = reference_parts[-2]
+                if nested_class_name != "Outputs":
+                    raise Exception(
+                        f"Failed to parse lazy reference: {value._get}. Outputs are the only node reference supported."
+                    )
+
+                node_class_name = ".".join(reference_parts[:-2])
+                for node in display_context.node_displays.keys():
```
dvargasfuertes, line 361: not clear to me whether to use `node_displays` or `global_node_displays`

noanflaherty: Global node displays is if you want nodes within a subworkflow to have access to nodes outside the subworkflow

dvargasfuertes: so we likely need to check both eventually

PR #900, opened 2025-02-05
Add test for circular lazy references

dvargas92495:
This PR defines the test we need to succeed. We begin impl on the next one


noanflaherty (commented):
✅

PR #901, opened 2025-02-05
[SC-6508] Add global output variable contexts to nested workflows

NgoHarrison:
Context: This is the fix for inline subworkflow nodes not getting codegen'ed discovered from QA'ing customer workflows. This was a regression from the output values I merged in and we did not catch this since we had no output_values on inline subworkflow nodes


noanflaherty (approved):

dvargas92495 (commented):
> This was a regression from the output values I merged in

From the test changes, I'm not seeing something that would've caught this. Could we fast follow with a new test/test update that would've protected this regression? Does this also solve for map nodes?

✅

PR #904, opened 2025-02-05
Console warn when receiving bad edge data and not adding errors to workflow context

NgoHarrison:
Context: Currently, workflow runs will fail if we have bad edge data since it runs in strict mode. This PR changes the error notification by instead console warning instead of adding to our workflow context errors which is what strict mode checks against. This essentially does the filter at processing time. Longer term, we probably want something similar to our error list for workflow context but with warnings instead. Alternative short term approach to this was to pre-process the edges before but it felt a little redundant

### DIFF
--- a/ee/codegen/src/generators/workflow.ts
+++ b/ee/codegen/src/generators/workflow.ts
@@ -17,6 +17,7 @@ import { BaseState } from "src/generators/base-state";
 import {
   BaseCodegenError,
   NodeNotFoundError,
+  NodePortGenerationError,
   WorkflowGenerationError,
 } from "src/generators/errors";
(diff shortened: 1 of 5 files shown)


ee/codegen/src/generators/workflow.ts
```diff
@@ -389,7 +390,7 @@ export class Workflow {
               this.workflowContext.getPortContextById(sourcePortId);
           } catch (e) {
             if (e instanceof BaseCodegenError) {
```
dvargasfuertes: [nit] Let's only warn on `NodePortGenerationError`, like how we made it more specific below

dvargas92495 (approved):

PR #906, opened 2025-02-06
Raise a better error if the user tries to Serialize nested references

dvargas92495:
We don't plan to support this anytime soon, so let's raise a better error


vincent0426:
✅

vincent0426 (approved):

marlenc3 (commented):
✅

NgoHarrison (commented):
✅

PR #907, opened 2025-02-06
Implement string lazy reference codegen

dvargas92495:
Need to split out some of this work - lots of generic node codegen not yet implemented

EDIT: Was going to split this out some more, but need to go to bed soon and want to have a release cut for Bit before I do

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -1,5 +1,21 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`WorkflowProjectGenerator > Nodes with forward references > should generate a proper Lazy Reference for the first node 1`] = `
+"from vellum.workflows.nodes.displayable import BaseNode
+from vellum.workflows.references import LazyReference
+
+
+class FirstNode(BaseNode):
+    forward = LazyReference("SecondNode.Outputs.output")
+
+    class NodeTrigger(BaseNode.Trigger):
+        merge_behavior = "AWAIT_ATTRIBUTES"
+
+    class Outputs(BaseNode.Outputs):
+        pass
+"
+`;
+
 exports[`WorkflowProjectGenerator > code execution node at project level > should not autoformat the script file 1`] = `
 "import foo, bar
 baz = foo + bar
--- a/ee/codegen/src/__test__/node-inputs/node-input-value-pointer-rules/node-input-value-pointer-rule.test.ts
+++ b/ee/codegen/src/__test__/node-inputs/node-input-value-pointer-rules/node-input-value-pointer-rule.test.ts
@@ -5,7 +5,8 @@ import {
   workflowContextFactory,
 } from "src/__test__/helpers";
 import { inputVariableContextFactory } from "src/__test__/helpers/input-variable-context-factory";
-import { WorkflowContext } from "src/context";
+import { genericNodeFactory } from "src/__test__/helpers/node-data-factories";
+import { WorkflowContext, createNodeContext } from "src/context";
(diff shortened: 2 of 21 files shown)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1,5 +1,21 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`WorkflowProjectGenerator > Nodes with forward references > should generate a proper Lazy Reference for the first node 1`] = `
+"from vellum.workflows.nodes.displayable import BaseNode
+from vellum.workflows.references import LazyReference
+
+
+class FirstNode(BaseNode):
+    forward = LazyReference("SecondNode.Outputs.output")
```
dvargasfuertes, line 9: Main goal of PR

ee/codegen/src/__test__/node-inputs/node-input-value-pointer-rules/node-input-value-pointer-rule.test.ts
```diff
@@ -43,11 +44,23 @@ describe("NodeInputValuePointerRule", () => {
   });
 
   it("should generate correct AST for NODE_OUTPUT pointer", async () => {
-    vi.spyOn(nodeContext.workflowContext, "getNodeContext").mockReturnValue({
-      nodeClassName: "TestNode",
-      path: ["nodes", "test-node-path"],
-      getNodeOutputNameById: vi.fn().mockReturnValue("my_output"),
-    } as unknown as BaseNodeContext<WorkflowDataNode>);
```
dvargasfuertes, line 50: We made a change to `NodeContext`'s interface, which caused a bunch of these spies to fail. I argue that the real thing is just as performant, and in some ways simpler, to just use instead

ee/codegen/src/__test__/project.test.ts
```diff
@@ -922,7 +922,7 @@ baz = foo + bar
       output_variables: [],
       runner_config: {},
     };
-    it.skip("should generate a proper Lazy Reference for the first node", async () => {
+    it("should generate a proper Lazy Reference for the first node", async () => {
```
dvargasfuertes, line 925: Other main goal of the PR

ee/codegen/src/context/node-context/base.ts
```diff
@@ -13,11 +13,13 @@ export declare namespace BaseNodeContext {
   interface Args<T extends WorkflowDataNode> {
     workflowContext: WorkflowContext;
     nodeData: T;
+    importOrder: readonly number[];
```
dvargasfuertes, line 16: Most important file to review.

We use a readonly array here to account for subworkflows

ee/codegen/src/project.ts
```diff
@@ -347,31 +369,28 @@ ${errors.slice(0, 3).map((err) => {
...
+    if (entrypointNodes.length > 1) {
+      throw new WorkflowGenerationError("Multiple entrypoint nodes found");
+    }
 
-          nodesToGenerate.push(nodeData);
+    const entrypointNode = entrypointNodes[0];
+    if (!entrypointNode) {
+      throw new WorkflowGenerationError("Entrypoint node not found");
+    }
+    this.workflowContext.addEntrypointNode(entrypointNode);
```
dvargasfuertes, line 384: needed the entrypoint set before defining import order of nodes

ee/codegen_integration/fixtures/simple_merge_node/code/display/nodes/__init__.py
```diff
@@ -6,8 +6,8 @@
 
 __all__ = [
```
dvargasfuertes, line 7: is there any harm in this always being alphabetized?

PR #908, opened 2025-02-06
Support for referencing generic node outputs

dvargas92495:
Noticed while implementing [lazy string references](https://github.com/vellum-ai/vellum-python-sdks/pull/907/files) that we were not supporting output references to generic nodes. Now we do


marlenc3:
✅

noanflaherty (commented):
🤘

PR #913, opened 2025-02-06
Fix graph gen case where there's an edge between sets

dvargas92495:
This would fail to compile a workflow without `Graph.from_set()`


marlenc3:
👍

PR #914, opened 2025-02-06
missing id for function call def fix

marlenc3:
🙃 
not seeing this work e2e was bugging me, figured out why


ee/vellum_ee/workflows/display/nodes/vellum/inline_prompt_node.py
```diff
@@ -39,7 +39,7 @@ def serialize(
             self._generate_prompt_block(block, input_variable_id_by_name, [i]) for i, block in enumerate(node_blocks)
         ]
         functions = (
-            [self._generate_function_tools(function) for function in function_definitions]
+            [self._generate_function_tools(function, i) for i, function in enumerate(function_definitions)]
```
siddseethepalli, line 42: [nit] Prefer using `idx` (or similar) over `i` for index variables since `i` is more easily misread as `1` or another character

siddseethepalli (commented):
✅

dvargas92495 (commented):
✅

PR #919, opened 2025-02-06
Fix reproducing multi branch into node test case in graph generation

dvargas92495:
We were previously generating a `None` in this case - this test case and implementation resolves that


marlenc3 (approved):

NgoHarrison (commented):
✅

PR #921, opened 2025-02-06
Prioritize terminal node data over workflow output values

NgoHarrison:
Context: We are receiving bad data on output_values from workflows. This was discovered when QAing a customer workflow and the subworkflow was referencing the outer workflow's output variables and not the inner one. This related PR (https://github.com/vellum-ai/vellum/pull/7874) fixes this change for workflows going forward but the current existing workflows are polluted with bad data. This change prioritizes the legacy terminal nodes temporarily so that it points to the right outputs.

This also fixes SC-6545 where the inner subworkflow starts referencing the outer workflow's outputs

### DIFF
--- a/ee/codegen/src/context/workflow-context/workflow-context.ts
+++ b/ee/codegen/src/context/workflow-context/workflow-context.ts
@@ -303,7 +303,10 @@ export class WorkflowContext {
       outputVariableId,
       outputVariableContext
     );
-    this.addUsedOutputVariableName(outputVariableContext.name);
+    // This was added from the workflow output context. We should remove this once terminal node data is removed from data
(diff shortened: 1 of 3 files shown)


marlenc3 (approved):

dvargas92495 (approved):
Let's fast follow with a test communicating that we now prioritize terminal nodes

noanflaherty:
@NgoHarrison @dvargas92495 just a heads up - I’ve been cooking up a PR that should ensure output values are always non-null and always correct/consistent with terminal nodes. My hope is that once that is in, we can stop reading from terminal nodes altogether when determining workflow output.

dvargas92495:
Will that cover the case where we may now have bad data as a result of: https://github.com/vellum-ai/vellum/pull/7874 ?

PR #925, opened 2025-02-07
[SC-6566] Codegen prompt deployment name instead of id

NgoHarrison:


ee/codegen/src/generators/nodes/prompt-deployment-node.ts
```diff
@@ -34,7 +34,7 @@ export class PromptDeploymentNode extends BaseSingleFileNode<
     statements.push(
       python.field({
         name: "deployment",
-        initializer: python.TypeInstantiation.str(nodeData.promptDeploymentId),
+        initializer: python.TypeInstantiation.str(nodeData.label),
```
noanflaherty: Wait I don't think we want this.

We should be hitting our api to get the name from the id when building contexts, like we do elsewhere for guardrail nodes and subworkflow deployment nodes.

noanflaherty (changes requested):

ee/codegen/src/generators/nodes/prompt-deployment-node.ts
noanflaherty: Note that we want the `name` of the deployment, not its `label`.

NgoHarrison: Oh gotcha good point, will update

ee/codegen/src/context/node-context/prompt-deployment-node.ts
```diff
@@ -25,4 +32,31 @@ export class PromptDeploymentNodeContext extends BaseNodeContext<PromptNode> {
...
+        this.nodeData.data.releaseTag,
+        this.nodeData.data.promptDeploymentId
+      );
+    } catch (error) {
+      if (
+        error instanceof VellumError &&
+        typeof error.body === "object" &&
+        error.body !== null &&
+        "detail" in error.body &&
+        typeof error.body.detail === "string"
```
noanflaherty: Do we have this same logic elsewhere? If so, would be nice to consolidate into a shared util.

ee/codegen/src/generators/nodes/prompt-deployment-node.ts
```diff
@@ -31,12 +34,21 @@ export class PromptDeploymentNode extends BaseSingleFileNode<
...
-      python.field({
-        name: "deployment",
-        initializer: python.TypeInstantiation.str(nodeData.promptDeploymentId),
-      })
-    );
+    if (!this.nodeContext.deploymentHistoryItem) {
+      this.workflowContext.addError(
+        new NodeAttributeGenerationError(`
+          Failed to generate attribute: ${this.nodeData.data.label}.deployment`)
+      );
```
noanflaherty: Rather than throwing an error, maybe we can just log a warning and fall back to using the id?

noanflaherty (approved):
Now we're talkin'!

ee/codegen/src/generators/nodes/prompt-deployment-node.ts
NgoHarrison: Another good point, will update!

ee/codegen/src/context/node-context/prompt-deployment-node.ts
NgoHarrison: Yes we do, updated to own util

ee/codegen/src/generators/nodes/prompt-deployment-node.ts
NgoHarrison: Added fallback and test for fallback

noanflaherty (approved):
Looks great – thanks for iterating

dvargas92495 (commented):
✅

PR #926, opened 2025-02-07
Raise proper node exception for PRompt Deployment API Errors

dvargas92495:
Was causing try-node-wrapped prompt nodes to raise workflow rejected errors. Separately, we should investigate why internal errors within try node is causing the workflow to reject


marlenc3 (approved):

NgoHarrison (commented):
✅

PR #927, opened 2025-02-07
Conditional Expression should accept Undef

marlenc3:
ahhhhh currently working on a test - but fix for now

### DIFF
--- a/src/vellum/workflows/expressions/contains.py
+++ b/src/vellum/workflows/expressions/contains.py
@@ -24,11 +24,11 @@ class ContainsExpression(BaseDescriptor[bool], Generic[LHS, RHS]):
         # Support any type that implements the in operator
         # https://app.shortcut.com/vellum/story/4658
         lhs = resolve_value(self._lhs, state)
-        if not isinstance(lhs, (list, tuple, set, dict, str, UNDEF)):
-            raise ValueError(f"Expected a LHS that supported contains, got: {lhs.__class__.__name__}")
         # assumes that lack of is also false
         if lhs is UNDEF:
             return False
+        if not isinstance(lhs, (list, tuple, set, dict, str)):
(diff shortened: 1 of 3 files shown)


NgoHarrison (approved):

src/vellum/workflows/expressions/contains.py
```diff
@@ -23,6 +24,9 @@ def resolve(self, state: "BaseState") -> bool:
         # Support any type that implements the in operator
         # https://app.shortcut.com/vellum/story/4658
         lhs = resolve_value(self._lhs, state)
+        # assumes that lack of is also false
+        if lhs is UNDEF:
+            return False
         if not isinstance(lhs, (list, tuple, set, dict, str)):
             raise ValueError(f"Expected a LHS that supported contains, got: {lhs.__class__.__name__}")
```
dvargasfuertes, line 31: [q] Was this the error that was being throws? Wondering if we should just return false here

dvargasfuertes: [oos] Actually, I think we still want to raise an error here for other non-undefined types but make it rise up to a user error instead of internal error. I'll patch that up now

dvargas92495 (commented):
✅

src/vellum/workflows/expressions/contains.py
marlenc3: Heyo! Sorry for the late reply, but the error is that the conditional was against an error that is undefined, due to there being no error inside of the try/catch and so the type error below was being thrown

marlenc3: If we don't want this here, we can probably make it so Error is always at least None?

dvargasfuertes: Yea I think we could keep this here, followed up with the other changes I had in mind

PR #928, opened 2025-02-07
Try Catch -- API Nodes

marlenc3:


src/vellum/workflows/nodes/displayable/bases/api_node/node.py
```diff
@@ -83,9 +84,13 @@ def _local_execute_api(self, data, headers, json, method, url):
...
         client_vellum_secret = ClientVellumSecret(name=bearer_token.name) if bearer_token else None
-        vellum_response = self._context.vellum_client.execute_api(
-            url=url, method=method.value, body=data, headers=headers, bearer_token=client_vellum_secret
-        )
+        try:
+            vellum_response = self._context.vellum_client.execute_api(
+                url=url, method=method.value, body=data, headers=headers, bearer_token=client_vellum_secret
+            )
+        except ApiError as e:
+            NodeException(f"Failed to prepare HTTP request: {e}", code=WorkflowErrorCode.PROVIDER_ERROR)
```
noanflaherty: We should reserve provider error for errors from LLM Providers

```suggestion
            NodeException(f"Failed to prepare HTTP request: {e}", code=WorkflowErrorCode.NODE_EXECUTION)
```

noanflaherty (approved):

src/vellum/workflows/nodes/displayable/bases/api_node/node.py
marlenc3: aight, I'll put in this change, but do we want to align the local api catch error as well? seems its also provider 👍  I can follow up with another PR

noanflaherty: Hm not sure what you mean by local, but probably?

src/vellum/workflows/nodes/displayable/bases/api_node/node.py
```diff
@@ -83,9 +84,13 @@ def _local_execute_api(self, data, headers, json, method, url):
...
         client_vellum_secret = ClientVellumSecret(name=bearer_token.name) if bearer_token else None
-        vellum_response = self._context.vellum_client.execute_api(
-            url=url, method=method.value, body=data, headers=headers, bearer_token=client_vellum_secret
-        )
+        try:
+            vellum_response = self._context.vellum_client.execute_api(
+                url=url, method=method.value, body=data, headers=headers, bearer_token=client_vellum_secret
+            )
+        except ApiError as e:
+            NodeException(f"Failed to prepare HTTP request: {e}", code=WorkflowErrorCode.NODE_EXECUTION)
```
dvargasfuertes, line 92: [blocking] we're missing the `raise` here. let's fast follow with a test that would've caught this :D

PR #932, opened 2025-02-08
Catch Invalid Expressions during port resolutions to avoid internal error

dvargas92495:
Followup to something I noticed in [this PR](https://github.com/vellum-ai/vellum-python-sdks/pull/927#discussion_r1947847023). We want all other instances of these _expected_ exceptions to not cause an internal error either

### DIFF
--- a/src/vellum/workflows/expressions/does_not_contain.py
+++ b/src/vellum/workflows/expressions/does_not_contain.py
@@ -1,6 +1,7 @@
 from typing import Generic, TypeVar, Union
 
 from vellum.workflows.descriptors.base import BaseDescriptor
+from vellum.workflows.descriptors.exceptions import InvalidExpressionException
 from vellum.workflows.descriptors.utils import resolve_value
 from vellum.workflows.state.base import BaseState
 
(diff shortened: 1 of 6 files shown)


src/vellum/workflows/expressions/does_not_contain.py
```diff
@@ -24,7 +25,9 @@ def resolve(self, state: "BaseState") -> bool:
         # https://app.shortcut.com/vellum/story/4658
         lhs = resolve_value(self._lhs, state)
         if not isinstance(lhs, (list, tuple, set, dict, str)):
-            raise ValueError(f"Expected a LHS that supported contains, got: {lhs.__class__.__name__}")
```
dvargasfuertes, line 27: Going to fast follow this PR to replace all other `ValueError`s in expressions with this new exception

tests/workflows/invalid_port_resolution/tests/test_workflow.py
```diff
@@ -0,0 +1,23 @@
...
+
+    # GIVEN a workflow with an invalid port description
+    workflow = InvalidPortResolutionWorkflow()
+
+    # WHEN the workflow is executed
+    final_event = workflow.run()
+
+    # THEN the workflow raises the correct error
+    assert final_event.name == "workflow.execution.rejected"
+    assert final_event.error.code == WorkflowErrorCode.INVALID_INPUTS
```
dvargasfuertes, line 19: Main takeaway here - no more INTERNAL_ERROR's

marlenc3:
✅️

PR #933, opened 2025-02-08
Replace all expression ValueError's with new InvalidExpressionException

dvargas92495:
Followup from https://github.com/vellum-ai/vellum-python-sdks/pull/932


marlenc3:
✅️

PR #940, opened 2025-02-09
[SC-4008] Auto-Inherit BaseNodeOutputs in Outputs classes

vincent0426:
allow users to define class Outputs in Node without inheriting explicitly, should also do the same for workflow output

test:
```py
from vellum.workflows.nodes.bases.base import BaseNode
from vellum.workflows.state.base import BaseState


class Inputs(BaseInputs):
    input: str


class BasicGenericNode(BaseNode):
    class Outputs:  # no need to inherit
        output = Inputs.input


class BasicGenericNodeWorkflow(BaseWorkflow[Inputs, BaseState]):
    graph = BasicGenericNode

    class Outputs(BaseWorkflow.Outputs):  # still need now
        output = BasicGenericNode.Outputs.output
```

```py
from tests.workflows.basic_generic_node.workflow import BasicGenericNodeWorkflow, Inputs


def test_workflow__happy_path():
    # GIVEN a workflow that just passes through its input to a terminal node
    workflow = BasicGenericNodeWorkflow()

    # WHEN we run the workflow
    terminal_event = workflow.run(inputs=Inputs(input="hello"))

    # THEN the workflow should be fulfilled
    assert terminal_event.name == "workflow.execution.fulfilled", terminal_event
    assert terminal_event.outputs == {"output": "hello"}

```

### DIFF
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -46,9 +46,17 @@ class BaseNodeMeta(type):
         if "Outputs" in dct:
             outputs_class = dct["Outputs"]
             if not any(issubclass(base, BaseOutputs) for base in outputs_class.__bases__):
+                parent_outputs_class = next(
+                    (base.Outputs for base in bases if hasattr(base, "Outputs")),
+                    BaseOutputs,  # Default to BaseOutputs only if no parent has Outputs
+                )
+
+                # Filter out object from bases while preserving other inheritance
+                filtered_bases = tuple(base for base in outputs_class.__bases__ if base is not object)
+
                 dct["Outputs"] = type(
                     f"{name}.Outputs",
-                    (BaseOutputs,) + outputs_class.__bases__,  # Add BaseOutputs as first base
+                    (parent_outputs_class,) + filtered_bases,
                     {**outputs_class.__dict__, "__module__": dct["__module__"]},
                 )
         else:
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -43,7 +43,15 @@ class BaseNodeMeta(type):
         # TODO: Inherit the inner Output classes from every base class.
         #   https://app.shortcut.com/vellum/story/4007/support-auto-inheriting-parent-node-outputs
 
-        if "Outputs" not in dct:
+        if "Outputs" in dct:
+            outputs_class = dct["Outputs"]
+            if not any(issubclass(base, BaseOutputs) for base in outputs_class.__bases__):
+                dct["Outputs"] = type(
+                    f"{name}.Outputs",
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/bases/base.py
```diff
@@ -43,7 +43,15 @@ def __new__(mcs, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any
         # TODO: Inherit the inner Output classes from every base class.
         #   https://app.shortcut.com/vellum/story/4007/support-auto-inheriting-parent-node-outputs
 
-        if "Outputs" not in dct:
+        if "Outputs" in dct:
+            outputs_class = dct["Outputs"]
+            if not any(issubclass(base, BaseOutputs) for base in outputs_class.__bases__):
+                dct["Outputs"] = type(
+                    f"{name}.Outputs",
+                    (BaseOutputs,) + outputs_class.__bases__,  # Add BaseOutputs as first base
```
dvargasfuertes: [nit] we should likely filter out this `<class 'object'>` from these bases as it'll be redundant with `BaseOutputs` already being a subclass of it
Screenshot 2025-02-09 at 12 10 33 PM

src/vellum/workflows/nodes/bases/tests/test_base_node.py
```diff
@@ -148,3 +149,21 @@ class SomeNode(BaseNode):
...
+            foo: str
+
+    # TEST that the Outputs class is a subclass of BaseOutputs
+    assert issubclass(MyNode.Outputs, BaseOutputs)
+
+    # TEST that the Outputs class has the correct attributes
+    assert hasattr(MyNode.Outputs, "foo")
+
+    # WHEN the node outputs is initialized
+    outputs = MyNode.Outputs(foo="bar")  # type: ignore
```
dvargasfuertes: [q] why do we need the type ignore here?

dvargasfuertes: [nit] let's actually make this WHEN condition a standard `node = MyNode()`, `outputs = node.run()`, since I think our custom type help is on the instance of node instead of the class. Will likely need to add a `run` method to `MyNode`

dvargas92495 (approved):
good stuff!

src/vellum/workflows/nodes/bases/tests/test_base_node.py
vincent0426: > [q] why do we need the type ignore here?

because we didn't explicit have `__init__`  or inherit in `MyNode.Outputs`? instead using implicit inherit
mypy would throw `Unexpected keyword argument "foo" for "Outputs"  [call-arg]`

dvargas92495 (approved):
🚢

marlenc3 (approved):

PR #943, opened 2025-02-10
Auto-Inherit BaseWorkflow Outputs in Outputs classes

vincent0426:


dvargas92495 (approved):
Amazing! Will add the codegen for this in your sprint too while we're at it since we are codegenning a `(BaseOutputs)` inheritance right now.

cc: @awlevin - starting with version 0.14.0, we will no longer need our `Outputs` classes to inherit from the parent class manually. This means all of our examples could specially just specify `class Outputs:`

dvargas92495:
Gonna merge this in to include in the 0.14.0 release

PR #945, opened 2025-02-11
Add checks for optional input strings with null default value

NgoHarrison:
This is for optional strings with default values of null


marlenc3 (approved):

ee/codegen/src/generators/vellum-variable.ts
```diff
@@ -40,7 +40,7 @@ export class VellumVariable extends AstNode {
   private generateInitializerIfDefault(
     variable: VellumVariableWithName
   ): AstNode | undefined {
-    return variable.default
+    return variable.default && variable.default.value
```
dvargasfuertes, line 43: [nit] this is going to accidentally skip `""`. I would use the `isNil` helper

dvargas92495 (approved):

PR #946, opened 2025-02-11
[SC-6587] Do not codegen authorization type input if that field is undefined

NgoHarrison:
Came up from QAing customer workflow, api nodes were not being codegen because this field was undefined causing errors and skipping of this node

### DIFF
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -999,6 +999,7 @@ export type ApiNodeFactoryProps = {
     key: NodeInput;
     value: NodeInput;
   }[];
+  useUndefinedAuthorizationTypeInputId?: boolean;
 };
 
 export function apiNodeFactory({
@@ -1006,6 +1007,7 @@ export function apiNodeFactory({
   bearerToken,
   apiKeyHeaderValue,
   additionalHeaders,
+  useUndefinedAuthorizationTypeInputId = false,
 }: ApiNodeFactoryProps = {}): ApiNode {
   const bearerTokenInput = bearerToken ?? {
     id: "931502c1-23a5-4e2a-a75e-80736c42f3c9",
@@ -1040,6 +1042,9 @@ export function apiNodeFactory({
       combinator: "OR",
     },
   };
+
+  const defaultAuthorizationTypeInputId =
+    "de330dac-05b1-4e78-bee7-7452203af3d5";
 
   const additionalHeaderInputs =
     additionalHeaders ??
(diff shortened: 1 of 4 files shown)


marlenc3 (approved):

ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -1153,7 +1158,9 @@ export function apiNodeFactory({
       methodInputId: "9bf086d4-feed-47ff-9736-a5a6aa3a11cc",
       urlInputId: "480a4c12-22d6-4223-a38a-85db5eda118c",
       bodyInputId: "74865eb7-cdaf-4d40-a499-0a6505e72680",
-      authorizationTypeInputId: "de330dac-05b1-4e78-bee7-7452203af3d5",
+      authorizationTypeInputId: useUndefinedAuthorizationTypeInputId
+        ? undefined
+        : defaultAuthorizationTypeInputId,
```
dvargasfuertes, line 1163: [nit] even if we do this, the `authorizationTypeInput` is in `node_inputs`. my preference would be to define the following on the factory:

```typescript
authorizationTypeInput?: NodeInput
```

then we could use `authorizationTypeInput.id` here if present and add it to the `inputs` list below if present

ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
```diff
@@ -231,3 +231,24 @@ class APINode(BaseAPINode):
...
+    method = APIRequestMethod.POST
+    json = {}
+    headers = {
+        "foo": Inputs.additional_header_value,
+        "bar": Inputs.additional_header_value_1,
+        "baz": Inputs.additional_header_value_1,
+    }
+    api_key_header_key = None
+    api_key_header_value = "<my-api-value>"
+    bearer_token_value = "<my-bearer-token>"
```
dvargasfuertes, line 252: [nit] I would expect our example that has authorization type as undefined would not have these authorization related attributes either

dvargasfuertes, line 252: [nit] I would expect our example that has authorization type as undefined would not have these authorization related attributes either

ee/codegen/src/__test__/helpers/node-data-factories.ts
dvargasfuertes: could use `undefined` vs `null` to disambiguate between using a default value so that we don't have to update a bunch of tests and a test opting into "ignore this field"

dvargas92495 (approved):

PR #947, opened 2025-02-11
[SC-6310] Adding Optional wrapper for required == false

vincent0426:
- regenerate snapshot
- modify fixture test case to make it `require = true`

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/inputs.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/inputs.test.ts.snap
@@ -2,12 +2,12 @@
 
 exports[`Inputs > write > should convert input variable names into valid python attributes 1`] = `
 "from vellum.workflows.inputs import BaseInputs
-from typing import Optional, Union
+from typing import Union
 
 
 class Inputs(BaseInputs):
-    my_input: Optional[str]
-    my_input_1: Optional[Union[float, int]]
+    my_input: str
+    my_input_1: Union[float, int]
 "
 `;
 
@@ -24,25 +24,24 @@ class Inputs(BaseInputs):
 
 exports[`Inputs > write > should generate correct code for complex input variables 1`] = `
 "from vellum.workflows.inputs import BaseInputs
-from typing import Optional, Union, List
+from typing import Union, List
 from vellum import ChatMessage, SearchResult
 
 
 class Inputs(BaseInputs):
-    query: Optional[str]
-    max_runtime: Optional[Union[float, int]]
-    previous_chat_history: Optional[List[ChatMessage]]
-    prior_results: Optional[List[SearchResult]]
+    query: str
+    max_runtime: Union[float, int]
+    previous_chat_history: List[ChatMessage]
+    prior_results: List[SearchResult]
 "
 `;
 
 exports[`Inputs > write > should generate correct code when Inputs has a custom name 1`] = `
 "from vellum.workflows.inputs import BaseInputs
-from typing import Optional
 
 
 class CustomInputs(BaseInputs):
-    input1: Optional[str]
+    input1: str
 "
 `;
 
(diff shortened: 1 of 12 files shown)


ee/codegen/src/generators/vellum-variable.ts
```diff
@@ -27,9 +28,13 @@ export class VellumVariable extends AstNode {
...
+
+    const baseType = getVellumVariablePrimitiveType(variable.type);
+    const finalType = !variable.required ? Type.optional(baseType) : baseType;
+
     this.field = python.field({
       name: variable.name ?? variable.key,
-      type: getVellumVariablePrimitiveType(variable.type),
+      type: finalType,
       initializer: variable.default
         ? this.generateInitializerIfDefault(variable)
```
marlenc3, line 42: assuming this is the base of the update 👍  cool!

marlenc3 (approved):

NgoHarrison (approved):
Nice!

ee/codegen/src/__test__/__snapshots__/inputs.test.ts.snap
```diff
@@ -2,79 +2,93 @@
...
 "from vellum.workflows.inputs import BaseInputs
-from typing import Union
+from typing import Optional, Union
 
 
 class Inputs(BaseInputs):
-    my_input: str
-    my_input_1: Union[float, int]
+    my_input: Optional[str]
+    my_input_1: Optional[Union[float, int]]
```
dvargasfuertes: [nit] I'm a bit worried about all of these unrelated test snapshots updating. were we not using `"required": true` in the other fixtures?
- If not, let's add to minimize unrelated tests updating
- If so, then this would hint at a bug in our impl to investigate

dvargas92495 (approved):
Would like us to check the source of the other test snapshots updating first

ee/codegen/src/__test__/nodes/__snapshots__/subworkflow-deployment-node.test.ts.snap
vincent0426, line 1: output optional

ee/codegen/src/generators/nodes/subworkflow-deployment-node.ts
```diff
@@ -102,7 +102,12 @@ export class SubworkflowDeploymentNode extends BaseSingleFileNode<
         if (outputName) {
           outputsClass.add(
             codegen.vellumVariable({
-              variable: { name: outputName, type: output.type, id: output.id },
+              variable: {
+                name: outputName,
+                type: output.type,
+                id: output.id,
+                required: output.required,
```
vincent0426, line 109: propagate required value

ee/codegen/src/__test__/__snapshots__/vellum-variable.test.ts.snap
```diff
@@ -31,7 +31,7 @@ exports[`VellumVariableField > JsonVellumVariable snapshot 1`] = `
 `;
 
 exports[`VellumVariableField > NullVellumVariable snapshot 1`] = `
-"test: None
+"test: Optional[None]
```
dvargasfuertes: [nit] This should likely just be `None`. `Optional[None]` resolves to `None | None` which is just `None`

ee/codegen/src/__test__/nodes/subworkflow-deployment-node.test.ts
```diff
@@ -29,8 +29,8 @@ describe("SubworkflowDeploymentNode", () => {
       ).mockResolvedValue({
         name: "test-deployment",
         outputVariables: [
-          { id: "1", key: "output-1", type: "STRING" },
-          { id: "2", key: "output-2", type: "NUMBER" },
+          { id: "1", key: "output-1", type: "STRING", required: true },
+          { id: "2", key: "output-2", type: "NUMBER", required: true },
```
dvargasfuertes: I led you astray here, check out this thread: https://vellum-ai.slack.com/archives/C06P13ABK0W/p1739308671509959?thread_ts=1739302036.746689&cid=C06P13ABK0W

Basically, the `undefined` behavior is dependent on the context. So for workflow inputs, undefined means required: false, and for workflow outputs, undefined means required: true. There are legacy reasons driving this confusion

dvargas92495 (approved):

dvargas92495 (approved):

PR #948, opened 2025-02-11
[SC-6593] Correctly unwrap nested inputs in array inputs for code execution inline nodes

NgoHarrison:
Fix handling of nested VellumValue objects in CodeExecutorInput arrays. Previously, when handling array inputs in CodeExecutorInput, we were wrapping the list itself but not properly unwrapping the inner VellumValue objects.

### DIFF
--- a/src/vellum/workflows/nodes/displayable/code_execution_node/tests/test_code_execution_node.py
+++ b/src/vellum/workflows/nodes/displayable/code_execution_node/tests/test_code_execution_node.py
@@ -2,7 +2,7 @@ import pytest
 import os
 from typing import Any
 
-from vellum import CodeExecutorResponse, NumberVellumValue, StringInput
+from vellum import CodeExecutorResponse, NumberVellumValue, StringInput, StringVellumValue
 from vellum.client.types.code_execution_package import CodeExecutionPackage
 from vellum.client.types.code_executor_secret_input import CodeExecutorSecretInput
 from vellum.client.types.function_call import FunctionCall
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
```diff
@@ -74,8 +75,25 @@ def run_code_inline(
...
+    VELLUM_TYPES = get_args(VellumValue)
+
+    def wrap_value(value):
+        if isinstance(value, list):
+            return ListWrapper(
+                [
+                    # Convert VellumValue to dict with its fields
+                    (
+                        item.model_dump()
+                        if isinstance(item, VELLUM_TYPES)
```
dvargasfuertes, line 87: [nit] I _think_ we could avoid the nested method by just adding `elif isinstance(item, VELLUM_TYPES)` as a case within `_clean_for_dict_wrapper`

src/vellum/workflows/nodes/displayable/code_execution_node/tests/test_code_execution_node.py
```diff
@@ -493,3 +493,36 @@ def main(word: str) -> dict:
...
+
+    # GIVEN a node that subclasses CodeExecutionNode that processes an array of VellumValues
+    class State(BaseState):
+        pass
+
+    class ExampleCodeExecutionNode(CodeExecutionNode[State, str]):
+        code = """\
+from typing import List, Dict
+def main(arg1: List[Dict]) -> str:
+    return arg1[0]["value"] + " " + arg1[1]["value"]
```
dvargasfuertes, line 509: [q] damn - is this a syntax that we saw with an existing code exec node?

dvargas92495 (approved):

src/vellum/workflows/nodes/displayable/code_execution_node/tests/test_code_execution_node.py
NgoHarrison: Yeah something very similar 😬

PR #949, opened 2025-02-11
Add check for union types for inline code execution nodes

NgoHarrison:
This came up from this PR https://github.com/vellum-ai/vellum-python-sdks/pull/948

I was testing around and instead of strings, I tried using numbers. I kept getting errors on code exec runs with arg1 + arg2 or if I added two values from a list of dicts like `input[index]['some-key']`. The main issue was that since we now make the numbers a union type, we need to account for it


dvargas92495 (approved):
gooood catch - let's get this in for 0.14.0 too

PR #955, opened 2025-02-12
Add defaultRequired to handle optional types under different context

vincent0426:
Add config for `defaultRequired`

### DIFF
--- a/ee/codegen/src/generators/vellum-variable.ts
+++ b/ee/codegen/src/generators/vellum-variable.ts
@@ -18,23 +18,37 @@ type VellumVariableWithName = (
     default?: VellumValueType | null;
   };
 
+// VellumVariable.defaultRequired:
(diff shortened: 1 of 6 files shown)


ee/codegen/src/generators/vellum-variable.ts
```diff
@@ -18,23 +18,37 @@ type VellumVariableWithName = (
...
   };
 
+// VellumVariable.defaultRequired:
+// Prompt Inputs: required: undefined == required: true,
+// Workflow Inputs: required: undefined == required: false
+// Workflow Outputs: required: undefined == required: true
 export declare namespace VellumVariable {
   interface Args {
     variable: VellumVariableWithName;
+    defaultRequired?: boolean;
```
vincent0426, line 28: should we assign default value?

dvargas92495 (approved):

PR #956, opened 2025-02-12
Fix jinja template to parse string like literal

vincent0426:
fix ` template = """{{- prompt_outputs | selectattr(\'type\', \'equalto\', \'FUNCTION_CALL\') | list | replace(\"\\n\",\",\") -}}"""`

### DIFF
--- a/src/vellum/utils/templating/render.py
+++ b/src/vellum/utils/templating/render.py
@@ -1,3 +1,4 @@
+import ast
 import json
 from typing import Any, Callable, Dict, Optional, Union
 
@@ -8,8 +9,16 @@ from vellum.workflows.state.encoder import DefaultStateEncoder
 
 
 def finalize(obj: Any) -> str:
-    if isinstance(obj, dict):
+    if isinstance(obj, (dict, list)):
         return json.dumps(obj, cls=DefaultStateEncoder)
+
+    if isinstance(obj, str):
+        try:
+            data = ast.literal_eval(obj)
+            if isinstance(data, (dict, list)):
+                return json.dumps(data, cls=DefaultStateEncoder)
+        except (ValueError, SyntaxError):
+            pass
 
     return str(obj)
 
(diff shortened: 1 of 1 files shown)


src/vellum/utils/templating/render.py
```diff
@@ -8,9 +9,17 @@
 
 
 def finalize(obj: Any) -> str:
-    if isinstance(obj, dict):
+    if isinstance(obj, (dict, list)):
         return json.dumps(obj, cls=DefaultStateEncoder)
 
+    if isinstance(obj, str):
+        try:
+            data = ast.literal_eval(obj)
```
dvargasfuertes, line 17: [blocking] this looks dangerous - why do we need to literally eval here? if we already have a str, we should just return

src/vellum/utils/templating/render.py
```diff
@@ -8,9 +9,17 @@
 
 
 def finalize(obj: Any) -> str:
-    if isinstance(obj, dict):
+    if isinstance(obj, (dict, list)):
```
dvargasfuertes, line 12: [blocking] Let's add a test to `templating_node/node.py` reproducing the template error you wrote in the PR description

My hunch is that this is the only impl change we'll actually need to get it to pass

src/vellum/utils/templating/render.py
```diff
@@ -22,6 +31,12 @@ def render_sandboxed_jinja_template(
     jinja_globals: Optional[Dict[str, Any]] = None,
 ) -> str:
     """Render a Jinja template within a sandboxed environment."""
+    prepared_input_values = {}
+    for key, value in input_values.items():
+        if isinstance(value, list):
+            prepared_input_values[key] = json.loads(json.dumps(value, cls=DefaultStateEncoder))
```
dvargasfuertes, line 37: [blocking] I would expect us to pass in the values as unaltered as possible, and perform any string formatting we need on the way out

PR #959, opened 2025-02-12
Create simple map node test

dvargas92495:
I was noticing a few issues with map nodes until realizing that we don't have map node tests. I assume bc it was a bit tricky to get one online, and I plan to followup with some cleanup to make it a bit easier

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/map-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/map-node.test.ts.snap
@@ -1,0 +1,50 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`MapNode > basic > map node display file 1`] = `
+"# flake8: noqa: F401, F403
+
+from uuid import UUID
+
+from vellum_ee.workflows.display.nodes import BaseMapNodeDisplay
+from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplayOverrides
+from vellum_ee.workflows.display.vellum import NodeDisplayData, NodeDisplayPosition
+
+from ....nodes.map_node import MapNode
+from .nodes import *
+from .workflow import *
+
+
+class MapNodeDisplay(BaseMapNodeDisplay[MapNode]):
+    label = "Map Node"
+    node_id = UUID("14fee4a0-ad25-402f-b942-104d3a5a0824")
+    target_handle_id = UUID("3fe4b4a6-5ed2-4307-ac1c-02389337c4f2")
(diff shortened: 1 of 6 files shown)


marlenc3 (approved):
🔥

ee/codegen/src/__test__/nodes/__snapshots__/map-node.test.ts.snap
```diff
@@ -0,0 +1,50 @@
...
+class MapNodeDisplay(BaseMapNodeDisplay[MapNode]):
+    label = "Map Node"
+    node_id = UUID("14fee4a0-ad25-402f-b942-104d3a5a0824")
+    target_handle_id = UUID("3fe4b4a6-5ed2-4307-ac1c-02389337c4f2")
+    node_input_ids_by_name = {"items": UUID("f34872c2-5c0e-45a3-b204-3af22d1028d3")}
+    output_display = {
+        MapNode.Outputs.final_output: NodeOutputDisplay(
+            id=UUID("edd5cfd5-6ad8-437d-8775-4b9aeb62a5fb"), name="final-output"
+        )
+    }
```
dvargasfuertes, line 26: This is bug one that I want to follow up with. We are hallucinating `final_output` here

ee/codegen/src/__test__/nodes/__snapshots__/map-node.test.ts.snap
```diff
@@ -0,0 +1,50 @@
...
+    items = [
+        "apple",
+        "banana",
+        "cherry",
+    ]
+    subworkflow = MapNodeWorkflow
+    max_concurrency = 4
+
+    class Outputs(BaseMapNode.Outputs):
+        pass
```
dvargasfuertes, line 48: This is a second followup. We no longer need to generate these outputs classes

ee/codegen/src/generators/nodes/bases/nested-workflow-base.ts
```diff
@@ -23,6 +23,10 @@ export abstract class BaseNestedWorkflowNode<
   protected abstract getInnerWorkflowData(): WorkflowRawData;
 
   constructor(args: BaseNode.Args<T, V>) {
+    console.log(
```
dvargasfuertes: damn, will remove

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -193,6 +193,7 @@ export class WorkflowContext {
       workflowRawEdges,
       codeExecutionNodeCodeRepresentationOverride:
         this.codeExecutionNodeCodeRepresentationOverride,
+      strict: this.strict,
```
dvargasfuertes, line 196: needed to prop drill this down

NgoHarrison (commented):
✅

PR #962, opened 2025-02-12
Setup test for map node displays

dvargas92495:
- [x] Found some issues in subworkflow output contexts that I want to clean up first

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/map-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/map-node.test.ts.snap
@@ -48,3 +48,55 @@ class MapNode(BaseMapNode):
         pass
 "
 `;
+
+exports[`MapNode > with additional output > map node display file 1`] = `
+"# flake8: noqa: F401, F403
+
+from uuid import UUID
+
+from vellum_ee.workflows.display.nodes import BaseMapNodeDisplay
+from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplayOverrides
+from vellum_ee.workflows.display.vellum import NodeDisplayData, NodeDisplayPosition
+
+from ....nodes.map_node import MapNode
+from .nodes import *
+from .workflow import *
+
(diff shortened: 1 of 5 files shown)


ee/codegen/src/context/workflow-output-context.ts
```diff
@@ -32,25 +31,6 @@ export class WorkflowOutputContext {
     this.name = this.getOutputVariableName();
   }
 
-  public getOutputNodeId(): string {
```
dvargasfuertes, line 35: no longer used

ee/codegen/src/__test__/nodes/__snapshots__/map-node.test.ts.snap
```diff
@@ -35,6 +35,58 @@ exports[`MapNode > basic > map node file 1`] = `
...
+from .workflow import *
+
+
+class MapNodeDisplay(BaseMapNodeDisplay[MapNode]):
+    label = "Map Node"
+    node_id = UUID("14fee4a0-ad25-402f-b942-104d3a5a0824")
+    target_handle_id = UUID("3fe4b4a6-5ed2-4307-ac1c-02389337c4f2")
+    node_input_ids_by_name = {"items": UUID("f34872c2-5c0e-45a3-b204-3af22d1028d3")}
+    output_display = {
+        MapNode.Outputs.first_output: NodeOutputDisplay(
```
dvargasfuertes, line 72: we were previously always generating `final_output` 😬

ee/codegen/src/generators/nodes/map-node.ts
```diff
@@ -158,43 +154,47 @@ export class MapNode extends BaseNestedWorkflowNode<
   }
 
   protected getOutputDisplay(): python.Field {
```
dvargasfuertes, line 156: this method is now exactly the same as `InlineSubworkflowNode.getOutputDisplay()`. May be worth pulling up in a fast follow

marlenc3:
✅

PR #965, opened 2025-02-12
[SC-6621] Add custom replace filter to Jinja templating and corresponding tests

vincent0426:
this could solve the replace problem, however, while running `re_act_agent` https://app.vellum.ai/workflow-sandboxes/795d7568-bac5-4430-bdf0-e01c97b17110 

there is another err that seems to be related to the map item, since changing `replace` to `tojson` will produce the same err
`Error: WorkflowError(message='list assignment index out of range', code=<WorkflowErrorCode.INTERNAL_ERROR: 'INTERNAL_ERROR'>)`

### DIFF
--- a/src/vellum/utils/templating/render.py
+++ b/src/vellum/utils/templating/render.py
@@ -50,10 +50,9 @@ def render_sandboxed_jinja_template(
             "cls": DefaultStateEncoder,
         }
 
+        environment.filters["replace"] = custom_replace
         if jinja_custom_filters:
             environment.filters.update(jinja_custom_filters)
-
-        environment.filters["replace"] = custom_replace
 
         jinja_template = environment.from_string(template)
 
--- a/src/vellum/utils/templating/render.py
+++ b/src/vellum/utils/templating/render.py
@@ -1,8 +1,9 @@
 import json
-from typing import Any, Callable, Dict, Optional, Union
+from typing import Any, Dict, Optional
 
 from jinja2.sandbox import SandboxedEnvironment
 
+from vellum.utils.templating.constants import FilterFunc
 from vellum.utils.templating.exceptions import JinjaTemplateError
 from vellum.workflows.state.encoder import DefaultStateEncoder
 
@@ -18,7 +19,7 @@ def render_sandboxed_jinja_template(
     *,
     template: str,
     input_values: Dict[str, Any],
-    jinja_custom_filters: Optional[Dict[str, Callable[[Union[str, bytes]], bool]]] = None,
+    jinja_custom_filters: Optional[Dict[str, FilterFunc]] = None,
     jinja_globals: Optional[Dict[str, Any]] = None,
 ) -> str:
     """Render a Jinja template within a sandboxed environment."""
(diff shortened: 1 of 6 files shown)


src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
```diff
@@ -193,3 +194,29 @@ class UnionTemplateNode(TemplatingNode[BaseState, Union[float, int]]):
 
     # THEN it should correctly parse as a float
     assert outputs.result == 42.5
+
+
+def test_templating_node__replace_filter():
+    # GIVEN a templating node that outputs a complex object
+    class ReplaceFilterTemplateNode(TemplatingNode[BaseState, Json]):
+        template = """{{- prompt_outputs | selectattr(\'type\', \'equalto\', \'FUNCTION_CALL\') \
```
vincent0426, line 202: splitting to avoid flake

vincent0426:
[blocking] `Error: WorkflowError(message='list assignment index out of range', code=<WorkflowErrorCode.INTERNAL_ERROR: 'INTERNAL_ERROR'>)`

src/vellum/utils/templating/render.py
```diff
@@ -14,6 +14,27 @@ def finalize(obj: Any) -> str:
     return str(obj)
 
 
+def custom_replace(s, old, new):
+    """
+    Custom replace filter that uses DefaultStateEncoder for conversion
+    """
+    import json
+
+    from vellum.workflows.state.encoder import DefaultStateEncoder
```
dvargasfuertes: [nit] Why are we importing here instead of using the ones above?

src/vellum/utils/templating/render.py
```diff
@@ -14,6 +14,27 @@ def finalize(obj: Any) -> str:
...
+
+    def encode(obj):
+        try:
+            return json.dumps(obj, cls=DefaultStateEncoder)
```
dvargasfuertes: [blocking] I don't think we _always_ want to do the json dumps - a string will go from `"hello"` to `'"hello"'`. We likely want to have the same `if not isinstance()` check

src/vellum/utils/templating/render.py
```diff
@@ -35,6 +56,8 @@ def render_sandboxed_jinja_template(
         if jinja_custom_filters:
             environment.filters.update(jinja_custom_filters)
 
+        environment.filters["replace"] = custom_replace
```
dvargasfuertes: [blocking] let's move this above `jinja_custom_filters` to let the user provide their own if so inclined

src/vellum/utils/templating/render.py
```diff
@@ -14,6 +14,25 @@ def finalize(obj: Any) -> str:
     return str(obj)
 
 
+def custom_replace(s, old, new):
+    """
+    Custom replace filter that uses DefaultStateEncoder for conversion
+    """
+
+    def encode(obj):
```
dvargasfuertes: [blocking] I think `str` returns `None` instead of itself

I'm surprised that mypy didn't ask for types here. I'd be a fan of this method being pulled out and adding type hints to both

vincent0426:
@dvargas92495 I found that there's a custom filter folder so I refactor to stay consistent with it, also add some test

dvargas92495 (approved):
💥

PR #969, opened 2025-02-13
Fix the output display for inline subworkflow nodes to drive off of output variable context's

dvargas92495:
Continuing work started here: https://github.com/vellum-ai/vellum-python-sdks/pull/968

I noticed that we were using the wrong output variable id's when defining subworkflow node output displays (for both subworkflow and map node outputs). So I started with the expected snapshot and worked backwards. In doing so that we have redundancy between `output-variable-context` and `workflow-output-context`s, so this cleanup will eventually remove the latter

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/inline-subworkflow-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/inline-subworkflow-node.test.ts.snap
@@ -6,7 +6,7 @@ exports[`InlineSubworkflowNode > basic > inline subworkflow node display file 1`
 from uuid import UUID
 
 from vellum_ee.workflows.display.nodes import BaseInlineSubworkflowNodeDisplay
-from vellum_ee.workflows.display.nodes.types import PortDisplayOverrides
+from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplayOverrides
 from vellum_ee.workflows.display.vellum import NodeDisplayData, NodeDisplayPosition
 
 from ....nodes.inline_subworkflow_node import InlineSubworkflowNode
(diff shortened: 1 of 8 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/inline-subworkflow-node.test.ts.snap
```diff
@@ -20,7 +20,11 @@ class InlineSubworkflowNodeDisplay(BaseInlineSubworkflowNodeDisplay[InlineSubwor
     target_handle_id = UUID("3fe4b4a6-5ed2-4307-ac1c-02389337c4f2")
     workflow_input_ids_by_name = {}
     node_input_ids_by_name = {}
-    output_display = {}
+    output_display = {
+        InlineSubworkflowNode.Outputs.final_output: NodeOutputDisplay(
+            id=UUID("edd5cfd5-6ad8-437d-8775-4b9aeb62a5fb"), name="final-output"
+        )
+    }
```
dvargasfuertes, line 27: This was the main goal of the PR:
- `id` matches that of the output variable
- `name` matches the original name in the UI
- `output_display[str]` matches the attribute name generated

ee/codegen/src/context/output-variable-context.ts
```diff
@@ -26,4 +36,26 @@ export class OutputVariableContext {
   public getRawName(): string {
     return this.outputVariableData.key;
   }
+
+  private generateSanitizedOutputVariableName(): string {
```
dvargasfuertes, line 40: Blindish copypasta of workflow-output-context. This is what clued me into the fact that we may remove that class once this clean up is done

ee/codegen/src/context/workflow-output-context.ts
```diff
@@ -77,28 +77,15 @@ export class WorkflowOutputContext {
...
-
-    // Deduplicate the output variable name if it's already in use
-    let sanitizedName = initialOutputVariableName;
-    let numRenameAttempts = 0;
-    while (this.workflowContext.isOutputVariableNameUsed(sanitizedName)) {
-      sanitizedName = `${initialOutputVariableName}${
-        initialOutputVariableName.endsWith("_") ? "" : "_"
-      }${numRenameAttempts + 1}`;
-      numRenameAttempts += 1;
-    }
```
dvargasfuertes, line 100: This is a temporary measure until we're fully cutover to just output-variable-context instead of `workflow-output-context` in the next PR

PR #972, opened 2025-02-13
Skip generating the outputs class for map node

dvargas92495:
Map node outputs are dynamic, so no need to try to create an outputs class


marlenc3:
✅

PR #973, opened 2025-02-13
Speed up codegen tests by passing disable formatting to workflow context

dvargas92495:
Before:
Screenshot 2025-02-12 at 11 56 48 PM

After:
Screenshot 2025-02-12 at 11 55 40 PM

The main source of the speedup is not running isort. We could get this down further if we stopped running `isort` in the batch of `project.test.ts` tests, but these affect our python files which we want to keep sorted. Am bullish on eventually removing this from `project.ts` and running this sort of python proc in django as @m-abboud has advised previously, and also just doing an initial isort pass in typescript which doesn't seem too bad


marlenc3:
✅

PR #975, opened 2025-02-13
Fix flaky expensive emitter test

dvargas92495:
This test failed three times on another pr: https://github.com/vellum-ai/vellum-python-sdks/actions/runs/13313436086/job/37181988700

Bumping up the sleep time but making sure it only does it once so as to not slow down tests


NgoHarrison (commented):
Nice ✅

PR #976, opened 2025-02-13
Refactor to reuse valid uuid function

vincent0426:
should we reuse function from utils which is same?


dvargas92495 (approved):
yup I've been meaning to refactor it 👍

PR #979, opened 2025-02-13
Don't Hard Fail on Missing Subworkflow Deployment Output

noanflaherty:
Here we warn instead of error when a downstream Node is pointing to an output on a Subworkflow Deployment Node that no longer exists.


NgoHarrison (approved):

noanflaherty:
I'm gonna close this one in favor of https://github.com/vellum-ai/vellum-python-sdks/pull/980

PR #980, opened 2025-02-13
Proposal: New Way of Handling Codegen Warnings

noanflaherty:
I find it helpful to collect warnings much like we do hard errors. This way, we can introduce "ultra strict modes" later, or just generally track issues in codegen of various severity levels.

This PR proposes a pattern for doing so that's inspired by log levels.

### DIFF
--- a/ee/codegen/src/context/node-context/base.ts
+++ b/ee/codegen/src/context/node-context/base.ts
@@ -106,7 +106,8 @@ export abstract class BaseNodeContext<T extends WorkflowDataNode> {
       ) {
         this.workflowContext.addError(
           new NodeOutputNotFoundError(
-            `Could not find subworkflow deployment output with id ${outputId}`
+            `Could not find subworkflow deployment output with id ${outputId}`,
+            "WARNING"
           )
         );
         return;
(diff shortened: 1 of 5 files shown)


ee/codegen/src/context/node-context/base.ts
```diff
@@ -106,7 +106,8 @@ export abstract class BaseNodeContext<T extends WorkflowDataNode> {
       ) {
         this.workflowContext.addError(
           new NodeOutputNotFoundError(
-            `Could not find subworkflow deployment output with id ${outputId}`
+            `Could not find subworkflow deployment output with id ${outputId}`,
+            "WARNING"
```
noanflaherty, line 110: This shows an example usage of setting a lower severity level.

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -444,9 +445,12 @@ export class WorkflowContext {
   }
 
   public addError(error: BaseCodegenError): void {
-    if (this.strict) {
+    if (this.strict && error.severity === "ERROR") {
       throw error;
     }
+    if (error.severity === "WARNING") {
+      error.log();
+    }
```
noanflaherty, line 453: Here we still log warnings as we encounter them

NgoHarrison (approved):
Been meaning to do this change eventually, love it!

dvargas92495 (commented):
I dig it!

✅

PR #983, opened 2025-02-14
Fix templating node output as a list of dataclasses

dvargas92495:
This was failing one of the customer workflows that we were qaing


vincent0426:
🥇

marlenc3 (approved):

NgoHarrison (commented):
✅

PR #987, opened 2025-02-14
Codegen Nodes Not in Graph

noanflaherty:
It's a valid case to have Nodes in a Workflow and for them to not be referenced in the graph, but for you to still want to codegen-them – for example, Note Nodes, or WIP nodes.

This PR:
1. Codegens nodes even if they're not in the graph
2. Adds test coverage
3. Cleans up some of our existing tests to share some expect helpers

Next Step: Add serialization support for nodes not in graph

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -73,7 +73,29 @@ baz = foo + bar
 "
 `;
 
-exports[`WorkflowProjectGenerator > inlude sandbox > should include a sandbox.py file when passed sandboxInputs 1`] = `
+exports[`WorkflowProjectGenerator > failure cases > should generate code even if a node fails to find invalid ports and target nodes 1`] = `
+"from vellum.workflows.nodes.displayable import TemplatingNode
+from vellum.workflows.state import BaseState
+
+
+class BadNode(TemplatingNode[BaseState, str]):
+    template = """foo"""
+    inputs = {}
+"
+`;
+
+exports[`WorkflowProjectGenerator > failure cases > should generate code even if a node fails to generate 1`] = `
+"from vellum.workflows.nodes.displayable import TemplatingNode
+from vellum.workflows.state import BaseState
+
+
+class BadNode(TemplatingNode[BaseState, str]):
+    template = """foo"""
+    inputs = {}
+"
+`;
+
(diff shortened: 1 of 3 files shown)


ee/codegen/src/__test__/project.test.ts
```diff
@@ -21,6 +21,42 @@ import { WorkflowProjectGenerator } from "src/project";
 describe("WorkflowProjectGenerator", () => {
   let tempDir: string;
 
+  function expectProjectFileToMatchSnapshot(
```
noanflaherty, line 24: Here are three new helpers to perform asserts and de-duplicate some code across our tests. I opted for parity with what was there.

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1,5 +1,56 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`WorkflowProjectGenerator > Nodes present but not in graph > should still generate a file for the second node 1`] = `
```
noanflaherty, line 3: In the cases where I'm now using snapshot testing rather than string-comparisons, I've confirmed that the new snapshots match the old strings.

ee/codegen/src/project.ts
```diff
@@ -522,6 +522,23 @@ ${errors.slice(0, 3).map((err) => {
...
 
+    // Include at the end nodes that are included in the workflow, but not referenced in the graph
+    // For example, Note Nodes.
+    const unprocessedNodes: WorkflowDataNode[] = rawData.nodes
+      .map<WorkflowDataNode | undefined>((node: WorkflowNode) => {
+        if (
+          processedNodeIds.has(node.id) ||
+          node.type === "ENTRYPOINT" ||
+          node.type === "TERMINAL"
+        ) {
```
dvargasfuertes: [nit] why not just filter on this condition haha

dvargasfuertes: oh prob bc of typecast. here's a what I usually do:
```typescript
.filter((node): node is WorkflowDataNode => ...)
```

I think there's a recent version of TypeScript that starts to do this automatically for array.filter

dvargas92495 (approved):

ee/codegen/src/project.ts
noanflaherty: Yeah it was for the type casting. I’ll try the suggested approach

PR #989, opened 2025-02-14
[SC-6686] Handle escape characters in input variable name

NgoHarrison:
Context: A customer has escape characters in the input variable names. This PR aims to codegen correct python code for that and allow that workflow to run

### DIFF
--- a/ee/codegen/src/context/input-variable-context.ts
+++ b/ee/codegen/src/context/input-variable-context.ts
@@ -2,7 +2,10 @@ import { isEmpty, isNil } from "lodash";
 import { VellumVariable } from "vellum-ai/api/types";
 
 import { WorkflowContext } from "src/context/workflow-context";
-import {removeEscapeCharacters, toPythonSafeSnakeCase} from "src/utils/casing";
+import {
+  removeEscapeCharacters,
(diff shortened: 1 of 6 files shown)


ee/codegen/src/context/input-variable-context.ts
```diff
@@ -39,7 +42,8 @@ export class InputVariableContext {
   }
 
   public getRawName(): string {
-    return this.inputVariableData.key;
+    // This is for an edge case where there are escape characters in the variable
+    return removeEscapeCharacters(this.inputVariableData.key);
```
dvargasfuertes, line 46: [q] Does this means on push we would lose the escape characters? that's likely fine, just want to make sure it was intentional

dvargas92495 (approved):

ee/codegen/src/context/input-variable-context.ts
NgoHarrison: Yeah I was doing this change with the assumption that users are escaping strings because to form valid json but they really wanted to just convey 'Bad "Input"' for example. Also this is still a weird edge case imo since a user is naming a variable like this and this change shouldn't affect other workflows or cause regressions *knocks on wood*

PR #992, opened 2025-02-15
Add unused to support non-graph nodes or graph in workflow

vincent0426:
Current workflow does not support unused nodes. However, if a user defines nodes that are not part of the graph, they will be lost during serialization.

- add unused attr to workflow
- tests

### DIFF
--- a/src/vellum/workflows/workflows/base.py
+++ b/src/vellum/workflows/workflows/base.py
@@ -117,12 +117,12 @@ class _BaseWorkflowMeta(type):
             return nodes
 
         graph_nodes = collect_nodes(dct.get("graph", set()))
-        unused_nodes = collect_nodes(dct.get("unused_graph", set()))
+        unused_nodes = collect_nodes(dct.get("unused_graphs", set()))
 
         overlap = graph_nodes & unused_nodes
         if overlap:
             node_names = [node.__name__ for node in overlap]
-            raise ValueError(f"Node(s) {', '.join(node_names)} cannot appear in both graph and unused_graph")
+            raise ValueError(f"Node(s) {', '.join(node_names)} cannot appear in both graph and unused_graphs")
 
         cls = super().__new__(mcs, name, bases, dct)
         workflow_class = cast(Type["BaseWorkflow"], cls)
@@ -136,7 +136,7 @@ GraphAttribute = Union[Type[BaseNode], Graph, Set[Type[BaseNode]], Set[Graph]]
 class BaseWorkflow(Generic[InputsType, StateType], metaclass=_BaseWorkflowMeta):
     __id__: UUID = uuid4_from_hash(__qualname__)
     graph: ClassVar[GraphAttribute]
-    unused_graph: ClassVar[Set[GraphAttribute]]  # nodes or graphs that are defined but not used in the graph
+    unused_graphs: ClassVar[Set[GraphAttribute]]  # nodes or graphs that are defined but not used in the graph
     emitters: List[BaseWorkflowEmitter]
     resolvers: List[BaseWorkflowResolver]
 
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/workflows/base.py
```diff
@@ -112,6 +112,7 @@ def __new__(mcs, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any
 class BaseWorkflow(Generic[InputsType, StateType], metaclass=_BaseWorkflowMeta):
     __id__: UUID = uuid4_from_hash(__qualname__)
     graph: ClassVar[GraphAttribute]
+    unused: ClassVar[List[GraphAttribute]]  # nodes or graphs that are defined but not used in the graph
```
vincent0426: Not sure if `unused` is clear

dvargasfuertes: [blocking]
- Let's make it a `Set[GraphAttribute]` so that syntax is consistent with the rest of GraphAttribute
- let's call it `unused_graphs`

src/vellum/workflows/workflows/base.py
```diff
@@ -196,6 +197,40 @@ def get_nodes(cls) -> Iterator[Type[BaseNode]]:
                     nodes.add(node)
                     yield node
 
+    @classmethod
+    def get_nodes_not_in_graph(cls) -> Iterator[Type[BaseNode]]:
```
dvargasfuertes: [nit] `get_unused_nodes(cls)` for consistency with the attribute

src/vellum/workflows/workflows/base.py
```diff
@@ -196,6 +197,40 @@ def get_nodes(cls) -> Iterator[Type[BaseNode]]:
...
+                        elif issubclass(subitem, BaseNode):
+                            if subitem not in nodes:
+                                nodes.add(subitem)
+                                yield subitem
+                elif issubclass(item, BaseNode):
+                    # Item is a node
+                    if item not in nodes:
+                        nodes.add(item)
+                        yield item
+
```
dvargasfuertes, line 257: [oos] Let's also add a `get_unused_edges(cls):` since serialization will need both. Can be deferred to the next PR

src/vellum/workflows/workflows/tests/test_base_workflow.py
```diff
@@ -78,3 +78,62 @@ class Outputs:
...
+    class NodeB(BaseNode):
+        pass
+
+    # WHEN we create a workflow with no unused nodes
+    class TestWorkflow(BaseWorkflow[BaseInputs, BaseState]):
+        graph = NodeA >> NodeB
+
+    # TEST that nodes_not_in_graph is empty
+    nodes = set(TestWorkflow.get_nodes_not_in_graph())
+    assert nodes == set()
```
dvargasfuertes, line 147: [oos] these tests are great! Let's add one more where the user puts the node in both `graph` and `unused`. It will force us to come to a stance on how we should handle that situation

dvargas92495 (approved):
love it!

src/vellum/workflows/workflows/base.py
```diff
@@ -112,7 +138,7 @@ def __new__(mcs, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any
 class BaseWorkflow(Generic[InputsType, StateType], metaclass=_BaseWorkflowMeta):
     __id__: UUID = uuid4_from_hash(__qualname__)
     graph: ClassVar[GraphAttribute]
-    unused: ClassVar[List[GraphAttribute]]  # nodes or graphs that are defined but not used in the graph
+    unused_graph: ClassVar[Set[GraphAttribute]]  # nodes or graphs that are defined but not used in the graph
```
dvargasfuertes: [nit] let's make this plural! could be several disconnected graphs

dvargas92495 (approved):

PR #993, opened 2025-02-15
Increase sleep time to avoid flaky cancel test

vincent0426:
Bump up sleep duration to prevent flaky tests for `test_workflow__cancel_run` and `test_workflow__cancel_stream`


dvargas92495 (approved):
yea I've noticed this too, nice 🔥

PR #995, opened 2025-02-16
Replace isort with simple sort logic

dvargas92495:
Here's the exec(isort) bug that finally tipped me over the edge here:
- Part 1 https://www.loom.com/share/6576163093784bbb8c8a6eb4f25f4e45
- Part 2 https://www.loom.com/share/eae686ab09c44012b812932feb152594

Reasons why I'm not worried about risk:
- We don't need to account for user config in codgen service - that could happen on their machine as part of `vellum workflows pull` or in django
- Because of ^, we don't need perfect sorting within codegen service, just something deterministic. The deterministic algo I chose was one that minimized changes from the snapshot generated from `project.test.ts`

Other Benefits:
- Tests run 50% faster now
- No divergence from preview and run
- No flaky exec(isort) failures referenced in video
- All of our node unit tests snapshots will now have consistent sorting with our project tests and are no longer sensitive to order of references within implementation.
    - This is where the vast majority of the LoC lives. The actual sorting algo is ~40 lines

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -20,8 +20,9 @@ class TestWorkflow(BaseWorkflow):
 
 exports[`Workflow > write > should generate correct code with Search Results as an output variable 1`] = `
 "from vellum.workflows import BaseWorkflow
-from .inputs import Inputs
 from vellum.workflows.state import BaseState
+
+from .inputs import Inputs
 from .nodes.final_output import FinalOutput
 
 
@@ -33,8 +34,9 @@ class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
 
 exports[`Workflow > write > should generate correct code with an output variable mapped to an input variable 1`] = `
 "from vellum.workflows import BaseWorkflow
-from .inputs import Inputs
 from vellum.workflows.state import BaseState
+
+from .inputs import Inputs
 
 
 class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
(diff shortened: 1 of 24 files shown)


ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -9,17 +9,6 @@ class TestWorkflow(BaseWorkflow):
 "
 `;
 
-exports[`Workflow > write > should generate correct code when there are input variables with escape characters 1`] = `
```
dvargasfuertes: looks unrelated and an issue on main - will merge through separately to keep this PR focused

ee/codegen/src/generators/base-persisted-file.ts
```diff
@@ -17,6 +20,128 @@ export declare namespace BasePersistedFile {
   }
 }
 
+const VELLUM_MODULES = new Set(["vellum", "vellum_ee"]);
+
+class ImportSortedPythonFile extends PythonFile {
+  writeImports({
```
dvargasfuertes, line 26: This method is mostly copy/pasted, exceptions are highlighted

ee/codegen/src/generators/base-persisted-file.ts
```diff
@@ -17,6 +20,128 @@ export declare namespace BasePersistedFile {
...
+    uniqueReferences,
+  }: {
+    writer: Writer;
+    uniqueReferences: Map<
+      string,
+      { modulePath: ModulePath; references: Reference[] }
+    >;
+  }) {
+    const sortedReferences = this.sortReferences(uniqueReferences);
+    let lastPriority: number | null = null;
```
dvargasfuertes, line 37: This is where we invoke our sort

ee/codegen/src/generators/base-persisted-file.ts
```diff
@@ -17,6 +20,128 @@ export declare namespace BasePersistedFile {
...
+      if (this.isDefinedInFile(refModulePath)) {
+        continue;
+      }
+
+      if (priority !== lastPriority) {
+        if (lastPriority !== null) {
+          writer.newLine();
+        }
+        lastPriority = priority;
+      }
```
dvargasfuertes, line 56: This handles newlines between groups

ee/codegen/src/generators/base-persisted-file.ts
```diff
@@ -17,6 +20,128 @@ export declare namespace BasePersistedFile {
...
+      writer.newLine();
+    }
+    if (Object.keys(uniqueReferences).length > 0) {
+      writer.newLine();
+    }
+  }
+
+  // The above logic is copied and pasted from the Fern PythonFile AST Library
+  // The below is our only contribution to the class
```
dvargasfuertes, line 100: Once/If we fix Python File on the fern side, the sort logic we need is just 40 lines, not the entirety of isort. That's bc the vast majority of isort is:
- custom config - which we'll handle in userland not on our end
- ast building and writing - which codegen is already doing

ee/codegen/src/generators/protected-python-file.ts
```diff
@@ -0,0 +1,309 @@
+// This file is a direct copy paste of
+// https://github.com/fern-api/fern/blob/main/generators/python-v2/ast/src/PythonFile.ts
+// But the only difference is that all private methods are now protected
```
dvargasfuertes, line 3: Would skip reviewing this file

dvargasfuertes: If we go forward with this, we should update the upstream fern package, but this allows us to be unblocked for the time being

noanflaherty (approved):
Wow, wild. Good find

siddseethepalli (commented):
Holy shit. Yeah this is great, I'll eat my words from Friday.

PR #997, opened 2025-02-16
Fix import paths for base node

dvargas92495:
Pulling workflows with base node will fail locally bc we had the wrong import path specified.  This PR fixes


vincent0426 (commented):
✅

PR #1000, opened 2025-02-17
Supplement undefined class to show better error messages in userland code

dvargas92495:
Users can interface with our `undefined` type within base nodes and code nodes. We want error messages that involve them to clearly indicate an undefined type


vincent0426:
✅

PR #1003, opened 2025-02-17
Triage all codegen warning/errors

dvargas92495:
Part 2 of the following 4 step process:

- Run workflows in non-vembda mode (started [here](https://github.com/vellum-ai/vellum-python-sdks/compare/vargas/triage-base-codegen-errors?expand=1))
- Audit existing BaseCodegenErrors and turn most that would not affect runtime into warnings instead
- Change strict to still error on both error/warnings
- non-strict codegen always errors on errors

### DIFF
--- a/ee/codegen/src/context/workflow-context/workflow-context.ts
+++ b/ee/codegen/src/context/workflow-context/workflow-context.ts
@@ -228,12 +228,19 @@ export class WorkflowContext {
       (n): n is EntrypointNode => n.type === "ENTRYPOINT"
     );
     if (entrypointNodes.length > 1) {
-      throw new WorkflowGenerationError("Multiple entrypoint nodes found");
+      this.addError(
+        new WorkflowGenerationError(
+          "Multiple entrypoint nodes found",
+          "WARNING"
+        )
+      );
     }
 
     const entrypointNode = entrypointNodes[0];
     if (!entrypointNode) {
-      throw new WorkflowGenerationError("Entrypoint node not found");
+      this.addError(
+        new WorkflowGenerationError("Entrypoint node not found", "WARNING")
+      );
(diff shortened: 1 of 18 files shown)


ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -309,8 +317,9 @@ export class WorkflowContext {
       this.findOutputVariableContextById(outputVariableId);
 
     if (!outputVariableContext) {
-      throw new WorkflowInputGenerationError(
-        `Output variable context not found for ID: ${outputVariableId}`
+      throw new WorkflowOutputGenerationError(
+        `Output variable context not found for ID: ${outputVariableId}`,
+        "WARNING"
       );
```
dvargasfuertes, line 347: most of the time we want to `addError`. We throw when we actually can't keep going. In this case, the return type assumes non-nullable

ee/codegen/src/generators/workflow-sandbox-file.ts
```diff
@@ -92,6 +93,11 @@ if __name__ != "__main__":
             this.workflowContext.findInputVariableContextByRawName(rawName);
 
           if (isNil(inputVariableContext)) {
+            this.workflowContext.addError(
+              new WorkflowInputGenerationWarning(
+                `Input variable context not found for raw name: ${rawName}`
+              )
```
dvargasfuertes, line 99: I accidentally deleted this one in a prev pr

dvargas92495 (commented):
Going to actually leave this guy in draft and use this as an opportunity to add tests for each case as I pull them out

dvargas92495:
closing this and will update warnings as sentry errors come up

PR #1005, opened 2025-02-18
Remove unused docstring

vincent0426:
Quick fix on doctring to match the fields

### DIFF
--- a/src/vellum/workflows/nodes/core/map_node/node.py
+++ b/src/vellum/workflows/nodes/core/map_node/node.py
@@ -42,7 +42,6 @@ class MapNode(BaseAdornmentNode[StateType], Generic[StateType, MapNodeItemType])
     Used to map over a list of items and execute a Subworkflow on each iteration.
 
     items: List[MapNodeItemType] - The items to map over
-    subworkflow: Type["BaseWorkflow[SubworkflowInputs, BaseState]"] - The Subworkflow to execute on each iteration
     max_concurrency: Optional[int] = None - The maximum number of concurrent subworkflow executions
     """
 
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -42,7 +42,6 @@ class MapNode(BaseAdornmentNode[StateType], Generic[StateType, MapNodeItemType])
     Used to map over a list of items and execute a Subworkflow on each iteration.
 
     items: List[MapNodeItemType] - The items to map over
-    subworkflow: Type["BaseWorkflow[SubworkflowInputs, BaseState]"] - The Subworkflow to execute on each iteration
```
dvargasfuertes, line 45: [blocking] this is on the base class `BaseAdornmentNode`

src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -18,9 +18,9 @@ class RetryNode(BaseAdornmentNode[StateType], Generic[StateType]):
     Used to retry a Subworkflow a specified number of times.
 
     max_attempts: int - The maximum number of attempts to retry the Subworkflow
-    delay: float - The number of seconds to wait between retries
-    retry_on_error_code: Optional[VellumErrorCode] = None - The error code to retry on
-    subworkflow: Type["BaseWorkflow[SubworkflowInputs, BaseState]"] - The Subworkflow to execute
```
dvargasfuertes, line 23: same here

src/vellum/workflows/nodes/core/try_node/node.py
```diff
@@ -17,8 +17,7 @@ class TryNode(BaseAdornmentNode[StateType], Generic[StateType]):
     """
     Used to execute a Subworkflow and handle errors.
 
-    on_error_code: Optional[VellumErrorCode] = None - The error code to handle
-    subworkflow: Type["BaseWorkflow"] - The Subworkflow to execute
```
dvargasfuertes, line 21: same here. Though the top two should likely update their types to this one

vincent0426: ah so we should likely keep docstring consistent with the base class, then I would update the type only

dvargas92495 (approved):

PR #1007, opened 2025-02-18
Strict mode throws warnings

dvargas92495:
Part 3 of the following 4 step process:

- Run workflows in non-vembda mode (started [here](https://github.com/vellum-ai/vellum-python-sdks/compare/vargas/triage-base-codegen-errors?expand=1))
- Audit existing BaseCodegenErrors and turn most that would not affect runtime into warnings instead (ongoing [here](https://github.com/vellum-ai/vellum-python-sdks/pull/1003))
- Change strict to still error on both error/warnings (this PR)
- non-strict codegen always errors on errors (after we finish the audit in step 2)

This is safe to do since we are no longer sending strict: true from vellum for our sandbox and deployment code paths


noanflaherty (commented):
✅

PR #1008, opened 2025-02-18
Add null check for function call id value

NgoHarrison:


ee/codegen/src/__test__/chat-message-content.test.ts
```diff
@@ -57,6 +57,25 @@ describe("ChatMessageContent", () => {
       expect(await writer.toStringFormatted()).toMatchSnapshot();
       expect(chatMessageContent.getReferences()).toHaveLength(2);
     });
+
+    it("should write a function call content with id as null correctly", async () => {
+      const chatMessageContent = new ChatMessageContent({
+        chatMessageContent: {
+          type: "FUNCTION_CALL",
+          value: {
+            id: null as unknown as string,
```
dvargasfuertes, line 66: [nit] this `id` field is expected to be nullable actually. Let's update our serializer/types to not need the type cast here

dvargas92495 (approved):

ee/codegen/src/__test__/chat-message-content.test.ts
NgoHarrison: Will fast follow this

PR #1009, opened 2025-02-18
Convert API node workspace secret entity not found error to warning

dvargas92495:
Part of the triage from: https://github.com/vellum-ai/vellum-python-sdks/pull/1003/files

### DIFF
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -1075,9 +1075,10 @@ export function apiNodeFactory({
           value: {
             rules: [
               {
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -1075,9 +1075,10 @@ export function apiNodeFactory({
           value: {
             rules: [
               {
-                type: "INPUT_VARIABLE",
+                type: "CONSTANT_VALUE",
```
dvargasfuertes, line 1078: Removed these so that tests don't need input variable contexts defined for unrelated attributes

PR #1012, opened 2025-02-18
Remove escape characters for quotes before processing chat message values

NgoHarrison:
Context: This came up from qa'ing a customer workflow. They had strings in the chat history that essentially represented sql strings and the nested quotes weren't being escaped correctly when passed into fern. This in turn generates a sandbox file that has syntax issues

### DIFF
--- a/ee/codegen/src/generators/chat-message-content.ts
+++ b/ee/codegen/src/generators/chat-message-content.ts
@@ -18,6 +18,7 @@ import {
 
 import { VELLUM_CLIENT_MODULE_PATH } from "src/constants";
 import { Json } from "src/generators/json";
+import { removeEscapeCharacters } from "src/utils/casing";
 import { assertUnreachable } from "src/utils/typing";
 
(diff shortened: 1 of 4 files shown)


ee/codegen/src/generators/chat-message-content.ts
```diff
@@ -37,7 +38,7 @@ class StringChatMessageContent extends AstNode {
       arguments_: [
         python.methodArgument({
           name: "value",
-          value: python.TypeInstantiation.str(value),
+          value: python.TypeInstantiation.str(removeEscapeCharacters(value)),
```
dvargasfuertes, line 41: [oos] I get the sense we're going to see something like this in all usages of `python.TypeInstantiation.str`. It may be worth opening a PR to that repo with a test that reproduces the bug on their end and implementing a fix that escapes properly rather than removes it. I have a creeping concern that removal could lead to data loss, which was fine for input names but seems less tolerant for input values

dvargas92495 (approved):

PR #1014, opened 2025-02-18
Add adornments to legacy node serialization

peicodes:
This PR creates a base adornment node display that is used by `RetryNode` and `TryNode` to backfill adornments on non-generic nodes.

While `MapNode` is inheriting from the same display, it does not actively use it right now. IIRC we're not supporting map node as an adornment yet.

Some thoughts:
- It seems that we can't tell the difference between a `TryNode` adornment and an explicitly defined `TryNode`. At least not in the `TryNodeDisplay`. They both have wrapped nodes.

### DIFF
--- a/ee/codegen_integration/fixtures/simple_node_with_try_wrap/display_data/simple_node_with_try_wrap.json
+++ b/ee/codegen_integration/fixtures/simple_node_with_try_wrap/display_data/simple_node_with_try_wrap.json
@@ -142,7 +142,8 @@
             "<adornment>"
           ],
           "name": "TryNode"
-        }
+        },
+        "adornments": []
       },
       {
         "id": "54803ff7-9afd-4eb1-bff3-242345d3443d",
--- a/ee/codegen_integration/fixtures/simple_node_with_try_wrap/display_data/simple_node_with_try_wrap.json
+++ b/ee/codegen_integration/fixtures/simple_node_with_try_wrap/display_data/simple_node_with_try_wrap.json
@@ -143,7 +143,7 @@
           ],
           "name": "TryNode"
         },
-        "adornments": []
+        "adornments": [{"id": "3344083c-a32c-4a32-920b-0fb5093448fa", "label": "TryNode", "base": {"name": "TryNode", "module": ["vellum", "workflows", "nodes", "core", "try_node", "node"]}, "attributes": [{"id": "ab2fbab0-e2a0-419b-b1ef-ce11ecf11e90", "name": "on_error_code", "value": {"type": "CONSTANT_VALUE", "value": {"type": "JSON", "value": null}}}]}]
       },
       {
         "id": "54803ff7-9afd-4eb1-bff3-242345d3443d",
--- a/ee/vellum_ee/workflows/display/nodes/vellum/base_adornment_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/base_adornment_node.py
@@ -1,14 +1,12 @@
 from uuid import UUID
 from typing import Any, Callable, Generic, Optional, TypeVar, cast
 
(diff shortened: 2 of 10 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/base_adornment_node.py
```diff
@@ -0,0 +1,45 @@
...
+from vellum_ee.workflows.display.nodes.base_node_vellum_display import BaseNodeVellumDisplay
+from vellum_ee.workflows.display.nodes.get_node_display_class import get_node_display_class
+from vellum_ee.workflows.display.nodes.utils import raise_if_descriptor
+from vellum_ee.workflows.display.types import WorkflowDisplayContext
+
+_BaseAdornmentNodeType = TypeVar("_BaseAdornmentNodeType", bound=BaseAdornmentNode)
+
+
+class BaseAdornmentNodeDisplay(BaseNodeVellumDisplay[_BaseAdornmentNodeType], Generic[_BaseAdornmentNodeType]):
+    def serialize(
```
noanflaherty, line 16: With regards to:
> It seems that we can't tell the difference between a TryNode adornment and an explicitly defined TryNode.

We should be able to given that we use `@BaseTryNodeDisplay.wrap(...)` in the case of the TryNode being used as an adornment vs being used as a standalone node.

Given that we don't have near-term plans to support Retry nor Try as standalone nodes, I'm down to defer, but want to make sure we're at a minimum ticketing and not losing track of this. Mostly want to point out that this should be possible given the above.

ee/codegen_integration/fixtures/simple_node_with_try_wrap/display_data/simple_node_with_try_wrap.json
```diff
@@ -142,7 +142,8 @@
             "<adornment>"
           ],
           "name": "TryNode"
-        }
+        },
+        "adornments": [{"id": "3344083c-a32c-4a32-920b-0fb5093448fa", "label": "TryNode", "base": {"name": "TryNode", "module": ["vellum", "workflows", "nodes", "core", "try_node", "node"]}, "attributes": [{"id": "ab2fbab0-e2a0-419b-b1ef-ce11ecf11e90", "name": "on_error_code", "value": {"type": "CONSTANT_VALUE", "value": {"type": "JSON", "value": null}}}]}]
```
noanflaherty, line 146: Can we add a test cases that shows this works as expected if a node has multiple adornment nodes stacked? For example, a TryNode that wraps a RetryNode that wraps some other node.

noanflaherty (approved):

ee/vellum_ee/workflows/display/nodes/vellum/base_adornment_node.py
dvargasfuertes: Line 28 below we are already telling the difference between adornment and standalone node

ee/vellum_ee/workflows/display/nodes/vellum/base_adornment_node.py
```diff
@@ -0,0 +1,45 @@
...
+
+        wrapped_node = get_wrapped_node(node)
+        if not wrapped_node:
+            subworkflow = raise_if_descriptor(node.subworkflow)
+            if not isinstance(subworkflow.graph, type) or not issubclass(subworkflow.graph, BaseNode):
+                raise NotImplementedError(
+                    "Unable to serialize Try Nodes that wrap subworkflows containing more than one Node."
+                )
+
+            wrapped_node = subworkflow.graph
```
dvargasfuertes: [oos] In fact, we should stop supporting the graph case here of adding adornments or error output id in this case. Not gonna mark blocking for this pr but ideally is a fast follow

dvargas92495 (approved):

PR #1017, opened 2025-02-19
Use same run logic for inline prompt node and prompt deployment nodes

NgoHarrison:
Context: Came from QAing customer workflow. The workflow had internal server errors where the root cause was that it was not getting any outputs from the events since our logic was specifically looking for string types


dvargas92495 (approved):
Let's reproduce the error with a test that returns just a function call and no string outputs. Cool with it being a fast follow

PR #1020, opened 2025-02-20
Update ast node to write nothing when chat history is an empty list

NgoHarrison:
Context: Came up from QAing a customer workflow. I was trying to debug a related issue by running the sandbox locally and noticed the sandbox file was having syntax issues from codegen. It was writing things like `Inputs(chat_history=,foo=bar`.

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/vellum-variable-value.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/vellum-variable-value.test.ts.snap
@@ -18,6 +18,8 @@ exports[`VellumValue > CHAT_HISTORY > should write a CHAT_HISTORY value with a s
 
 exports[`VellumValue > CHAT_HISTORY > should write a CHAT_HISTORY value with just text 1`] = `"[ChatMessage(role="USER", text="Hello, AI!"),]"`;
 
+exports[`VellumValue > CHAT_HISTORY > should write nothing when CHAT_HISTORY value is empty list 1`] = `"None"`;
+
 exports[`VellumValue > CHAT_HISTORY > should write nothing when CHAT_HISTORY value is null 1`] = `"None"`;
 
 exports[`VellumValue > ERROR > should write a ERROR value correctly 1`] = `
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/vellum-variable-value.ts
```diff
@@ -75,7 +75,7 @@ class JsonVellumValue extends AstNode {
 }
 
 class ChatHistoryVellumValue extends AstNode {
-  private astNode: AstNode | undefined;
+  public astNode: AstNode | undefined;
```
dvargasfuertes: [nit] why do we need to make this public?

ee/codegen/src/generators/vellum-variable-value.ts
```diff
@@ -501,15 +503,14 @@ export class VellumValue extends AstNode {
         assertUnreachable(vellumValue);
     }
 
-    this.inheritReferences(this.astNode);
+    this.inheritReferences(this.astNode ?? undefined);
```
dvargasfuertes: [nit] is this not redundant?

dvargasfuertes: oh we set to `null` on L481. I'd just set to undefined if you disagree with the comment I left in the snapshot test file

ee/codegen/src/__test__/__snapshots__/vellum-variable-value.test.ts.snap
```diff
@@ -18,6 +18,8 @@ exports[`VellumValue > CHAT_HISTORY > should write a CHAT_HISTORY value with a s
 
 exports[`VellumValue > CHAT_HISTORY > should write a CHAT_HISTORY value with just text 1`] = `"[ChatMessage(role="USER", text="Hello, AI!"),]"`;
 
+exports[`VellumValue > CHAT_HISTORY > should write nothing when CHAT_HISTORY value is empty list 1`] = `"None"`;
```
dvargasfuertes: [nit] I would actually expect this to be `"[]"` - is that not possible from the fern sdk?

dvargas92495 (approved):

ee/codegen/src/generators/vellum-variable-value.ts
NgoHarrison: On the logic block around L481, I need to use this for that logic

ee/codegen/src/__test__/__snapshots__/vellum-variable-value.test.ts.snap
NgoHarrison: Definitely possible, let me update

ee/codegen/src/generators/vellum-variable-value.ts
NgoHarrison: But this was because I was trying to force it to write nothing, I can undo this once I make it write empty list instead

PR #1021, opened 2025-02-20
Support union output type of dataclasses for code exec node

dvargas92495:
This PR forces us to make out code node output type resolution to be recursive, which will enable the fix to the actual customer facing issue we're facing


vincent0426 (commented):
✅

marlenc3:
✅

PR #1022, opened 2025-02-20
Support chat history output type in code nodes

dvargas92495:
Chat history output type was failing for a customer - this fixes

### DIFF
--- a/src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
+++ b/src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
@@ -79,6 +79,7 @@ def _get_type_name(obj: Any) -> str:
 
 
 def _cast_to_output_type(result: Any, output_type: Any) -> Any:
+    is_valid_output_type = isinstance(output_type, type)
     if get_origin(output_type) is Union:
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
```diff
@@ -79,22 +79,27 @@ def _get_type_name(obj: Any) -> str:
...
             try:
                 return _cast_to_output_type(result, allowed_type)
             except NodeException:
                 continue
-    elif issubclass(output_type, BaseModel) and not isinstance(result, output_type):
+    elif get_origin(output_type) is list:
+        allowed_item_type = get_args(output_type)[0]
+        if isinstance(result, list):
+            return [_cast_to_output_type(item, allowed_item_type) for item in result]
+    elif is_valid_output_type and issubclass(output_type, BaseModel) and not isinstance(result, output_type):
```
dvargasfuertes, line 94: Wait a second. I feel like this whole method could just be replaced with a dynamic pydantic model and using `model_validate()`... will do this refactor in the next PR

marlenc3 (approved):

PR #1023, opened 2025-02-20
Refactor workflow get graph and get unused graph

vincent0426:
refactor get graph and get unused graph method
- get graph
- get edges
- get nodes

`_get_edges_from_subgraphs` and `_get_nodes_from_subgraphs` still look similar, but don't think grouping them will further improve readability which would become something like:
```python
@staticmethod
def _get_unique_items_from_subgraphs(subgraphs: Iterable[Graph], attr: str) -> Iterator[Any]:
    seen = set()
    for subgraph in subgraphs:
        for item in getattr(subgraph, attr):
            if item not in seen:
                seen.add(item)
                yield item

@classmethod
def get_edges(cls) -> Iterator[Edge]:
    """
    Returns an iterator over the edges in the workflow. We use a set to
    ensure uniqueness, and the iterator to preserve order.
    """
    return cls._get_unique_items_from_subgraphs(cls.get_subgraphs(), "edges")


@classmethod
def get_nodes(cls) -> Iterator[Type[BaseNode]]:
    """
    Returns an iterator over the nodes in the workflow. We use a set to
    ensure uniqueness, and the iterator to preserve order.
    """
    return cls._get_unique_items_from_subgraphs(cls.get_subgraphs(), "nodes")
```

### DIFF
--- a/src/vellum/workflows/workflows/base.py
+++ b/src/vellum/workflows/workflows/base.py
@@ -179,7 +179,7 @@ class BaseWorkflow(Generic[InputsType, StateType], metaclass=_BaseWorkflowMeta):
         return self._context
 
     @staticmethod
-    def _resolve_graph(graph: Any) -> List[Graph]:
+    def _resolve_graph(graph: GraphAttribute) -> List[Graph]:
         """
         Resolves a single graph source to a list of Graph objects.
         """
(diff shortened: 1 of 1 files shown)


vincent0426:
Follow up PR will add test case for get nodes and get edges!

dvargas92495 (approved):

NgoHarrison (commented):
✅

PR #1024, opened 2025-02-20
[SC-6715] Add serialization support for coalesce expression

NgoHarrison:
Context: Came from QAing react workflow. We had a lambda for the lazy reference


marlenc3 (approved):

dvargas92495 (approved):
Merging to include in 0.14.6

PR #1027, opened 2025-02-20
Make workflow expression values nullable and workflow output value descriptors nullable

NgoHarrison:
Follow up for this PR:
https://github.com/vellum-ai/vellum/pull/8382

### DIFF
--- a/ee/codegen/src/generators/workflow-value-descriptor.ts
+++ b/ee/codegen/src/generators/workflow-value-descriptor.ts
@@ -5,10 +5,6 @@ import { isNil } from "lodash";
 
 import { WorkflowContext } from "src/context";
 import { BaseNodeContext } from "src/context/node-context/base";
-import {
-  NodeAttributeGenerationError,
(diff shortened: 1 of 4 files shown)


ee/codegen/src/generators/workflow-value-descriptor.ts
```diff
@@ -107,40 +107,37 @@ export class WorkflowValueDescriptor extends AstNode {
...
-          `This operator: ${operator} is not supported`
-        );
-      }
-      return value;
-    } else {
-      throw new NodeAttributeGenerationError(
-        `Operators should exist on expression and not be null`
-      );
+    if (!this.isExpression(workflowValueDescriptor)) {
+      return "equals"; // default operator if not an expression
```
dvargasfuertes, line 111: [oos] long term, we'll probably want an identity operator to use as the default. maybe that's the same as the group expression.

dvargas92495 (approved):
Merging to include in 0.14.6

PR #1034, opened 2025-02-21
[SC-6781] Inputs validation on required values

NgoHarrison:
Context: Came from a customer issue where the required input variable was required but not defined. This is part 1 to error here. Part 2 is to handle this in vembda and return a 400 from vembda server

### DIFF
--- a/src/vellum/workflows/inputs/base.py
+++ b/src/vellum/workflows/inputs/base.py
@@ -43,20 +43,19 @@ class BaseInputs(metaclass=_BaseInputsMeta):
 
     def __init__(self, **kwargs: Any) -> None:
         for name, field_type in self.__class__.__annotations__.items():
-            if name not in kwargs:
+            if name not in kwargs and name not in vars(self.__class__):
                 origin = get_origin(field_type)
                 args = get_args(field_type)
-                has_default = name in vars(self.__class__)
-                if not has_default and not (origin is Union and type(None) in args):
+                if not (origin is Union and type(None) in args):
                     raise WorkflowInitializationException(
                         message="Required input variables should have defined value",
                         code=WorkflowErrorCode.INVALID_INPUTS,
                     )
 
-        # Then validate and set the provided values
         for name, value in kwargs.items():
             field_type = self.__class__.__annotations__.get(name)
-            self._validate_input(value, field_type)
+            if field_type:
+                self._validate_input(value, field_type)
             setattr(self, name, value)
 
     def __iter__(self) -> Iterator[Tuple[InputReference, Any]]:
--- a/src/vellum/workflows/inputs/base.py
+++ b/src/vellum/workflows/inputs/base.py
@@ -1,9 +1,11 @@
-from typing import Any, Iterator, Tuple, Type
+from typing import Any, Iterator, Tuple, Type, Union, get_args, get_origin
 from typing_extensions import dataclass_transform
 
 from pydantic import GetCoreSchemaHandler
 from pydantic_core import core_schema
 
+from vellum.workflows.errors.types import WorkflowErrorCode
+from vellum.workflows.exceptions import NodeException
 from vellum.workflows.references import ExternalInputReference, WorkflowInputReference
 from vellum.workflows.references.input import InputReference
 from vellum.workflows.types.utils import get_class_attr_names, infer_types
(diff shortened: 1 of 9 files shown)


src/vellum/workflows/inputs/base.py
```diff
@@ -41,13 +43,34 @@ class BaseInputs(metaclass=_BaseInputsMeta):
...
+                )
+        if isinstance(value, str) and value == "":
+            origin = get_origin(field_type)
+            args = get_args(field_type)
+            if not (origin is Union and type(None) in args):
+                raise NodeException(
+                    message="Empty string not allowed for required string input",
+                    code=WorkflowErrorCode.INVALID_INPUTS,
+                )
+
```
NgoHarrison: This is to account for the case where the UI is sending "" for a required input that is null

vincent0426:
👍

src/vellum/workflows/inputs/tests/test_inputs.py
```diff
@@ -0,0 +1,52 @@
...
+
+def test_base_inputs_empty_string():
+    # GIVEN some input class with required and optional string fields
+    class TestInputs(BaseInputs):
+        required_string: str
+        optional_string: Optional[str]
+
+    # WHEN we try to assign an empty string to a required string field
+    with pytest.raises(WorkflowInitializationException) as exc_info:
+        TestInputs(required_string="", optional_string="ok")
```
dvargasfuertes: [blocking] I think this should be valid

src/vellum/workflows/inputs/tests/test_inputs.py
```diff
@@ -0,0 +1,52 @@
...
+
+def test_base_inputs_empty_value():
+    # GIVEN some input class with required and optional string fields
+    class TestInputs(BaseInputs):
+        required_string: str
+        optional_string: Optional[str]
+
+    # WHEN we try to assign None to a required field
+    with pytest.raises(WorkflowInitializationException) as exc_info:
+        TestInputs(required_string=None, optional_string="ok")  # type: ignore
```
dvargasfuertes: [blocking] when python uses `Optional`, what it actually means is `Nullable`. An arg is optional if we add a default value with `required_string: str = "hello"`.

so the test here should be `TestInputs(optional_string="ok")` - looking at the implementation of `__init__` I think this test will catch a bug

NgoHarrison: Updated the test case and added handling to ensure required vars are given

src/vellum/workflows/inputs/tests/test_inputs.py
NgoHarrison: Removed test case based on discussions

src/vellum/workflows/inputs/base.py
```diff
@@ -40,14 +42,38 @@ class BaseInputs(metaclass=_BaseInputsMeta):
     __parent_class__: Type = type(None)
 
     def __init__(self, **kwargs: Any) -> None:
+        for name, field_type in self.__class__.__annotations__.items():
+            if name not in kwargs and name not in vars(self.__class__):
```
dvargasfuertes, line 46: [nit] we already do some of this magic in the metaclass, we should be able to just do `for input_descriptor in self.__class__:` like in the `__iter__` method below

src/vellum/workflows/inputs/base.py
```diff
@@ -40,14 +42,38 @@ class BaseInputs(metaclass=_BaseInputsMeta):
...
+                origin = get_origin(field_type)
+                args = get_args(field_type)
+                if not (origin is Union and type(None) in args):
+                    raise WorkflowInitializationException(
+                        message="Required input variables should have defined value",
+                        code=WorkflowErrorCode.INVALID_INPUTS,
+                    )
+
         for name, value in kwargs.items():
```
dvargasfuertes, line 55: [nit] we could do the `setattr` calls above and not need this loop. Doing it this way also avoids an unexpected kwarg from being set to the class

tests/workflows/basic_ports/elif_with_no_if/tests/test_workflow.py
```diff
@@ -1,9 +1,9 @@
-from tests.workflows.basic_ports.elif_with_no_if.workflow import ElifWithNoIfWorkflow
+from tests.workflows.basic_ports.elif_with_no_if.workflow import ElifWithNoIfWorkflow, Inputs
 
 
 def test_run_workflow():
     workflow = ElifWithNoIfWorkflow()
-    terminal_event = workflow.run()
+    terminal_event = workflow.run(inputs=Inputs(value="foo"))
```
dvargasfuertes, line 6: oh interesting - I guess this means the validation works haha

dvargas92495 (approved):

PR #1035, opened 2025-02-21
Fix failure to serialize Display Data attribtue

dvargas92495:
Solves a serialization <> codegen mismatch bug by consolidating our two display data classes

### DIFF
--- a/ee/vellum_ee/workflows/display/vellum.py
+++ b/ee/vellum_ee/workflows/display/vellum.py
@@ -41,10 +41,6 @@ class NodeDisplayData(UniversalBaseModel):
     comment: Optional[NodeDisplayComment] = None
 
 
-class GenericNodeDisplayData(UniversalBaseModel):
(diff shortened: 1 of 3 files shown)


ee/vellum_ee/workflows/display/vellum.py
```diff
@@ -41,10 +41,6 @@ class NodeDisplayData(UniversalBaseModel):
     comment: Optional[NodeDisplayComment] = None
 
 
-class GenericNodeDisplayData(UniversalBaseModel):
-    position: NodeDisplayPosition = Field(default_factory=NodeDisplayPosition)
```
dvargasfuertes, line 45: @peicodes any insight to share on why we created a separate Display data class?

peicodes: I think the legacy ones had fields that generic nodes did not, but I don't remember which fields

vincent0426:
✅

PR #1036, opened 2025-02-22
[SC-6595] Add init command to initialize a new Vellum project with templates

vincent0426:
sc: https://app.shortcut.com/vellum/story/6595/create-a-vellum-workflows-init-cli-command

Update: merged (https://github.com/vellum-ai/vellum/pull/8408)
~~to test now~~:
<strike>
- in vellum repo find list_examples under `vellum/django/app/api/workflow_sandboxes/views.py`
- add `authentication_classes` and `permission_classes` in action
</strike>

```python
@action(
    detail=False,
    methods=["get"],
    url_path="examples",
    url_name="list-examples",
    pagination_class=LimitOffsetPagination,
    authentication_classes=[APIKeyAuthentication, CustomJWTAuthentication],
    permission_classes=[HasAPIKey | (IsAuthenticated & HasWorkspacePermission)],
)
```

After above changes:
- `VELLUM_API_URL=http://localhost:8000 poetry run vellum workflows init`
- should see sandbox templates
- choose a number in the range
- it should pull down the template

https://github.com/user-attachments/assets/32c85f7b-5657-4443-a7e9-2b5a2a9d4cfc

### TODO:
- update vellum repo `authentication_classes` and `permission_classes`
- add [blank workspace](https://app.vellum.ai/workflow-sandboxes/e5fe5f20-fe0b-4b2a-b893-d192ba5c349b) to vellum repo
- group same logic between pull and init into `utils.py`
- follow PR in vellum repo will add blank project
- discuss which sandbox needs to be added as templates

### DIFF
--- a/ee/vellum_cli/__init__.py
+++ b/ee/vellum_cli/__init__.py
@@ -4,6 +4,7 @@ import click
 
 from vellum_cli.aliased_group import ClickAliasedGroup
 from vellum_cli.image_push import image_push_command
+from vellum_cli.init import init_command
 from vellum_cli.ping import ping_command
 from vellum_cli.pull import pull_command
 from vellum_cli.push import push_command
@@ -329,6 +330,13 @@ def images() -> None:
 def image_push(image: str, tag: Optional[List[str]] = None) -> None:
     """Push Docker image to Vellum"""
     image_push_command(image, tag)
+
+
+@workflows.command(name="init")
+def workflows_init() -> None:
+    """Initialize a new Vellum project"""
+
+    init_command()
 
 
 if __name__ == "__main__":
--- a/ee/vellum_cli/__init__.py
+++ b/ee/vellum_cli/__init__.py
@@ -334,7 +334,7 @@ def image_push(image: str, tag: Optional[List[str]] = None) -> None:
 
 @workflows.command(name="init")
 def workflows_init() -> None:
-    """Initialize a new Vellum project"""
+    """Initialize a new Vellum Workflow using a predefined template"""
 
     init_command()
 
--- a/ee/vellum_cli/init.py
+++ b/ee/vellum_cli/init.py
@@ -1,7 +1,6 @@
 import io
 import json
 import os
-from pathlib import Path
 import zipfile
 from typing import Optional
 
@@ -58,6 +57,11 @@ def init_command():
     if not pk:
         raise ValueError("No workflow sandbox ID found in project to pull from.")
 
+    target_dir = os.path.join(os.getcwd(), workflow_config.module)
+    if os.path.exists(target_dir):
+        click.echo(click.style(f"{target_dir} already exists.", fg="red"))
+        return
+
     logger.info(f"Pulling workflow into {workflow_config.module}...")
 
     query_parameters = {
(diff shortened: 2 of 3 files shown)


ee/vellum_cli/__init__.py
```diff
@@ -331,5 +332,12 @@ def image_push(image: str, tag: Optional[List[str]] = None) -> None:
     image_push_command(image, tag)
 
 
+@workflows.command(name="init")
+def workflows_init() -> None:
+    """Initialize a new Vellum project"""
```
dvargasfuertes:
```suggestion
    """Initialize a new Vellum Workflow using a predefined template"""
```

ee/vellum_cli/init.py
```diff
@@ -0,0 +1,141 @@
...
+    click.echo(click.style("Available Templates", bold=True, fg="green"))
+    for idx, template in enumerate(templates, 1):
+        click.echo(f"{idx}. {template.label}")
+
+    choice = click.prompt(
+        f"Please select a template number (1-{len(templates)})", type=click.IntRange(1, len(templates))
+    )
+    selected_template = templates[choice - 1]
+
+    click.echo(click.style(f"\nYou selected: {selected_template.label}\n", bold=True, fg="cyan"))
```
dvargasfuertes, line 42: this style stuff is cool!

ee/vellum_cli/init.py
```diff
@@ -0,0 +1,141 @@
...
+            if not workflow_config.module and pull_contents_metadata.label:
+                workflow_config.module = snake_case(pull_contents_metadata.label)
+
+        if not workflow_config.module:
+            raise ValueError(f"Failed to resolve a module name for Workflow {pk}")
+
+        target_dir = os.path.join(os.getcwd(), *workflow_config.module.split("."))
+
+        # Delete files in target_dir that aren't in the zip file
+        if os.path.exists(target_dir):
```
dvargasfuertes: [nit] I would think that we wouldn't need this block. If we are init-ing on a pre-existing directory, we likely just overwrite the directory entirely, or init in a new directory (`blank_2`), or fail if the target directory already has contents. Would be a good test to include in a follow-on PR to take a stance in one of the three options

ee/vellum_cli/tests/test_init.py
```diff
@@ -0,0 +1,178 @@
...
+    assert result.exit_code == 0
+
+    # AND `vellum_client.workflows.pull` is called with the selected template ID
+    vellum_client.workflows.pull.assert_called_once_with(
+        "template-1",
+        request_options={"additional_query_parameters": {"include_sandbox": True}},
+    )
+
+    # AND the `workflow.py` file should be created in the correct module directory
+    workflow_py = os.path.join(temp_dir, "example_workflow", "workflow.py")
```
dvargasfuertes, line 77: [oos] I could imagine a second input prompting the user for the target directory, where one inferred from the template (in this case, `example_workflow`) is the default

dvargas92495 (approved):
good stuff!

ee/vellum_cli/init.py
vincent0426: I moved the check forward to fail immediately if the folder already exists, prevents the user from:
- Coincidentally (even if unlikely) having the same name as the template.
- Init the same template twice (possibly after modifying the first one), which would force them to rename the original folder, giving them control over the process.

dvargas92495 (approved):

PR #1037, opened 2025-02-22
Nested set graph gen test case

dvargas92495:
This is a reduced test case that is currently failing a customer workflow. Will solve in the next pr


vincent0426:
✅

PR #1039, opened 2025-02-24
Validate external input types in WorkflowRunner

vincent0426:
inspired by #1034 

Validate external input type in workflow runner, which I think is the earliest point that gets the value passed in with types?

### DIFF
--- a/tests/workflows/basic_external_input/tests/test_workflow.py
+++ b/tests/workflows/basic_external_input/tests/test_workflow.py
@@ -1,3 +1,10 @@
+import pytest
+from typing import Union
+
+from vellum.workflows.exceptions import NodeException
+from vellum.workflows.nodes.bases.base import BaseNode
+from vellum.workflows.workflows.base import BaseWorkflow
+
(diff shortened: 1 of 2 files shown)


tests/workflows/basic_external_input/tests/test_workflow.py
```diff
@@ -29,3 +36,93 @@ def test_workflow__happy_path_multi_stop():
     # THEN we should get workflow in FULFILLED state
     assert final_terminal_event.name == "workflow.execution.fulfilled"
     assert final_terminal_event.outputs.final_value == "sunny"
+
+
+def test_workflow__happy_path_multi_stop_invalid_external_input():
+    """
+    Runs the non-streamed execution of a Workflow with a single Nodes defining ExternalInputs.
+    """
```
dvargasfuertes: [nit] Let's update these descriptions to indicate that we are validating an error case

dvargas92495 (approved):
love it

NgoHarrison (approved):
Nice!

PR #1041, opened 2025-02-24
[SC-6683] Convert hyphenated output names from fulfilled events to underscore

NgoHarrison:
Context: For this ticket, the workflow that flagged this issue had a final output name with a hyphen in it. Main vellum was using this key in the exec config (i.e `some-final_output`) but when we try to match that here, it doesn't find the attr and uses undefined. Codegen uses all underscores by sanitizing the key

### DIFF
--- a/src/vellum/workflows/outputs/base.py
+++ b/src/vellum/workflows/outputs/base.py
@@ -32,7 +32,7 @@ class BaseOutput(Generic[_Delta, _Accumulated]):
         if value is not undefined and delta is not undefined:
             raise ValueError("Cannot set both value and delta")
 
-        self._name = name
+        self._name = name.replace("-", "_")  # Convert hyphens to underscores for valid python variable names
         self._value = value
         self._delta = delta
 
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/outputs/base.py
```diff
@@ -32,7 +32,7 @@ def __init__(
         if value is not undefined and delta is not undefined:
             raise ValueError("Cannot set both value and delta")
 
-        self._name = name
+        self._name = name.replace("-", "_")  # Convert hyphens to underscores for valid python variable names
```
marlenc3, line 35: is it possible for hyphens to be a wanted state? like a user names it final-output? does this happen because codegen just uses hyphens for names?

NgoHarrison: Yeah we are codegen'ing underscores from sanitizing these outputs like `final-output` for variable naming so python won't complain

marlenc3: Ahhhh okok makes sense

marlenc3 (approved):

src/vellum/workflows/outputs/base.py
dvargasfuertes, line 35: [oos] We likely need to reproduce the full sanitization method that we have in codegen here. Probably as its own util. 

[way oos] Once we move display context from vembda initiated -> workflow initiated, I think we should be able to use that mapping instead of reproducing the logic. This is a refactor that I'm tracking this sprint

dvargas92495 (approved):

PR #1042, opened 2025-02-24
[SC-6794] Handle edge case where workflow init has two unconnected edges of entrypoint and terminal

NgoHarrison:


marlenc3 (approved):

ee/codegen/src/project.ts
```diff
@@ -445,6 +445,18 @@ ${errors.slice(0, 3).map((err) => {
...
     const rawData = this.workflowVersionExecConfig.workflowRawData;
+
+    // Edge case: Workflow init only has two nodes of ENTRYPOINT and TERMINAL with no edge between them
+    if (
+      rawData.edges.length === 0 &&
+      rawData.nodes.length === 2 &&
+      rawData.nodes.some((node) => node.type === "ENTRYPOINT")
+    ) {
+      return rawData.nodes.filter(
+        (node): node is WorkflowDataNode => node.type === "TERMINAL"
```
dvargasfuertes: [q] why do we filter out terminal nodes here?

ee/codegen/src/project.ts
```diff
@@ -445,6 +445,18 @@ ${errors.slice(0, 3).map((err) => {
    */
   private getOrderedNodes(): WorkflowDataNode[] {
     const rawData = this.workflowVersionExecConfig.workflowRawData;
+
+    // Edge case: Workflow init only has two nodes of ENTRYPOINT and TERMINAL with no edge between them
+    if (
+      rawData.edges.length === 0 &&
```
dvargasfuertes: [q] What was the error being thrown? Something tells me that no edges is the sufficient condition. eg, what happens when I drag out a third node and hit save?

dvargas92495 (approved):

ee/codegen/src/project.ts
NgoHarrison: Ahh very good point. The error is that node id not found which is being caused by this method in here not having the right info. But yeah this wouldn't work if  we drag out n number of nodes. Will update

ee/codegen/src/project.ts
NgoHarrison: I was only considering the case of the workflow init e.g two nodes and had to return terminal to start the processing for the bfs. But I have updated now to just filter and take all the ndoes that aren't entrypoint

PR #1044, opened 2025-02-25
Support template name as argument in workflows init cli

vincent0426:
Allow user to specify template name directly

currently only support following templates
- customer_support_q_a
- trust_center_q_a

test:
`vellum workflows init customer_support_q_a`


dvargas92495 (approved):
amazing 🎉

PR #1045, opened 2025-02-25
Support `--target-dir` option for vellum init cli 

vincent0426:
Allow user to specify --target-dir when init

test:
- `vellum workflows init --target-dir="dir1"`
- `vellum workflows init trust_center_q_a --target-dir="dir2"`

### DIFF
--- a/ee/vellum_cli/tests/test_init.py
+++ b/ee/vellum_cli/tests/test_init.py
@@ -353,3 +353,221 @@ def test_init_command_with_nonexistent_template_name(vellum_client, mock_module,
         with open(vellum_lock_json) as f:
             lock_data = json.load(f)
             assert lock_data["workflows"] == []
+
+
+@pytest.mark.parametrize(
+    "base_command",
+    [
+        ["workflows", "init"],
+    ],
+    ids=["workflows_init"],
+)
+def test_init__with_target_dir(vellum_client, mock_module, base_command):
+    # GIVEN a module on the user's filesystem
+    temp_dir = mock_module.temp_dir
+    mock_module.set_pyproject_toml({"workflows": []})
+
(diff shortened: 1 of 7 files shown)


ee/vellum_cli/tests/test_init.py
```diff
@@ -353,3 +336,197 @@ def test_init_command_with_nonexistent_template_name(vellum_client, mock_module,
...
             assert lock_data["workflows"] == []
+
+
+@pytest.mark.parametrize(
+    "base_command",
+    [
+        ["workflows", "init"],
+    ],
+    ids=["workflows_init"],
+)
```
dvargasfuertes: [nit] no need for `parametrize` if there's only one paramter

ee/vellum_cli/tests/test_init.py
```diff
@@ -353,3 +336,197 @@ def test_init_command_with_nonexistent_template_name(vellum_client, mock_module,
...
+    assert not os.path.exists(default_module_path)
+
+    # AND the vellum.lock.json file should be created in the original directory
+    vellum_lock_json = os.path.join(temp_dir, "vellum.lock.json")
+    assert os.path.exists(vellum_lock_json)
+    with open(vellum_lock_json) as f:
+        lock_data = json.load(f)
+        assert lock_data["workflows"] == [
+            {
+                "module": "example_workflow",
```
dvargasfuertes, line 343: [blocking] I missed this in the previous pull pr - I would expect the target dir to be in this module path now.

Either that or a new field in this vellum lock file, but that might be higher LoE bc it will have `push` implications

dvargas92495 (approved):

dvargas92495 (approved):

PR #1046, opened 2025-02-25
allow execution-context from init 

marlenc3:
turns out vembda did *not* play well with the wrapped context 
** moving back to using context to init workflow run with execution_context

### DIFF
--- a/src/vellum/workflows/runner/runner.py
+++ b/src/vellum/workflows/runner/runner.py
@@ -7,7 +7,7 @@ from uuid import UUID
 from typing import TYPE_CHECKING, Any, Dict, Generic, Iterable, Iterator, Optional, Sequence, Set, Tuple, Type, Union
 
 from vellum.workflows.constants import undefined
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/runner/runner.py
```diff
@@ -76,6 +76,7 @@ def __init__(
         cancel_signal: Optional[ThreadingEvent] = None,
         node_output_mocks: Optional[MockNodeExecutionArg] = None,
         max_concurrency: Optional[int] = None,
+        current_execution_context: Optional[ExecutionContext] = None,
```
dvargasfuertes: [nit] I'd just call this `execution_context`

I think event if we get the wrapper to work, we'll want to keep this arg

dvargas92495 (approved):

src/vellum/workflows/runner/runner.py
marlenc3: it just clashes with the wrapper - with execution_context() 😭  I might change it to init_execution_context

PR #1053, opened 2025-02-25
[SC-6767] Keep track of class names used for subworkflow nodes and nested nodes

NgoHarrison:
This helps prevent name collision by keep track of all the class names

### DIFF
--- a/ee/codegen/src/generators/nodes/bases/nested-workflow-base.ts
+++ b/ee/codegen/src/generators/nodes/bases/nested-workflow-base.ts
@@ -76,6 +76,7 @@ export abstract class BaseNestedWorkflowNode<
         parentNode: this,
         workflowClassName: nestedWorkflowClassName,
         workflowRawEdges: innerWorkflowData.edges,
+        classNames: this.workflowContext.classNames,
       });
 
     return new Map([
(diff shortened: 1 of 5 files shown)


ee/codegen/src/generators/nodes/bases/nested-workflow-base.ts
```diff
@@ -76,6 +76,7 @@ export abstract class BaseNestedWorkflowNode<
         parentNode: this,
         workflowClassName: nestedWorkflowClassName,
         workflowRawEdges: innerWorkflowData.edges,
+        classNames: this.workflowContext.classNames,
```
dvargasfuertes, line 79: [q] Why do we need nodes at different subworkflow levels to be deduped?

dvargas92495 (approved):

ee/codegen/src/generators/nodes/bases/nested-workflow-base.ts
NgoHarrison: When we codegen the classes, fern is using this className value we used. So the import path gets messed up and we end up importing the wrong node and then calling a non existent attribute on it e.g calling function_args on a subworkflow when really it is the nested templating node that has this

dvargasfuertes: Oh does fern's reference importing assume there isn't multiple classes of the same name at different directory levels?

NgoHarrison: I don't believe so. This is similar to the issue from the other month with us not having a handle on the import names in the `__all__` for the init displays and we had to modify that logic

PR #1055, opened 2025-02-26
Add codegen for unused graph

vincent0426:
for unused graph
- add two attr to GraphAttribute
  - ~~`isUnused`: default to false for normal graph~~
  - `unusedEdges`: since we check if there are unused edges right after we enter `addUnusedGraphs` and return directly without adding the `unused_graph` field, we don't want to recompute unused edges that will be used 
  - separate logic in generateGraphMutableAst with`unusedEdges` field

for unused graph (updated)
- maintain unused edges
- keep parsing next edges -> get all nodes that belongs to that edge
- find root node (if cycle return any)
- construct graph based on root node

~~Here, we didn't wrap the return value into set since its already a set~~
```py
const unusedGraphsField = python.field({
  name: "unused_graphs",
  initializer: new GraphAttribute({
    isUnused: true, // This flag tells GraphAttribute to handle unused nodes
    unusedEdges,
  }),
});
```

### DIFF
--- a/ee/codegen/src/__test__/unused-graph-attribute.test.ts
+++ b/ee/codegen/src/__test__/unused-graph-attribute.test.ts
@@ -129,5 +129,54 @@ describe("Workflow", () => {
         [disconnectedNode5, disconnectedNode6],
       ]);
     });
+
+    it("should identify multiple disconnected graphs with flipped edges", async () => {
+      const connectedNode = genericNodeFactory({
+        label: "ConnectedNode",
+      });
+
+      const disconnectedNode1 = genericNodeFactory({
+        label: "DisconnectedNode1",
+      });
+
+      const disconnectedNode2 = genericNodeFactory({
+        label: "DisconnectedNode2",
+      });
+
+      const disconnectedNode3 = genericNodeFactory({
+        label: "DisconnectedNode3",
+      });
+
+      await runUnusedGraphsWorkflowTest([
+        [entrypointNode, connectedNode],
+        [disconnectedNode2, disconnectedNode3],
+        [disconnectedNode1, disconnectedNode2],
+      ]);
+    });
+
+    it("should handle circular graphs", async () => {
+      const connectedNode = genericNodeFactory({
+        label: "ConnectedNode",
+      });
+
+      const disconnectedNode1 = genericNodeFactory({
+        label: "DisconnectedNode1",
+      });
+
+      const disconnectedNode2 = genericNodeFactory({
+        label: "DisconnectedNode2",
+      });
+
+      const disconnectedNode3 = genericNodeFactory({
+        label: "DisconnectedNode3",
+      });
+
+      await runUnusedGraphsWorkflowTest([
+        [entrypointNode, connectedNode],
+        [disconnectedNode1, disconnectedNode2],
+        [disconnectedNode2, disconnectedNode3],
+        [disconnectedNode3, disconnectedNode1],
+      ]);
+    });
   });
 });
--- a/ee/codegen/src/__test__/unused-graph-attribute.test.ts
+++ b/ee/codegen/src/__test__/unused-graph-attribute.test.ts
@@ -1,5 +1,4 @@
 import { Writer } from "@fern-api/python-ast/core/Writer";
-import { v4 as uuidv4 } from "uuid";
 
 import { workflowContextFactory } from "./helpers";
 import {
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/unused-graph-attribute.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      });
+
+      await runUnusedGraphsWorkflowTest([
+        [entrypointNode, connectedNode1],
+        [connectedNode1, connectedNode2],
+      ]);
+    });
+
+    it("should identify simple disconnected graph", async () => {
+      const connectedNode = templatingNodeFactory();
```
dvargasfuertes: [nit] let's use `genericNodeFactory` as our default node factory going forward

ee/codegen/src/generators/workflow.ts
```diff
@@ -506,6 +507,57 @@ export class Workflow {
...
+      });
+    }
+
+    // Get all unreachable edges
+    const unreachableEdges = this.getEdges().filter(
+      (edge) => !reachableEdges.has(edge.id)
+    );
+
+    // Filter out edges that reference nonexistent nodes
+    return unreachableEdges.filter((edge) => {
```
dvargasfuertes: [nit] does this not already happen in `GraphAttribute` ?

ee/codegen/src/generators/workflow.ts
```diff
@@ -530,6 +582,34 @@ export class Workflow {
...
+
+    try {
+      // We don't need to do any edge analysis here since GraphAttribute will handle it
+      // Just create a field with a single GraphAttribute that knows it's for unused nodes
+      const unusedGraphsField = python.field({
+        name: "unused_graphs",
+        initializer: new GraphAttribute({
+          workflowContext: this.workflowContext,
+          isUnused: true, // This flag tells GraphAttribute to handle unused nodes
+          unusedEdges,
```
dvargasfuertes: [nit] Untested idea that may be able to simplify both this interface and the redundant logic above:
- `workflow.ts` maintains its own list of unused edges
- `GraphAttribute` has a public method after construction of "edges used"
- We use the public method to filter down the current list of unused edges
- Keep constructing new graph attributes on a loop until we've used all of the edges, with each of these new ones considered part of unused_graphs.

Thoughts?

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -110,6 +149,157 @@ export class GraphAttribute extends AstNode {
...
+        edge.targetNodeId,
+        (incomingEdges.get(edge.targetNodeId) || 0) + 1
+      );
+      if (!incomingEdges.has(edge.sourceNodeId)) {
+        incomingEdges.set(edge.sourceNodeId, 0);
+      }
+    });
+
+    // Return nodes with zero incoming edges
+    return Array.from(incomingEdges.entries())
```
dvargasfuertes: [nit] Do we need to do this over `incomingEdges`? Or can we just iterate over nodes left that have no incoming edge?

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -110,6 +149,157 @@ export class GraphAttribute extends AstNode {
...
+        });
+      }
+
+      components.push(component);
+    }
+
+    return components;
+  }
+
+  private findRootNodes(componentEdges: WorkflowEdge[]): string[] {
```
dvargasfuertes: [oos] does this method handle loops? `A >> B >> C >> A`

ee/codegen/src/__test__/unused-graph-attribute.test.ts
```diff
@@ -0,0 +1,153 @@
...
+        sourceHandleId: uuidv4(),
+        targetHandleId: uuidv4(),
+      });
+
+      await runUnusedGraphsWorkflowTest([
+        [entrypointNode, connectedNode],
+        [disconnectedNode1, disconnectedNode2],
+        [disconnectedNode1, disconnectedNode3],
+        [disconnectedNode4, disconnectedNode5],
+        [disconnectedNode5, disconnectedNode6],
```
dvargasfuertes, line 129: [oos] I _think_ there might be a bug if these two were flipped in order, but the `findRootNodeIds` may account for it

dvargas92495 (approved):
good stuff - Ideas for improvement are up to your discretion

ee/codegen/src/__test__/unused-graph-attribute.test.ts
vincent0426: yeah I keep this part and return any node if there is a cycle

dvargas92495 (approved):

PR #1058, opened 2025-02-26
Serialization of Stacked Adornments

peicodes:
The reason the test was failing was because our code for collecting node displays only ever went one level deep with adorned nodes.


dvargas92495 (approved):

vincent0426:
✅

PR #1059, opened 2025-02-27
[SC-4658] Support custom types compression and handle incompatible types

vincent0426:
sc: https://app.shortcut.com/vellum/story/4658/support-any-class-with-relevant-dunder-methods-for-expression-binary-desciptors

Add support for any types that implement expression
- add error handling
- add test to validate custom type can override expression

### DIFF
--- a/src/vellum/workflows/expressions/greater_than.py
+++ b/src/vellum/workflows/expressions/greater_than.py
@@ -1,9 +1,20 @@
-from typing import Generic, TypeVar, Union
+from typing import Any, Generic, Protocol, TypeVar, Union, runtime_checkable
+from typing_extensions import TypeGuard
 
 from vellum.workflows.descriptors.base import BaseDescriptor
 from vellum.workflows.descriptors.exceptions import InvalidExpressionException
 from vellum.workflows.descriptors.utils import resolve_value
 from vellum.workflows.state.base import BaseState
+
+
+@runtime_checkable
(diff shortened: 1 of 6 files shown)


src/vellum/workflows/expressions/greater_than.py
```diff
@@ -21,14 +21,10 @@ def __init__(
...
-        if not isinstance(lhs, (int, float)):
-            raise InvalidExpressionException(f"Expected a numeric lhs value, got: {lhs.__class__.__name__}")
-
         rhs = resolve_value(self._rhs, state)
-        if not isinstance(rhs, (int, float)):
-            raise InvalidExpressionException(f"Expected a numeric rhs value, got: {rhs.__class__.__name__}")
 
-        return lhs > rhs
+        try:
+            return lhs > rhs  # type: ignore[operator]
```
dvargasfuertes: [oos] here's a technique for ignoring type ignore's throughout the PR - feel free to leave as a followup if preferred
Screenshot 2025-02-27 at 6 24 21 AM

dvargas92495 (approved):

src/vellum/workflows/expressions/greater_than.py
vincent0426: we could probably use python default error if the comparison can't be executed which is pretty clear

dvargas92495 (approved):
🎉

PR #1061, opened 2025-02-27
[SC-4850] Support json parse for descriptors

vincent0426:
sc: https://app.shortcut.com/vellum/story/4850/add-a-json-parse-operation-to-descriptors

Allow user to call `parse_json`() on descriptor
- comply with json.load types
- add serialization test to make sure it assert UI not supported yet

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -27,6 +27,7 @@ from vellum.workflows.expressions.is_not_undefined import IsNotUndefinedExpressi
 from vellum.workflows.expressions.is_null import IsNullExpression
 from vellum.workflows.expressions.is_undefined import IsUndefinedExpression
 from vellum.workflows.expressions.not_between import NotBetweenExpression
+from vellum.workflows.expressions.parse_json import ParseJsonExpression
 from vellum.workflows.nodes.bases.base import BaseNode
 from vellum.workflows.nodes.utils import get_wrapped_node
 from vellum.workflows.ports import Port
@@ -386,7 +387,7 @@ class BaseNodeDisplay(Generic[NodeType], metaclass=BaseNodeDisplayMeta):
                 "node_id": str(node_class_display.node_id),
             }
 
-        if value.__class__.__name__ == "ParseJsonExpression":
+        if isinstance(value, ParseJsonExpression):
             raise ValueError("ParseJsonExpression is not supported in the UI")
 
         if not isinstance(value, BaseDescriptor):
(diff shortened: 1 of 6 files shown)


src/vellum/workflows/expressions/parse_json.py
```diff
@@ -0,0 +1,30 @@
...
+_T = TypeVar("_T")
+
+
+class ParseJsonExpression(BaseDescriptor[Any], Generic[_T]):
+    def __init__(
+        self,
+        *,
+        expression: Union[BaseDescriptor[_T], _T],
+    ) -> None:
+        super().__init__(name=f"parse_json({expression})", types=(Any,))  # type: ignore[arg-type]
```
vincent0426, line 18: `type: ignore` allow `types=(Any,)` which is same type from `json.loads`

ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -386,6 +386,9 @@ def serialize_value(self, display_context: "WorkflowDisplayContext", value: Base
                 "node_id": str(node_class_display.node_id),
             }
 
+        if value.__class__.__name__ == "ParseJsonExpression":
```
dvargasfuertes: [nit] why not use `isinstance(value, ParseJsonExpression)` like the rest of the conditions here?

src/vellum/workflows/expressions/parse_json.py
dvargasfuertes: [oos] we're going to need to extend the definition of `types` to include `Any` as an option. Last I checked though this was a bit tricky

src/vellum/workflows/expressions/parse_json.py
```diff
@@ -0,0 +1,30 @@
...
+        self._expression = expression
+
+    def resolve(self, state: "BaseState") -> Any:
+        value = resolve_value(self._expression, state)
+
+        if not isinstance(value, str):
```
dvargasfuertes: [nit] we should support bytes and bytearray here too
Screenshot 2025-02-27 at 7 03 35 AM

src/vellum/workflows/expressions/parse_json.py
```diff
@@ -0,0 +1,30 @@
...
+            raise InvalidExpressionException(f"Expected a string, but got {value} of type {type(value)}")
+
+        try:
+            return json.loads(value)
+        except json.JSONDecodeError:
+            raise InvalidExpressionException(f"Failed to parse JSON: {value}")
```
dvargasfuertes: [nit] let's not put the `value` in the error message - these can often be giant payloads. Instead, let's just add the exception from JSONDecodeError

dvargas92495 (approved):
Nice!

NgoHarrison (approved):

vincent0426:
Refactor success tests to `utils.py` and only test error ones in `test_parse_json.py`

dvargas92495 (approved):

PR #1060, opened 2025-02-27
Raise exception in api node when api fails

vincent0426:
slack: https://vellum-ai.slack.com/archives/C06P13ABK0W/p1740633672761709

Raise the exception when api fail
- add test to make sure it raise

PR #1064, opened 2025-02-27
Allow empty state for base node for testing purpose

vincent0426:
context: when trying to remove non-used state https://github.com/vellum-ai/vellum-python-sdks/pull/1060#discussion_r1973436555, there's an error when args are empty since we access args[0] without checking it

As the comment mentions `# Instantiate a default state class if one is not provided, for ease of testing`, this should not affect anything besides test?

- Check if args have value first before assigning. Otherwise, it would cause index error if no state is provided


dvargas92495 (approved):
oh nice 👍

PR #1065, opened 2025-02-27
Support workflow streaming events for final output node's

dvargas92495:
This is a feature we had for legacy workflows that we're currently only supporting for streaming. Adds the same logic for fulfilled. Future iterations include:
- consolidation of the logic between our two workflow streaming emission sites
- support final output node streaming when prompt node is streaming, which workflows supports today


dvargas92495:
test failures are valid - need to decide whether we want to double emit a workflow streaming event when the output fulfills

PR #1067, opened 2025-02-28
Support single orphaned terminal node referenced by workflow outputs

dvargas92495:
Solves an issue a customer ran into where they had an orphaned terminal node that wouldn't codegen as an unused graph


marlenc3 (commented):
✅

NgoHarrison (commented):
✅

vincent0426:
✅ 
do some refactor in #1079

PR #1070, opened 2025-02-28
Remove the add Workflow Edges method on workflow context

dvargas92495:
Last refactor needed before I tackle: https://github.com/vellum-ai/vellum-python-sdks/pull/1067

This method on workflow context was only ever used in tests. It's a dangerous method for consumers to interact with as it distorts the original user supplied data. So this PR removes it

### DIFF
--- a/ee/codegen/src/context/workflow-context/workflow-context.ts
+++ b/ee/codegen/src/context/workflow-context/workflow-context.ts
@@ -429,10 +429,6 @@ export class WorkflowContext {
     return mlModel.name;
   }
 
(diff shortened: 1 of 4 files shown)


ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -429,10 +429,6 @@ export class WorkflowContext {
     return mlModel.name;
   }
 
-  public addWorkflowEdges(edges: WorkflowEdge[]): void {
-    this.workflowRawData.edges.push(...edges);
-  }
-
```
dvargasfuertes, line 435: Main goal of the PR

PR #1071, opened 2025-02-28
Forward 4xx errors from execute code API instead of internal error

dvargas92495:
A bug in the user's code should not be flagged by us as an internal server error


marlenc3 (commented):
✅

NgoHarrison (commented):
✅

PR #1072, opened 2025-02-28
Ignore serializing keys that are mapped to the undefined value

dvargas92495:
This was causing odd serialization failures on the Vellum side, bc `undefined` was serializing as `"undefined"`, which is not ideal. We now treat it exactly how javascript treats it


marlenc3 (commented):
✅

PR #1073, opened 2025-02-28
Move WorkflowEventDisplayContext into open source directory for use within workflow.execution.initiated

dvargas92495:
This moves us one step closer to moving the initiated event for the workflow server to the entry of the method

### DIFF
--- a/ee/vellum_ee/workflows/display/types.py
+++ b/ee/vellum_ee/workflows/display/types.py
@@ -1,9 +1,8 @@
 from dataclasses import dataclass, field
-from uuid import UUID
 from typing import TYPE_CHECKING, Dict, Generic, Tuple, Type, TypeVar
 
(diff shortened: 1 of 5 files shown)


ee/vellum_ee/workflows/display/types.py
```diff
@@ -1,9 +1,8 @@
 from dataclasses import dataclass, field
-from uuid import UUID
 from typing import TYPE_CHECKING, Dict, Generic, Tuple, Type, TypeVar
 
-from vellum.client.core import UniversalBaseModel
 from vellum.workflows.descriptors.base import BaseDescriptor
+from vellum.workflows.events.workflow import WorkflowEventDisplayContext  # noqa: F401
```
dvargasfuertes, line 5: Keeping this here to prevent a breaking change during a patch release

marlenc3 (approved):

PR #1077, opened 2025-02-28
Add function call obj handling for templating nodes

NgoHarrison:
This adds more support for function calling. This came from the change where if we reference array outputs instead of regular outputs on function calling, we don't need to access attributes the legacy way like `{{ json.loads(some_function_call[0].tool_id) }}`. We can now do templating like `{{ output[0] }}` when given array outputs

### DIFF
--- a/src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
+++ b/src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
@@ -234,3 +234,52 @@ def test_templating_node__last_chat_message():
 
     # THEN the output is the expected JSON
     assert outputs.result == [ChatMessage(role="USER", text="Hello")]
+
+
+def test_templating_node__function_call_value_input():
+    # GIVEN a templating node that receives a FunctionCallVellumValue
+    class FunctionCallTemplateNode(TemplatingNode[BaseState, FunctionCall]):
+        template = """{{ function_call }}"""
+        inputs = {
+            "function_call": FunctionCallVellumValue(
+                type="FUNCTION_CALL",
+                value=FunctionCall(name="test_function", arguments={"key": "value"}, id="test_id", state="FULFILLED"),
+            )
+        }
+
+    # WHEN the node is run
+    node = FunctionCallTemplateNode()
+    outputs = node.run()
+
+    # THEN the output is the expected function call
+    assert outputs.result == FunctionCall(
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
```diff
@@ -234,3 +234,52 @@ class LastChatMessageTemplateNode(TemplatingNode[BaseState, List[ChatMessage]]):
...
+    node = FunctionCallTemplateNode()
+    outputs = node.run()
+
+    # THEN the output is the expected function call
+    assert outputs.result == FunctionCall(
+        name="test_function", arguments={"key": "value"}, id="test_id", state="FULFILLED"
+    )
+
+
+def test_templating_node__function_call_as_json():
```
NgoHarrison, line 260: This test represents the real situation I encountered when debugging this issue. We wanted the output as json

src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
```diff
@@ -234,3 +234,52 @@ class LastChatMessageTemplateNode(TemplatingNode[BaseState, List[ChatMessage]]):
 
     # THEN the output is the expected JSON
     assert outputs.result == [ChatMessage(role="USER", text="Hello")]
+
+
+def test_templating_node__function_call_value_input():
```
NgoHarrison, line 239: This represents another situation where I was playing around with the UI and specified `FunctionCall` as the desired output type. This was also causing a bug so I added extra support here

marlenc3 (approved):

src/vellum/workflows/nodes/utils.py
```diff
@@ -124,9 +128,16 @@ def parse_type_from_str(result_as_str: str, output_type: Any) -> Any:
     if issubclass(output_type, BaseModel):
         try:
             data = json.loads(result_as_str)
+            # If we got a FunctionCallVellumValue extract FunctionCall,
+            if (
+                hasattr(output_type, "__name__")
+                and output_type.__name__ == "FunctionCall"
+                and isinstance(data, dict)
+                and "value" in data
+            ):
```
dvargasfuertes, line 137: [oos] we're going to want to generalize these two checks to the full schema. I was about to write a comment on how to do so, but it got gnarlier than expected, so happy to defer while we get this change in first

dvargas92495 (approved):

PR #1078, opened 2025-03-01
[SC-6270] Default api node should only codegen url field

vincent0426:
sc: https://app.shortcut.com/vellum/story/6270/skip-codegenning-of-default-api-node-fields

Default api node (GET & empty body & empty headers) should only generate url field
- add test for each conditions
  - should only generate url field
  - should generate method field when method is POST
  - should generate headers field when headers are provided
  - ~~should generate body field when body is provided~~ -> should not generate body field when body is empty
  - should generate all fields when all conditions are provided

since current API node factory didn't accept some fields, follow up PR could address this

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
@@ -94,6 +94,15 @@ class APINode(BaseAPINode):
 "
 `;
 
+exports[`ApiNode > basic api node > should only generate url field 1`] = `
+"from vellum.workflows.nodes.displayable import APINode as BaseAPINode
+
+
+class APINode(BaseAPINode):
+    url = "fasdfadsf"
+"
+`;
+
 exports[`ApiNode > basic auth secret node > secret ids should show names 1`] = `
 "from vellum.workflows.constants import APIRequestMethod, AuthorizationType
 from vellum.workflows.nodes.displayable import APINode as BaseAPINode
--- a/ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
@@ -94,6 +94,75 @@ class APINode(BaseAPINode):
 "
 `;
 
+exports[`ApiNode > basic api node > should generate API key header fields when API key is provided 1`] = `
+"from vellum.workflows.constants import APIRequestMethod, AuthorizationType
+from vellum.workflows.nodes.displayable import APINode as BaseAPINode
+
+
+class APINode(BaseAPINode):
+    url = "https://api.example.com/data"
+    method = APIRequestMethod.GET
+    api_key_header_key = "X-API-KEY"
+    authorization_type = AuthorizationType.API_KEY
+    api_key_header_value = "my-api-key"
+"
+`;
+
+exports[`ApiNode > basic api node > should generate all fields when all conditions are provided 1`] = `
+"from vellum.workflows.constants import APIRequestMethod, AuthorizationType
+from vellum.workflows.nodes.displayable import APINode as BaseAPINode
+
+
+class APINode(BaseAPINode):
+    url = "https://api.example.com/data"
+    method = APIRequestMethod.POST
+    json = {}
+    headers = {
+        "Content-Type": "application/json",
+    }
+    api_key_header_key = "X-API-KEY"
+    authorization_type = AuthorizationType.API_KEY
+    api_key_header_value = "my-api-key"
+"
+`;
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
```diff
@@ -94,6 +94,84 @@ class APINode(BaseAPINode):
...
+
+exports[`ApiNode > basic api node > should generate body field when body is provided 1`] = `
+"from vellum.workflows.constants import APIRequestMethod
+from vellum.workflows.nodes.displayable import APINode as BaseAPINode
+
+
+class APINode(BaseAPINode):
+    url = "fasdfadsf"
+    method = APIRequestMethod.GET
+    json = {}
```
dvargasfuertes: [nit] I'd expect `json = {}` to be skipped, particularly on GET requests. let's:
- update this to a POST example
- add a `"foo": "bar"` to the `json`

ee/codegen/src/__test__/nodes/__snapshots__/api-node.test.ts.snap
```diff
@@ -94,6 +94,84 @@ class APINode(BaseAPINode):
...
+"
+`;
+
+exports[`ApiNode > basic api node > should generate all fields when all conditions are provided 1`] = `
+"from vellum.workflows.constants import APIRequestMethod, AuthorizationType
+from vellum.workflows.nodes.displayable import APINode as BaseAPINode
+
+
+class APINode(BaseAPINode):
+    url = "fasdfadsf"
```
dvargasfuertes: [nit] let's use something that looks like a url

oh looks like we copied this from `apiNodeFactory` - let's have both use something url-looking

ee/codegen/src/generators/nodes/api-node.ts
```diff
@@ -27,16 +27,50 @@ export class ApiNode extends BaseSingleFileNode<ApiNodeType, ApiNodeContext> {
       })
     );
 
+    const methodInput = this.nodeInputsByKey.get("method");
+    if (!methodInput) {
+      throw new NodeAttributeGenerationError(
+        'Node input "method" is required but not found.'
+      );
```
dvargasfuertes: [blocking] in general, we've been moving to having these all be `"WARNING"`s (see: https://github.com/vellum-ai/vellum-python-sdks/pull/1003) so that we don't kill the _entire_ codegen process.

So, let's:
- add `"WARNING"` params to all of these
- wrap this in `this.workflowContext.addError()`

ee/codegen/src/generators/nodes/api-node.ts
```diff
@@ -27,16 +27,50 @@ export class ApiNode extends BaseSingleFileNode<ApiNodeType, ApiNodeContext> {
...
+    // If all four conditions are met (GET, no body, no headers, no API key)
+    // return with just the URL field
+    if (isGetMethod && !hasBody && !hasHeaders && !hasApiKeyHeader) {
+      return [
+        python.field({
+          name: "url",
+          initializer: urlInput,
+        }),
+      ];
+    }
```
dvargasfuertes: [blocking] instead of early returning we simply want to iterate over each attribute, and only add to `statements` if it's non default

ee/codegen/src/generators/nodes/api-node.ts
vincent0426: I removed this check for method since it now follows the original method handling

ee/codegen/src/generators/nodes/api-node.ts
```diff
@@ -27,16 +27,17 @@ export class ApiNode extends BaseSingleFileNode<ApiNodeType, ApiNodeContext> {
...
+        python.field({
+          name: "method",
+          initializer: methodValue,
+        })
+      );
+    }
     const body = this.nodeInputsByKey.get("body");
-
-    if (body) {
+    if (body && body.toString() !== "{}") {
```
vincent0426, line 40: should not generate `body` field when its empty

ee/codegen/src/generators/nodes/api-node.ts
```diff
@@ -45,12 +46,13 @@ export class ApiNode extends BaseSingleFileNode<ApiNodeType, ApiNodeContext> {
       );
     }
 
-    if (this.nodeData.data.additionalHeaders) {
+    const additionalHeaders = this.nodeData.data.additionalHeaders;
+    if (additionalHeaders) {
```
dvargasfuertes: [nit] we should also check that the length of `additionalHeaders` > 0

dvargas92495 (approved):
🎉

dvargas92495 (approved):

PR #1079, opened 2025-03-01
Refactor filter logic for unused edges

vincent0426:
Refactor filter edge logic from `addUnusedGraphs()` to `addGraph()` which is similar to adding node

### DIFF
--- a/ee/codegen/src/generators/workflow.ts
+++ b/ee/codegen/src/generators/workflow.ts
@@ -545,7 +545,16 @@ export class Workflow {
       const allEdges = this.getEdges();
       allEdges.forEach((edge) => {
         if (!usedEdges.has(edge)) {
-          this.unusedEdges.add(edge);
+          if (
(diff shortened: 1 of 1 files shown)


ee/codegen/src/generators/workflow.ts
```diff
@@ -545,7 +545,16 @@ export class Workflow {
       const allEdges = this.getEdges();
       allEdges.forEach((edge) => {
         if (!usedEdges.has(edge)) {
-          this.unusedEdges.add(edge);
+          if (
+            this.workflowContext.globalNodeContextsByNodeId.has(
```
dvargasfuertes: [nit] Instead of `this.workflowContext.globalNodeContextsByNodeId` (which is a derived field), let's use the original data by creating a set of node ids with `this.workflowContext.workflowRawData.nodes` below

dvargas92495 (approved):

dvargas92495 (approved):

PR #1083, opened 2025-03-02
Refactor unused nodes and edges marking logic into separate methods

vincent0426:
Extracts unused logic function into dedicated methods
- refactor only, no changes to logic


dvargas92495 (approved):

m-abboud (commented):
✅

PR #1092, opened 2025-03-05
Support returning arrays in  mode for code nodes

dvargas92495:
Found while QA-ing a customer's workflow - the ForwardRef in the `Json` type was causing problems.

I could've sworn I switched this to use pydantic. Will likely repursue that soon


marlenc3:
✅

m-abboud (commented):
✅

vincent0426:
✅

PR #1094, opened 2025-03-05
Clean up image push logs

m-abboud:
These warnings happen during architecture detection and don't matter too much since it can fallback to a different method


dvargas92495 (commented):
✅

PR #1095, opened 2025-03-06
[SC-6923] Resile AND combinator

vincent0426:
sc: https://app.shortcut.com/vellum/story/6923/make-codegen-resilient-to-and-s-in-the-combinator

Normalize `AND` combinator to `OR` before parsing for resilience


ee/codegen/src/__test__/project.test.ts
```diff
@@ -1526,4 +1526,191 @@ baz = foo + bar
...
+        },
+      });
+
+      await project.generateCode();
+
+      expectProjectFileToExist(project, ["nodes", "final_output.py"]);
+      expectProjectFileToMatchSnapshot(project, ["nodes", "final_output.py"]);
+    });
+
+    it("should handle a workflow with mixed combinators", async () => {
```
dvargasfuertes: [q] what's different about this test from the above test?

ee/codegen/src/project.ts
```diff
@@ -191,6 +191,26 @@ ${errors.slice(0, 3).map((err) => {
     }
   }
 
+  // preprocessWorkflowConfig is used to modify the workflow config before it is used to generate the project.
+  // - Changes the combinator of any AND conditions to OR conditions, this should happen rarely and only for
+  //   resiliency reasons.
```
dvargasfuertes: [nit] I'd give this a try from `ee/codegen/src/serializers/vellum.ts` on `NodeInputSerializer`:

```typescript
    combinator: undiscriminatedUnionSchema([
      stringLiteralSchema("OR"),
      stringLiteralSchema("AND"),
    ]).transform({
      transform: () => "OR" as const,
      untransform: () => "OR" as const,
    }),
```

It allows us to not need to do a preprocess step on unknown jsons, keeping everything strictly typed

dvargasfuertes: you'll also need to update the `combinator` annotation for `NodeInputSerializer.Raw`

dvargas92495 (approved):

ee/codegen/src/__test__/project.test.ts
vincent0426: maybe this one could be removed, testing if it receives both AND and OR combinator but its redundant actually

dvargas92495 (approved):

PR #1096, opened 2025-03-06
Use command line docker for getting image hash instead of python docker client

m-abboud:
For some reason when Vargas pushed an image it gave a sha that was not the sha that had been pushed to the internal docker repo. I had Vargas run the docker inspect command locally and it gave the right sha so I'm assuming this will work.

### DIFF
--- a/ee/vellum_cli/image_push.py
+++ b/ee/vellum_cli/image_push.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import re
 import subprocess
 from typing import List, Optional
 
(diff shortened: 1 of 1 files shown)


ee/vellum_cli/image_push.py
```diff
@@ -72,9 +73,29 @@ def image_push_command(image: str, tags: Optional[List[str]] = None) -> None:
...
                 continue
 
-    logger.info("Updating Vellum metadata and validating image works in our system...")
-    image_details = docker_client.api.inspect_image(image)
-    sha = image_details["Id"]
+    result = subprocess.run(
+        ["docker", "inspect", "--format='{{index .RepoDigests 0}}'", image],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
```
dvargasfuertes, line 80: [oos] would be nice to add tests - down to defer if hard to mock out the response to the docker inspect command

dvargas92495 (approved):

ee/vellum_cli/image_push.py
m-abboud: Ya I can followup with this and ya I defered on it since basically every line here needs to get mocked out.

vincent0426:
✅

PR #1099, opened 2025-03-06
[SC-6977] Resile port name context

vincent0426:
sc: https://app.shortcut.com/vellum/story/6977/generic-ports-are-codegen-ing-port-ids-instead-of-port-names
slack: https://vellum-ai.slack.com/archives/C06P13ABK0W/p1741215085018699

Add `toPythonSafeSnakeCase` to transform port name

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
@@ -277,6 +277,14 @@ exports[`Workflow > graph > should handle a simple edge of generic nodes 1`] = `
 "
 `;
 
+exports[`Workflow > graph > should properly format UUID-like port names in graph references 1`] = `
+"{
+    GenericNode.Ports._7d813792_04c2_4d26_a126_16de4c4cebaf >> FirstNode,
+    GenericNode.Ports._6f909cc6_887c_46d8_966b_46547900fc9c >> SecondNode,
+}
+"
+`;
+
 exports[`Workflow > graph > should support an edge between two sets 1`] = `
 "{
     TopLeftNode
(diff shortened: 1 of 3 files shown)


noanflaherty (approved):
This can be done in a separate PR, but can we also please add a test for when two ports have the same name from the api?

vincent0426:
> This can be done in a separate PR, but can we also please add a test for when two ports have the same name from the api?

Should we expect them to have different names? its same now

vincent0426:
> This can be done in a separate PR, but can we also please add a test for when two ports have the same name from the api?

node ports will have same value, don't think this is what we want

can do something like this for nodeport
`name: ${toPythonSafeSnakeCase(port.name)}_${idx + 1},`
```py
class Ports(BaseNode.Ports):
    same_port_name_1 = Port.on_if("test-value-1")
    same_port_name_2 = Port.on_elif("test-value-2")
    unique_name_3 = Port.on_else(None)
```
but if the name is not duplicated, it will also have the suffix, is this fine or should we add a check to handle only same value?
```py
class Ports(BaseNode.Ports):
    if_port_1 = Port.on_if(Inputs.count.is_null())
    elif_port_2 = Port.on_elif(
        TestNode.Outputs.my_output.equals("expected-value").equals(
            "another-expected-value"
        )
    )
    else_port_3 = Port.on_else(None)
```

dvargas92495 (approved):
> but if the name is not duplicated, it will also have the suffix

I would look at how we handle this for output/attribute/input names: https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/codegen/src/context/output-variable-context.ts#L40

In summary, we:
- generate the python safe name case
- are tracking all generated names within a given class
- if and only if we detect the dupe, then we generate the suffix

As Noa mentioned, we could tackle this in a followup

PR #1102, opened 2025-03-06
Use dynamic pydantic models to workaround all future code exec node output type problems

dvargas92495:
Defer all type validation work to folks who are already thinking about this problem


marlenc3 (commented):
🔥

vincent0426:
✅

PR #1103, opened 2025-03-06
Add mock node output parsing support

dvargas92495:
For context, check out this PR in our Workflow Server for how we anticipate using this function: https://github.com/vellum-ai/vembda-service/pull/154

I view this PR as the quickest path to solving Node Output Mocks in the SDK that doesn't involve making big legacy workflow runner changes + data migrations. We parse the raw json that the FE/Django maintains today, and translate it into our SDK native shape. We also outline a migration path that allows us to delete mock support in old workflows and only support in SDK workflows, incentivizing more gradual adoption.

Still TODO:
- Support using output ids reassigned by output display classes
- Cut a release to merge the Vembda PR

PR #1104, opened 2025-03-06
Simplify iterators in BaseOutputs

dvargas92495:
Simplification I noticed in https://github.com/vellum-ai/vellum-python-sdks/pull/1103

### DIFF
--- a/src/vellum/workflows/outputs/base.py
+++ b/src/vellum/workflows/outputs/base.py
@@ -204,13 +204,11 @@ class BaseOutputs(metaclass=_BaseOutputsMeta):
         if not isinstance(other, dict):
             return super().__eq__(other)
 
-        outputs = {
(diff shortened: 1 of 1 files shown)


src/vellum/workflows/outputs/base.py
```diff
@@ -204,13 +204,11 @@ def __eq__(self, other: object) -> bool:
         if not isinstance(other, dict):
             return super().__eq__(other)
 
-        outputs = {
-            name: value for name, value in vars(self).items() if not name.startswith("_") and value is not undefined
-        }
+        outputs = {ref.name: value for ref, value in self if value is not undefined}
```
dvargasfuertes, line 207: I want to remove the need to do ` if value is not undefined` one day, but still need to audit which use cases that negatively affects

m-abboud (commented):
✅

PR #1112, opened 2025-03-08
Implement pattern of fern extensions for non-blocking custom functionality

dvargas92495:
There's a bug in fern's block string AST Node that we don't want to wait for upstream changes for. I foresee this becoming a pattern that we should become increasingly comfortable with.

This PR creates a dedicated directory of changes that we should eventually upstream to Fern, but don't have the bandwidth or patience to do so now


vincent0426 (approved):

noanflaherty (commented):
✅

PR #1118, opened 2025-03-09
Solve Internal Server Error initiated by Map Node index access

dvargas92495:
Here is the sentry error I was debugging: https://vellum.sentry.io/issues/6360534819/?alert_rule_id=15639643&alert_type=issue&notification_uuid=5d198d7e-0177-433f-8a37-405afcfc54e8&project=4508362683514880

The bug was that map node was confusing workflow events from internal subworkflows as its own. We likely have this issue with all subworkflowable nodes, so will follow up this PR with tests/fixes for those nodes as well

### DIFF
--- a/src/vellum/workflows/nodes/core/map_node/node.py
+++ b/src/vellum/workflows/nodes/core/map_node/node.py
@@ -107,11 +107,17 @@ class MapNode(BaseAdornmentNode[StateType], Generic[StateType, MapNodeItemType])
                 subworkflow_event = map_node_event[1]
                 self._context._emit_subworkflow_event(subworkflow_event)
 
-                if subworkflow_event.name == "workflow.execution.initiated":
+                if (
+                    subworkflow_event.name == "workflow.execution.initiated"
+                    and subworkflow_event.workflow_definition == self.subworkflow
+                ):
                     for output_name in mapped_items.keys():
                         yield BaseOutput(name=output_name, delta=(None, index, "INITIATED"))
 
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -104,19 +107,36 @@ def run(self) -> Iterator[BaseOutput]:
...
 
-                elif subworkflow_event.name == "workflow.execution.fulfilled":
-                    workflow_output_vars = vars(subworkflow_event.outputs)
-
-                    for output_name in workflow_output_vars:
-                        output_mapped_items = mapped_items[output_name]
-                        output_mapped_items[index] = workflow_output_vars[output_name]
+                elif (
+                    subworkflow_event.name == "workflow.execution.fulfilled"
+                    and subworkflow_event.workflow_definition == self.subworkflow
```
dvargasfuertes, line 119: this was the fix

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -104,19 +107,36 @@ def run(self) -> Iterator[BaseOutput]:
...
+                ):
+                    for output_reference, output_value in subworkflow_event.outputs:
```
dvargasfuertes, line 121: the other changes here were unrelated, but should build some more resilience and avoid use of `vars` since our outputs class is iterable

vincent0426 (approved):

PR #1120, opened 2025-03-09
Fix bug related to nested subworkflows inside a subworkflow node

dvargas92495:
Continuing the work from here: https://github.com/vellum-ai/vellum-python-sdks/pull/1119, but now for inline subworkflow nodes. Last PR in the series will be retry node


vincent0426:
🔥

PR #1121, opened 2025-03-10
[WRK-4] Add support for adornments in inline subworkflow nodes

vincent0426:
Retry Nodes fail to generate when wrapping Inline Subworkflows

test workflow: 3224375e-e94c-459d-b97c-5b70c5fc5fce

Screenshot 2025-03-09 at 6 30 27 PM
Screenshot 2025-03-09 at 6 30 35 PM

### DIFF
--- a/ee/codegen/src/serializers/vellum.ts
+++ b/ee/codegen/src/serializers/vellum.ts
@@ -752,27 +752,6 @@ export declare namespace SubworkflowNodeDataSerializer {
     | DeploymentSubworkflowNodeDataSerializer.Raw;
 }
 
-export const SubworkflowNodeSerializer: ObjectSchema<
-  SubworkflowNodeSerializer.Raw,
-  Omit<SubworkflowNode, "type">
-> = objectSchema({
-  id: stringSchema(),
-  data: SubworkflowNodeDataSerializer,
-  inputs: listSchema(NodeInputSerializer),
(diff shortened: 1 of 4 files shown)


ee/codegen/src/serializers/vellum.ts
```diff
@@ -1938,6 +1917,28 @@ export declare namespace PromptNodeSerializer {
...
+  id: stringSchema(),
+  data: SubworkflowNodeDataSerializer,
+  inputs: listSchema(NodeInputSerializer),
+  displayData: propertySchema(
+    "display_data",
+    NodeDisplayDataSerializer.optional()
+  ),
+  base: CodeResourceDefinitionSerializer.optional(),
+  definition: CodeResourceDefinitionSerializer.optional(),
+  adornments: listSchema(AdornmentNodeSerializer).optional(),
```
dvargasfuertes, line 1933: [nit] Oh was the fix just adding the adornments serializer? let's add it to all of the nodes then (no need for tests for the rest of them)

dvargas92495 (approved):

ee/codegen/src/serializers/vellum.ts
vincent0426: ah yeah, I think that's the only change that generates the correct test; I will do it on the next PR since it may need some reordering!

PR #1122, opened 2025-03-10
Add adornments support for various workflow node serializers

vincent0426:
follow up: https://github.com/vellum-ai/vellum-python-sdks/pull/1121

Add adornment field for all node serializer
- reorder a lot

I think some may not need the adornment field? e.g., `EntrypointNodeSerializer`, `FinalOutputNodeSerializer`, `NoteNodeSerializer`, `ErrorNodeSerializer` though I have added for all of them now

### DIFF
--- a/ee/codegen/src/serializers/vellum.ts
+++ b/ee/codegen/src/serializers/vellum.ts
@@ -638,6 +638,201 @@ export declare namespace NodeDisplayDataSerializer {
   }
 }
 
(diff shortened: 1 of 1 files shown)


dvargas92495 (approved):

dvargas92495:
yes you're correct that we don't strictly need them for those nodes, but if it doesn't hurt, adding it helps us normalize them all to base node's in the future

PR #1126, opened 2025-03-10
Custom class comments

vincent0426:
Init attempt for custom class, since docs accept a string type, I think we couldn't simply customize its writer instead overriding the whole class_, but this may introduce more issues 

there are two errors now:
- comments will now have a new line which may be caused by ` writer.newLine();` under `if (this.docs) { `. However, if we remove that one, the children will just concat after """
- there is a white space for single line comment `"""\\"Hello" "World\\" """`

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -237,3 +237,32 @@ class Workflow(BaseWorkflow):
     graph = TemplatingNode
 "
 `;
+
+exports[`WorkflowProjectGenerator > should escape special characters > should correctly handle multi-line comments 1`] = `
+"from vellum.workflows.nodes.displayable import TemplatingNode as BaseTemplatingNode
+from vellum.workflows.state import BaseState
+
+
+class TemplatingNode(BaseTemplatingNode[BaseState, str]):
+    """
+    "Hello"
+    "World"
+    """
+
+    template = """test template"""
+    inputs = {}
+"
+`;
+
+exports[`WorkflowProjectGenerator > should escape special characters > should correctly handle node comments with quotes 1`] = `
+"from vellum.workflows.nodes.displayable import TemplatingNode as BaseTemplatingNode
+from vellum.workflows.state import BaseState
+
+
+class TemplatingNode(BaseTemplatingNode[BaseState, str]):
+    """\\"Hello" "World\\" """
+
+    template = """test template"""
+    inputs = {}
+"
+`;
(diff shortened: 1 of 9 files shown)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -237,3 +237,32 @@ class Workflow(BaseWorkflow):
...
+"from vellum.workflows.nodes.displayable import TemplatingNode as BaseTemplatingNode
+from vellum.workflows.state import BaseState
+
+
+class TemplatingNode(BaseTemplatingNode[BaseState, str]):
+    """
+    "Hello"
+    "World"
+    """
+
```
vincent0426, line 251: here's an additional line

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -237,3 +237,32 @@ class Workflow(BaseWorkflow):
...
+`;
+
+exports[`WorkflowProjectGenerator > should escape special characters > should correctly handle node comments with quotes 1`] = `
+"from vellum.workflows.nodes.displayable import TemplatingNode as BaseTemplatingNode
+from vellum.workflows.state import BaseState
+
+
+class TemplatingNode(BaseTemplatingNode[BaseState, str]):
+    """\\"Hello" "World\\" """
+
```
vincent0426, line 264: and here

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -237,3 +237,32 @@ class Workflow(BaseWorkflow):
...
+"
+`;
+
+exports[`WorkflowProjectGenerator > should escape special characters > should correctly handle node comments with quotes 1`] = `
+"from vellum.workflows.nodes.displayable import TemplatingNode as BaseTemplatingNode
+from vellum.workflows.state import BaseState
+
+
+class TemplatingNode(BaseTemplatingNode[BaseState, str]):
+    """\\"Hello" "World\\" """
```
vincent0426, line 263: an additional white space not sure where it comes

ee/codegen/src/generators/extensions/custom-class.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    writer.write(":");
+    writer.newLine();
+    writer.indent();
+
+    if (this.docs) {
+      const comment = new CustomComment({ docs: this.docs });
+      comment.write(writer);
+      writer.newLine();
```
vincent0426, line 76: if we remove this newLine, comments will concat with children, e.g., """comment"""template =

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -337,7 +338,7 @@ export abstract class BaseNode<
       });
     }
 
-    const nodeClass = python.class_({
+    const nodeClass = customPython.class_({
       name: nodeContext.nodeClassName,
       extends_: [nodeBaseClass],
       docs: this.generateNodeComment(),
```
vincent0426, line 344: since `docs` accepts a string, we need to override the whole python.class_

vincent0426:
fixed here: https://github.com/vellum-ai/vellum-python-sdks/pull/1139

PR #1128, opened 2025-03-10
[APO-7] Handle empty string to list type conversion in templating node

vincent0426:
cast empty string into empty array

### DIFF
--- a/src/vellum/workflows/nodes/utils.py
+++ b/src/vellum/workflows/nodes/utils.py
@@ -95,6 +95,10 @@ def parse_type_from_str(result_as_str: str, output_type: Any) -> Any:
         return bool(result_as_str)
 
     if get_origin(output_type) is list:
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/utils.py
```diff
@@ -95,6 +95,10 @@ def parse_type_from_str(result_as_str: str, output_type: Any) -> Any:
         return bool(result_as_str)
 
     if get_origin(output_type) is list:
+        # Handle empty string case for list types by returning an empty list
+        if not result_as_str.strip():
+            return []
+
         try:
             data = json.loads(result_as_str)
```
dvargasfuertes, line 103: [oos] as a followup, could we also make invalid json result in a `NodeException` instead of a `ValueError`, the latter of which hits us in sentry?

dvargas92495 (approved):

PR #1130, opened 2025-03-11
Consolidate all of the base display wraps

dvargas92495:
In this PR, we consolidate all of the `Base[Adornment]Display.wrap` classes to a single method. This will allow us a centralized place for future adornment changes that we need for event monitoring across all adornments.

We also clean up some codegen for Adornments. Attribute values are only needed for the node classes themselves, not the adornment classes

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/base_adornment_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/base_adornment_node.py
@@ -1,9 +1,13 @@
+import re
+import types
 from uuid import UUID
-from typing import Any, Callable, Generic, Optional, TypeVar, cast
+from typing import Any, Callable, Generic, Optional, Type, TypeVar, cast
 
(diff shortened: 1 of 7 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/base_adornment_node.py
```diff
@@ -37,3 +41,46 @@ def serialize(
         serialized_wrapped_node["adornments"] = adornments + [adornment] if adornment else adornments
 
         return serialized_wrapped_node
+
+    @classmethod
+    def wrap(cls, **kwargs: Any) -> Callable[..., Type[BaseNodeDisplay]]:
```
dvargasfuertes, line 46: We lose a mypy benefit here, which is fine since:
- Display classes are codegen only for now
- We are deferring all mypy improvements to at least Q3

PR #1132, opened 2025-03-11
Reduce variant union type

vincent0426:
Variant is redundant here? since they are also defined under data type


dvargas92495 (approved):
yup

PR #1133, opened 2025-03-11
Return mocked flag from SDK

dvargas92495:
https://www.loom.com/share/4aa5596ba06e4ed9a3e6966189387cdd

### DIFF
--- a/src/vellum/workflows/events/tests/test_event.py
+++ b/src/vellum/workflows/events/tests/test_event.py
@@ -328,6 +328,7 @@ mock_node_uuid = str(uuid4_from_hash(MockNode.__qualname__))
                             "name": "default",
                         }
                     ],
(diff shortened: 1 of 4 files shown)


marlenc3 (approved):

src/vellum/workflows/events/tests/test_event.py
```diff
@@ -328,6 +328,7 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
                             "name": "default",
                         }
                     ],
+                    "mocked": None,
```
dvargasfuertes, line 331: We should clean this up and not have it appear on serialization: https://linear.app/vellum/issue/APO-146/skip-default-values-in-event-serialization-like-mocked-none

vincent0426:
✅

PR #1134, opened 2025-03-11
Filter out default values for codegen adornment

vincent0426:
I think this needs to be done case by case as we don't have knowledge whats the default value of each adornment in codegen?

### DIFF
--- a/ee/codegen/src/generators/nodes/bases/base.ts
+++ b/ee/codegen/src/generators/nodes/bases/base.ts
@@ -294,6 +294,17 @@ export abstract class BaseNode<
         continue;
       }
 
+      // Filter out attributes that match their default values
+      const filteredAttributes = adornment.attributes.filter((attr) => {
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -294,6 +294,17 @@ export abstract class BaseNode<
         continue;
       }
 
+      // Filter out attributes that match their default values
+      const filteredAttributes = adornment.attributes.filter((attr) => {
+        if (adornment.base.name === "RetryNode") {
```
dvargasfuertes: [nit] let's create a mapping in `ee/codegen/src/generators/nodes/defaults.ts` or `ee/codegen/src/generators/nodes/constants.ts` where we start tracking these

dvargas92495 (approved):

dvargas92495 (approved):

PR #1136, opened 2025-03-11
Allow users to use the client SDK with pydantic v1

dvargas92495:
Unfortunately this is tough to test pre-release. [tox](https://vellum-ai.slack.com/archives/C06P13ABK0W/p1741716824246289) may be a compelling option for us in that regard long term.

For now, this is a quick fix to unblock pydantic v1 customers


marlenc3:
✅

vincent0426:
✅

PR #1139, opened 2025-03-11
[APO-108] Add post process to temporary handle docstring escape issue

### LINKED ISSUE #16640: Unexpected space in docstring near escaped quote
### Description

Ruff formatter modifies the spacing in a string within a docstring when the string contains an escaped quote. Specifically, it adds an unwanted space between the escaped quote and the closing triple quotes of the docstring.

### Steps to Reproduce

1. Create a Python class with the following content:
   ```python
   class Sample:
       """Hello "World\""""
   
       def __init__(self, name):
           self.name = name

2. Run `ruff format` on the file
3. Observe that the formatter modifies the docstring to:
    ```python
    class Sample:
        """Hello "World\" """
    
        def __init__(self, name):
            self.name = name

### Expected Behavior
Docstring remain unchanged after formatting as its escaped already

`"""Hello "World\""""`

### Version

ruff 0.9.10 (0dfa810e9 2025-03-07)

vincent0426:
linear: https://linear.app/vellum/issue/APO-108/node-comments-that-end-in-a-fail-to-generate-artifact-due-to-invalid

revisit when upstream fixed https://github.com/astral-sh/ruff/issues/16640

Handle extra space added by ruff formatted by post process

https://github.com/user-attachments/assets/51a95da0-fc88-4c7d-b1ef-56dab74cdcb4

### DIFF
--- a/ee/codegen/src/generators/base-persisted-file.ts
+++ b/ee/codegen/src/generators/base-persisted-file.ts
@@ -218,7 +218,7 @@ export abstract class BasePersistedFile extends AstNode {
 
   private postprocessDocstrings(content: string): string {
     // Find an escaped quote followed by space(s) right before closing triple quotes
-    const fixPattern = /(\\")(\s+)(""")/g;
+    const fixPattern = /(\\")(\s)(""")/g;
 
     // Remove the extra space between escaped quote and closing triple quotes
     return content.replace(fixPattern, "$1$3");
--- a/ee/codegen/src/generators/base-persisted-file.ts
+++ b/ee/codegen/src/generators/base-persisted-file.ts
@@ -207,11 +207,20 @@ export abstract class BasePersistedFile extends AstNode {
     let contents: string;
     try {
       contents = await writer.toStringFormatted({ line_width: 120 });
+      contents = this.postprocessDocstrings(contents);
     } catch (error) {
       console.error("Error formatting", fileName, "with error", error);
       contents = writer.toString();
     }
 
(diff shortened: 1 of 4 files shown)


ee/codegen/src/generators/base-persisted-file.ts
```diff
@@ -207,11 +207,20 @@ export abstract class BasePersistedFile extends AstNode {
...
       console.error("Error formatting", fileName, "with error", error);
       contents = writer.toString();
     }
 
     await writeFile(filepath, contents);
   }
+
+  private postprocessDocstrings(content: string): string {
+    // Find an escaped quote followed by space(s) right before closing triple quotes
+    const fixPattern = /(\\")(\s+)(""")/g;
```
dvargasfuertes: [nit] I would only do `\s` instead of `\s+`, to really hypertarget the formatting case we care about
[nit] Could we add a link to your github issue?
[oos] If a user intentionally added a space to the end of a block string, the current impl wipes it. I think this is the lesser evil than the alternative at least until ruff patches. I think in the meantime, I think we could use `writer.toString();` to detect if there even is a `""""` sequence in the unformatted string, and skip the method if so

dvargas92495 (approved):

ee/codegen/src/generators/base-persisted-file.ts
vincent0426: > Could we add a link to your github issue?

ah I think its linked in description already!

PR #1141, opened 2025-03-12
[APO-130] Add support for ParseJsonExpression in workflow serialization

vincent0426:
linear: https://linear.app/vellum/issue/APO-130/serialization-for-parse-json-expression

Add serialization for `parse_json`

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -326,6 +326,7 @@ class BaseNodeDisplay(Generic[NodeType], metaclass=BaseNodeDisplayMeta):
                 IsNotNilExpression,
                 IsUndefinedExpression,
                 IsNotUndefinedExpression,
+                ParseJsonExpression,
             ),
         ):
             lhs = self.serialize_value(display_context, condition._expression)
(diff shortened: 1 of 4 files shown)


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -404,7 +404,11 @@ def serialize_value(self, display_context: "WorkflowDisplayContext", value: Base
             }
 
         if isinstance(value, ParseJsonExpression):
-            raise ValueError("ParseJsonExpression is not supported in the UI")
+            json_string = self.serialize_value(display_context, value._expression)
+            return {
+                "type": "PARSE_JSON",
+                "json_string": json_string,
```
dvargasfuertes: [blocking] as discussed, we'll want to model this as a `"UNARY_EXPRESSION"` so that the frontend can you the same primitives it's using to represent things like `notNull`. both are functions from `(T) -> O`, where `parse_json` is `String -> JSON` and `notNull` is `Any -> Boolean`

ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_ports_serialization.py
```diff
@@ -1003,3 +1003,70 @@ def test_serialize_node__or_then_and(serialize_node):
         serialized_node,
         ignore_order=True,
     )
+
+
+class ParseJsonGenericNode(BaseNode):
```
dvargasfuertes, line 1008: [nit] I like to keep single test use nodes within the body of the tests

dvargas92495 (approved):

ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_ports_serialization.py
vincent0426: ah I will refactor in the next PR, the whole file needs to change

PR #1142, opened 2025-03-12
Fix Templating node event input display

dvargas92495:
This PR started as a way to fix node event input display, but I realized unfortunately that our display classes currently generate:
```python
node_input_ids_by_name = {
    "foo": UUID
}
```

but our templating node events will emit:
```python
node_input_ids_by_name = {
    "inputs.foo": UUID
}
```

The latter is more consistent with "attribute paths" - meaning it will be more extensible to dynamic attributes at any level. So in the meantime, this PR just supports generating input_display when there's no display classes defined, and we'll update vellum side handling to do both


vincent0426:
✅

PR #1144, opened 2025-03-12
[APO-181] Add default api method for api node

vincent0426:
linear: https://linear.app/vellum/issue/APO-181/investigate-api-node-server-error-with-sdk-enabled

Default to GET method if not provided


dvargas92495 (approved):
Nice

NgoHarrison (commented):
✅

PR #1145, opened 2025-03-12
Allow node id annotation of the inner wrapped node in adornment setups

dvargas92495:
~Gonna block this guy on this one first: https://github.com/vellum-ai/vellum-python-sdks/pull/1130~

Before this PR, we were returning:
```js
display context = {
    nodes: {
        inner_node_id: {
            nodes: {
                outer_node_id
            }
        }
    }
}
```

This unblocked our monitoring in the short term, but now prevents us from showing separate traces invoked by an adornment. By putting it in the correct order, we can now treat adornments as any other subworkflow paradigm.

We'll need to update vellum-side logic before this goes live with version 0.14.15


vincent0426:
✅

PR #1146, opened 2025-03-12
upgrade pyyaml (Closes #1143)

### LINKED ISSUE #1143: pyyaml version
Got this error today trying to deploy. 
We'd ideally not downgrade pyyaml just for vellum. 

Is there a reason it needs to be that exact version, or should that be a minimum?

The conflict is caused by:
    The user requested pyyaml==6.0.2
    vellum-ai 0.14.13 depends on pyyaml==6.0.1

dvargas92495:
We need to remember to apply this to the client generator too


vincent0426:
✅

PR #1152, opened 2025-03-13
Support pydantic <2.5

dvargas92495:
`SchemaKind` was introduced later than some of our users' versions, and we're not really using it. So just redefining inline.

NOTE: I was trying to repro this in our prototype repo locally and was unable to get my venv to work. I'd encourage whoever takes this over the line to manually ensure none of the other pydantic imports in this file will cause problems

### DIFF
--- a/src/vellum/plugins/pydantic.py
+++ b/src/vellum/plugins/pydantic.py
@@ -1,16 +1,19 @@
 from functools import lru_cache
 from typing import Any, Dict, Literal, Optional, Tuple, Union
+from typing_extensions import TypeAlias
 
 from pydantic.fields import FieldInfo
 from pydantic.plugin import (
     PydanticPluginProtocol,
-    SchemaKind,
     SchemaTypePath,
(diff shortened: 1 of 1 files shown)


src/vellum/plugins/pydantic.py
```diff
@@ -81,7 +91,7 @@ def new_schema_validator(
         self,
         schema: CoreSchema,
         schema_type: Any,
-        schema_type_path: SchemaTypePath,
+        schema_type_path: SchemaTypePath,  # type: ignore
```
NgoHarrison, line 94: Mypy was complaining about `Argument 3 of "new_schema_validator" is incompatible with supertype "PydanticPluginProtocol"; supertype defines the argument type as "SchemaTypePath"  [override]`. I think it is confused since the super class is defining the same class at its level and we redefine that class here. Should be fine since this is the same class definition and struct

NgoHarrison (commented):
✅

PR #1155, opened 2025-03-13
Add docker mock client for test image push

vincent0426:
We might need to mock docker client? 

which will fail with the following error before this PR
```sh
>       assert result.exit_code == 0, result.output
E       AssertionError: Self hosted install detected...
E         Pre-validating image...
E         
E       assert 1 == 0
E        +  where 1 = <Result DockerException("Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))")>.exit_code
```

### DIFF
--- a/ee/vellum_cli/tests/test_image_push.py
+++ b/ee/vellum_cli/tests/test_image_push.py
@@ -1,5 +1,5 @@
 import subprocess
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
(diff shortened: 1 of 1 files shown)


m-abboud (approved):
Oh oops my bad! Thank you! Weird CI didn't catch this on my PR 🤔

marlenc3:
✅

PR #1156, opened 2025-03-13
[APO-1] Add none log output ID handling for CodeExecutionNode

vincent0426:
linear: https://linear.app/vellum/issue/APO-1/try-node-wrapped-code-node-display-is-code-genning-none-for-log-output

Resilience for `none` in log output id

### DIFF
--- a/ee/codegen/src/__test__/nodes/code-execution-node.test.ts
+++ b/ee/codegen/src/__test__/nodes/code-execution-node.test.ts
@@ -281,7 +281,7 @@ describe("CodeExecutionNode", () => {
     );
   });
   describe("log output id", () => {
-    it("should not generate log output id", async () => {
+    it("should not generate log output id if not given", async () => {
       const nodeData = codeExecutionNodeFactory();
 
       // GIVEN a node without a log output id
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/nodes/code-execution-node.test.ts
```diff
@@ -280,4 +280,25 @@ describe("CodeExecutionNode", () => {
       }
     );
   });
+  describe("log output id", () => {
+    it("should not generate log output id", async () => {
```
NgoHarrison: [nit] `it("should not generate log output id if not given")`

ee/codegen/src/__test__/nodes/code-execution-node.test.ts
```diff
@@ -280,4 +280,25 @@ describe("CodeExecutionNode", () => {
...
+      const nodeData = codeExecutionNodeFactory();
+
+      // GIVEN a node without a log output id
+      nodeData.data.logOutputId = undefined;
```
NgoHarrison: [nit] Generally we have been passing / constructing these tests by adding these attributes to the factory class and then using that object. Maybe we should do it here also? 1/5 on this

vincent0426: ah yeah I try to make them similar, however, its hard to differentiate passing in undefined explicitly or just not passing in (this should be the default case or we may need to specify log output for other test)

NgoHarrison: Can we not use a default of true for the log output and if that default is true, define it with the hard coded id otherwise it's undefined?

NgoHarrison (approved):

dvargas92495 (commented):
✅

PR #1160, opened 2025-03-14
[APO-8] Resile input validation

vincent0426:
workflow initialization should not trigger sentry alert

### DIFF
--- a/src/vellum/workflows/inputs/base.py
+++ b/src/vellum/workflows/inputs/base.py
@@ -5,7 +5,7 @@ from pydantic import GetCoreSchemaHandler
 from pydantic_core import core_schema
 
 from vellum.workflows.errors.types import WorkflowErrorCode
-from vellum.workflows.exceptions import WorkflowInitializationException
+from vellum.workflows.exceptions import NodeException
 from vellum.workflows.references import ExternalInputReference, WorkflowInputReference
 from vellum.workflows.references.input import InputReference
 from vellum.workflows.types.utils import get_class_attr_names, infer_types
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/inputs/base.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, **kwargs: Any) -> None:
                 origin = get_origin(field_type)
                 args = get_args(field_type)
                 if not (origin is Union and type(None) in args):
-                    raise WorkflowInitializationException(
+                    raise NodeException(
```
dvargasfuertes: [blocking] This is not an exception raised by a node, but rather by the Workflow, so we don't want to raise a Node exception. Rather, we want the consumer of the sdk to not trigger a sentry alert when we detect workflow initialization exceptions

src/vellum/workflows/runner/runner.py
```diff
@@ -332,6 +332,18 @@ def initiate_node_streaming_output(output: BaseOutput) -> None:
...
+                NodeExecutionRejectedEvent(
+                    trace_id=node.state.meta.trace_id,
+                    span_id=span_id,
+                    body=NodeExecutionRejectedBody(
+                        node_definition=node.__class__,
+                        error=e.error,
+                    ),
+                    parent=parent_context,
+                )
+            )
```
dvargasfuertes, line 346: [oos] Is this for the case of subworkflows? I would actually expect the subworkflow nodes to catch and reraise as a node exception. Could do as a followup for now and depri

dvargas92495 (approved):

PR #1161, opened 2025-03-14
Harrison/upgrade vellum version

NgoHarrison:
The original intent was to bump the vellum version to `0.14.16` for new `parse_json` logical operator. However, we now need to add support for new `Document` vellum type. I have created this ticket https://linear.app/vellum/issue/APO-189/add-codegen-support-for-new-document-variable-type where we can fast follow. I think my changes should be fine if we don't currently have active usage on this field but if we do, then I would add that support in this PR so nothing breaks

### DIFF
--- a/ee/codegen/src/generators/chat-message-content.ts
+++ b/ee/codegen/src/generators/chat-message-content.ts
@@ -289,7 +289,7 @@ export namespace ChatMessageContent {
 }
 
 export class ChatMessageContent extends AstNode {
-  private astNode: AstNode;
+  private astNode: AstNode | undefined;
 
   public constructor({
     chatMessageContent,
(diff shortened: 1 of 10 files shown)


ee/codegen/src/generators/chat-message-content.ts
```diff
@@ -342,6 +342,10 @@ export class ChatMessageContent extends AstNode {
         );
         break;
       }
+      // TODO: Implement DocumentMessageContent support
+      // https://linear.app/vellum/issue/APO-189/add-codegen-support-for-new-document-variable-type
+      case "DOCUMENT":
+        break;
```
noanflaherty, line 348: It is indeed possible for users to provide a Document to a Workflow as part of a Chat History variable. While it doesn't have to be part of this PR, we should probably follow up with support asap. It's schema very closely follows that of IMAGE and AUDIO

noanflaherty (approved):

PR #1163, opened 2025-03-14
[APO-193] Serialization support for vellum document value

NgoHarrison:


src/vellum/workflows/nodes/experimental/openai_chat_completion_node/node.py
```diff
@@ -251,6 +252,9 @@ def _create_message_content(
                 }
 
                 content.append(audio_content_item)
+            elif block.content.type == "DOCUMENT":
+                # TODO: Fill out
+                break
```
NgoHarrison, line 257: Currently, OpenAI does not support Document as a completion type:
https://platform.openai.com/docs/api-reference/chat

We can try extracting the pdf and sending it as a text source though but not sure how much of a lift it is to. I am mostly thinking about the work around fetching these docs from a url and the work needed there. Thoughts?
cc @noanflaherty @dvargas92495

src/vellum/workflows/nodes/experimental/openai_chat_completion_node/node.py
```diff
@@ -251,6 +252,9 @@ def _create_message_content(
...
             else:
```
NgoHarrison: Currently, OpenAI does not support Document as a completion type:
https://platform.openai.com/docs/api-reference/chat

We can try extracting the pdf and sending it as a text source tho but not sure how much of a lift it is to do. I am mostly thinking about the work around fetching from a url and the work needed there. I think the parsing of the content should be straight forward tho. Thoughts?
cc @noanflaherty @dvargas92495

src/vellum/workflows/nodes/experimental/openai_chat_completion_node/node.py
```diff
@@ -202,7 +203,7 @@ def _create_message_content(
                 json_content_item: ChatCompletionContentPartTextParam = {"type": "text", "text": json.dumps(json_value)}
                 content.append(json_content_item)
             elif block.content.type == "IMAGE":
-                image_value = cast(VellumAudio, block.content.value)
+                image_value = cast(VellumImage, block.content.value)
```
NgoHarrison, line 206: Seems like a typo from before

src/vellum/workflows/nodes/experimental/openai_chat_completion_node/node.py
noanflaherty: Yeah let's just raise a Node Exception in this case

noanflaherty (approved):

dvargas92495 (commented):
✅

PR #1165, opened 2025-03-15
[APO-6] Add prompt block input validation

vincent0426:
linear: https://linear.app/vellum/issue/APO-6/invalid-key-error-on-execute-prompt-endpoint

This PR should handle `KeyError: 'some_key'` from sentry

- Validate prompt inputs have all the keys that blocks defined


src/vellum/workflows/nodes/displayable/bases/base_prompt_node/node.py
```diff
@@ -38,6 +42,7 @@ def run(self) -> Iterator[BaseOutput]:
 
     def _process_prompt_event_stream(self) -> Generator[BaseOutput, None, Optional[List[PromptOutput]]]:
         try:
+            self._validate()
```
dvargasfuertes: [nit] no need to have this nested within the try/except, which we should usually try to keep as targeted as possible

src/vellum/workflows/nodes/displayable/prompt_deployment_node/node.py
```diff
@@ -37,6 +37,9 @@ class Outputs(BasePromptDeploymentNode.Outputs):
 
         text: str
 
+    def _validate(self) -> None:
+        pass
```
dvargasfuertes: [nit] no need to make this method abstract if downstream consumers are just going to skip anyway

src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/tests/test_inline_prompt_node.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # GIVEN a BaseInlinePromptNode
+    class TestNode(BaseInlinePromptNode):
+        ml_model = "test-model"
+        blocks = test_blocks
+        prompt_inputs = {}
+
+    # WHEN the _extract_required_input_variables method is called
+    node = TestNode()
+    required_vars = node._extract_required_input_variables(test_blocks)
```
dvargasfuertes: [nit] in general, we should not need to be testing private methods, since it makes it harder to update implementation by requiring updates to these tests.

Instead, we should be able to call run and our assertion on the exception should give us the same information as the assertion below. Same feedback for all of the tests below

src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
```diff
@@ -59,6 +60,32 @@ class BaseInlinePromptNode(BasePromptNode[StateType], Generic[StateType]):
     class Trigger(BasePromptNode.Trigger):
         merge_behavior = MergeBehavior.AWAIT_ANY
 
+    def _extract_required_input_variables(self, blocks: Union[List[PromptBlock], List[RichTextChildBlock]]) -> Set[str]:
+        required_variables = set()
+
+        for block in blocks:
+            if block.block_type == "VARIABLE":
+                if hasattr(block, "input_variable"):
```
dvargasfuertes: [nit] is this check not redundant?

src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
```diff
@@ -59,6 +60,32 @@ class BaseInlinePromptNode(BasePromptNode[StateType], Generic[StateType]):
...
+
+    def _validate(self) -> None:
+        required_variables = self._extract_required_input_variables(self.blocks)
+        provided_variables = set(self.prompt_inputs.keys() if self.prompt_inputs else set())
+
+        missing_variables = required_variables - provided_variables
+        if missing_variables:
+            missing_vars_str = ", ".join(f"'{var}'" for var in missing_variables)
+            raise NodeException(
+                message=f"Missing required input variables: {missing_vars_str}",
```
dvargasfuertes: [nit] I'd make this message specific to block. Something like:
`f"Missing inputs referenced by `VariableBlock`s: {missing_vars_str}"

tests/workflows/basic_inline_prompt_node_validation/tests/test_workflow.py
```diff
@@ -0,0 +1,37 @@
+from vellum.workflows.errors.types import WorkflowErrorCode
+
+from tests.workflows.basic_inline_prompt_node_validation.workflow import BasicInlinePromptWorkflow, WorkflowInputs
+
+
+def test_run_workflow__missing_required_input_variable():
```
dvargasfuertes: [nit] If we are adding unit tests above, then these workflow-level tests are not validating anything new. A good rule of thumb I notice is that if there are no changes to `workflows/runner.py`, then we probably don't need a workflow level test

dvargas92495 (approved):

dvargas92495 (approved):

PR #1166, opened 2025-03-17
Fix workflow generic node port name

vincent0426:
This PR solves the issue that workflow generic port name does not use the sanitized name but genric node does
- handle sanitization at context level (same as input/output)
- use the sanitized name when writing workflow

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
@@ -279,8 +279,8 @@ exports[`Workflow > graph > should handle a simple edge of generic nodes 1`] = `
 
 exports[`Workflow > graph > should properly format UUID-like port names in graph references 1`] = `
 "{
-    GenericNode.Ports._7d813792_04c2_4d26_a126_16de4c4cebaf >> FirstNode,
-    GenericNode.Ports._6f909cc6_887c_46d8_966b_46547900fc9c >> SecondNode,
+    GenericNode.Ports.port_7d813792_04c2_4d26_a126_16de4c4cebaf >> FirstNode,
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
```diff
@@ -279,8 +279,8 @@ exports[`Workflow > graph > should handle a simple edge of generic nodes 1`] = `
 
 exports[`Workflow > graph > should properly format UUID-like port names in graph references 1`] = `
 "{
-    GenericNode.Ports._7d813792_04c2_4d26_a126_16de4c4cebaf >> FirstNode,
-    GenericNode.Ports._6f909cc6_887c_46d8_966b_46547900fc9c >> SecondNode,
+    GenericNode.Ports.port_7d813792_04c2_4d26_a126_16de4c4cebaf >> FirstNode,
+    GenericNode.Ports.port_6f909cc6_887c_46d8_966b_46547900fc9c >> SecondNode,
```
vincent0426, line 283: this is the main change that workflow uses the sanitized name

ee/codegen/src/context/node-context/base.ts
```diff
@@ -129,38 +127,14 @@ export abstract class BaseNodeContext<T extends WorkflowDataNode> {
...
-        initialPortName.endsWith("_") ? "" : "_"
-      }${numRenameAttempts + 1}`;
-      numRenameAttempts += 1;
-    }
-
-    this.addUsedPortName(sanitizedName);
-
-    return sanitizedName;
-  }
-
```
vincent0426, line 163: move this to the port-context level (same as input/output variable)

dvargas92495 (approved):

PR #1167, opened 2025-03-18
Pass the adornment's id as that node's id

dvargas92495:
Uses the display class for our adornments to track the node id for that specific adornment. Allows for the stable node ids of adornments themselves


NgoHarrison (approved):

marlenc3:
✅

vincent0426:
✅

PR #1168, opened 2025-03-18
[APO-185] Add codegen support for handling json output for prompt nodes

NgoHarrison:
This is part 1 for supporting new json output on prompt nodes. I will do a follow up PR for the serialization side

### DIFF
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -557,8 +557,8 @@ export function inlinePromptNodeDataInlineVariantFactory({
       },
     ],
     // Add adornments if provided
-    ...(adornments && { adornments }),
-    ...(outputs && { outputs }),
+    adornments,
+    outputs,
   };
   return nodeData;
 }
--- a/ee/codegen/src/__test__/helpers/node-data-factories.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factories.ts
@@ -493,12 +493,14 @@ export function inlinePromptNodeDataInlineVariantFactory({
   parameters,
   defaultBlock,
   adornments,
+  outputs,
 }: {
   blockType?: string;
   errorOutputId?: string;
   parameters?: PromptParameters;
   defaultBlock?: PromptTemplateBlock;
   adornments?: AdornmentNode[];
+  outputs?: NodeOutput[];
 }): PromptNode {
   const block = defaultBlock ?? generateBlockGivenType(blockType ?? "JINJA");
   const nodeData: PromptNode = {
@@ -556,6 +558,7 @@ export function inlinePromptNodeDataInlineVariantFactory({
     ],
     // Add adornments if provided
     ...(adornments && { adornments }),
+    ...(outputs && { outputs }),
   };
   return nodeData;
 }
--- a/ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
@@ -1266,32 +1266,3 @@ class PromptNodeDisplay(BaseInlinePromptNodeDisplay[PromptNode]):
     )
 "
 `;
-
-exports[`InlinePromptNode > with json output id defined > getNodeFile 1`] = `
-"from vellum import JinjaPromptBlock, PromptParameters
-from vellum.workflows.nodes.displayable import InlinePromptNode
-
-from ..inputs import Inputs
-
-
-class PromptNode(InlinePromptNode):
-    ml_model = "gpt-4o-mini"
-    blocks = [
-        JinjaPromptBlock(template="""Summarize what this means {{ INPUT_VARIABLE }}"""),
-    ]
-    prompt_inputs = {
-        "text": Inputs.text,
-    }
-    parameters = PromptParameters(
-        stop=[],
-        temperature=0,
-        max_tokens=1000,
-        top_p=1,
-        top_k=0,
-        frequency_penalty=0,
-        presence_penalty=0,
-        logit_bias={},
-        custom_parameters={},
-    )
-"
-`;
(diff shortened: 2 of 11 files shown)


ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -556,6 +558,7 @@ export function inlinePromptNodeDataInlineVariantFactory({
     ],
     // Add adornments if provided
     ...(adornments && { adornments }),
+    ...(outputs && { outputs }),
```
dvargasfuertes: [nit] I'm _pretty_ sure this is the same as just:
```typescript
adornments,
outputs,
```

ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
```diff
@@ -1223,3 +1223,73 @@ class PromptNode(InlinePromptNode):
...
+
+
+class PromptNodeDisplay(BaseInlinePromptNodeDisplay[PromptNode]):
+    label = "Prompt Node"
+    node_id = UUID("7e09927b-6d6f-4829-92c9-54e66bdcaf80")
+    output_id = UUID("2d4f1826-de75-499a-8f84-0a690c8136ad")
+    array_output_id = UUID("771c6fba-5b4a-4092-9d52-693242d7b92c")
+    target_handle_id = UUID("3feb7e71-ec63-4d58-82ba-c3df829a2948")
+    prompt_input_ids_by_name = {"text": UUID("7b8af68b-cf60-4fca-9c57-868042b5b616")}
+    json_output_id = UUID("some-json-output-id")
```
dvargasfuertes: [blocking] instead of a new PromptNodeDisplay-specific attribute, we want to progressively get more and more generic to one day treat all of these nodes as generic nodes.

so I would expect a new entry in `output_display` instead of this field here

ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
```diff
@@ -1223,3 +1223,73 @@ class PromptNode(InlinePromptNode):
...
+            id=UUID("dd8397b1-5a41-4fa0-8c24-e5dffee4fb98")
+        )
+    }
+    display_data = NodeDisplayData(
+        position=NodeDisplayPosition(x=0, y=0), width=None, height=None
+    )
+"
+`;
+
+exports[`InlinePromptNode > with json output id defined > getNodeFile 1`] = `
```
dvargasfuertes: [nit] we had no changes to inline prompt node impl, so this test/snapshot doesn't tell us anything new, which we could remove.

Instead, we should have a test of another node _referencing_ the new json output

dvargasfuertes: oh you have the second one below, nice! we could remove this one outright then

ee/codegen/src/__test__/nodes/multiple-nodes.test.ts
```diff
@@ -362,3 +364,84 @@ describe("Non-existent Subworkflow Deployment Node referenced by Templating Node
...
+      workflowContext,
+      nodeData: promptNode,
+    })) as InlinePromptNodeContext;
+
+    const templatingNode = templatingNodeFactory({
+      id: "46e221ab-a749-41a2-9242-b1f5bf31f3a5",
+      sourceHandleId: "6ee2c814-d0a5-4ec9-83b6-45156e2f22c4",
+      targetHandleId: "3960c8e1-9baa-4b9c-991d-e399d16a45aa",
+      inputs: [
+        {
```
dvargasfuertes: [nit] we have a `nodeInputFactory` to make some of this easier

ee/codegen/src/__test__/nodes/multiple-nodes.test.ts
```diff
@@ -362,3 +364,84 @@ describe("Non-existent Subworkflow Deployment Node referenced by Templating Node
...
+
+    (await createNodeContext({
+      workflowContext,
+      nodeData: promptNode,
+    })) as InlinePromptNodeContext;
+
+    const templatingNode = templatingNodeFactory({
+      id: "46e221ab-a749-41a2-9242-b1f5bf31f3a5",
+      sourceHandleId: "6ee2c814-d0a5-4ec9-83b6-45156e2f22c4",
+      targetHandleId: "3960c8e1-9baa-4b9c-991d-e399d16a45aa",
```
dvargasfuertes: [nit] I think we could omit these

ee/codegen/src/context/node-context/inline-prompt-node.ts
```diff
@@ -12,12 +12,17 @@ export class InlinePromptNodeContext extends BaseNodeContext<InlinePromptNodeTyp
   baseNodeDisplayClassName = "BaseInlinePromptNodeDisplay";
 
   protected getNodeOutputNamesById(): Record<string, string> {
+    const jsonOutput = this.nodeData.outputs?.find(
```
dvargasfuertes, line 15: [oos] long term, when we're done normalizing all of our nodes, we'll be able to remove all of these methods from the node class specific contexts and just return `this.nodeData.outputs.map((output) => [output.id, output.name])`

src/vellum/workflows/nodes/displayable/inline_prompt_node/node.py
```diff
@@ -30,9 +30,11 @@ class Outputs(BaseInlinePromptNode.Outputs):
         The outputs of the InlinePromptNode.
 
         text: str - The result of the Prompt Execution
+        json: Optional[Dict[Any, Any]] - The result of the Prompt Execution in JSON format
```
dvargasfuertes, line 34: [nit] I think we want to start using `undefined` instead of `None` to represent "this output is missing"

dvargas92495 (commented):
Nice! only blocking comment is the display class attribute

ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
NgoHarrison: Deleted

ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
NgoHarrison: Updated

dvargas92495 (approved):
🔥

PR #1170, opened 2025-03-18
[APO-125] Fix merge all behavior

vincent0426:
Fix merge all conditions to make sure every deps have invoked the `MERGE_ALL` node

### DIFF
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -297,10 +297,8 @@ class BaseNode(Generic[StateType], metaclass=BaseNodeMeta):
                 when all of its dependencies have invoked this node.
                 """
                 # Check if all dependencies have invoked this node
-                all_deps_invoked = all(
-                    dep in state.meta.node_execution_cache._dependencies_invoked.get(node_span_id, set())
-                    for dep in dependencies
-                )
+                dependencies_invoked = state.meta.node_execution_cache._dependencies_invoked.get(node_span_id, set())
+                all_deps_invoked = all(dep in dependencies_invoked for dep in dependencies)
 
                 return all_deps_invoked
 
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/nodes/bases/base.py
```diff
@@ -293,15 +293,17 @@ def should_initiate(
...
+                when all of its dependencies have invoked this node.
                 """
-                current_node_execution_count = state.meta.node_execution_cache.get_execution_count(cls.node_class)
-                return all(
-                    state.meta.node_execution_cache.get_execution_count(dep) == current_node_execution_count + 1
+                # Check if all dependencies have invoked this node
+                all_deps_invoked = all(
+                    dep in state.meta.node_execution_cache._dependencies_invoked.get(node_span_id, set())
                     for dep in dependencies
                 )
```
vincent0426: returns `True` only if every dependency has directly triggered this execution

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -293,15 +293,17 @@ def should_initiate(
...
-                when all of its dependencies have been executed N times.
+                A node utilizing an AWAIT_ALL merge strategy will only be considered ready
+                when all of its dependencies have invoked this node.
                 """
-                current_node_execution_count = state.meta.node_execution_cache.get_execution_count(cls.node_class)
-                return all(
-                    state.meta.node_execution_cache.get_execution_count(dep) == current_node_execution_count + 1
+                # Check if all dependencies have invoked this node
+                all_deps_invoked = all(
+                    dep in state.meta.node_execution_cache._dependencies_invoked.get(node_span_id, set())
```
dvargasfuertes: [nit] looks like we could initialize `state.meta.node_execution_cache._dependencies_invoked.get(node_span_id, set())` once outside of this loop

dvargas92495 (approved):
wow, didn't realize we already had the primitives to solve this!

PR #1172, opened 2025-03-18
Generate output ids for adornment nodes

dvargas92495:
This PR finally makes it such that we emit output ids in the display context for adornment nodes, which will enable those events to start appearing for adornment nodes again


marlenc3 (approved):
🔥

vincent0426:
✅

PR #1180, opened 2025-03-19
Templating node serialization should support the inputs. prefix

dvargas92495:
Templating node inputs have keys that look like `'foo'` in the UI.

However, they have keys that look like `'inputs.foo'` in the SDK.

Prompt, Code exec, and Subworkflow nodes all have a similar disparity. This disparity prevents inputs from rendering in our monitoring.

In order to align, our serialization needs read from the correct spot. Next PR will update our codegen. `node_input_ids_by_name` will now be a map of attribute path names to node input ids for legacy nodes.

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/tests/test_templating_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/tests/test_templating_node.py
@@ -1,0 +1,96 @@
+import pytest
+from uuid import UUID
+
+from vellum.workflows import BaseWorkflow
+from vellum.workflows.nodes.core.templating_node.node import TemplatingNode
+from vellum_ee.workflows.display.nodes.vellum.templating_node import BaseTemplatingNodeDisplay
+from vellum_ee.workflows.display.workflows.get_vellum_workflow_display_class import get_workflow_display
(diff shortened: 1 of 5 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/tests/test_templating_node.py
```diff
@@ -0,0 +1,96 @@
...
+    ["GetDisplayClass", "expected_input_id"],
+    [
+        (_no_display_class, "d3519cec-590c-416d-8eb1-96051aed5ddd"),
+        (_display_class_with_node_input_ids_by_name, "fba6a4d5-835a-4e99-afb7-f6a4aed15110"),
+        (_display_class_with_node_input_ids_by_name_with_inputs_prefix, "fba6a4d5-835a-4e99-afb7-f6a4aed15110"),
+    ],
+    ids=[
+        "no_display_class",
+        "display_class_with_node_input_ids_by_name",
+        "display_class_with_node_input_ids_by_name_with_inputs_prefix",
```
dvargasfuertes, line 40: Handle both for templating nodes

vincent0426 (approved):

PR #1183, opened 2025-03-19
Support codegen for prefixed templating node inputs

dvargas92495:
Last piece needed on the SDK side to solve the templating node inputs bug.

We use codegen to solidify `node_input_ids_by_name` being a mapping from attribute name to node input id


marlenc3:
✅

PR #1186, opened 2025-03-20
[APO-180] Add timeout for retry node

vincent0426:
- Support timeout feature in retry node
- Add workflow error code NODE_TIMEOUT


src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -53,60 +56,90 @@ def run(self) -> BaseNode.Outputs:
...
+                            return o
+                except multiprocessing.TimeoutError:
                     last_exception = NodeException(
-                        event.error.message,
-                        code=event.error.code,
+                        code=WorkflowErrorCode.NODE_TIMEOUT,
+                        message=f"Subworkflow timed out after {self.timeout} seconds on attempt {attempt_number}",
                     )
                     if self.delay:
                         time.sleep(self.delay)
```
vincent0426, line 76: I think the last iter does not need to sleep, could fix in the follow PR

src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -53,60 +56,90 @@ def run(self) -> BaseNode.Outputs:
...
 
-            if exception:
-                raise exception
+        raise last_exception
 
-            if node_outputs:
-                return node_outputs
+    def _run_subworkflow(
+        self, subworkflow_stream: WorkflowEventStream, attempt_number: int
+    ) -> Union[BaseNode.Outputs, NodeException]:
```
vincent0426, line 88: since we need to capture both outputs or the last exception, not sure if this is the best way to do so

ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_adornments_serialization.py
```diff
@@ -110,6 +112,7 @@ class InnerRetryGenericNodeDisplay(BaseNodeDisplay[InnerRetryGenericNode]):
             ],
         },
         serialized_node,
+        ignore_order=True,
```
dvargasfuertes: [nit] In general, I like to avoid `ignore_order` if possible bc:
- makes diffs harder to read
- could obfuscate nested out of order bugs

src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -22,13 +23,15 @@ class RetryNode(BaseAdornmentNode[StateType], Generic[StateType]):
 
     max_attempts: int - The maximum number of attempts to retry the Subworkflow
     delay: float = None - The number of seconds to wait between retries
+    timeout: float = None - The number of seconds to wait for the Subworkflow to complete
```
dvargasfuertes, line 26: agh I reread the original slack thread , and I should've interjected earlier. the `timeout` setting referenced by Noa was surrounding this prompt timeout: https://github.com/vellum-ai/vellum/blob/2fa6eca28683fbcccb5d63de66059d5157baee38/django/app/domain_models/types/ml_model.py#L240

Strictly speaking, we may not need this attribute, the original ticket was more focused on adding delay to the default retry node attributes. We could finish SDK support through this PR, but let's defer any serialization/codegen support to focus on the higher urgency items we have for this week

src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -53,60 +56,90 @@ def run(self) -> BaseNode.Outputs:
...
-                        message=f"""Unexpected rejection on attempt {attempt_number}: {event.error.code.value}.
-Message: {event.error.message}""",
-                    )
-                elif self.retry_on_condition and not resolve_value(self.retry_on_condition, self.state):
-                    exception = NodeException(
-                        code=WorkflowErrorCode.INVALID_OUTPUTS,
-                        message=f"""Rejection failed on attempt {attempt_number}: {event.error.code.value}.
-Message: {event.error.message}""",
-                    )
+            if self.timeout is None:
```
dvargasfuertes: [nit] we likely also want to make sure timeout is > 0

src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -53,60 +56,90 @@ def run(self) -> BaseNode.Outputs:
...
+                o = self._run_subworkflow(subworkflow_stream, attempt_number)
+                if isinstance(o, NodeException):
+                    last_exception = o
                 else:
+                    return o
+            else:
+                try:
+                    with multiprocessing.pool.ThreadPool(processes=1) as pool:
+                        async_result = pool.apply_async(
+                            self._run_subworkflow, args=(subworkflow_stream, attempt_number)
```
dvargasfuertes, line 63: [oos] The subworkflow will already have a thread that initiates, I wonder if we could pass the timeout directly to the workflow

dvargas92495 (commented):
I think it's more likely we add support for this through an adornment than on a node-by-node basis. Happy to huddle on this tm morning, but in any case, I think the only high pri part of the ticket is the missing attribute in the Vellum frontend for default retry node attribtues

vincent0426:
> I think it's more likely we add support for this through an adornment than on a node-by-node basis. Happy to huddle on this tm morning, but in any case, I think the only high pri part of the ticket is the missing attribute in the Vellum frontend for default retry node attribtues

Ah yeah, timeout adornment seems better!

PR #1187, opened 2025-03-21
[APO-2] Add code merging support for generic node

vincent0426:
PR in codegen-service: https://github.com/vellum-ai/codegen-service/pull/74

https://www.loom.com/share/8dcfa94d8e384698aa8899984e178ce1?sid=a88f33de-1512-4492-aa93-e974d1b36078

### DIFF
--- a/ee/codegen/src/__test__/project.test.ts
+++ b/ee/codegen/src/__test__/project.test.ts
@@ -1784,4 +1784,163 @@ baz = foo + bar
       ]);
     });
   });
+  describe("Get generic node files", () => {
+    it("should get generic node files", async () => {
+      const displayData = {
+        workflow_raw_data: {
+          nodes: [
+            {
+              id: "entry",
+              type: "ENTRYPOINT",
+              data: {
+                label: "Entrypoint",
+                source_handle_id: "entry_source",
+                target_handle_id: "entry_target",
+              },
+              inputs: [],
+            },
+            {
+              id: "generic-node",
+              type: "GENERIC",
+              label: "Generic Node",
+              attributes: [
+                {
+                  id: uuidv4(),
+                  name: "forward",
+                  value: {
+                    type: "NODE_OUTPUT",
+                    node_id: "generic-node",
+                    node_output_id: "output",
+                  },
+                },
+              ],
+              trigger: {
+                id: "generic-node-trigger",
+                merge_behavior: "AWAIT_ATTRIBUTES",
+              },
+              ports: [
+                {
+                  id: "generic-node-default-port",
+                  name: "default",
+                  type: "DEFAULT",
+                },
+              ],
+              base: {
+                name: "BaseNode",
+                module: ["vellum", "workflows", "nodes", "bases", "base"],
+              },
+              outputs: [],
+            },
+          ],
+          edges: [
+            {
+              source_node_id: "entry",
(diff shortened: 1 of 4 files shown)


ee/codegen/src/project.ts
```diff
@@ -779,4 +779,15 @@ ${errors.slice(0, 3).map((err) => {
       sandboxInputs: this.sandboxInputs ?? [],
     });
   }
+
+  /**
+   * Gets the generic node file paths that have been tracked during code generation
+   * @returns Object mapping file paths to booleans
+   */
+  public getGenericNodeFiles(): Record<string, boolean> {
```
dvargasfuertes: [nit] would love to update an existing test we have in `project.test.ts` that has generic nodes to assert that this is now returning a non-empty. and also one _without_ generic node's to show that it is empty

ee/codegen/src/project.ts
```diff
@@ -779,4 +779,15 @@ ${errors.slice(0, 3).map((err) => {
...
+    const genericNodeFiles = this.workflowContext.getGenericNodeFiles();
+    return Object.fromEntries(
+      Array.from(genericNodeFiles).map((file) => [file, true])
```
dvargasfuertes: [nit] a `Dict<string, bool>` is actually just a `Set<string>`

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -94,4 +94,22 @@ export class GenericNode extends BaseSingleFileNode<
   protected getErrorOutputId(): string | undefined {
     return undefined;
   }
+
+  private getModulePath(): string[] {
+    return this.nodeContext.nodeModulePath;
+  }
+
+  public async persist(): Promise<void> {
```
dvargasfuertes, line 102: [q] is there a way to reuse the logic in `super().persist()`?

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -114,6 +114,9 @@ export class WorkflowContext {
   private readonly mlModelNamesById: Record<string, string> = {};
   private readonly errors: BaseCodegenError[] = [];
 
+  // Track generic node files for code merging
+  private readonly genericNodeFiles: Set<string> = new Set();
```
dvargasfuertes: [q] can we name this to something specific to needing python code merging rather than just generic node files? As mentioned, I think this is a set that we will look to decrease over time and drive most of it through codegen the more we enable editing in the UI

dvargas92495 (approved):

ee/codegen/src/__test__/project.test.ts
```diff
@@ -1784,4 +1784,163 @@ baz = foo + bar
...
+        },
+      });
+
+      await project.generateCode();
+      const pythonCodeMergeableNodeFiles =
+        project.getPythonCodeMergeableNodeFiles();
+      expect(pythonCodeMergeableNodeFiles).toEqual(
+        new Set(["nodes/generic_node.py"])
+      );
+    });
```
vincent0426, line 1867: main test part

dvargas92495 (approved):
Feel free to kick off the `0.14.19` release cut on `vellum-client-generator`!

PR #1194, opened 2025-03-22
Serialize code inputs with proper input path prefix

dvargas92495:
Same PR as https://github.com/vellum-ai/vellum-python-sdks/pull/1180, but for code exec node inputs. This will eventually build up to proper initiated event display in Vellum


vincent0426:
✅

marlenc3 (commented):
✅

PR #1195, opened 2025-03-22
Hoist up edge display utilities to the root BaseWorkflowDisplay

dvargas92495:
We are eventually building up to [this](https://github.com/vellum-ai/vellum-python-sdks/pull/1191) PR, where we deprecate the other three Edge*Display classes. Long term, we are looking to remove all of those classes including VellumWorkflowDisplay to consolidate around just BaseWorkflowDisplay.

In this PR, we are simply hoisting up the utility methods from VellumWorkflowDisplay -> BaseWorkflowDisplay, removing some generics along the way


vincent0426:
✅

PR #1199, opened 2025-03-22
Remove faa exclusion

dvargas92495:
Noticed that we had this fixture still excluded, which is preventing some further cleanup work. Most LoC comes from regenerating the JSON once I confirmed that the only changes left would be reorders/missing definition/base fields


marlenc3 (commented):
✅

PR #1200, opened 2025-03-22
Codegen EdgeDisplay instead of EdgeVellumDisplayOverrides

dvargas92495:
We eventually building up to [this](https://github.com/vellum-ai/vellum-python-sdks/pull/1191/files) PR. We are now only codegenning EdgeDisplay and marking the other classes for deprecation


marlenc3 (commented):
✅

PR #1202, opened 2025-03-23
Harrison/apo 200 codegen terminal node outputs attribute

NgoHarrison:
This handles codegen and serialization of outputs on terminal nodes.

Main changes in this PR:

2. Final output nodes are being codegen'd based on the outputs field now
3. Serialization of the outputs field on terminal nodes


ee/vellum_ee/workflows/display/nodes/types.py
```diff
@@ -8,6 +11,14 @@ class NodeOutputDisplay:
     name: str
 
 
+@dataclass
+class NodeInputDisplay:
+    id: UUID
+    name: str
+    type: Optional[str] = None
+    value: Optional[WorkflowValueDescriptor] = None
```
dvargasfuertes: [blocking] I'm not following why we need this? there's no new data that this is tracking afaict. I have a sync with a customer from 4-430p, let's sync after

ee/vellum_ee/workflows/display/nodes/vellum/final_output_node.py
```diff
@@ -47,6 +47,14 @@ def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> J
             "display_data": self.get_display_data().dict(),
             "base": self.get_base().dict(),
             "definition": self.get_definition().dict(),
+            "outputs": [
+                {
+                    "id": str(self._get_output_id()),
+                    "name": "node_input",
+                    "type": inferred_type,
+                    "value": self._create_workflow_value_descriptor(),
```
dvargasfuertes: ah, I think the disconnect is that we should be able to get this value directly from `self.Outputs.value` and serialize that descriptor instead of creating a new one to track

ee/vellum_ee/workflows/display/vellum.py
```diff
@@ -209,3 +210,124 @@ class NodeInput(UniversalBaseModel):
     id: str
     key: str
     value: NodeInputValuePointer
+
+
+class LogicalOperator(str, Enum):
```
dvargasfuertes: I wouldn't expect to need any of these after we clean of `NodeInputDisplay`

ee/codegen/src/__test__/nodes/__snapshots__/final-output-node.test.ts.snap
```diff
@@ -43,45 +43,3 @@ class FinalOutputNode(BaseFinalOutputNode[BaseState, str]):
         value = "<my-output>"
 "
 `;
-
-exports[`FinalOutputNode > should codegen successfully without node input > getNodeDisplayFile 1`] = `
```
dvargasfuertes: [blocking] let's get this test back

ee/codegen/src/generators/nodes/final-output-node.ts
```diff
@@ -42,12 +43,19 @@ export class FinalOutputNode extends BaseSingleFileNode<
       ],
     });
 
-    const nodeInput = this.getNodeInputByName("node_input");
+    // There is only one value in the outputs list so we grab the first one
+    const descriptor = this.nodeData.outputs?.map((output) => output.value)[0];
```
dvargasfuertes: [nit] we should just generate all of the node outputs that are in here, using `output.name` -> `python.field({ name:})`, so that it makes it easier to replace this file as just a generic node in the future.

ee/codegen/src/generators/nodes/final-output-node.ts
```diff
@@ -42,12 +43,19 @@ export class FinalOutputNode extends BaseSingleFileNode<
       ],
     });
 
-    const nodeInput = this.getNodeInputByName("node_input");
```
dvargasfuertes, line 45: [nit] let's fallback to this logic in case `this.nodeData.outputs` is null/undefined/empty for whatever reason. In the future once we have confidence that we've migrated all nodes as generic/base nodes, we can axe.

ee/codegen_integration/fixtures/faa_q_and_a_bot/display_data/faa_q_and_a_bot.json
```diff
@@ -2029,7 +2029,46 @@
             "nodes",
             "final_output_2"
           ]
-        }
+        },
+        "outputs": [
+          {
+            "id": "8c6e5464-8916-4039-b911-cf707855d372",
+            "name": "node_input",
```
dvargasfuertes: [blocking] I would expect all of these to match the outputs class attribute, so `"name": "value"`

ee/vellum_ee/workflows/display/nodes/vellum/final_output_node.py
```diff
@@ -47,6 +48,14 @@ def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> J
             "display_data": self.get_display_data().dict(),
             "base": self.get_base().dict(),
             "definition": self.get_definition().dict(),
+            "outputs": [
+                {
+                    "id": str(self._get_output_id()),
+                    "name": "node_input",
```
dvargasfuertes:
```suggestion
                    "name": "value",
```

addresses the blocking comment above. In fact, I've started using direct references, eg `FinalOutputNode.Outputs.value.name`

ee/vellum_ee/workflows/display/nodes/vellum/final_output_node.py
```diff
@@ -47,6 +48,14 @@ def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> J
...
+                    "type": inferred_type,
+                    "value": self.serialize_value(display_context, cast(BaseDescriptor, node.Outputs.value.instance)),
```
dvargasfuertes: [nit] shouldn't need the cast here

dvargas92495 (approved):
blocking comments are quick

PR #689, opened 2025-01-24
Vargas/handle id bug

dvargas92495:
[Context:](https://vellum-ai.slack.com/archives/C06P13ABK0W/p1737670741488019)
Screenshot 2025-01-23 at 11 57 57 PM

Looks like we had disconnected logic between what our edges and what our base nodes used for handle ids. This PR tackles this.

PR #1204, opened 2025-03-24
Hoist target handle id to BaseNodeDisplay

dvargas92495:
We are eventually building up to [this](https://github.com/vellum-ai/vellum-python-sdks/pull/1191) PR.

We want to remove dependency on the `[source|target]_node_(handle)?_id` attributes from our deprecated `EdgeVellumDisplay` class by using the node/handle ids that we already have in context. This pushed us to finally hoist up `target_handle_id` from `BaseNodeVellumDisplay` to `BaseNodeDisplay`, the former being something we want to eventually delete so that everything just inherits from BaseNodeDisplay directly


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -255,7 +259,24 @@ def get_node_port_display(self, port: Port) -> PortDisplay:
         return PortDisplay(id=port_id, node_id=self.node_id)
 
     def get_trigger_id(self) -> UUID:
-        return uuid4_from_hash(f"{self.node_id}|trigger")
+        return self.get_target_handle_id()
```
dvargasfuertes, line 262: Most of the file deltas below are driven from this change. This is actually a reversion to what we had previously, and undoes a lot of the diffs introduced here: https://github.com/vellum-ai/vellum-python-sdks/pull/689

PR #1205, opened 2025-03-24
Rework file loader to handle inspect.getsource and lazy references

dvargas92495:
https://linear.app/vellum/issue/APO-227/unblock-react-workflow

We rework our Virtual File Finder and Virtual File Loader so that:
- we only have one instance of the loader
- that one instance has access to the full files dict
- that one loader could resolve any other file's code

### DIFF
--- a/ee/vellum_ee/workflows/server/virtual_file_loader.py
+++ b/ee/vellum_ee/workflows/server/virtual_file_loader.py
@@ -1,42 +1,72 @@
 import importlib
+import re
+from typing import Optional
 
 
 class VirtualFileLoader(importlib.abc.Loader):
-    def __init__(self, code: str, is_package: bool):
-        self.code = code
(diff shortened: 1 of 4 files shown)


ee/vellum_ee/workflows/server/virtual_file_loader.py
```diff
@@ -1,42 +1,72 @@
...
+        `inspect` module uses this method to get the source code of a module.
+        """
+
+        module_info = self._resolve_module(fullname)
+        if module_info:
+            return module_info[1]
+
+        return None
+
+    def _resolve_module(self, fullname: str) -> Optional[tuple[str, str]]:
```
dvargasfuertes, line 33: Most of the logic from `VirtualFileFinder.find_spec` was moved to here, so that if could work with any `inspect` call invoked within the module

PR #1210, opened 2025-03-24
Generate edges from display context instead of EdgeVellumDisplay

dvargas92495:
We are eventually building up to [this](https://github.com/vellum-ai/vellum-python-sdks/pull/1191) PR.

In this PR, we read from our node/port displays directly to build serialized edges instead of using these [fields](https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/vellum_ee/workflows/display/vellum.py#L106-L114) which we are deprecating


marlenc3 (commented):
✅

PR #1212, opened 2025-03-24
[APO-235] Add default return based on output type for codeexec

vincent0426:
Return default value if code exec function returns None

### DIFF
--- a/src/vellum/workflows/nodes/tests/test_utils.py
+++ b/src/vellum/workflows/nodes/tests/test_utils.py
@@ -136,11 +136,30 @@ def test_parse_type_from_str_error_cases(input_str, output_type, expected_except
     assert expected_message in str(excinfo.value)
 
 
-def test_cast_to_output_type_none_value():
-    assert cast_to_output_type(None, str) == ""  # String
-    assert cast_to_output_type(None, int) == 0  # Number
-    assert cast_to_output_type(None, float) == 0.0  # Number
-    assert cast_to_output_type(None, Any) is None  # Json
-    assert cast_to_output_type(None, FunctionCallType) == {}  # FunctionCall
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/tests/test_utils.py
```diff
@@ -131,3 +134,13 @@ def test_parse_type_from_str_error_cases(input_str, output_type, expected_except
...
+
+
+def test_cast_to_output_type_none_value():
+    assert cast_to_output_type(None, str) == ""  # String
+    assert cast_to_output_type(None, int) == 0  # Number
+    assert cast_to_output_type(None, float) == 0.0  # Number
+    assert cast_to_output_type(None, Any) is None  # Json
+    assert cast_to_output_type(None, FunctionCallType) == {}  # FunctionCall
+    assert cast_to_output_type(None, List[ChatMessage]) == []  # Chat History
+    assert cast_to_output_type(None, List[VellumValue]) == []  # Array
```
dvargasfuertes: [nit] prefer using `pytest.mark.parametrize`

src/vellum/workflows/nodes/utils.py
```diff
@@ -195,7 +196,34 @@ def _clean_output_type(output_type: Any) -> Any:
...
+        return 0.0
+    elif output_type is str:
+        return ""
+    elif origin is list:
+        return []
+    elif output_type is FunctionCall:
+        return {}
+    elif origin is Union:
+        # Recursively find first type with a non-None default
+        for inner_type in args:
```
dvargasfuertes: [nit] I think we just always use the first arg, since `None` is a valid default return value

dvargas92495 (approved):

PR #1214, opened 2025-03-25
Deprecate legacy API Node display attributes

dvargas92495:
Similar to the line of work for other nodes here: https://github.com/vellum-ai/vellum-python-sdks/pull/1198

API Node body's are not showing up in inputs. This cleans that up, while also removing display attributes that we don't need since we have `node_input_ids_by_name`. Deleting these fields helps unify all node displays eventually to `BaseNodeDisplay`

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/api_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/api_node.py
@@ -14,15 +14,6 @@ _APINodeType = TypeVar("_APINodeType", bound=APINode)
 
 
 class BaseAPINodeDisplay(BaseNodeVellumDisplay[_APINodeType], Generic[_APINodeType]):
-    url_input_id: ClassVar[Optional[UUID]] = None
-    method_input_id: ClassVar[Optional[UUID]] = None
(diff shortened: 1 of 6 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/api_node.py
```diff
@@ -14,15 +14,6 @@
...
-    method_input_id: ClassVar[Optional[UUID]] = None
-    body_input_id: ClassVar[Optional[UUID]] = None
-
-    authorization_type_input_id: ClassVar[Optional[UUID]] = None
-    bearer_token_value_input_id: ClassVar[Optional[UUID]] = None
-    api_key_header_key_input_id: ClassVar[Optional[UUID]] = None
-    api_key_header_value_input_id: ClassVar[Optional[UUID]] = None
-
     # A mapping between node input keys and their ids for inputs representing additional header keys
     additional_header_key_input_ids: ClassVar[Optional[Dict[str, UUID]]] = None
```
dvargasfuertes, line 18: Deferring these guys to a followup. Need more time to think about how to handle it

marlenc3 (approved):

PR #1218, opened 2025-03-25
[APO-245] Fix api node

vincent0426:
- Define custom to json to handle undefined value
- Fix API node body from data to json

### DIFF
--- a/src/vellum/workflows/nodes/displayable/bases/api_node/node.py
+++ b/src/vellum/workflows/nodes/displayable/bases/api_node/node.py
@@ -57,7 +57,7 @@ class BaseAPINode(BaseNode, Generic[StateType]):
             if isinstance(headers[header], VellumSecret):
                 vellum_instance = True
         if vellum_instance or bearer_token:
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/nodes/displayable/bases/api_node/node.py
```diff
@@ -57,7 +57,7 @@ def _run(
             if isinstance(headers[header], VellumSecret):
                 vellum_instance = True
         if vellum_instance or bearer_token:
-            return self._vellum_execute_api(bearer_token, data, headers, method, url)
+            return self._vellum_execute_api(bearer_token, json, headers, method, url)
```
dvargasfuertes, line 60: [oos] Could we fast follow a test that would've caught this bug? We should be able to edit an existing test to verify that we are sending the correct body to Vellum

dvargas92495 (approved):

marlenc3 (approved):

dvargas92495:
Merging this through to rip a release

PR #1225, opened 2025-03-25
[APO-250] Handle unused graph where final output is not connected

vincent0426:
Error caused by generating an empty unused_graph = {}

After fixed:
Screenshot 2025-03-25 at 2 45 23 PM


dvargas92495 (approved):
How did the former produce an error on `issubclass`? May be worth a followup to protect against that too

dvargas92495:
sneaking this one into the release

PR #1229, opened 2025-03-25
Add subclass error handler

vincent0426:
follow up: https://github.com/vellum-ai/vellum-python-sdks/pull/1225

if we pass in something wrong then it would raise an error (expected). The subclass error is due to we pass in {} and it call `issubclass` on it

### DIFF
--- a/src/vellum/workflows/workflows/base.py
+++ b/src/vellum/workflows/workflows/base.py
@@ -113,8 +113,7 @@ class _BaseWorkflowMeta(type):
             elif inspect.isclass(graph_item) and issubclass(graph_item, BaseNode):
                 nodes.add(graph_item)
             else:
-                # We should never get here, but just in case
-                raise ValueError(f"Unexpected graph type: {graph_item.__class__}")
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/workflows/base.py
```diff
@@ -110,9 +110,10 @@ def collect_nodes(graph_item: Union[GraphAttribute, Set[GraphAttribute]]) -> Set
                         nodes.update(node for node in item.nodes)
                     elif inspect.isclass(item) and issubclass(item, BaseNode):
                         nodes.add(item)
-            elif issubclass(graph_item, BaseNode):
+            elif inspect.isclass(graph_item) and issubclass(graph_item, BaseNode):
                 nodes.add(graph_item)
             else:
+                # We should never get here, but just in case
```
dvargasfuertes: SDK first users could actually get here by doing dumb stuff like:
```python
class Workflow(BaseWorkflow):
    graph = 5
```

which is why I like us now falling through to a user facing error case here. Let's solidify with a quick unit test!

dvargas92495 (approved):

PR #1239, opened 2025-03-26
[APO-253] Add validation for prompt deployment

vincent0426:
Current Vellum sandbox is still showing INVALID TEMPLATES, follow up PR will fix that

Screenshot 2025-03-26 at 3 23 13 PM


NgoHarrison (approved):

src/vellum/workflows/nodes/displayable/prompt_deployment_node/tests/test_node.py
```diff
@@ -94,3 +96,22 @@ def generate_prompt_events(*args: Any, **kwargs: Any) -> Iterator[ExecutePromptE
...
     ]
+
+
+def test_prompt_deployment_node_missing_input_variable():
+    """Test that PromptDeploymentNode correctly raises error for missing input variables."""
+
+    # GIVEN a Prompt Deployment Node that is missing an input variable
+    class TestPromptDeploymentNode(PromptDeploymentNode):
+        deployment = "test-deployment"
+        prompt_inputs = {"fruits": None}
```
dvargasfuertes, line 107: [blocking] we're actually ok with `None` if the input variable from the deployment is optional. What was the response from the prompt deployment? We should be mocking the response from that API in this test. See `test_inline_prompt_node__api_error__invalid_inputs_node_exception`

PR #1240, opened 2025-03-26
[APO-254] Init output ids during dynamic output cls generation of inline subworkflows

NgoHarrison:
The original problem that surfaced was that we were getting key errors of final output when subworkflows have an adornment

### DIFF
--- a/src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
+++ b/src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
@@ -13,6 +13,7 @@ from vellum.workflows.state.base import BaseState
 from vellum.workflows.state.context import WorkflowContext
 from vellum.workflows.types.core import EntityInputsInterface
 from vellum.workflows.types.generics import InputsType, StateType
+from vellum.workflows.utils.uuids import uuid4_from_hash
 from vellum.workflows.workflows.event_filters import all_workflow_event_filter
 
 if TYPE_CHECKING:
@@ -140,3 +141,8 @@ class InlineSubworkflowNode(
         # Subclasses of InlineSubworkflowNode can override this method to provider their own
         # approach to annotating the outputs class based on the `subworkflow.Outputs`
         setattr(outputs_class, reference.name, reference)
+
+        if not hasattr(cls, "__output_ids__"):
+            cls.__output_ids__ = {}
+
+        cls.__output_ids__[reference.name] = uuid4_from_hash(f"{cls.__id__}|{reference.name}")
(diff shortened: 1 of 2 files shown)


marlenc3 (approved):

src/vellum/workflows/nodes/core/inline_subworkflow_node/tests/test_node.py
```diff
@@ -116,3 +116,30 @@ class TestNode(InlineSubworkflowNode):
...
+        subworkflow = TestSubworkflow
+
+    # THEN the wrapped node should have the correct output IDs
+    assert "final_output" in TestNode.__output_ids__
+
+    # AND when we run the node
+    node = TestNode()
+    events = list(node.run())
+
+    assert any(e.name == "final_output" and e.value == "hello" for e in events)
```
dvargasfuertes: [nit] let's just assert the last event here `events[-1]`

Also these are technically outputs that are emitted from the node, events are emitted from a Workflow

src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
```diff
@@ -140,3 +141,8 @@ def __annotate_outputs_class__(cls, outputs_class: Type[BaseOutputs], reference:
         # Subclasses of InlineSubworkflowNode can override this method to provider their own
         # approach to annotating the outputs class based on the `subworkflow.Outputs`
         setattr(outputs_class, reference.name, reference)
+
+        if not hasattr(cls, "__output_ids__"):
```
dvargasfuertes: [nit] should be safe to do `if cls.__output_ids__ is None`

src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
```diff
@@ -140,3 +141,8 @@ def __annotate_outputs_class__(cls, outputs_class: Type[BaseOutputs], reference:
...
+            cls.__output_ids__ = {}
+
+        cls.__output_ids__[reference.name] = uuid4_from_hash(f"{cls.__id__}|{reference.name}")
```
dvargasfuertes, line 148: oh snap, this is bc at the moment that BaseNodeMeta runs, this output class isn't defined. Nice catch.

dvargas92495 (approved):

src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
NgoHarrison: Updated

PR #1242, opened 2025-03-27
Centralize node display enrichment to single utils method

dvargas92495:
Takes the two loops under get_nodes() and get_unused_nodes() and centralized them for future simplification

### DIFF
--- a/ee/vellum_ee/workflows/display/types.py
+++ b/ee/vellum_ee/workflows/display/types.py
@@ -22,6 +22,12 @@ if TYPE_CHECKING:
 
 WorkflowDisplayType = TypeVar("WorkflowDisplayType", bound="BaseWorkflowDisplay")
 
+NodeDisplays = Dict[Type[BaseNode], BaseNodeDisplay]
+NodeOutputDisplays = Dict[OutputReference, Tuple[Type[BaseNode], NodeOutputDisplay]]
+WorkflowOutputDisplays = Dict[BaseDescriptor, WorkflowOutputDisplay]
+EdgeDisplays = Dict[Tuple[Port, Type[BaseNode]], EdgeDisplay]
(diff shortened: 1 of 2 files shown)


ee/vellum_ee/workflows/display/types.py
```diff
@@ -22,6 +22,12 @@
 
 WorkflowDisplayType = TypeVar("WorkflowDisplayType", bound="BaseWorkflowDisplay")
 
+NodeDisplays = Dict[Type[BaseNode], BaseNodeDisplay]
+NodeOutputDisplays = Dict[OutputReference, Tuple[Type[BaseNode], NodeOutputDisplay]]
+WorkflowOutputDisplays = Dict[BaseDescriptor, WorkflowOutputDisplay]
+EdgeDisplays = Dict[Tuple[Port, Type[BaseNode]], EdgeDisplay]
+PortDisplays = Dict[Port, PortDisplay]
```
dvargasfuertes, line 29: We're going to be editing some of these types. Since they are also redefined elsewhere, we centralize their spec

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -190,43 +197,35 @@ def display_context(
...
 
-        global_node_displays: Dict[Type[BaseNode], BaseNodeDisplay] = (
+        global_node_displays: NodeDisplays = (
             copy(self._parent_display_context.global_node_displays) if self._parent_display_context else {}
         )
 
-        port_displays: Dict[Port, PortDisplay] = {}
+        port_displays: PortDisplays = {}
 
         for node in self._workflow.get_nodes():
```
dvargasfuertes, line 212: Centralizing the loop body here and below was the main goal of the PR

marlenc3 (approved):
✅

PR #1243, opened 2025-03-27
Workaround for inline code node codegen escaping problem

m-abboud:
When running codegen in vembda mode with a code node with typescript with a bunch of escaping and regex fern is generating incorrectly escaped multiline strings for some reason. To sidestep this problem without having to get up to speed for fern internals I am switching it to use a single line when generating code in Inline mode which is really only useful for vembda.  I'm also disabling the formatting for this case because it triggers another fern bug. Another approach might be to add another layer of escaping before passing the string to fern and unescaping it in the runner by that is a bit gross or doing something to the string to make fern not maim it.

We should go forward with this less than ideal fix because it is actively impacting customers.

### DIFF
--- a/ee/codegen/src/__test__/nodes/code-execution-node.test.ts
+++ b/ee/codegen/src/__test__/nodes/code-execution-node.test.ts
@@ -1,3 +1,6 @@
+import { mkdir, readFile } from "fs/promises";
+import { join } from "path";
+
 import { Writer } from "@fern-api/python-ast/core/Writer";
 import { v4 as uuid } from "uuid";
 import { SecretTypeEnum, WorkspaceSecretRead } from "vellum-ai/api";
@@ -9,6 +12,7 @@ import {
   codeExecutionNodeFactory,
   nodeInputFactory,
 } from "src/__test__/helpers/node-data-factories";
+import { makeTempDir } from "src/__test__/helpers/temp-dir";
 import { createNodeContext, WorkflowContext } from "src/context";
 import { CodeExecutionContext } from "src/context/node-context/code-execution-node";
 import { NodeAttributeGenerationError } from "src/generators/errors";
@@ -16,13 +20,16 @@ import { CodeExecutionNode } from "src/generators/nodes/code-execution-node";
 import { NodeInput, NodeInputValuePointerRule } from "src/types/vellum";
 
 describe("CodeExecutionNode", () => {
+  let tempDir: string;
   let workflowContext: WorkflowContext;
   let writer: Writer;
   let node: CodeExecutionNode;
 
-  beforeEach(() => {
+  beforeEach(async () => {
     workflowContext = workflowContextFactory();
     writer = new Writer();
+    tempDir = makeTempDir("code-node-test");
+    await mkdir(tempDir, { recursive: true });
   });
 
   describe("basic", () => {
@@ -70,7 +77,15 @@ describe("CodeExecutionNode", () => {
     });
   });
 
-  describe("code representation override", () => {
+  describe.each([
+    ["Base case", "print('Hello, World!')"],
+    [
+      "Escaped case",
+      "async function main(inputs: {\n  question: string\n}) {\n  inputs = {\n" +
+        '"question": "{\\"text_explanation\\":\\"First, \\\\\\\\(\\\\\\\\frac{1}{40}\\\\\\\\).\\\\"\n  }\n' +
+        "const test = \"\\frac\".replace(/\\\\frac/g, '\\\\dfrac');\n  return {};\n} \n",
+    ],
+  ])("code representation override: %s", (_, code) => {
     it.each<"INLINE" | "STANDALONE" | undefined>([
       undefined,
       "STANDALONE",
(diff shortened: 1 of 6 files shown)


ee/codegen/src/__test__/nodes/code-execution-node.test.ts
```diff
@@ -80,7 +82,14 @@ describe("CodeExecutionNode", () => {
...
         });
 
-        const nodeData = codeExecutionNodeFactory();
+        const nodeData = codeExecutionNodeFactory({
+          // This code triggers some things with the way fern does escapes that we need to test is escaping correctly
+          // as it sometimes does not for inline mode which is used by vembda.
+          code:
+            "async function main(inputs: {\n  question: string\n}) {\n  inputs = {\n    " +
+            '"question": "{\\"text_explanation\\":\\"First, \\\\\\\\(\\\\\\\\frac{1}{40}\\\\\\\\).\\\\"\n  }\n  ' +
+            "const test = \"\\frac\".replace(/\\\\frac/g, '\\\\dfrac');\n  return {};\n} \n",
```
dvargasfuertes: [nit] could we parametrize the test case further so that we have a base case code (the Hello World) and this complex case?

[nit] Could we make the string _here_ multiline instead of the newlines to make the user provided code slightly more readable?

ee/codegen/src/generators/base-persisted-file.ts
```diff
@@ -206,8 +206,14 @@ export abstract class BasePersistedFile extends AstNode {
...
     try {
-      contents = await writer.toStringFormatted({ line_width: 120 });
-      contents = this.postprocessDocstrings(contents);
+      if (skipFormatting) {
+        // Sometimes we need to disable formatting because of fern problems, this isn't strictly necessary because
+        // of the catch statement below but it will let us avoid an unneeded error log and sentry report.
+        contents = writer.toString();
+      } else {
+        contents = await writer.toStringFormatted({ line_width: 120 });
+        contents = this.postprocessDocstrings(contents);
```
dvargasfuertes, line 215: [oos] a midterm approach here is to give ourselves some hook here for pre-processing before it's fed into the formatter

ee/codegen/src/generators/nodes/code-execution-node.ts
```diff
@@ -95,7 +101,10 @@ export class CodeExecutionNode extends BaseSingleFileNode<
         python.field({
           name: CODE_INPUT_KEY,
           initializer: python.TypeInstantiation.str(this.scriptFileContents, {
-            multiline: true,
+            // Fern will garble the escaping when using python multiline for inline so we disable it.
+            // Inline code generation is currently only used by vembda and likely to remain that way
+            // so skipping over pretty formatting is acceptable.
+            multiline: false,
```
dvargasfuertes, line 107: [blocking] could we make our disabling of multiline here more targeted? eg if we detect `"\\"` in the `scriptFileContents`

same for the skipping of formatting above

ee/codegen/src/__test__/nodes/code-execution-node.test.ts
```diff
@@ -93,7 +102,13 @@ describe("CodeExecutionNode", () => {
         });
 
         node.getNodeFile().write(writer);
-        expect(await writer.toStringFormatted()).toMatchSnapshot();
+        const output = await (override === "INLINE"
+          ? // Fern will error out when using toStringFormatted with our approach to inline code nodes
+            // with a bizarre syntax error, there's a try catch in our persisted file code and we also
+            // pass in a skipFormatting flag to it for this case
+            writer.toString()
+          : writer.toStringFormatted());
```
dvargasfuertes: [nit] this feels like we are letting implementation details bleed into tests. is it possible to do `node.persist()` to some temp file and then use toMatchSnapshot on the contents?

dvargas92495 (commented):
Would like to align on targeting first (the last comment)

ee/codegen/src/generators/nodes/code-execution-node.ts
m-abboud: I don't think this is really blocking because this case is really only for internal vembda usage. Users are very unlikely to want inline code since it makes it unreadable with all the escapes.

m-abboud: Also this way might catch other cases we might not have thought about

dvargasfuertes: ok I'll downgrade to a [nit] - in general I want to strive for the gap between vembda representation and local representations to be as minimal as possible, and one day be zero

dvargas92495 (approved):

ee/codegen/src/generators/nodes/code-execution-node.ts
m-abboud: I would say in this case we would need to be able to use the script files directly to match 1:1 I forget where we landed on that to make it happen? Don't think that would be to bad actually just write a temp file for it to load and a special if check somewhere to know where to read it from

dvargasfuertes: Yea I think we're currently blocked on the fact that we are using `os` and some bugs with `inspect`. We are tracking solving both here https://linear.app/vellum/issue/APO-239/remove-disparity-between-local-code-node-resolution-and-vembda-code, which we could consider [oos] of this PR

PR #1244, opened 2025-03-27
Consolidate Entrypoint and Workflow Meta displays into BaseWorkflowDisplay

dvargas92495:
The goal of this PR is to consolidate our `EntrypointDisplay`s and `WorkflowMetaDisplay`s into `BaseWorkflowDisplay`. It discovered a bunch of unneccesary circular dependencies in the meantime. 

WIP - going to start splitting these out into separate

PR #1246, opened 2025-03-27
Hoist up NodeDisplayData to WorkflowMetaDisplay

dvargas92495:
PR we are eventually working towards: https://github.com/vellum-ai/vellum-python-sdks/pull/1244

This PR just hoists up the last field from `WorkflowMetaVellumDisplayOverrides` to `WorkflowMetaDisplay`, which will allow us to cutover in the next PR in codegen

### DIFF
--- a/ee/vellum_ee/workflows/display/editor/types.py
+++ b/ee/vellum_ee/workflows/display/editor/types.py
@@ -1,0 +1,22 @@
+from typing import Optional
+
+from pydantic import Field
+
+from vellum.client.core.pydantic_utilities import UniversalBaseModel
+
+
+class NodeDisplayPosition(UniversalBaseModel):
+    x: float = 0.0
+    y: float = 0.0
+
+
+class NodeDisplayComment(UniversalBaseModel):
+    value: Optional[str] = None
+    expanded: Optional[bool] = None
+
+
+class NodeDisplayData(UniversalBaseModel):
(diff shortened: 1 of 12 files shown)


ee/vellum_ee/workflows/display/editor/types.py
```diff
@@ -0,0 +1,22 @@
+from typing import Optional
```
dvargasfuertes, line 1: putting this in `ee/vellum_ee/workflows/display/nodes/types.py` introduced more circ deps bc the `__init__.py` in that package does a full re-export of all the node classes, which it needs to do for Node Display registration. So we started a new package to represent types needed for the sandbox editor experience

ee/vellum_ee/workflows/display/utils/vellum.py
```diff
@@ -39,23 +42,66 @@
 from vellum.workflows.utils.vellum_variables import primitive_type_to_vellum_variable_type
 from vellum.workflows.vellum_client import create_vellum_client
 from vellum_ee.workflows.display.utils.expressions import get_child_descriptor
-from vellum_ee.workflows.display.vellum import (
```
dvargasfuertes, line 42: needed to move these guys to resolve another angle of the circ deps. This should also allow us to delete `vellum_ee.workflows.display.vellum` entirely in the `0.15.0` release

marlenc3 (commented):
✅

PR #1252, opened 2025-03-27
[APO-269] Add log attr to guard rail node outputs

NgoHarrison:
[We started exposing `log` for guardrail node output in addition to score](https://github.com/vellum-ai/vellum/pull/8648)

### DIFF
--- a/src/vellum/workflows/nodes/displayable/guardrail_node/node.py
+++ b/src/vellum/workflows/nodes/displayable/guardrail_node/node.py
@@ -34,6 +34,7 @@ class GuardrailNode(BaseNode[StateType], Generic[StateType]):
 
     class Outputs(BaseOutputs):
         score: float
+        log: Optional[str]
 
     def run(self) -> Outputs:
         metric_execution = self._context.vellum_client.metric_definitions.execute_metric_definition(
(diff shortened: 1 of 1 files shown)


src/vellum/workflows/nodes/displayable/guardrail_node/node.py
```diff
@@ -52,8 +53,21 @@ def run(self) -> Outputs:
...
+        log: Optional[str] = None
+        raw_log = metric_outputs.get("log")
+
+        if raw_log:
+            if not isinstance(raw_log, str):
+                raise NodeException(
+                    message="Metric execution log output must be of type 'str'",
+                    code=WorkflowErrorCode.INVALID_OUTPUTS,
+                )
+            log = raw_log
```
NgoHarrison, line 65: This weird reassignment was to make mypy happy

src/vellum/workflows/nodes/displayable/guardrail_node/node.py
```diff
@@ -34,6 +34,7 @@ class Trigger(BaseNode.Trigger):
 
     class Outputs(BaseOutputs):
         score: float
+        log: Optional[str]
```
NgoHarrison, line 37: Made this optional since I think older workflows might not have this metric

dvargas92495 (approved):
This reminds me - Metric definitions could have any number of outputs. What happens if I define a Metric definition with a non-score output? I imagine the same error?

NgoHarrison:
> This reminds me - Metric definitions could have any number of outputs. What happens if I define a Metric definition with a non-score output? I imagine the same error?

I think the only reason log was causing an issue for the customer was because this field started getting exposed. If other outputs started getting exposed in the frontend, then I would imagine the same error since codegen will try and reference those values

PR #1254, opened 2025-03-28
Thread through vellum API environment to all clients

dvargas92495:
Needed this for local debugging, but we will also need for VPC


m-abboud (commented):
✅

PR #1264, opened 2025-03-30
Fix inline prompt node infinite loop in httpclient serialization

m-abboud:
from the workflow here https://vellum-ai.slack.com/archives/C08GM7FUT39/p1743148454956489


marlenc3 (approved):
Damn, that's brutal ✅️

dvargas92495:
Oh wow, great find.

What was causing the infinite loop? A circular reference?

We have a couple of other nodes where we'll need to do this too (subworkflow deployment nodes)

✅

PR #1265, opened 2025-03-30
Fix vellumApiEnvironment types

m-abboud:
VellumEnvironment has static string types as it's = `typeof VellumEnvironment.Production` and isn't useable by codegen service


dvargas92495 (approved):
yikes my b

m-abboud:
> yikes my b

I forgive you

PR #1270, opened 2025-03-31
Simplify execution context serialization across relevant nodes

dvargas92495:
cleans up our execution context serialization to just model dumping from the jump


m-abboud (commented):
✅

PR #1276, opened 2025-03-31
[APO-271] Test that shows error with loading ts file in virtual filemap for code exec

NgoHarrison:
This is just a test to mimic the failure of when we have the ts file loaded into the virtual file mapping. The followups from this will have to address the problems with inspect and read from file and then likely the usage of these virtual file maps instead of inlined code attributes

### DIFF
--- a/ee/vellum_ee/workflows/tests/test_server.py
+++ b/ee/vellum_ee/workflows/tests/test_server.py
@@ -1,3 +1,4 @@
+import pytest
 import sys
 from uuid import uuid4
 from typing import Type, cast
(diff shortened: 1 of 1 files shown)


ee/vellum_ee/workflows/tests/test_server.py
```diff
@@ -69,3 +70,55 @@ class Outputs(BaseNode.Outputs):
...
+    sys.meta_path.append(VirtualFileFinder(files, namespace))
+
+    # WHEN the workflow is loaded
+    Workflow = BaseWorkflow.load_from_module(namespace)
+    workflow = Workflow()
+
+    # THEN the workflow is successfully initialized
+    assert workflow
+
+    workflow.run()
```
dvargasfuertes: [nit] I'd at least assert that the final event `.name` here is `workflow.execution.fulfilled`

dvargas92495 (approved):
Dope, sorry for only now getting to this. I can sync up at 630p est for some pair programming

PR #1278, opened 2025-03-31
[APO-299] Wrap lhs as a constant value reference if node input pointer is a constant value

NgoHarrison:
Context: Workflows with a constant value reference codegen incorrectly. This PR wraps constant values with our ConstantValueReference obj

### DIFF
--- a/ee/codegen/src/generators/expression.ts
+++ b/ee/codegen/src/generators/expression.ts
@@ -1,3 +1,4 @@
+// eslint-disable-next-line import/order
 import { python } from "@fern-api/python-ast";
 import { AstNode } from "@fern-api/python-ast/core/AstNode";
 import { Writer } from "@fern-api/python-ast/core/Writer";
@@ -5,6 +6,9 @@ import { Writer } from "@fern-api/python-ast/core/Writer";
 import { VELLUM_WORKFLOW_CONSTANTS_PATH } from "src/constants";
 import { NodeAttributeGenerationError } from "src/generators/errors";
 import { WorkflowValueDescriptorReference } from "src/generators/workflow-value-descriptor-reference/workflow-value-descriptor-reference";
+// eslint-disable-next-line import/order
+import { NodeInput } from "src/generators/node-inputs";
+import { isNilOrEmpty } from "src/utils/typing";
 
 export declare namespace Expression {
   interface Args {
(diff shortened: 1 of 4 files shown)


ee/codegen/src/generators/expression.ts
```diff
@@ -1,10 +1,14 @@
+// eslint-disable-next-line import/order
 import { python } from "@fern-api/python-ast";
 import { AstNode } from "@fern-api/python-ast/core/AstNode";
 import { Writer } from "@fern-api/python-ast/core/Writer";
 
 import { VELLUM_WORKFLOW_CONSTANTS_PATH } from "src/constants";
 import { NodeAttributeGenerationError } from "src/generators/errors";
 import { WorkflowValueDescriptorReference } from "src/generators/workflow-value-descriptor-reference/workflow-value-descriptor-reference";
+// eslint-disable-next-line import/order
+import { NodeInput } from "src/generators/node-inputs";
```
dvargasfuertes, line 10: [nit] why do we have this import order issue?

ee/codegen/src/generators/expression.ts
```diff
@@ -101,6 +113,17 @@ export class Expression extends AstNode {
...
 
+  private isConstantValuePointer(lhs: AstNode): boolean {
+    return (
+      lhs instanceof NodeInput &&
+      !isNilOrEmpty(
+        lhs.nodeInputValuePointer.nodeInputValuePointerData.rules
+      ) &&
+      lhs.nodeInputValuePointer.nodeInputValuePointerData.rules[0]?.type ===
+        "CONSTANT_VALUE"
+    );
```
dvargasfuertes, line 124: [oos] this `rules[0]` is a bit spooky. I would've expected that we solved this at a per rule level instead of at the full value. I can't think of a counterexample though atm, so let's just keep a close 👀

dvargas92495 (approved):

ee/codegen/src/generators/expression.ts
NgoHarrison: I think it's causing a circular dependency with the import directly above `WorkflowValueDescriptorReference`. I can  clean this up in a fast follow but wanted to get the fix out for now

dvargasfuertes: Cool yea let's fast follow - a manner in which we could possible clean this up is figuring out how we could avoid the `instanceof NodeInput` reference below in the first place. I think in most of our nodes, we pass _down_ input data and build _up_ AstNode's but I could be mistaken here.

PR #1281, opened 2025-04-01
Test for Missing Lazy References

marlenc3:
if you remove the wrapper - it passes, with it it fails.
 
Missing references are causing loop failures (since the new reference is never brought in)

### DIFF
--- a/tests/workflows/basic_inline_prompt_node_with_functions_and_dependencies/node.py
+++ b/tests/workflows/basic_inline_prompt_node_with_functions_and_dependencies/node.py
@@ -1,0 +1,14 @@
+from vellum.workflows.nodes import BaseNode
+
+from tests.workflows.basic_inline_prompt_node_with_functions_and_dependencies.inputs import WorkflowInputs
+
+
+class StartNode(BaseNode):
+    noun = WorkflowInputs.noun
+
+    class Outputs(BaseNode.Outputs):
+        final_noun: str
+
+    def run(self) -> Outputs:
+        final_value = "animal"
+        return self.Outputs(final_noun=final_value)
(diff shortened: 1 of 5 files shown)


tests/workflows/basic_inline_prompt_node_with_functions_and_dependencies/prompt_node.py
```diff
@@ -0,0 +1,32 @@
...
+    prompt_inputs = {
+        "noun": LazyReference("StartNode.Outputs.final_noun").coalesce(WorkflowInputs.noun),
+    }
+    functions = [
+        FunctionDefinition(
+            name="favorite_noun",
+            description="Returns the favorite noun of the user",
+            parameters={},
+        ),
+    ]
```
dvargasfuertes, line 32: [nit] are the functions relevant for this test? if not, let's remove

tests/workflows/basic_inline_prompt_node_with_functions_and_dependencies/node.py
```diff
@@ -0,0 +1,14 @@
...
+
+class StartNode(BaseNode):
+    noun = WorkflowInputs.noun
+
+    class Outputs(BaseNode.Outputs):
+        final_noun: str
+
+    def run(self) -> Outputs:
+        final_value = "animal"
+        return self.Outputs(final_noun=final_value)
```
dvargasfuertes, line 14: [nit] can just do:
```python
        final_noun = "animal"
```

dvargas92495 (approved):
love it - Gonna be pairing with @NgoHarrison at around 2pest today, will reach out to pair with you on this after

PR #1291, opened 2025-04-01
[APO-271] Fix node inputs key name for inline prompt nodes

NgoHarrison:


ee/vellum_ee/workflows/display/nodes/vellum/inline_prompt_node.py
```diff
@@ -94,7 +94,7 @@ def _generate_node_and_prompt_inputs(
                 input_name=variable_name,
                 value=variable_value,
                 display_context=display_context,
-                input_id=self.node_input_ids_by_name.get(variable_name),
+                input_id=self.node_input_ids_by_name.get(f"{InlinePromptNode.prompt_inputs.name}.{variable_name}"),
```
dvargasfuertes: [blocking] let's get a `or self.node_input_ids_by_name.get(variable_name)` at the end here like we do in the other nodes so that we are backwards compatible in the short term

dvargasfuertes: [nit] a test confirming this logic as a fast follow would be great - Id take a look at `ee/vellum_ee/workflows/display/nodes/vellum/tests/test_code_execution_node.py` as an example

dvargas92495 (approved):

ee/vellum_ee/workflows/display/nodes/vellum/inline_prompt_node.py
NgoHarrison: Added the test and appended the `or` snippet logic

PR #1292, opened 2025-04-01
Add support for any future parent context

dvargas92495:
Last time we introduce a new parent context, it took down (live) workflows.

This should make all _future_ added parent contexts, which we know we need to soon add (test suite, test case, metric), so that we are always backwards compatible and casting to some safer `UNKNOWN` parent context, that uses just the span id.

### DIFF
--- a/src/vellum/workflows/events/tests/test_event.py
+++ b/src/vellum/workflows/events/tests/test_event.py
@@ -4,6 +4,7 @@ from uuid import UUID
 
 from deepdiff import DeepDiff
 
+from vellum.client.core.pydantic_utilities import UniversalBaseModel
 from vellum.workflows.constants import undefined
 from vellum.workflows.errors.types import WorkflowError, WorkflowErrorCode
 from vellum.workflows.events.node import (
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/events/tests/test_event.py
```diff
@@ -419,3 +420,30 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
...
+        "parent": {
+            "type": "SOME_FUTURE_ENTITY",
+            "span_id": "123e4567-e89b-12d3-a456-426614174000",
+            "some_randome_field": "some_random_value",
+            "parent": None,
+        },
+    }
+
+    # AND a dataclass that references the parent context
+    class MyData(UniversalBaseModel):
```
dvargasfuertes, line 438: Doing a custom dataclass here bc one of the classes that uses `ParentContext` is in vembda

marlenc3 (approved):
✅  hell yeah

noanflaherty (commented):
Very cool

PR #1293, opened 2025-04-01
Prevent future prompt node regressions

dvargas92495:
Context: https://vellum-ai.slack.com/archives/C04CQ2DULL8/p1743536676556849?thread_ts=1743535475.529549&cid=C04CQ2DULL8
Screenshot 2025-04-01 at 4 24 12 PM


src/vellum/client/core/jsonable_encoder.py
```diff
@@ -45,7 +45,7 @@ def jsonable_encoder(obj: Any, custom_encoder: Optional[Dict[Any, Callable[[Any]
             encoder = getattr(obj.__config__, "json_encoders", {})  # type: ignore # Pydantic v1
         if custom_encoder:
             encoder.update(custom_encoder)
-        obj_dict = obj.model_dump(mode="json")
+        obj_dict = obj.dict()
```
dvargasfuertes, line 48: this change needs to be reverted in client generator as well

marlenc3 (commented):
✅

PR #1294, opened 2025-04-01
Fast follow - clean up import order ignore in expression generator

NgoHarrison:
Follow up from this comment:
https://github.com/vellum-ai/vellum-python-sdks/pull/1278#discussion_r2021877029

### DIFF
--- a/ee/codegen/src/generators/expression.ts
+++ b/ee/codegen/src/generators/expression.ts
@@ -1,4 +1,3 @@
-// eslint-disable-next-line import/order
 import { python } from "@fern-api/python-ast";
 import { AstNode } from "@fern-api/python-ast/core/AstNode";
 import { Writer } from "@fern-api/python-ast/core/Writer";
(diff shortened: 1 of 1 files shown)


ee/codegen/src/generators/expression.ts
```diff
@@ -19,6 +15,17 @@ export declare namespace Expression {
   }
 }
 
+// This is to replace the usage of instanceof when checking for NodeInput
+// due to a circular dependency
+function isNodeInput(node: AstNode): boolean {
```
dvargasfuertes:
```suggestion
NodeInput = AstNode & {
      nodeInputValuePointer: {
        nodeInputValuePointerData?: {
          rules?: Array<{ type: string }>;
        };
      };
    }

function isNodeInput(node: AstNode): node is NodeInput {
```

does this let us remove the cast below?

dvargas92495 (approved):

ee/codegen/src/generators/expression.ts
NgoHarrison: Yupppp

PR #1297, opened 2025-04-02
Missing Lazy Reference Fix

marlenc3:
** should fix https://linear.app/vellum/issue/APO-242/try-node-adorned-nodes-appear-to-have-stale-values and possibly https://linear.app/vellum/issue/EPD-2023/investigate-regression-in-dronedeploy-workflows-with-empty-chat


tests/workflows/basic_inline_prompt_node_with_functions_and_dependencies/tests/test_workflow.py
```diff
@@ -95,13 +92,7 @@ def generate_prompt_events(*args: Any, **kwargs: Any) -> Iterator[ExecutePromptE
...
             ),
         ],
-        functions=[
-            FunctionDefinition(
-                name="favorite_noun",
-                description="Returns the favorite noun of the user",
-                parameters={},
-            ),
-        ],
+        functions=None,
```
dvargasfuertes, line 92: [nit] let's also remove the `FunctionCallVellumValue` output

dvargas92495 (approved):

PR #1299, opened 2025-04-02
Make ExecutionContext::trace_id non-optional

dvargas92495:
Our Workflow Server has two references to trace_id in the request and we could delete one of them by making this required

### DIFF
--- a/src/vellum/workflows/context.py
+++ b/src/vellum/workflows/context.py
@@ -1,7 +1,7 @@
 from contextlib import contextmanager
 from dataclasses import field
 import threading
-from uuid import UUID, uuid4
+from uuid import UUID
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/context.py
```diff
@@ -8,8 +9,8 @@
 
 
 class ExecutionContext(UniversalBaseModel):
+    trace_id: UUID = field(default_factory=lambda: UUID("00000000-0000-0000-0000-000000000000"))
```
dvargasfuertes, line 12: we use UUID 0 to replace `None` so that we always have a `UUID` type here

NgoHarrison (commented):
✅

PR #1300, opened 2025-04-02
Create a pass through store so that workflow server doesn't store events

dvargas92495:
This enables our Python Workflow Server to opt out of storing events in memory, far decreasing the memory footprint of each workflow execution

### DIFF
--- a/tests/workflows/pass_through_store/tests/test_workflow.py
+++ b/tests/workflows/pass_through_store/tests/test_workflow.py
@@ -1,0 +1,20 @@
+from vellum.workflows.state.store import PassThroughStore
+
+from tests.workflows.pass_through_store.workflow import WorkflowWithPassThroughStore
+
+
+def test_run_workflow__happy_path():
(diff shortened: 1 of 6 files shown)


tests/workflows/pass_through_store/tests/test_workflow.py
```diff
@@ -0,0 +1,20 @@
+from vellum.workflows.state.store import PassThroughStore
+
+from tests.workflows.pass_through_store.workflow import WorkflowWithPassThroughStore
+
+
+def test_run_workflow__happy_path():
+    # GIVEN a trivial workflow with access to a pass through store
+    workflow = WorkflowWithPassThroughStore(
+        store=PassThroughStore(),
```
dvargasfuertes: this is what usage looks like

dvargas92495 (commented):
On second thought, I'm going to rename to just `EmptyStore()`

NgoHarrison (commented):
✅

PR #1302, opened 2025-04-02
[APO-218] Add codegen support for ml model attribute on inline prompt node

NgoHarrison:
This is part 1 of sdk support on transitioning inline prompt node `ml_model` attr derivation. I will do another PR for the serialization side

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
@@ -1297,3 +1297,32 @@ class PromptNodeDisplay(BaseInlinePromptNodeDisplay[PromptNode]):
     )
 "
 `;
+
+exports[`InlinePromptNode > with ml model node attribute defined > getNodeFile 1`] = `
+"from vellum import JinjaPromptBlock, PromptParameters
+from vellum.workflows.nodes.displayable import InlinePromptNode
+
+from ..inputs import Inputs
+
+
+class PromptNode(InlinePromptNode):
+    ml_model = "gpt-4"
+    blocks = [
+        JinjaPromptBlock(template="""Summarize what this means {{ INPUT_VARIABLE }}"""),
+    ]
+    prompt_inputs = {
+        "text": Inputs.text,
+    }
+    parameters = PromptParameters(
+        stop=[],
+        temperature=0,
+        max_tokens=1000,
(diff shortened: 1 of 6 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
```diff
@@ -1297,3 +1297,32 @@ class PromptNodeDisplay(BaseInlinePromptNodeDisplay[PromptNode]):
...
+
+
+class PromptNode(InlinePromptNode):
+    ml_model = "gpt-4"
+    blocks = [
+        JinjaPromptBlock(template="""Summarize what this means {{ INPUT_VARIABLE }}"""),
+    ]
+    prompt_inputs = {
+        "text": Inputs.text,
+    }
```
dvargasfuertes, line 1315: [oos] we should clean up these tests eventually, such that these don't generate in the default factory case, and that way we could focus each of these tests on the incremental functionality added

ee/codegen/src/__test__/nodes/inline-prompt-node.test.ts
```diff
@@ -232,4 +233,42 @@ describe("InlinePromptNode", () => {
       expect(await writer.toStringFormatted()).toMatchSnapshot();
     });
   });
+
+  describe("with ml model node attribute defined", async () => {
+    let node: InlinePromptNode;
+    beforeEach(async () => {
+      const randomMlModelAttrId = "some-ml-mlmodel-attr-id";
```
dvargasfuertes: [nit] let's keep these `uuid4()`s

ee/codegen/src/generators/nodes/inline-prompt-node.ts
```diff
@@ -47,12 +48,32 @@ export class InlinePromptNode extends BaseSingleFileNode<
           block.blockType === "FUNCTION_DEFINITION"
       );
 
-    statements.push(
-      python.field({
-        name: "ml_model",
-        initializer: python.TypeInstantiation.str(nodeData.mlModelName),
-      })
+    const mlModelAttr = this.nodeData.attributes?.find(
+      (attr) => attr.name === "ml_model"
```
dvargasfuertes, line 52: [nit] let's start defining these constants towards the top of the file, we may source them form elsewhere in the future

dvargas92495 (approved):

PR #1305, opened 2025-04-02
Stand up test for conditional node cast edge case

dvargas92495:
This is preventing users from doing simple APINode status code output = 200 checks. This PR simply sets up the test reproducing the failure, my plan for the fix is a bit intricate so want to space it out


marlenc3 (approved):

NgoHarrison (commented):
✅

PR #1306, opened 2025-04-02
[APO-316] Explicitly check for undefined when checking node output descriptors

NgoHarrison:
Context: This came up from a backfill issue but surfaced this bug: when we have "falsy" values like empty strings/0/empty list, we return an exception when the final output value should still resolve to those values.


src/vellum/workflows/descriptors/tests/test_utils.py
```diff
@@ -162,3 +163,34 @@ class Outputs(BaseNode.Outputs):
...
+    # WHEN we run the node
+    empty_string_output = empty_string_node.run()
+
+    # Assert that the output has the correct value
+    assert empty_string_output.value == ""
+
+    # Define a FinalOutputNode subclass with a zero value
+    class ZeroOutput(FinalOutputNode[FixtureState, int]):
+        class Outputs(FinalOutputNode.Outputs):
+            value = 0
```
dvargasfuertes: [nit] Typically preferred to use a `@pytest.mark.parameterize` rather than have two tests within 1

src/vellum/workflows/descriptors/tests/test_utils.py
```diff
@@ -162,3 +163,34 @@ class Outputs(BaseNode.Outputs):
 def test_resolve_value__happy_path(descriptor, expected_value):
     actual_value = resolve_value(descriptor, FixtureState())
     assert actual_value == expected_value
+
+
+def test_resolve_value__for_falsy_values():
```
dvargasfuertes: [nit] Let's also move this to `src/vellum/workflows/nodes/bases/tests/test_base_node.py` since I think this is an issue that affects all nodes not just final output nodes and is not utility specific

dvargas92495 (approved):

vincent0426 (approved):

PR #1308, opened 2025-04-03
Cast RHS based on the inferred equality type of LHS

dvargas92495:
We use new type inference logic to be able to infer when we should cast our right hand sides to support numeric equality. We scope out inferring node outputs, since that will be its own challenge

### DIFF
--- a/ee/codegen/src/generators/conditional-node-port.ts
+++ b/ee/codegen/src/generators/conditional-node-port.ts
@@ -4,6 +4,7 @@ import { OperatorType } from "@fern-api/python-ast/OperatorType";
 import { AstNode } from "@fern-api/python-ast/core/AstNode";
 import { Writer } from "@fern-api/python-ast/core/Writer";
 import { isNil } from "lodash";
+import { VellumVariableType } from "vellum-ai/api";
 
 import { NodePortGenerationError, ValueGenerationError } from "./errors";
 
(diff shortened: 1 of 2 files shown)


ee/codegen/src/generators/conditional-node-port.ts
```diff
@@ -272,6 +266,92 @@ export class ConditionalNodePort extends AstNode {
...
+      const nodeContext = this.portContext.workflowContext.findNodeContext(
+        rule.data.nodeId
+      );
+      if (nodeContext) {
+        const outputName = nodeContext.getNodeOutputNameById(
+          rule.data.outputId
+        );
+        if (outputName) {
+          // TODO: We need to flesh out the node output types
+          return "JSON";
```
dvargasfuertes, line 289: To be done in the next PR

ee/codegen/src/generators/conditional-node-port.ts
```diff
@@ -272,6 +266,92 @@ export class ConditionalNodePort extends AstNode {
...
+      if (inputVariableContext) {
+        return inputVariableContext.getInputVariableData().type;
+      }
+    }
+
+    if (rule.type === "WORKSPACE_SECRET") {
+      return "STRING";
+    }
+
+    return "JSON";
```
dvargasfuertes, line 312: `"JSON"`s are basically our any type

vincent0426:
✅

marlenc3:
✅

PR #1310, opened 2025-04-03
[APO-271] Add dynamic file loader to workflow context

NgoHarrison:
This is to help with loading code execution script files on vembda. The next followup will be to build this file loader on vembda

### DIFF
--- a/ee/vellum_ee/workflows/tests/test_server.py
+++ b/ee/vellum_ee/workflows/tests/test_server.py
@@ -101,11 +101,6 @@ from .code_execution_node import CodeExecutionNode
 __all__ = ["CodeExecutionNode"]
 """,
         "nodes/code_execution_node/__init__.py": """
-from .code_execution_node import CodeExecutionNode
-
-__all__ = ["CodeExecutionNode"]
-""",
-        "nodes/code_execution_node/code_execution_node.py": """
 from typing import Any
 
 from vellum.workflows.nodes.displayable import CodeExecutionNode as BaseCodeExecutionNode
@@ -122,9 +117,8 @@ class CodeExecutionNode(BaseCodeExecutionNode[BaseState, int]):
 
     namespace = str(uuid4())
 
-    finder = VirtualFileFinder(files, namespace)
-
     # AND the virtual file loader is registered
+    finder = VirtualFileFinder(files, namespace)
     sys.meta_path.append(finder)
 
     # AND we know what the Code Execution Node will respond with
@@ -136,7 +130,7 @@ class CodeExecutionNode(BaseCodeExecutionNode[BaseState, int]):
 
     # WHEN the workflow is loaded
     Workflow = BaseWorkflow.load_from_module(namespace)
-    workflow = Workflow(context=WorkflowContext(dynamic_file_loader=finder.loader))
+    workflow = Workflow(context=WorkflowContext(dynamic_files=files))
 
     # THEN the workflow is successfully initialized
     assert workflow
@@ -145,4 +139,4 @@ class CodeExecutionNode(BaseCodeExecutionNode[BaseState, int]):
     assert event.name == "workflow.execution.fulfilled"
 
     # AND we get the code execution result
-    assert event.body.outputs.final_output == 5.0  # type: ignore
+    assert event.body.outputs == {"final_output": 5.0}
--- a/ee/vellum_ee/workflows/tests/test_server.py
+++ b/ee/vellum_ee/workflows/tests/test_server.py
@@ -1,4 +1,3 @@
-import pytest
 import sys
 from uuid import uuid4
 from typing import Type, cast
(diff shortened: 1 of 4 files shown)


ee/vellum_ee/workflows/tests/test_server.py
```diff
@@ -91,19 +85,27 @@ def test_load_from_module__ts_code_in_file_loader(
...
+        "nodes/__init__.py": """
+from .code_execution_node import CodeExecutionNode
+
+__all__ = ["CodeExecutionNode"]
+""",
+        "nodes/code_execution_node/__init__.py": """
 from .code_execution_node import CodeExecutionNode
 
 __all__ = ["CodeExecutionNode"]
 """,
```
dvargasfuertes: [nit] we actually don't code gen this today and put the node in the __init__.py file currently

ee/vellum_ee/workflows/tests/test_server.py
```diff
@@ -120,11 +122,10 @@ class CodeExecutionNode(BaseCodeExecutionNode[BaseState, int]):
 
     namespace = str(uuid4())
 
-    # AND the virtual file loader is registered
-    sys.meta_path.append(VirtualFileFinder(files, namespace))
+    finder = VirtualFileFinder(files, namespace)
```
dvargasfuertes: [nit] let's keep this line under the ` # AND the virtual file loader is registered`

ee/vellum_ee/workflows/tests/test_server.py
```diff
@@ -135,16 +136,13 @@ class CodeExecutionNode(BaseCodeExecutionNode[BaseState, int]):
...
-    assert event.name == "workflow.execution.fulfilled", event.model_dump_json()
-
-    # AND we pass in the correct file path to the open function
-    assert mock_open.call_args[0][0] == f"{namespace}/nodes/./script.ts"
+    assert event.name == "workflow.execution.fulfilled"
 
-    # AND we invoke the Code Execution Node with the correct code
-    assert vellum_client.execute_code.call_args.kwargs["code"] == ts_code
+    # AND we get the code execution result
+    assert event.body.outputs.final_output == 5.0  # type: ignore
```
dvargasfuertes: [nit] can avoid the type ignore for now with `event.body.outputs == {"final_output": 5.0}`

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
```diff
@@ -1,19 +1,31 @@
...
+    node_filepath: str, script_filepath: str, context: Optional[WorkflowContext] = None
+) -> Union[str, None]:
     node_filepath_dir = os.path.dirname(node_filepath)
     full_filepath = os.path.join(node_filepath_dir, script_filepath)
 
+    # If dynamic file loader is present, try and read the code from there
+    if context and context.dynamic_file_loader:
+        normalized_path = os.path.normpath(full_filepath)
+
+        code = context.dynamic_file_loader._get_code(normalized_path)
```
dvargasfuertes: [nit] sorry was being a bit misleading when I called this `dynamic_file_loader` - I think we _only_ want to pass through the file map (or some other class that wraps it) and keep the VirtualFileLoader itself away from the end user.

src/vellum/workflows/state/context.py
```diff
@@ -8,6 +8,7 @@
 from vellum.workflows.outputs.base import BaseOutputs
 from vellum.workflows.references.constant import ConstantValueReference
 from vellum.workflows.vellum_client import create_vellum_client
+from vellum_ee.workflows.server.virtual_file_loader import VirtualFileLoader
```
dvargasfuertes: [blocking] ah yea then I'll double down on my nit from above. As a hard rule (that I want to set up a lint rule for), `src/vellum` (free license) should _never_ import from `vellum_ee` (commercial license)

NgoHarrison: Updated

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
NgoHarrison: Ah gotcha. I updated the context to pass in the actual file mapping dict instead

src/vellum/workflows/nodes/displayable/code_execution_node/utils.py
```diff
@@ -14,17 +14,20 @@
 def read_file_from_path(
     node_filepath: str, script_filepath: str, context: Optional[WorkflowContext] = None
 ) -> Union[str, None]:
-    node_filepath_dir = os.path.dirname(node_filepath)
-    full_filepath = os.path.join(node_filepath_dir, script_filepath)
```
dvargasfuertes: [nit] still looks like we could keep these two centralized here and strip out namespace within the if block

dvargas92495 (approved):

PR #1311, opened 2025-04-03
[APO-315] Codegen uuid on prompt deployment id fallback

NgoHarrison:


dvargas92495 (approved):
Let's fast follow with test edit/addition for this case 🙏

NgoHarrison:
> Let's fast follow with test edit/addition for this case 🙏

@dvargas92495 
The existing test is here:
https://github.com/vellum-ai/vellum-python-sdks/blob/3c80976a7fcf146685657fdd55dcc17e9db79dd4/ee/codegen/src/__test__/nodes/prompt-deployment-node.test.ts

this change just updated the snapshot so that the codegen is now codegening a uuid instead of str

PR #1312, opened 2025-04-03
Fix codegen to generate correct ImageChatMessageContent object

NgoHarrison:
Context: While debugging https://vellum-ai.slack.com/archives/C081CFHPESE/p1743701472326479 I encountered a pydantic issue when running the workflow locally

Before our codegen was codegening this shape:

```
ImageChatMessageContent(
    src="foo",
      metadata={
            "bar": "baz"  },
       ),
)
```

but the shape for this object is:
https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/client/types/image_chat_message_content.py
```
ImageChatMessageContent(
   value=VellumImage(
     src="foo",
     metadata={
            "bar": "baz"  },
     ),
)
)
```


dvargas92495 (approved):
good shit

PR #1316, opened 2025-04-03
[APO-320] Add error handling for invalid response

vincent0426:
this should only happen if users send a request to `app.vellum.ai` instead of the valid API endpoint, other endpoint will return error like

`  <p>The requested URL <code>/v1/workflows/f1477885-2a17-4c60-ab6e-6ce3aec94822/pull</code> was not found on this server.  <ins>That’s all we know.</ins>`

but `app.vellum.ai` will send back 200 with an invalid zip file

### DIFF
--- a/ee/vellum_cli/pull.py
+++ b/ee/vellum_cli/pull.py
@@ -151,68 +151,71 @@ def pull_command(
 
     error_content = ""
 
-    with zipfile.ZipFile(zip_buffer) as zip_file:
-        if METADATA_FILE_NAME in zip_file.namelist():
-            metadata_json: Optional[dict] = None
-            with zip_file.open(METADATA_FILE_NAME) as source:
-                metadata_json = json.load(source)
+    try:
+        with zipfile.ZipFile(zip_buffer) as zip_file:
+            if METADATA_FILE_NAME in zip_file.namelist():
+                metadata_json: Optional[dict] = None
+                with zip_file.open(METADATA_FILE_NAME) as source:
+                    metadata_json = json.load(source)
 
-            pull_contents_metadata = PullContentsMetadata.model_validate(metadata_json)
+                pull_contents_metadata = PullContentsMetadata.model_validate(metadata_json)
 
-            if pull_contents_metadata.runner_config:
-                workflow_config.container_image_name = pull_contents_metadata.runner_config.container_image_name
-                workflow_config.container_image_tag = pull_contents_metadata.runner_config.container_image_tag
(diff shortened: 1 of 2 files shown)


ee/vellum_cli/pull.py
```diff
@@ -151,68 +151,71 @@ def pull_command(
...
+                        continue
+                    if file_name == METADATA_FILE_NAME:
+                        continue
+
+                    target_file = os.path.join(target_dir, file_name)
+                    os.makedirs(os.path.dirname(target_file), exist_ok=True)
+                    with open(target_file, "w") as target:
+                        logger.info(f"Writing to {target_file}...")
+                        target.write(content)
+    except Exception:
```
dvargasfuertes: [nit] let's make this targeted to the specific ZipFileException or whatever was thrown

ee/vellum_cli/pull.py
```diff
@@ -151,68 +151,71 @@ def pull_command(
...
+        raise Exception("Please check if API URL is set correctly.")
```
dvargasfuertes:
```suggestion
        raise Exception("The API we tried to pull from returned an invalid zip file. Please make sure your `VELLUM_API_URL` environment variable is set correctly.")
```

ee/vellum_cli/pull.py
```diff
@@ -151,68 +151,71 @@ def pull_command(
...
-                    continue
-                if file_name == METADATA_FILE_NAME:
-                    continue
-
-                target_file = os.path.join(target_dir, file_name)
-                os.makedirs(os.path.dirname(target_file), exist_ok=True)
-                with open(target_file, "w") as target:
-                    logger.info(f"Writing to {target_file}...")
-                    target.write(content)
+    try:
```
dvargasfuertes, line 154: [q] did the Response's content type header return different answers for the valid and invalid case? If so, we could work with Fern to return that info to us via the generated SDK and use that to check earlier

dvargas92495 (approved):

ee/vellum_cli/pull.py
vincent0426: for `app.vellum.ai` I print before and get Content-Type: text/html; charset=utf-8

https://github.com/vellum-ai/vellum-python-sdks/blob/c9b4bbdc04fc84e30fbf776057b97447e4a11c9f/src/vellum/client/resources/workflows/client.py#L71

for others
it raises the error here https://github.com/vellum-ai/vellum-python-sdks/blob/c9b4bbdc04fc84e30fbf776057b97447e4a11c9f/src/vellum/client/resources/workflows/client.py#L87

PR #1323, opened 2025-04-03
[APO-271] Update codegen to now always persist standalone script files for code execution nodes

NgoHarrison:
This is the final follow piece for https://github.com/vellum-ai/vellum-python-sdks/pull/1310

Now we can rip out the logic of standalone vs inline and just have codegen always persist the script file as standalone and vembda can read from this


dvargas92495 (approved):
legooo

PR #1324, opened 2025-04-03
Add error handling for ping

vincent0426:


ee/vellum_cli/ping.py
```diff
@@ -10,8 +11,13 @@ def ping_command():
...
 
-    workspace = client.workspaces.workspace_identity()
-    organization = client.organizations.organization_identity()
+    try:
+        workspace = client.workspaces.workspace_identity()
+        organization = client.organizations.organization_identity()
+    except ApiError:
+        raise Exception(
+            "The API we tried to ping returned an invalid response. Please make sure your `VELLUM_API_URL` environment variable is set correctly."  # noqa: E501
+        )
```
dvargasfuertes, line 20: [oos] let's see if we could be more granular here. Are we returning a `401` if the api key is missing? we should raise an exception talking about the `VELLUM_API_KEY` in that case

dvargas92495 (approved):

PR #1325, opened 2025-04-03
Add 403 error handling for ping

vincent0426:
follow up https://github.com/vellum-ai/vellum-python-sdks/pull/1324

### DIFF
--- a/ee/vellum_cli/ping.py
+++ b/ee/vellum_cli/ping.py
@@ -16,8 +16,10 @@ def ping_command():
         organization = client.organizations.organization_identity()
     except ApiError as e:
         # If user did not provide an API key, we will get a 403 error
-        if e.status_code == 403:
-            raise e
+        if e.status_code == 403 or e.status_code == 401:
+            raise Exception(
+                "Please make sure your `VELLUM_API_KEY` environment variable is set correctly."  # noqa: E501
+            ) from e
         raise Exception(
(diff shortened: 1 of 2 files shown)


ee/vellum_cli/tests/test_ping.py
```diff
@@ -65,3 +63,22 @@ def test_ping__error_path(vellum_client):
...
     )
+
+
+def test_ping__403_error_path(vellum_client):
+    # GIVEN a cli
+    runner = CliRunner()
+
+    # GIVEN a 403 error with the error message from the API
+    vellum_client.workspaces.workspace_identity.side_effect = ApiError(
+        status_code=403, body={"detail": "Authentication credentials were not provided."}
```
vincent0426, line 74: this message is what will be returned if user don't provide API key

ee/vellum_cli/ping.py
```diff
@@ -14,7 +14,10 @@ def ping_command():
     try:
         workspace = client.workspaces.workspace_identity()
         organization = client.organizations.organization_identity()
-    except ApiError:
+    except ApiError as e:
+        # If user did not provide an API key, we will get a 403 error
+        if e.status_code == 403:
+            raise e
```
dvargasfuertes: [nit] can we include copy in the error message instructing the user to set their `VELLUM_API_KEY`?

We should also catch 401 here too - the fact that our django server returns a 403 here is a bug

dvargas92495 (approved):

PR #1327, opened 2025-04-04
Re codegen

vincent0426:
these tests should be removed in this PR? https://github.com/vellum-ai/vellum-python-sdks/pull/1323

CI does not catch this as npm test will not trigger a failure?

https://github.com/vellum-ai/vellum-python-sdks/blob/62c01ab03da678cb879287a28f1d6a18ded1d395/ee/codegen/src/__test__/nodes/__snapshots__/code-execution-node.test.ts.snap#L115-L198


NgoHarrison (approved):

dvargas92495:
hm yea we should see how npm test could flag us of this

PR #1332, opened 2025-04-04
Bump vellum version to 0.14.32

NgoHarrison:


dvargas92495 (approved):
did you npm install? I would've expected package-lock changes too

vincent0426 (approved):

NgoHarrison:
> did you npm install? I would've expected package-lock changes too

Blah I forgot to include it, thanks

package-lock.json
dvargasfuertes, line 1: make sure we npm install in `ee/codegen` instead of the root 😆 

fern wiped this out on the 0.14.33 release, so no further action needed

PR #1333, opened 2025-04-04
Fix/Deduplicate api node test

vincent0426:
We have two API node tests that are doing similar things; this PR keeps the one serialization test used. Also, find a bug where our serialization is not getting the correct body value since we define it as `body` field which should be `json`


tests/workflows/basic_vellum_api_node/workflow.py
```diff
@@ -1,27 +0,0 @@
...
-from vellum.workflows.references.vellum_secret import VellumSecretReference
-
-
-class SimpleAPINode(APINode):
-    method = APIRequestMethod.POST
-    url = "https://api.vellum.ai"
-    authorization_type = AuthorizationType.API_KEY
-    api_key_header_key = "CUSTOM_API_KEY"
-    api_key_header_value = VellumSecretReference("MY_SECRET")
-    body = {
```
vincent0426, line 13: this should be json

https://github.com/vellum-ai/vellum-python-sdks/blob/808e8ebc798926d0421232fcc4cf622686a97609/ee/codegen/src/generators/nodes/api-node.ts#L45-L53

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_api_node_serialization.py
```diff
@@ -99,7 +99,7 @@ def test_serialize_workflow(vellum_client):
                     "id": "c3a17ceb-e201-4025-b18a-9162aac7705e",
                     "key": "body",
                     "value": {
-                        "rules": [{"type": "CONSTANT_VALUE", "data": {"type": "JSON", "value": None}}],
+                        "rules": [{"type": "CONSTANT_VALUE", "data": {"type": "JSON", "value": {"key": "value"}}}],
```
vincent0426, line 102: after we change to `json` field

dvargas92495 (approved):

PR #1335, opened 2025-04-04
[APO-352] Pass in generated files to nested workflow contexts

NgoHarrison:
I will fast follow with tests. Wanted to just get the fix out now

### DIFF
--- a/src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
+++ b/src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
@@ -75,7 +75,9 @@ class InlineSubworkflowNode(
         with execution_context(parent_context=get_parent_context()):
             subworkflow = self.subworkflow(
                 parent_state=self.state,
-                context=WorkflowContext(vellum_client=self._context.vellum_client),
+                context=WorkflowContext(
+                    vellum_client=self._context.vellum_client, generated_files=self._context.generated_files
+                ),
             )
             subworkflow_stream = subworkflow.stream(
                 inputs=self._compile_subworkflow_inputs(),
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
```diff
@@ -75,7 +75,9 @@ def run(self) -> Iterator[BaseOutput]:
         with execution_context(parent_context=get_parent_context()):
             subworkflow = self.subworkflow(
                 parent_state=self.state,
-                context=WorkflowContext(vellum_client=self._context.vellum_client),
+                context=WorkflowContext(
```
dvargasfuertes, line 78: [blocking] we're forgetting map node

src/vellum/workflows/nodes/core/try_node/node.py
```diff
@@ -32,7 +32,9 @@ def run(self) -> Iterator[BaseOutput]:
         with execution_context(parent_context=parent_context):
             subworkflow = self.subworkflow(
                 parent_state=self.state,
-                context=WorkflowContext(vellum_client=self._context.vellum_client),
+                context=WorkflowContext(
+                    vellum_client=self._context.vellum_client, generated_files=self._context.generated_files
+                ),
```
dvargasfuertes, line 37: [oos] We should make a static method that allows us to do:

```python
WorkflowContext.from(self._context)
```

so we only have to update one place in the future

dvargas92495 (approved):
map node, then 🚢

src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
NgoHarrison: Good catch, added

src/vellum/workflows/nodes/core/try_node/node.py
NgoHarrison: Oh yeah I like this idea. Can follow up after the tests

PR #1340, opened 2025-04-06
Add verification for snapshot change in CI

vincent0426:
Add `git diff` in CI and use `npm run test:update` to check if snapshot change

CI should fail since we remove a test, later would remove the comment to see if it pass

### DIFF
--- a/.github/workflows/ci-codegen.yaml
+++ b/.github/workflows/ci-codegen.yaml
@@ -52,8 +52,14 @@ jobs:
         working-directory: ee/codegen
 
       - name: Test
-        run: npm test
+        run: npm run test:update
         working-directory: ee/codegen
+
+      - name: Fail if snapshot changes were not committed
+        run: |
+          git diff --exit-code -- 'ee/codegen/**/*.snap' || (
+            echo "Snapshot files were modified. Run 'npm run test:update' and commit the changes." && exit 1
+          )
 
   publish:
     permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-codegen.yaml
```diff
@@ -52,9 +52,15 @@ jobs:
         working-directory: ee/codegen
 
       - name: Test
-        run: npm test
+        run: npm run test:update
```
dvargasfuertes, line 55: [blocking] I'm spooked by introducing this since it passes failing tests, and then we're relying on a global git diff below instead of knowing exactly which tests failed.

Does this achieve the same thing for us?
Screenshot 2025-04-07 at 8 23 57 AM

vincent0426: mm, i dont think that option exists, the other way is to check if there is a warning , or we could skip this one as I think it would only happen if some tests are removed which should be rare and could be fixed easily?

Screenshot 2025-04-07 at 3 32 56 PM

```sh
set -o pipefail
OUTPUT=$(npm run test -- --reporter=verbose 2>&1)
echo "$OUTPUT"
if echo "$OUTPUT" | grep -q "obsolete"; then
  echo "Obsolete snapshots found!"
  exit 1
fi
```

dvargasfuertes: I'd lean closer to failing if we detect "obsolete" in the normal test output than changing our test command. I would think that vitest would have native support for something like this, but if not, happy to diy

PR #1341, opened 2025-04-07
Add tests for nested code execution node for vembda

NgoHarrison:
Follow up for https://github.com/vellum-ai/vellum-python-sdks/pull/1335

### DIFF
--- a/ee/vellum_ee/workflows/tests/test_server.py
+++ b/ee/vellum_ee/workflows/tests/test_server.py
@@ -12,7 +12,6 @@ from vellum_ee.workflows.server.virtual_file_loader import VirtualFileFinder
 
 
 def test_load_workflow_event_display_context():
-    # DEPRECATED: Use `vellum.workflows.events.workflow.WorkflowEventDisplayContext` instead. Will be removed in 0.15.0
     from vellum.workflows.events.workflow import WorkflowEventDisplayContext
 
     # We are actually just ensuring there are no circular dependencies when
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/displayable/code_execution_node/tests/test_code_execution_node.py
```diff
@@ -653,7 +654,9 @@ def main():
         node.run()
 
     # AND the error should contain the execution error details
-    assert exc_info.value.message == "Expected an output of type 'int | float', but received 'str'"
+    assert re.match(
+        r"Expected an output of type '(int \| float|float \| int)', but received 'str'", exc_info.value.message
+    )
```
NgoHarrison, line 660: I have no idea why this is failing now. The actual error message has the union types flipped. I just added this regex to account for both orders. Strangely my pycharm test run is fine without this change

dvargasfuertes: maybe different python versions have different union orderings, interesting

ee/vellum_ee/workflows/tests/test_server.py
```diff
@@ -13,7 +13,7 @@
 
 def test_load_workflow_event_display_context():
     # DEPRECATED: Use `vellum.workflows.events.workflow.WorkflowEventDisplayContext` instead. Will be removed in 0.15.0
```
dvargasfuertes: looks like we could remove this comment if we update the comment on `ee/vellum_ee/workflows/display/types.py::L5` to say it should be removed in `0.15.0`

dvargas92495 (approved):

PR #1342, opened 2025-04-07
Make static method for creating workflow contexts on nested workflows

NgoHarrison:


src/vellum/workflows/state/context.py
```diff
@@ -74,3 +74,7 @@ def _register_node_output_mocks(self, node_output_mocks: MockNodeExecutionArg) -
 
     def _get_all_node_output_mocks(self) -> List[MockNodeExecution]:
         return [mock for mocks in self._node_output_mocks_map.values() for mock in mocks]
+
+    @classmethod
+    def from_(cls, context):
```
dvargasfuertes: [q] why the `_` as a suffix? if it's bc `from` is a reserved keyword, could we choose one that isn't reserved? `of(...)`, `create(...)`, `propagate(...)`, etc

dvargas92495 (approved):

src/vellum/workflows/state/context.py
NgoHarrison: Yeah it's reserved, changed to create_from

PR #1347, opened 2025-04-08
Add test for incorrect lazy refernce

vincent0426:


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1,5 +1,24 @@
...
+from vellum.workflows.nodes.displayable import APINode as BaseAPINode
+from vellum.workflows.references import LazyReference, VellumSecretReference
+
+from .parallel_node_2 import ParallelNode2
+
+
+class APINode(BaseAPINode):
+    url = ParallelNode2.Outputs.result
+    method = APIRequestMethod.POST
+    json = LazyReference("RequestBodyNode.Outputs.result")
```
vincent0426: next PR will fix the lazy reference issue

dvargasfuertes: [nit] I like to leave the snapshot in the "desired end state". so I'd `json = RequestBodyNode.Outputs.result` here and `it.skip()` the test

dvargas92495 (approved):

PR #1348, opened 2025-04-08
Fix lazy reference

vincent0426:
This PR fix the issues in `getOrderedNodes` and fix this test https://github.com/vellum-ai/vellum-python-sdks/pull/1347. Will rebase after merging the upstream one first.

If we have parallel nodes 1 and 2, using BFS will get the order as below. `request-body-node` will appear to be behind the API node, which leads to generating lazy reference.

```python
graph = (
    {
        parallel-node-1 (1)
        >> seq-node-1 (2)
        >> seq-node-2 (3)
        >> request-body-node, (4)
        parallel-node-2 (1)
    }
    >> MergeNode (2)
    >> APINode (3)
    >> FinalOutput
)
```


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1,16 +1,38 @@
...
+from vellum.workflows.references import VellumSecretReference
+
+from .parallel_node_2 import ParallelNode2
+from .request_body_node import RequestBodyNode
+
+
+class APINode(BaseAPINode):
+    url = ParallelNode2.Outputs.result
+    method = APIRequestMethod.POST
+    json = RequestBodyNode.Outputs.result
```
vincent0426, line 15: the fix

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1,16 +1,38 @@
...
+exports[`WorkflowProjectGenerator > Nodes present but not in graph > should generate unused_graphs if final output is not used 1`] = `
 "from vellum.workflows import BaseWorkflow
 
+from .nodes.final_output import FinalOutput
 from .nodes.generic_node import GenericNode
 
 
 class Workflow(BaseWorkflow):
     graph = GenericNode
+    unused_graphs = {FinalOutput}
```
vincent0426, line 32: We are omitting final output before thie PR if we have a node connected.

https://www.loom.com/share/db0c2f579d8f42749a13e1e8d78afef0?sid=a7805ec3-d904-4960-9b3d-1fcee12b5f61

Final output is included in the base case because we early return.

https://github.com/vellum-ai/vellum-python-sdks/blob/4250dfad9c6016ea745227f010b709aeff2c6303/ee/codegen/src/project.ts#L431-L435

but filter later
https://github.com/vellum-ai/vellum-python-sdks/blob/4250dfad9c6016ea745227f010b709aeff2c6303/ee/codegen/src/project.ts#L495-L503

dvargasfuertes: woah nice find!

ee/codegen_integration/fixtures/faa_q_and_a_bot/code/nodes/final_output_2.py
```diff
@@ -1,16 +1,16 @@
...
 from .prompt_node_14 import PromptNode14
+from .prompt_node_18 import PromptNode18
 from .prompt_node_19 import PromptNode19
+from .prompt_node_9 import PromptNode9
 
 
 class FinalOutput2(FinalOutputNode[BaseState, str]):
     class Outputs(FinalOutputNode.Outputs):
         value = (
-            LazyReference("PromptNode9.Outputs.text")
```
dvargasfuertes, line 12: oh hey, another bug fix!

ee/codegen_integration/fixtures/faa_q_and_a_bot/code/nodes/__init__.py
```diff
@@ -19,15 +19,15 @@
...
     "FAADocumentStore",
+    "PromptNode16",
     "PromptNode19",
-    "PromptNode14",
-    "TemplatingNode15",
+    "SubworkflowNode",
     "FormattedSearchResults",
-    "FinalOutput2",
-    "APINode",
+    "TemplatingNode15",
```
dvargasfuertes: [oos] we should really just alphabetize these `__all__` attributes

ee/codegen/src/project.ts
```diff
@@ -426,79 +404,84 @@ ${errors.slice(0, 3).map((err) => {
...
+
+      if (!graph.has(sourceId)) {
+        return;
       }
 
-      const sourceNode = nodesById[edge.sourceNodeId];
-      if (!sourceNode) {
-        continue;
+      if (graph.has(targetId)) {
+        graph.get(sourceId)?.push(targetId);
```
dvargasfuertes: [nit] looks like we could set `graph.get(sourceId)` once, read it twice (the if check and here)

ee/codegen/src/project.ts
```diff
@@ -426,79 +404,84 @@ ${errors.slice(0, 3).map((err) => {
...
+    });
 
-    while (edgesQueue.length > 0) {
-      const edge = edgesQueue.shift();
-      if (!edge) {
-        continue;
+    // Add edges to graph and count in-degrees
+    rawData.edges.forEach((edge) => {
+      const sourceId = edge.sourceNodeId;
+      const targetId = edge.targetNodeId;
```
dvargasfuertes:
```suggestion
      const sourceNodeId = edge.sourceNodeId;
      const targetNodeId = edge.targetNodeId;
```

[nit] to disambiguate from the other ids that could be in play here (handles, edges, etc.) The algo here is a bit tricky, so will be a little more pedantic here trying to make it easier to follow

ee/codegen/src/project.ts
```diff
@@ -426,79 +404,84 @@ ${errors.slice(0, 3).map((err) => {
...
+    const nodesById = Object.fromEntries(nodes.map((node) => [node.id, node]));
 
-    const edgesQueue = this.workflowContext.getEntrypointNodeEdges();
-    const edgesByPortId = this.workflowContext.getEdgesByPortId();
-    const processedEdges = new Set<WorkflowEdge>();
-    const processedNodeIds = new Set<string>();
+    // Create an adjacency list representation of the graph
+    const graph: Map<string, string[]> = new Map();
+    // Track in-degree (number of incoming edges) for each node
+    const inDegree: Map<string, number> = new Map();
```
dvargasfuertes: [q] Correct me if I'm wrong, but aren't we only using `inDegree` to decide which source nodes to start the algorithm at? If so, shouldn't we only need to calculate the initial source nodes?

dvargasfuertes: Oh! I missed this bit:

`if (currentDegree - 1 === 0) { ... }`

I see, we are only adding it to our array of nodes to process if all of it's parents are processed. Neat! I wonder if our graph generation algo would simplify using this same approach now that I think about it...

dvargasfuertes: [blocking] Wait, does this algorithm handle loops?

> A >> B >> C >> D >> B

dvargasfuertes: I think the answer is not really, bc it will just add them at the end as `unprocessedNodes`

dvargas92495 (commented):
I'd like for us to confirm the loop case before 🚢

vincent0426:
found that topo sort should be hard to consider dependency, switch to dependency + DFS and add tests for get order nodes

ee/codegen/src/__test__/ordered-nodes.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    //   parallelNode1
+    //   >> seqNode1
+    //   >> seqNode2
+    //   >> requestBodyNode
+    //   parallelNode2
+    // }
+    // >> mergeNode
+    // >> apiNode
+    // >> finalOutput
+    const orderedNodes = runOrderedNodesTest([
```
vincent0426, line 170: this is the github reviewer test

ee/codegen/src/project.ts
```diff
@@ -421,88 +399,85 @@ ${errors.slice(0, 3).map((err) => {
...
+    rawData.edges.forEach((edge) => {
+      const sourceNodeId = edge.sourceNodeId;
+      const targetNodeId = edge.targetNodeId;
 
-      const sourceNode = nodesById[edge.sourceNodeId];
-      if (!sourceNode) {
-        continue;
+      const targetDependencies = dependencyGraph.get(targetNodeId);
+      const sourceDependencies = dependencyGraph.get(sourceNodeId);
+      if (targetDependencies && sourceDependencies) {
```
dvargasfuertes: [nit] do we need to check for the existence of `sourceDependencies`? but it's not used in the block

ee/codegen/src/project.ts
```diff
@@ -421,88 +399,85 @@ ${errors.slice(0, 3).map((err) => {
...
+    });
 
-      const targetNode = nodesById[edge.targetNodeId];
-      if (!targetNode) {
-        continue;
+    const orderedNodes: WorkflowDataNode[] = [];
+    const visited = new Set<string>();
+    const tempVisited = new Set<string>(); // Used to detect cycles
+
+    // DFS function that ensures dependencies are processed first
```
dvargasfuertes, line 437: Ok I think this works - nice!

dvargas92495 (approved):
good stuff here 💪

NgoHarrison (approved):
Niceee 🔥

ee/codegen_integration/fixtures/faa_q_and_a_bot/code/nodes/__init__.py
vincent0426: will fix in later PR

PR #1357, opened 2025-04-09
[APO-387] Generate | instead of or in condition node

vincent0426:
`branch_1 = Port.on_if(Inputs.text.equals("aaa") or Inputs.text.equals("ddd"))`

`or` operator will only return the first condition that evaluates to True, because Python checks for truthiness and short-circuits.

We manually wrote the pipe condition using `|`, since Fern doesn't support the `|` operator.

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
@@ -33,10 +33,8 @@ from vellum.workflows.references.constant import ConstantValueReference
 class ConditionalNode(BaseConditionalNode):
     class Ports(BaseConditionalNode.Ports):
         branch_1 = Port.on_if(
-            (
-                ConstantValueReference("text").equals("val-1")
-                | ConstantValueReference("text").equals("val-2")
-            )
+            ConstantValueReference("text").equals("val-1")
+            | ConstantValueReference("text").equals("val-2")
         )
 "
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
```diff
@@ -24,6 +24,23 @@ class ConditionalNode(BaseConditionalNode):
...
+exports[`Conditional Node with OR combinator generates pipe operator > should generate code with pipe operator for OR conditions 1`] = `
+"from vellum.workflows.nodes.displayable import ConditionalNode as BaseConditionalNode
+from vellum.workflows.ports import Port
+from vellum.workflows.references.constant import ConstantValueReference
+
+
+class ConditionalNode(BaseConditionalNode):
+    class Ports(BaseConditionalNode.Ports):
+        branch_1 = Port.on_if(
+            (
```
dvargasfuertes: [nit] these extra params look redundant. Are we generating them or is the formatter doing it?

dvargas92495 (approved):

ee/codegen/src/__test__/nodes/__snapshots__/conditional-node.test.ts.snap
vincent0426: ah my bad

NgoHarrison (commented):
✅

PR #1358, opened 2025-04-09
[APO-388] Support json output on prompt deployment

NgoHarrison:


ee/codegen/src/__test__/nodes/__snapshots__/multiple-nodes.test.ts.snap
```diff
@@ -414,3 +414,55 @@ class ConditionalNode(BaseConditionalNode):
...
+from vellum_ee.workflows.display.nodes import BaseTemplatingNodeDisplay
+from vellum_ee.workflows.display.nodes.types import (
+    NodeOutputDisplay,
+    PortDisplayOverrides,
+)
+
+from ...nodes.templating_node import TemplatingNode
+
+
+class TemplatingNodeDisplay(BaseTemplatingNodeDisplay[TemplatingNode]):
```
dvargasfuertes: [nit] were there any changes here? looks like we could omit this snapshot

ee/codegen/src/__test__/nodes/multiple-nodes.test.ts
```diff
@@ -433,3 +433,92 @@ describe("InlinePromptNode json output referenced by TemplatingNode", () => {
...
+      id: "some-id",
+      deploymentId: "947cc337-9a53-4c12-9a38-4f65c04c6317",
+      name: "some-unique-deployment-name",
+      outputs: [
+        {
+          id: uuidv4(),
+          name: "json",
+          type: "JSON",
+        },
+      ],
```
dvargasfuertes: [blocking] our deployment api should not be returning this field - we could reference `DeploymentHistoryItem` to see the shape

src/vellum/workflows/nodes/displayable/prompt_deployment_node/node.py
```diff
@@ -61,3 +70,6 @@ def run(self) -> Iterator[BaseOutput]:
 
         value = "\n".join(string_outputs)
         yield BaseOutput(name="text", value=value)
+
+        if json_output:
+            yield BaseOutput(name="json", value=json_output)
```
dvargasfuertes, line 75: [oos] let's fast follow a test for support here

dvargas92495 (approved):
pending blocking comment cleanup

ee/codegen/src/__test__/nodes/multiple-nodes.test.ts
NgoHarrison: Ah good catch. Removed

ee/codegen/src/__test__/nodes/__snapshots__/multiple-nodes.test.ts.snap
NgoHarrison: Removed

vincent0426:
✅

PR #1359, opened 2025-04-09
Add normalized score field to guard rail node output

NgoHarrison:
Context: While investigating metric node event translation, I would get this error about normalized_score being an unrecognized attribute

### DIFF
--- a/src/vellum/workflows/nodes/displayable/guardrail_node/node.py
+++ b/src/vellum/workflows/nodes/displayable/guardrail_node/node.py
@@ -34,6 +34,7 @@ class GuardrailNode(BaseNode[StateType], Generic[StateType]):
 
     class Outputs(BaseOutputs):
         score: float
+        normalized_score: Optional[float]
         log: Optional[str]
 
     def run(self) -> Outputs:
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/displayable/guardrail_node/node.py
```diff
@@ -34,6 +34,7 @@ class Trigger(BaseNode.Trigger):
 
     class Outputs(BaseOutputs):
         score: float
+        normalized_score: Optional[float]
```
NgoHarrison, line 37: [q] Not sure if we want to keep doing this manually? This is the second time we've had to append an attribute like this and I believe it was said previously that guard rail nodes could have a lot of other types of outputs. Not sure how we want to handle going forward

dvargasfuertes: While we'll want to handle this in a more first class way down the line, `score`, `normalized_score` and `log` are the three we originally hardcoded, so I think it's fine to hardcode those three for now.

Another question we could ask was "what was the down stream system that failed because this was missing?" and see if we want to make that piece more resilient

dvargas92495 (approved):

PR #1360, opened 2025-04-09
Eric/workflow sandbox id in pull deployment

elacava06:
Adds the workflow_sandbox_id to the lockfile if it exists in the metadata.json when pulling a deployment

Will be putting up a complimentary PR in the main repo to include that field in the metadata


dvargas92495 (approved):
🚀

ee/vellum_cli/pull.py
```diff
@@ -198,6 +199,9 @@ def pull_command(
                     workflow_config.container_image_tag = pull_contents_metadata.runner_config.container_image_tag
                     if workflow_config.container_image_name and not workflow_config.container_image_tag:
                         workflow_config.container_image_tag = "latest"
+                workflow_config.workflow_sandbox_id = (
+                    workflow_config.workflow_sandbox_id or pull_contents_metadata.workflow_sandbox_id
```
dvargasfuertes: CI/mypy is complaining bc left hand side is technically str | None and the new `pull_contents_metadata.workflow_sandbox_id` is `UUID | None`. Should be resolved with a `str()` cast

vincent0426: [q] do we return `workflow_sandbox_id` from the server in `metadata.json`?

elacava06: Not yet. Putting up a second pr in the other repo that handles that.

vincent0426 (approved):

dvargas92495:
looks like github had an incident preventing tests from rerunning - feel free to merged!

PR #1367, opened 2025-04-10
[APO-277] Cutover prompt deployments to use new api

NgoHarrison:


ee/codegen/src/__test__/nodes/prompt-deployment-node.test.ts
```diff
@@ -124,20 +184,43 @@ describe("PromptDeploymentNode", () => {
...
+          id: "mocked-environment-id",
+          name: "mocked-environment-name",
+          label: "mocked-environment-label",
+        },
+        createdBy: {
+          id: "mocked-created-by-id",
+          email: "mocked-created-by-email",
+        },
+        promptVersion: {
+          id: "mocked-prompt-release-id",
```
dvargasfuertes, line 202: [nit] this isn't true right?

dvargas92495 (approved):

ee/codegen/src/__test__/nodes/prompt-deployment-node.test.ts
NgoHarrison: Wdym?

dvargasfuertes: as in, there's no `id` field in the `promptVersion` field in this response correct?

PR #1368, opened 2025-04-10
[APO-390] Add circular loop test that would fail without fix

vincent0426:
Test for circular loop that will fail before fix


tests/workflows/circular_loop/tests/test_workflow.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from tests.workflows.circular_loop.workflow import CircularLoopWorkflow
+
+
+@pytest.mark.skip(reason="This test will fail because of circular loop")
```
dvargasfuertes: [nit] let's link the ticket in the `reason`

dvargas92495 (approved):

PR #1376, opened 2025-04-11
[APO-338 APO-339 APO-351] Init tool calling node

vincent0426:
Draft PR for tool calling node

### DIFF
--- a/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
+++ b/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
@@ -1,0 +1,126 @@
+from collections.abc import Callable
+from typing import ClassVar, List, Optional
+
+from vellum import ChatMessage, FunctionDefinition, PromptBlock
+from vellum.workflows.context import execution_context, get_parent_context
+from vellum.workflows.errors.types import WorkflowErrorCode
+from vellum.workflows.exceptions import NodeException
+from vellum.workflows.graph.graph import Graph
+from vellum.workflows.inputs.base import BaseInputs
+from vellum.workflows.nodes.bases import BaseNode
+from vellum.workflows.nodes.experimental.tool_calling_node.utils import (
+    ToolRouterNode,
+    create_function_node,
+    create_tool_router_node,
+)
+from vellum.workflows.outputs.base import BaseOutputs
+from vellum.workflows.state.base import BaseState
+from vellum.workflows.state.context import WorkflowContext
+from vellum.workflows.types.core import EntityInputsInterface, MergeBehavior
+from vellum.workflows.workflows.base import BaseWorkflow
+
+
+class ToolCallingNode(BaseNode):
+    """
+    A Node that allows for tool calling while dynamically generating a graph based on provided functions.
+
+    Attributes:
+        ml_model: str - The model to use for tool calling (e.g., "gpt-4o-mini")
+        blocks: List[PromptBlock] - The prompt blocks to use (same format as InlinePromptNode)
+        functions: List[FunctionDefinition] - The functions that can be called
+        prompt_inputs: Optional[EntityInputsInterface] - Mapping of input variable names to values
+    """
+
+    ml_model: ClassVar[str] = "gpt-4o-mini"
+    blocks: ClassVar[List[PromptBlock]] = []
+    functions: ClassVar[List[FunctionDefinition]] = []
+    function_callables: ClassVar[List[Callable]] = []
+    prompt_inputs: ClassVar[Optional[EntityInputsInterface]] = None
+
+    class Trigger(BaseNode.Trigger):
+        merge_behavior = MergeBehavior.AWAIT_ANY
+
+    class Outputs(BaseOutputs):
+        """
+        The outputs of the ToolCallingNode.
+
+        text: The final text response after tool calling
+        chat_history: The complete chat history including tool calls
+        """
+
+        text: str = ""
+        chat_history: List[ChatMessage] = []
+
+    def run(self) -> Outputs:
+        """
+        Run the tool calling workflow.
+
+        This dynamically builds a graph with router and function nodes,
+        then executes the workflow.
+        """
+        # TODO: We should parse input values and add them to the state
+        self._build_graph()
+
+        with execution_context(parent_context=get_parent_context()):
+
+            class ToolCallingState(BaseState):
+                chat_history: List[ChatMessage] = []
+
+            class ToolCallingWorkflow(BaseWorkflow[BaseInputs, ToolCallingState]):
+                graph = self._graph
+
+                class Outputs(BaseWorkflow.Outputs):
+                    text: str = ToolRouterNode.Outputs.text
+                    chat_history: List[ChatMessage] = ToolCallingState.chat_history
+
+            subworkflow = ToolCallingWorkflow(
+                parent_state=self.state,
+                context=WorkflowContext(
+                    vellum_client=self._context.vellum_client, generated_files=self._context.generated_files
+                ),
+            )
+
+            terminal_event = subworkflow.run()
+
+            if terminal_event.name == "workflow.execution.paused":
+                raise NodeException(
+                    code=WorkflowErrorCode.INVALID_OUTPUTS,
+                    message="Subworkflow unexpectedly paused",
+                )
+            elif terminal_event.name == "workflow.execution.fulfilled":
+                node_outputs = self.Outputs()
+
+                for output_descriptor, output_value in terminal_event.outputs:
+                    setattr(node_outputs, output_descriptor.name, output_value)
+
+                return node_outputs
+            elif terminal_event.name == "workflow.execution.rejected":
+                raise Exception(f"Workflow execution rejected: {terminal_event.error}")
+
+            raise Exception(f"Unexpected workflow event: {terminal_event.name}")
+
+    def _build_graph(self) -> None:
+        self.tool_router_node = create_tool_router_node(
+            model_name=self.ml_model,
+            prompt_blocks=self.blocks,
+            functions=self.functions,
+            input_values=self.prompt_inputs,
+        )
+
+        self._function_nodes = {function.name: create_function_node(function) for function in self.functions}
+
+        graph_set = set()
+
+        # Add connections from ports of router to function nodes and back to router
+        for function_name, FunctionNodeClass in self._function_nodes.items():
+            # TODO: We should think about how to handle this
+            if function_name is None:
+                raise ValueError("Function name is required")
+            router_port = getattr(self.tool_router_node.Ports, function_name)
+            edge_graph = router_port >> FunctionNodeClass >> self.tool_router_node
+            graph_set.add(edge_graph)
+
+        default_port = getattr(self.tool_router_node.Ports, "default")
+        graph_set.add(default_port)
+
+        self._graph = Graph.from_set(graph_set)
--- a/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
+++ b/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
@@ -1,5 +1,5 @@
 from collections.abc import Callable
-from typing import ClassVar, List, Optional
+from typing import ClassVar, Dict, List, Optional
 
 from vellum import ChatMessage, FunctionDefinition, PromptBlock
 from vellum.client.types.chat_message_request import ChatMessageRequest
@@ -17,7 +17,7 @@ from vellum.workflows.nodes.experimental.tool_calling_node.utils import (
 from vellum.workflows.outputs.base import BaseOutputs
 from vellum.workflows.state.base import BaseState
 from vellum.workflows.state.context import WorkflowContext
-from vellum.workflows.types.core import EntityInputsInterface, MergeBehavior
+from vellum.workflows.types.core import EntityInputsInterface
 from vellum.workflows.workflows.base import BaseWorkflow
 
 
@@ -36,7 +36,7 @@ class ToolCallingNode(BaseNode):
     ml_model: ClassVar[str] = "gpt-4o-mini"
     blocks: ClassVar[List[PromptBlock]] = []
     functions: ClassVar[List[FunctionDefinition]] = []
-    function_callables: ClassVar[List[Callable]] = []
+    function_callables: ClassVar[Dict[str, Callable]] = {}
     prompt_inputs: ClassVar[Optional[EntityInputsInterface]] = None
     # TODO: https://linear.app/vellum/issue/APO-342/support-tool-call-max-retries
     max_tool_calls: ClassVar[int] = 1
@@ -122,8 +122,12 @@ class ToolCallingNode(BaseNode):
             prompt_inputs=self.prompt_inputs,
         )
 
+        # TODO: fix mypy error
         self._function_nodes = {
-            function.name: create_function_node(function)
+            function.name: create_function_node(
+                function=function,
+                function_callable=self.function_callables[function.name]  # type: ignore
+            )
             for function in self.functions
             if function.name is not None
         }
--- a/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
+++ b/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
@@ -39,9 +39,6 @@ class ToolCallingNode(BaseNode):
     function_callables: ClassVar[List[Callable]] = []
     prompt_inputs: ClassVar[Optional[EntityInputsInterface]] = None
 
-    class Trigger(BaseNode.Trigger):
-        merge_behavior = MergeBehavior.AWAIT_ANY
-
     class Outputs(BaseOutputs):
         """
         The outputs of the ToolCallingNode.
--- a/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
+++ b/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
@@ -23,7 +23,7 @@ from vellum.workflows.workflows.base import BaseWorkflow
 
 class ToolCallingNode(BaseNode):
     """
-    A Node that allows for tool calling while dynamically generating a graph based on provided functions.
+    A Node that dynamically invokes the provided functions to the underlying Prompt
 
     Attributes:
         ml_model: str - The model to use for tool calling (e.g., "gpt-4o-mini")
--- a/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
+++ b/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
@@ -2,6 +2,7 @@ from collections.abc import Callable
 from typing import ClassVar, List, Optional
 
 from vellum import ChatMessage, FunctionDefinition, PromptBlock
+from vellum.client.types.chat_message_request import ChatMessageRequest
 from vellum.workflows.context import execution_context, get_parent_context
 from vellum.workflows.errors.types import WorkflowErrorCode
 from vellum.workflows.exceptions import NodeException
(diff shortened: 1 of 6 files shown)


dvargas92495:
Looking for an initial round of feedback or still more to iterate before we look?

src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -0,0 +1,127 @@
...
+        chat_history: List[ChatMessage] = []
+
+    def run(self) -> Outputs:
+        """
+        Run the tool calling workflow.
+
+        This dynamically builds a graph with router and function nodes,
+        then executes the workflow.
+        """
+        # TODO: We should parse input values and add them to the state
```
vincent0426: How should we get these values? I think they come from `def _compile_prompt_inputs` in prompt node which can't be accessed here

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -0,0 +1,89 @@
...
+            ChatMessage(
+                role="tool", text=f"Result from {self.function.name}: The current temperature is 22°C (71.6°F)."
+            )
+        )
+
+        return self.Outputs()
+
+
+class ToolRouterNode(InlinePromptNode):
+    # TODO: We should include chat history in our next prompt input
```
vincent0426: How should we include chat history in our next prompt node? I'm thinking of modifying prompt_inputs `chat_history` with a block that will fetch it during compile_prompt. However, prompt node is already created in this stage, so I think we could expose some methods like set_chat_history and prompt node will update its prompt_inputs attribute with the given chat_history?

vincent0426:
> Looking for an initial round of feedback or still more to iterate before we look?

I think it is good for the first round of feedback, I included some questions before moving forward!

src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -0,0 +1,127 @@
...
+from vellum.workflows.outputs.base import BaseOutputs
+from vellum.workflows.state.base import BaseState
+from vellum.workflows.state.context import WorkflowContext
+from vellum.workflows.types.core import EntityInputsInterface, MergeBehavior
+from vellum.workflows.workflows.base import BaseWorkflow
+
+
+class ToolCallingNode(BaseNode):
+    """
+    A Node that allows for tool calling while dynamically generating a graph based on provided functions.
```
dvargasfuertes:
```suggestion
    A Node that dynamically invokes the provided functions to the underlying Prompt
```

src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+
+    ml_model: ClassVar[str] = "gpt-4o-mini"
+    blocks: ClassVar[List[PromptBlock]] = []
+    functions: ClassVar[List[FunctionDefinition]] = []
+    function_callables: ClassVar[List[Callable]] = []
+    prompt_inputs: ClassVar[Optional[EntityInputsInterface]] = None
+
+    class Trigger(BaseNode.Trigger):
+        merge_behavior = MergeBehavior.AWAIT_ANY
```
dvargasfuertes: [nit] can remove - we want all nodes going forward to be AWAIT_ATTRIBUTES by default

src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -0,0 +1,127 @@
...
+        functions: List[FunctionDefinition] - The functions that can be called
+        function_callables: List[Callable] - The callables that can be called
+        prompt_inputs: Optional[EntityInputsInterface] - Mapping of input variable names to values
+    """
+
+    ml_model: ClassVar[str] = "gpt-4o-mini"
+    blocks: ClassVar[List[PromptBlock]] = []
+    functions: ClassVar[List[FunctionDefinition]] = []
+    function_callables: ClassVar[List[Callable]] = []
+    prompt_inputs: ClassVar[Optional[EntityInputsInterface]] = None
```
dvargasfuertes, line 40: [oos] let's make sure we are tracking a ticket for `max_tool_calls` or `max_loops` - https://www.notion.so/vellum-ai/PRD-ToolCallingNode-1c39035c57bd8084b8a9defa1a767b31?pvs=4#1c39035c57bd8046a578d029cf5e5c0a

src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
dvargasfuertes: [q] why do we need to compile them?

If we do, then we could split out a PR that splits out the logic used by `_compile_prompt_inputs` as its own util method that can be used in both places.

src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -0,0 +1,127 @@
...
+                raise Exception(f"Workflow execution rejected: {terminal_event.error}")
+
+            raise Exception(f"Unexpected workflow event: {terminal_event.name}")
+
+    def _build_graph(self) -> None:
+        self.tool_router_node = create_tool_router_node(
+            model_name=self.ml_model,
+            prompt_blocks=self.blocks,
+            functions=self.functions,
+            input_values=self.prompt_inputs,
```
dvargasfuertes: [nit] I'd keep all of these attributes/args consistently named

src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -0,0 +1,127 @@
...
+
+                class Outputs(BaseWorkflow.Outputs):
+                    text: str = ToolRouterNode.Outputs.text
+                    chat_history: List[ChatMessage] = ToolCallingState.chat_history
+
+            subworkflow = ToolCallingWorkflow(
+                parent_state=self.state,
+                context=WorkflowContext(
+                    vellum_client=self._context.vellum_client, generated_files=self._context.generated_files
+                ),
```
dvargasfuertes: [nit] `WorkflowContext` should now have a static method for initiating new workflow context's

src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+
+            terminal_event = subworkflow.run()
```
dvargasfuertes, line 96: [oos] we'll want to loop over the stream method to ensure the events are emitted through the root, similar to our other subworkflow nodes

src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -0,0 +1,127 @@
...
+
+        self._function_nodes = {function.name: create_function_node(function) for function in self.functions}
+
+        graph_set = set()
+
+        # Add connections from ports of router to function nodes and back to router
+        for function_name, FunctionNodeClass in self._function_nodes.items():
+            # TODO: We should think about how to handle this
+            if function_name is None:
+                raise ValueError("Function name is required")
```
dvargasfuertes: [nit] I would expect `function_name` to be non-null if it's the key to a dict

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -0,0 +1,89 @@
...
+from vellum.workflows.types.core import EntityInputsInterface
+
+
+class FunctionNode(BaseNode):
+    """Node that executes a specific function."""
+
+    function: FunctionDefinition
+
+    def run(self) -> BaseNode.Outputs:
+        # TODO: We should think about how to execute the function
```
dvargasfuertes: [oos] this should already be ticketed in you project, surrounding the invocations of custom inline functions.

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -0,0 +1,89 @@
...
+        self.state.chat_history.append(
+            ChatMessage(
+                role="tool", text=f"Result from {self.function.name}: The current temperature is 22°C (71.6°F)."
```
dvargasfuertes:
```suggestion
                role="FUNCTION", text=f"Result from {self.function.name}: The current temperature is 22°C (71.6°F)."
```

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
dvargasfuertes: I think we just append `prompt_inputs = {**prompt_inputs, "chat_history": State.chat_history}`

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -0,0 +1,89 @@
...
+                ToolRouterNode.Outputs.results[0]["type"].equals("FUNCTION_CALL")
+                & ToolRouterNode.Outputs.results[0]["value"]["name"].equals(function_name)
+            )
+        )
+        port = Port.on_if(port_condition)
+        setattr(Ports, function_name, port)
+
+    setattr(Ports, "default", Port.on_else())
+    node = type(
+        "ToolRouterNode",
```
dvargasfuertes, line 86: [nit] we may want to fully define the class within here - this way we could easily pass in the chat_history prop mentioned above and just edit `ToolRouterNode.Ports` after the fact

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -0,0 +1,89 @@
...
+            "__module__": __name__,
+        },
+    )
+
+    return node
+
+
+def create_function_node(function: FunctionDefinition) -> Type[FunctionNode]:
+    node = type(
+        f"FunctionNode.{function.name}",
```
dvargasfuertes: [nit] I don't think a class name could have a `.` within it

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    Test that the GetCurrentWeatherWorkflow returns the expected outputs.
+    """
+
+    def generate_prompt_events(*args, **kwargs) -> Iterator[ExecutePromptEvent]:
+        execution_id = str(uuid4())
+
+        call_count = vellum_adhoc_prompt_client.adhoc_execute_prompt_stream.call_count
+        expected_outputs: List[Union[FunctionCallVellumValue, StringVellumValue]]
+        if call_count == 1:
```
dvargasfuertes, line 38: cool pattern

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -0,0 +1,256 @@
...
+def test_get_current_weather_workflow(vellum_adhoc_prompt_client, mock_uuid4_generator):
+    """
+    Test that the GetCurrentWeatherWorkflow returns the expected outputs.
+    """
+
+    def generate_prompt_events(*args, **kwargs) -> Iterator[ExecutePromptEvent]:
+        execution_id = str(uuid4())
+
+        call_count = vellum_adhoc_prompt_client.adhoc_execute_prompt_stream.call_count
+        expected_outputs: List[Union[FunctionCallVellumValue, StringVellumValue]]
```
dvargasfuertes:
```suggestion
        expected_outputs: List[PromptOutput]
```

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -0,0 +1,256 @@
...
+
+    # WHEN the workflow is executed
+    terminal_event = workflow.run()
+
+    # THEN the workflow is executed successfully
+    assert terminal_event.name == "workflow.execution.fulfilled"
+    assert terminal_event.outputs.text == "The current temperature in San Francisco is 22°C (71.6°F)."
+    assert terminal_event.outputs.chat_history == [
+        ChatMessage(
+            text='{\n    "arguments": {\n        "location": "Miami",\n        "unit": "metric"\n    },\n    "id": "call_7115tNTmEACTsQRGwKpJipJK",\n    "name": "get_current_weather",\n    "state": "FULFILLED"\n}',  # noqa: E501
```
dvargasfuertes: [nit] I would expect this ChatMessage to contain a `FUNCTION_CALL` content rather than text

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -0,0 +1,256 @@
...
+                        state=None,
+                        cache_config=None,
+                        blocks=[
+                            VariablePromptBlock(
+                                block_type="VARIABLE", state=None, cache_config=None, input_variable="question"
+                            )
+                        ],
+                    )
+                ],
+            ),
```
dvargasfuertes, line 184: [nit] Going with the chat history approach I mentioned above, I would expect a variable block here that references the chat_history variable

tests/workflows/basic_tool_calling_node/workflow.py
```diff
@@ -0,0 +1,88 @@
...
+from vellum.client.types.function_definition import FunctionDefinition
+from vellum.client.types.plain_text_prompt_block import PlainTextPromptBlock
+from vellum.client.types.rich_text_prompt_block import RichTextPromptBlock
+from vellum.client.types.variable_prompt_block import VariablePromptBlock
+from vellum.workflows.nodes.experimental.tool_calling_node import ToolCallingNode
+from vellum.workflows.workflows.base import BaseWorkflow
+
+
+# A simple tool definition
+def get_current_weather():
```
dvargasfuertes: this function will literally get the current weather. it should have `location: str` as a parameter and do a mock api call to return some reasonable response

dvargas92495 (commented):
Again, feel free to start chopping this up so that we could start merging pieces in incrementally! Usually a warning sign to have three tickets in the subject line of the PR

vincent0426:
should we define a new `FunctionDefinition` type that overrides `name` as `str` instead of `str | None` after filtering? since mypy will yell as it doesn't detect that we filter out None name

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # Create a class-level wrapper that calls the original function
+    def execute_function(self) -> BaseNode.Outputs:
+        outputs = self.state.meta.node_outputs.get(ToolRouterNode.Outputs.text)
+        # first parse into json
+        outputs = json.loads(outputs)
+        arguments = outputs["arguments"]
+
+        # Call the original function directly with the arguments
+        result = function_callable(**arguments)
```
vincent0426, line 116: call the function with its arguments

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -0,0 +1,323 @@
...
+                            type="FUNCTION_CALL",
+                            value=FunctionCallChatMessageContentValue(
+                                name="get_current_weather",
+                                arguments={"location": "Miami", "unit": "metric"},
+                                id="call_7115tNTmEACTsQRGwKpJipJK",
+                            ),
+                        ),
+                    ),
+                    ChatMessage(
+                        text="The current weather in Miami is sunny with a temperature of 70 degrees metric.",
```
vincent0426: This combines `def get_current_weather` in test and the argument provided by mock arguments={"location": "Miami", "unit": "metric"}, which indicates that we call the function with the correct argument

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -0,0 +1,323 @@
...
+
+    # THEN the workflow is executed successfully
+    assert terminal_event.name == "workflow.execution.fulfilled"
+    assert (
+        terminal_event.outputs.text
+        == "Based on the function call, the current temperature in Miami is 70 degrees metric."
+    )
+    assert terminal_event.outputs.chat_history == [
+        ChatMessage(text="Hello, how are you?", role="USER"),
+        ChatMessage(text="I'm good, thank you!", role="ASSISTANT"),
```
vincent0426, line 88: I think the only chat history we missed here is to include the value from blocks `ChatMessagePromptBlock` and combine with `prompt_inputs`?

tests/workflows/basic_tool_calling_node/workflow.py
```diff
@@ -0,0 +1,84 @@
...
+                    blocks=[
+                        VariablePromptBlock(
+                            input_variable="question",
+                        ),
+                    ],
+                ),
+            ],
+        ),
+    ]
+    function_callables = {"get_current_weather": get_current_weather}
```
dvargasfuertes, line 50: [oos] we shouldn't need this attribute -  `functions` should already support callables

tests/workflows/basic_tool_calling_node/workflow.py
```diff
@@ -0,0 +1,84 @@
...
+            },
+            forced=False,
+        ),
+    ]
+    prompt_inputs = {
+        "question": "What's the weather like in San Francisco?",
+        "chat_history": [
+            ChatMessage(role="USER", text="Hello, how are you?"),
+            ChatMessage(role="ASSISTANT", text="I'm good, thank you!"),
+        ],
```
dvargasfuertes, line 71: [nit] Currently this prompt compiles with two consecutive user messages. I'd start this one as empty

tests/workflows/basic_tool_calling_node/workflow.py
dvargasfuertes: we also shouldn't need the `FunctionDefinition` below. Should be as simple as:

`functions = [get_current_weather]`

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -0,0 +1,323 @@
...
+            PromptRequestStringInput(key="question", type="STRING", value="What's the weather like in San Francisco?"),
+            PromptRequestChatHistoryInput(
+                key="chat_history",
+                type="CHAT_HISTORY",
+                value=[
+                    ChatMessage(text="Hello, how are you?", role="USER"),
+                    ChatMessage(text="I'm good, thank you!", role="ASSISTANT"),
+                    ChatMessage(
+                        text=None,
+                        role="FUNCTION",
```
dvargasfuertes, line 223: [oos] This should be `"ASSISTANT"`

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
dvargasfuertes: why not celsius or Fahrenheit instead of metric lol

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -0,0 +1,323 @@
...
+            VellumVariable(
+                id=str(first_call_input_id_2),
+                key="chat_history",
+                type="CHAT_HISTORY",
+                required=None,
+                default=None,
+                extensions=None,
+            ),
+        ],
+        "parameters": DEFAULT_PROMPT_PARAMETERS,
```
dvargasfuertes, line 143: [nit] feel free to only have assertions for:
- `input_values`
- `input_variables`
- `blocks`

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -0,0 +1,134 @@
...
+def create_function_node(function: FunctionDefinition, function_callable: Callable[..., Any]) -> Type[FunctionNode]:
+    """
+    Create a FunctionNode class for a given function.
+
+    This ensures the callable is properly registered and can be called with the expected arguments.
+    """
+
+    # Create a class-level wrapper that calls the original function
+    def execute_function(self) -> BaseNode.Outputs:
+        outputs = self.state.meta.node_outputs.get(ToolRouterNode.Outputs.text)
```
dvargasfuertes, line 110: [oos] on the class below, we should be able to do:

`"function_call": ToolRouterNode.Outputs.results[0]`

and then here we can replace this line and the two below with

`arguments = self.function_call.arguments`

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+class ToolRouterNode(InlinePromptNode):
+    def run(self) -> Iterator[BaseOutput]:
+        self.prompt_inputs = {**self.prompt_inputs, "chat_history": self.state.chat_history}  # type: ignore
+        generator = super().run()
+        for output in generator:
+            if output.name == "results" and output.value:
+                values = cast(List[Any], output.value)
+                if values and len(values) > 0:
+                    if isinstance(values[0], StringVellumValue):
```
dvargasfuertes: [nit] would do `values[0].type == 'STRING'` to avoid the import

dvargas92495 (approved):
This is a great start!

I suggest we merge as in as the size of the PR is getting harder to rereview - it should all be safe in `experimental` anyway. then we start with the initial set of feedback in followup PRs

PR #1378, opened 2025-04-12
Remove unused flake comment

vincent0426:
Exploring ruff and found that we actually only need `noqa` on top and it should be 403 if we want to state it explicitly https://flake8.pycqa.org/en/latest/user/error-codes.html#:~:text=%E2%80%98from%20module%20import%20*%E2%80%99%20used%3B%20unable%20to%20detect%20undefined%20names

Do we have plans to migrate to ruff? its so fast haha

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/__init__.py
+++ b/ee/vellum_ee/workflows/display/nodes/__init__.py
@@ -1,8 +1,8 @@
-# flake8: noqa: F401
+# flake8: noqa
 
 # Force an import to ensure that all display classes are registered with the BaseNodeDisplay registry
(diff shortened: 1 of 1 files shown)


ee/vellum_ee/workflows/display/nodes/__init__.py
```diff
@@ -1,8 +1,8 @@
-# flake8: noqa: F401
+# flake8: noqa
```
dvargasfuertes: let's leave an explicit rule (403 if that's the correct one)

dvargas92495 (approved):

PR #1382, opened 2025-04-14
[APO-419] Add parse for input int type for guardrail node

vincent0426:
- Add int type
- Add test for inputs


dvargas92495 (approved):
I think we should apply some of the changes we added to inline prompt node here as well as inevitable future issues

PR #1369, opened 2025-04-11
[APO-389] Handle prompt deployment fallback models

NgoHarrison:
Final PR to handle running fallbacks for prompt deployments

PR #1384, opened 2025-04-14
[APO-411] Handle provider errors with prompt deployment model fallbacks

NgoHarrison:
Follow up from this:
https://github.com/vellum-ai/vellum-python-sdks/pull/1369#discussion_r2039819701

### DIFF
--- a/src/vellum/workflows/nodes/displayable/bases/prompt_deployment_node.py
+++ b/src/vellum/workflows/nodes/displayable/bases/prompt_deployment_node.py
@@ -87,52 +87,72 @@ class BasePromptDeploymentNode(BasePromptNode, Generic[StateType]):
             request_options=request_options,
         )
 
-    def _process_prompt_event_stream(self) -> Generator[BaseOutput, None, Optional[List[PromptOutput]]]:
+    def _process_prompt_event_stream(
+        self, prompt_event_stream: Optional[Iterator[ExecutePromptEvent]] = None
+    ) -> Generator[BaseOutput, None, Optional[List[PromptOutput]]]:
         """Override the base prompt node _process_prompt_event_stream()"""
         self._validate()
-        try:
-            prompt_event_stream = self._get_prompt_event_stream()
-            next(prompt_event_stream)
-        except ApiError as e:
-            if (
-                e.status_code
-                and e.status_code < 500
-                and self.ml_model_fallbacks is not OMIT
-                and self.ml_model_fallbacks is not None
-            ):
-                for ml_model_fallback in self.ml_model_fallbacks:
-                    try:
-                        prompt_event_stream = self._get_prompt_event_stream(ml_model_fallback=ml_model_fallback)
-                        next(prompt_event_stream)
-                        break
-                    except ApiError:
-                        continue
+
+        if prompt_event_stream is None:
+            try:
+                prompt_event_stream = self._get_prompt_event_stream()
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/displayable/prompt_deployment_node/tests/test_node.py
```diff
@@ -278,3 +279,55 @@ def generate_successful_stream():
...
     body_params = second_call_kwargs["request_options"]["additional_body_parameters"]
     assert body_params["overrides"]["ml_model_fallback"] == "fallback_model_1"
+
+
+def test_prompt_deployment_node_provider_error_with_fallbacks(vellum_client):
+    # GIVEN a Prompt Deployment Node with fallback models
+    class TestPromptDeploymentNode(PromptDeploymentNode):
+        deployment = "test_deployment"
+        prompt_inputs = {}
+        ml_model_fallbacks = ["gpt-40", "gemini-1.5-flash-latest"]
```
dvargasfuertes:
```suggestion
        ml_model_fallbacks = ["gpt-4o", "gemini-1.5-flash-latest"]
```

src/vellum/workflows/nodes/displayable/prompt_deployment_node/tests/test_node.py
```diff
@@ -278,3 +279,55 @@ def generate_successful_stream():
...
+    def generate_fallback_events():
+        execution_id = str(uuid4())
+        expected_outputs: List[PromptOutput] = [StringVellumValue(value="Fallback response")]
+        events = [
+            InitiatedExecutePromptEvent(execution_id=execution_id),
+            FulfilledExecutePromptEvent(execution_id=execution_id, outputs=expected_outputs),
+        ]
+        return iter(events)
+
+    vellum_client.execute_prompt_stream.side_effect = [generate_primary_events(), generate_fallback_events()]
```
dvargasfuertes, line 317: [oos] can we add parameterized/new test that generates a provider error for _both_ the primary model and the first fallback? I think I see a bug in the iteration logic

dvargasfuertes: I think I see another one when the first one fails with an APIError and the second one with ProviderError or vice versa

src/vellum/workflows/nodes/displayable/bases/prompt_deployment_node.py
```diff
@@ -87,53 +87,73 @@ def _get_prompt_event_stream(self, ml_model_fallback: Optional[str] = None) -> I
...
+                    yield BaseOutput(name="results", value=event.outputs)
+                elif event.state == "REJECTED":
+                    if (
+                        event.error
+                        and event.error.code == WorkflowErrorCode.PROVIDER_ERROR.value
+                        and self.ml_model_fallbacks is not OMIT
+                        and self.ml_model_fallbacks is not None
+                    ):
+                        try:
+                            fallback_stream = self._retry_prompt_stream_with_fallbacks()
```
dvargasfuertes, line 129: [oos] Something tells me that we're going to want to try to consolidate these two retry calls together. I mention two test cases below that we should iterate on so that we validate that they work as expected, that may be down stream of these retry calls not being consolidated

dvargas92495 (approved):
Additional test cases could be done as followups 👍

PR #1386, opened 2025-04-15
Consolidate CodeResourceDefinition classes

dvargas92495:
Eventually working towards this Nested State Deserialization PR: https://github.com/vellum-ai/vellum-python-sdks/pull/1385

This PR splits out our CodeResourceDefinition classes into its own file to help avoid circular deps issues. We also consolidate the other two places that we call `CodeResourceDefinition`s

### DIFF
--- a/src/vellum/workflows/types/definition.py
+++ b/src/vellum/workflows/types/definition.py
@@ -1,0 +1,39 @@
+from uuid import UUID
+from typing import Annotated, Any, Dict, Union
+
+from pydantic import BeforeValidator
+
+from vellum.client.types.code_resource_definition import CodeResourceDefinition as ClientCodeResourceDefinition
+
+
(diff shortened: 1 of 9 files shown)


src/vellum/workflows/types/definition.py
```diff
@@ -0,0 +1,39 @@
+from uuid import UUID
+from typing import Annotated, Any, Dict, Union
+
+from pydantic import BeforeValidator
+
+from vellum.client.types.code_resource_definition import CodeResourceDefinition as ClientCodeResourceDefinition
```
dvargasfuertes, line 6: In the future, I'd like to consolidate it down to just the type, but need to see:
- What is our appetite for extending the client types from client-generator
- Once we could start reliably generating UUID's from our schema

NgoHarrison (commented):
✅

PR #1387, opened 2025-04-15
fix link to reference page readme

awlevin:


src/vellum/client/README.md
```diff
@@ -20,7 +20,7 @@ pip install vellum-ai
 
 ## Reference
 
-A full reference for this library is available [here](./reference.md).
+A full reference for this library is available [here](/reference.md).
```
dvargasfuertes: hm no I think we want to move `reference.md` into this directory since it's mostly only relevant for the client library

awlevin: ah kk makes sense

awlevin: is this file manually maintained or are there other things that would need to be changed? only references i find are [here](https://github.com/vellum-ai/vellum-python-sdks/blob/9741a69e750bfd9d9751d46fcfbacba78886f35a/.pre-commit-config.yaml#L9-L11) - seems like those need to point to this new specific path too?

dvargasfuertes: It's automatically managed by fern, I'll make the changes from client generator on the next release to make sure it's permanent.

The pre commit should be okay with the new filepath

PR #1391, opened 2025-04-15
Add more handling around provider errors for prompt deployment fallbacks

NgoHarrison:
Follow up from https://github.com/vellum-ai/vellum-python-sdks/pull/1384#discussion_r2044235737

### DIFF
--- a/src/vellum/workflows/nodes/displayable/bases/prompt_deployment_node.py
+++ b/src/vellum/workflows/nodes/displayable/bases/prompt_deployment_node.py
@@ -165,12 +165,6 @@ class BasePromptDeploymentNode(BasePromptNode, Generic[StateType]):
                     )
                 )
 
-        self._handle_api_error(
-            ApiError(
-                body={"detail": "No fallback models available"},
-                status_code=400,
-            )
-        )
         return None
 
     def _compile_prompt_inputs(self) -> List[PromptDeploymentInputRequest]:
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/displayable/prompt_deployment_node/tests/test_node.py
```diff
@@ -331,3 +333,133 @@ def generate_fallback_events():
...
+        generate_fallback1_events(),
+    ]
+
+    # WHEN we run the node
+    with pytest.raises(NodeException) as exc_info:
+        node = TestPromptDeploymentNode()
+        list(node.run())
+
+    # THEN we should get an exception
+    assert exc_info.value.message == "Failed to execute prompts with these fallbacks: ['gpt-4o']"
```
dvargasfuertes, line 458: [nit] we should show the error of the last model

src/vellum/workflows/nodes/displayable/bases/prompt_deployment_node.py
```diff
@@ -154,6 +165,14 @@ def _retry_prompt_stream_with_fallbacks(self):
                     )
                 )
 
+        self._handle_api_error(
+            ApiError(
+                body={"detail": "No fallback models available"},
+                status_code=400,
+            )
+        )
```
dvargasfuertes: [nit] when does this get hit? not seeing it come up in the test cases below. 

If it's a valid code path, we should instead show what the latest error was (with status code)

dvargas92495 (approved):

src/vellum/workflows/nodes/displayable/bases/prompt_deployment_node.py
NgoHarrison: Ah good point, removed it

PR #1394, opened 2025-04-15
remove trace_id from StateMeta

dvargas92495:
To simplify `State` deserialization, and to eventually build up to [this](https://github.com/vellum-ai/vellum-python-sdks/pull/1301) PR, this PR removes `trace_id` from `StateMeta`. It was originally added there before we had most of our execution context infra built out, which we could now use instead

### DIFF
--- a/src/vellum/workflows/state/base.py
+++ b/src/vellum/workflows/state/base.py
@@ -231,7 +231,6 @@ def default_datetime_factory() -> datetime:
 class StateMeta(UniversalBaseModel):
     workflow_definition: Type["BaseWorkflow"] = field(default_factory=import_workflow_class)
     id: UUID = field(default_factory=uuid4_default_factory)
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/state/base.py
```diff
@@ -231,7 +231,6 @@ def default_datetime_factory() -> datetime:
 class StateMeta(UniversalBaseModel):
     workflow_definition: Type["BaseWorkflow"] = field(default_factory=import_workflow_class)
     id: UUID = field(default_factory=uuid4_default_factory)
-    trace_id: UUID = field(default_factory=uuid4_default_factory)
```
dvargasfuertes, line 234: goal of the PR

marlenc3 (approved):

NgoHarrison (commented):
✅

PR #1395, opened 2025-04-15
Add some resilience to trying to deserialize workflow definition in State

dvargas92495:
This will be the last of the State deserialization PRs before release. This new Code Resource Definition decoding is a bit risky, so adding some resilience so that things could run smoothly in case we receive data we are not expecting


marlenc3 (commented):
✅

NgoHarrison (commented):
✅

PR #1398, opened 2025-04-16
Generate new span ids for Run From Node state's

dvargas92495:
Each workflow execution should be its own span id. However, bc Run From Node stored the span id, we were seeing the same span id across runs. This PR fixes that so workflows executed by the Run From Node flow have their own span id's


tests/workflows/basic_external_input/tests/test_workflow.py
```diff
@@ -37,6 +37,9 @@ def test_workflow__happy_path_multi_stop():
     assert final_terminal_event.name == "workflow.execution.fulfilled"
     assert final_terminal_event.outputs.final_value == "sunny"
 
+    # AND the workflow execution should have the same span id
```
dvargasfuertes, line 40: for pause/resume flows, we _want_ the same span id

vincent0426 (approved):

PR #1399, opened 2025-04-16
Add codegen support for node ports

NgoHarrison:


ee/codegen/src/serializers/vellum.ts
NgoHarrison, line 1: Added node ports to all the nodes even though some might not have it (I am assuming we will never get bad upstream data e.g terminal nodes getting ports defined in their json)

ee/codegen/src/generators/node-port.ts
```diff
@@ -3,24 +3,25 @@ import { AstNode } from "@fern-api/python-ast/core/AstNode";
 import { Writer } from "@fern-api/python-ast/core/Writer";
 
 import { WorkflowContext } from "src/context";
-import { GenericNodeContext } from "src/context/node-context/generic-node";
+import { BaseNodeContext } from "src/context/node-context/base";
```
dvargasfuertes, line 6: love it 👍 

in general, we want to keep progressing from things being just GenericNode -> All Nodes (eg, BaseNode)

ee/codegen/src/serializers/vellum.ts
dvargasfuertes: 👍

ee/codegen/src/serializers/vellum.ts
```diff
@@ -1571,6 +1658,7 @@ export const ConditionalNodeSerializer: ObjectSchema<
   base: CodeResourceDefinitionSerializer.optional(),
   definition: CodeResourceDefinitionSerializer.optional(),
   adornments: listSchema(AdornmentNodeSerializer).optional(),
+  ports: listSchema(NodePortSerializer).optional(),
```
dvargasfuertes, line 1661: [oos] How does this work with conditional nodes? can we follow up with a test to ensure that we don't generate two node port classes here?

dvargas92495 (approved):

vincent0426 (approved):

PR #1403, opened 2025-04-17
Prevent codegen of ports on conditional nodes if node port attribute is present

NgoHarrison:
Follow up to https://github.com/vellum-ai/vellum-python-sdks/pull/1399

### DIFF
--- a/ee/codegen/src/__test__/nodes/conditional-node.test.ts
+++ b/ee/codegen/src/__test__/nodes/conditional-node.test.ts
@@ -1043,7 +1043,7 @@ describe("Conditional Node with NodePorts defined", () => {
             type: "CONSTANT_VALUE",
             value: {
               type: "STRING",
-              value: "hello",
+              value: "new hello",
             },
           },
           operator: "null",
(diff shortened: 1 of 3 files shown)


ee/codegen/src/__test__/nodes/conditional-node.test.ts
```diff
@@ -1022,3 +1024,101 @@ describe("Conditional Node with OR combinator generates pipe operator", () => {
...
+        {
+          id: "f804da22-b3d2-43d0-8c6f-321434c28ead",
+          key: "895d82e8-e6d4-4033-9ae4-581e8c2dc45c.field",
+          value: {
+            rules: [
+              {
+                type: "CONSTANT_VALUE",
+                data: {
+                  type: "STRING",
+                  value: "hello",
```
dvargasfuertes: [nit] Can we make the two have different values? bc they are both `"hello"`, it's hard to see which one is winning from the test

dvargas92495 (approved):

PR #1404, opened 2025-04-17
Reduce number of workflow snapshot events emitted during prompt streaming

dvargas92495:
Prompt Streaming was emitting three events during output streaming:
- once when the stream started
- once when the text output is emitted
- once when the results output is emitted

We only need to emit once for when the node fully fulfills for now. Streaming state snapshots will require a longer conversation around how often we emit these events. For Run from Node purposes, we only need the fulfilled node events and their outputs, so let's just emit once for now.

Long term, I'm thinking we may not need `workflow.execution.snapshotted` events at all, and just store state deltas, which node/workflow events _already_ are.... Gonna discuss with @siddseethepalli during our 1-1 tm

### DIFF
--- a/src/vellum/workflows/state/tests/test_state.py
+++ b/src/vellum/workflows/state/tests/test_state.py
@@ -1,16 +1,13 @@
 import pytest
-from collections import defaultdict
 from copy import deepcopy
 import json
 from queue import Queue
-from typing import Dict
+from typing import Dict, cast
 
 from vellum.workflows.nodes.bases import BaseNode
 from vellum.workflows.outputs.base import BaseOutputs
 from vellum.workflows.state.base import BaseState
 from vellum.workflows.state.encoder import DefaultStateEncoder
-
-snapshot_count: Dict[int, int] = defaultdict(int)
 
 
 @pytest.fixture()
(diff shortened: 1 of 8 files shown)


src/vellum/workflows/state/tests/test_state.py
```diff
@@ -1,17 +1,14 @@
...
 from queue import Queue
-from typing import Dict
+from typing import Dict, cast
 
 from vellum.workflows.nodes.bases import BaseNode
 from vellum.workflows.outputs.base import BaseOutputs
 from vellum.workflows.state.base import BaseState
 from vellum.workflows.state.encoder import DefaultStateEncoder
 
-snapshot_count: Dict[int, int] = defaultdict(int)
```
dvargasfuertes, line 13: I think making this non-global will relieve some flakiness we've been seeing

tests/workflows/multi_output_node_streaming/tests/test_workflow.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_run_workflow__happy_path():
+    # GIVEN a simple workflow that references a node with a streaming node that emits 3 events
+    workflow = MultiOutputNodeStreaming()
+
+    # WHEN the workflow is run
+    inputs = Inputs(foo="Hello")
+    events = list(workflow.stream(event_filter=all_workflow_event_filter, inputs=inputs))
+
+    # THEN there should only be 1 snapshot event
```
dvargasfuertes, line 14: Main goal of the PR. This was previously 3

marlenc3:
✅

PR #1405, opened 2025-04-17
Fix trace ids on node events

dvargas92495:
I noticed that node events were emitting `UUID(0)`, probably as a result of some of my recent changes of removing off of StateMeta and us not having an assertion protecting that regrssion. This PR fixes both


dvargas92495:
Gonna cut a release for this to stop the events bleeding. Luckily, because of parent context and span ids, this is recoverable data if we ever need to in the future

marlenc3 (commented):
✅

NgoHarrison (commented):
✅

PR #1408, opened 2025-04-17
Allow workflows to push a workflow even if not in lockfile or in pyproject toml

dvargas92495:
I want to be able to start a workflow from scratch locally, then just do `vellum workflows push path.to.that.workflow`


marlenc3 (approved):

vincent0426 (commented):
✅

PR #1409, opened 2025-04-17
Simplify get_node_display_class utility

dvargas92495:
I was working on pushing a customer workflow to Vellum that utilizes a subclass of a generic node before I ran into:

```
  File "/Users/dvargas/vellum-ai/workflows-as-code-runner-prototype/venv/lib/python3.11/site-packages/vellum_ee/workflows/display/nodes/get_node_display_class.py", line 30, in get_node_display_class
    NodeDisplayBaseClass = base_node_display_class[node_class]  # type: ignore[index]
                           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/typing.py", line 351, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/typing.py", line 1826, in __class_getitem__
    _check_generic(cls, params, len(cls.__parameters__))
  File "/Users/dvargas/vellum-ai/workflows-as-code-runner-prototype/venv/lib/python3.11/site-packages/typing_extensions.py", line 2922, in _check_generic
    raise TypeError(f"{cls} is not a generic class")
TypeError: <class 'vellum_ee.workflows.display.nodes.base_node_display.MockNetworkingClientDisplay'> is not a generic class
```

So before I get started on tackling this issue, I wanted to continue some display cleanup work I've been doing by simplifying our `get_node_display_class` utility. We can remove two of the three args on the assumption that all displays will eventually inherit from `BaseNodeDisplay`. This also forces us to move the registry from `BaseNodeDisplay`, which was dangerous bc it allowed users to subclass and mess with that internal state.

PR #1411, opened 2025-04-17
Simplify get_workflow_display

dvargas92495:
Continuing with display simplification.

This PR deprecates two params to `get_workflow_display`, which are essentially the same two params we removed from `get_node_display_class`: https://github.com/vellum-ai/vellum-python-sdks/pull/1409

We have to deprecate for now instead of remove bc this is a user facing method that's used by vembda and the cli. Will start removing consumers on the next PR

### DIFF
--- a/ee/vellum_ee/workflows/display/workflows/get_vellum_workflow_display_class.py
+++ b/ee/vellum_ee/workflows/display/workflows/get_vellum_workflow_display_class.py
@@ -1,33 +1,45 @@
-from typing import Optional, Type
+import types
+from typing import TYPE_CHECKING, Optional, Type
 
 from vellum.workflows.types.generics import WorkflowType
-from vellum_ee.workflows.display.types import WorkflowDisplayContext, WorkflowDisplayType
+from vellum_ee.workflows.display.types import WorkflowDisplayContext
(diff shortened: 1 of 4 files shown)


ee/vellum_ee/workflows/display/workflows/get_vellum_workflow_display_class.py
```diff
@@ -1,33 +1,45 @@
...
 from vellum.workflows.types.generics import WorkflowType
-from vellum_ee.workflows.display.types import WorkflowDisplayContext, WorkflowDisplayType
+from vellum_ee.workflows.display.types import WorkflowDisplayContext
+from vellum_ee.workflows.display.utils.registry import get_from_workflow_display_registry
+
+if TYPE_CHECKING:
+    from vellum_ee.workflows.display.workflows import BaseWorkflowDisplay
+
+
+def _get_workflow_display_class(*, workflow_class: Type[WorkflowType]) -> Type["BaseWorkflowDisplay"]:
```
dvargasfuertes, line 12: This method now looks very similar to `get_node_display`. Part of me wants to move both methods to the `registry` module, but again, that would be a breaking change here unfortunately so not sure if it's worth it yet

vincent0426:
✅

vincent0426 (commented):
✅

PR #1412, opened 2025-04-17
remove deprecated args from get_workflow_display

dvargas92495:
This PR removes all consumers in the SDK repo using these deprecated args: https://github.com/vellum-ai/vellum-python-sdks/pull/1411

We are close to having no more reads of `VellumWorkflowDisplay`. Just need to update codegen to round it out


vincent0426 (commented):
wow super clean!

PR #1413, opened 2025-04-17
Change codegen from VellumWorkflowDisplay to BaseWorkflowDisplay

dvargas92495:
This removes the last reference to `VellumWorkflowDisplay` from the SDK repo. On the next release, will do the same to vembda and prototype repo


vincent0426 (commented):
✅

PR #1414, opened 2025-04-17
Support serialization of inherited node display classes automatically

dvargas92495:
This PR enables users to serialize nodes that extend for another user-defined base node


ee/vellum_ee/workflows/display/workflows/tests/test_workflow_display.py
```diff
@@ -279,3 +279,28 @@ class Outputs(BaseWorkflow.Outputs):
 
     # THEN the workflow output display should be included
     assert display_context.workflow_outputs.keys() == {"foo", "bar"}
+
+
+def test_serialize_workflow__inherited_node_display_class_not_registered():
```
dvargasfuertes, line 284: need to follow up with the same fix for Workflow's too

vincent0426 (commented):
✅

PR #1415, opened 2025-04-17
Fix bug surrounding BaseOutputs inheritance

dvargas92495:
This is a bug that was blocking me here: https://github.com/vellum-ai/vellum-python-sdks/pull/1414

Subworkflow nodes used to map all of the base outputs to the inherited outputs directly upon inheritance. This felt neat at the time I introduced it way back when, but breaks serialization logic, especially when one node inherits from an existing node. Instead this automapping should be the job of the workflow/node, and instead, inheriting just points outputs to the same default value.

This shouldn't affect live workflows, since we always codegen our outputs explicitly

### DIFF
--- a/tests/workflows/basic_conditional_node/workflow_with_try_node.py
+++ b/tests/workflows/basic_conditional_node/workflow_with_try_node.py
@@ -32,7 +32,7 @@ class FinalNode(BaseNode):
 
 class TestConditionalNode(ConditionalNode):
     class Ports(ConditionalNode.Ports):
-        branch_1 = Port.on_if(Node.Outputs.error.contains("Provider Error"))
+        branch_1 = Port.on_if(Node.Outputs.error.is_not_nil())
         branch_2 = Port.on_else()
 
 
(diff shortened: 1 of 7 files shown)


tests/workflows/basic_conditional_node/workflow_with_try_node.py
```diff
@@ -32,7 +32,7 @@ class Outputs(BaseNode.Outputs):
 
 class TestConditionalNode(ConditionalNode):
     class Ports(ConditionalNode.Ports):
-        branch_1 = Port.on_if(Node.Outputs.error.contains("Provider Error"))
+        branch_1 = Port.on_if(Node.Outputs.error.is_not_nil())
```
dvargasfuertes, line 35: this test actually had a bug, that this bug hid

tests/workflows/basic_retry_node/workflow.py
```diff
@@ -20,15 +20,15 @@ def run(self) -> Outputs:
 class Subworkflow(BaseWorkflow[RetryNode.SubworkflowInputs, BaseState]):
     graph = StartNode
 
-    class Outputs(StartNode.Outputs):
-        pass
+    class Outputs(BaseWorkflow.Outputs):
+        execution_count = StartNode.Outputs.execution_count
```
dvargasfuertes, line 24: we have a ticket for workflows auto inferring outputs from their last nodes, which I think could help clean this up

vincent0426 (commented):
✅

PR #1416, opened 2025-04-17
Support serialization of Inherited workflows

dvargas92495:
The workflow display version of https://github.com/vellum-ai/vellum-python-sdks/pull/1414/files

Admittedly, workflow inheritance will be a super advanced feature, but I wanted to keep the to `get_[node|workflow]_display_class` utilities in sync in the meantime


vincent0426 (commented):
✅

PR #1417, opened 2025-04-17
Skip serialization of legacy node inputs pointing to unsupported reference types

dvargas92495:
The customer workflow I'm trying to push to Vellum has legacy node inputs referencing State. This PR unblocks the failure that occurs during serialization


vincent0426 (commented):
✅

PR #1420, opened 2025-04-17
Serialize node ids in the node execution cache instead of node module paths

dvargas92495:
Hopefully the last piece I need from the SDK for run from node.

In order to show which previous nodes we are rendering results from the previous execution, we need the node execution cache to return the relevant data in terms of node id's instead of node module paths. This will allow the UI, which is node id driven, to do easy lookups.

Will need to followup to do the same for the output ids

### DIFF
--- a/src/vellum/workflows/workflows/tests/test_base_workflow.py
+++ b/src/vellum/workflows/workflows/tests/test_base_workflow.py
@@ -1,5 +1,5 @@
 import pytest
-from uuid import uuid4
+from uuid import UUID, uuid4
 
 from vellum.workflows.edges.edge import Edge
 from vellum.workflows.inputs.base import BaseInputs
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/workflows/tests/test_base_workflow.py
```diff
@@ -384,8 +389,57 @@ class TestWorkflow(BaseWorkflow[Inputs, State]):
...
+        class Outputs(BaseNode.Outputs):
+            foo: str
+
+    # AND a workflow that uses both
+    class TestWorkflow(BaseWorkflow[BaseInputs, State]):
+        graph = NodeA
+
+    # WHEN we deserialize the state that had a legacy node execution cache format
+    last_span_id = str(uuid4())
+    node_a_full_name = "vellum.workflows.workflows.tests.test_base_workflow.test_base_workflow__deserialize_legacy_node_execution_cache.<locals>.NodeA"  # noqa: E501
```
dvargasfuertes, line 416: since we have records in clickhouse with this old way, we will forever maintain both. This also leaves he door open in the future if we want to relitigate absolute module paths as identifiers instead of ids

PR #1424, opened 2025-04-17
Update port context logic to prioritize node port attributes and fallback on source handle id

NgoHarrison:
Needed this for serialization so decided to break up the PRs


ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/node-ports.test.ts.snap
NgoHarrison, line 1: Realized we were doing `Port.on_else(None)` and that doesn't look good

dvargasfuertes: and is incorrect, `on_else()` doesn't even take in an arg haha

dvargas92495 (approved):

PR #1427, opened 2025-04-17
Fix function definition callable

vincent0426:
Pass in function only! Which will be normalized with magic `compile_function_definition`

### DIFF
--- a/src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
+++ b/src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
@@ -51,16 +51,12 @@ class ToolRouterNode(InlinePromptNode):
 def create_tool_router_node(
     ml_model: str,
     blocks: List[PromptBlock],
-    functions: List[FunctionDefinition],
+    functions: List[Callable[..., Any]],
     prompt_inputs: Optional[EntityInputsInterface],
 ) -> Type[ToolRouterNode]:
     Ports = type("Ports", (), {})
     for function in functions:
-        if function.name is None:
-            # We should not raise an error here since we filter out functions without names
-            raise ValueError("Function name is required")
-
-        function_name = function.name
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -113,14 +109,14 @@ def execute_function(self) -> BaseNode.Outputs:
...
-        result = function_callable(**arguments)
+        result = function(**arguments)
 
         self.state.chat_history.append(ChatMessage(role="FUNCTION", text=result))
 
         return self.Outputs()
 
     node = type(
-        f"FunctionNode_{function.name}",
+        f"FunctionNode_{function.__name__}",
```
vincent0426, line 119: thoughts on better naming for this?

dvargasfuertes: I was having a touch time coming up with one too the first time. I think what we have here is reasonable enough. I had thought of some sort of `f"FunctionNode{camelCase(function.__name__)}"`, but this would technically allow for collisions for two functions of the same name

tests/workflows/basic_tool_calling_node/workflow.py
```diff
@@ -47,22 +46,7 @@ class GetCurrentWeatherNode(ToolCallingNode):
...
-                "properties": {
-                    "location": {"type": "string", "description": "City and state, e.g. San Francisco, CA"},
-                    "unit": {"type": "string", "enum": ["celsius", "fahrenheit"]},
-                },
-                "required": ["location"],
-            },
-            forced=False,
-        ),
-    ]
+    functions = [get_current_weather]
```
dvargasfuertes, line 49: beautiful

dvargas92495 (approved):

PR #1431, opened 2025-04-17
Update port displays to read from node ports if defined and fallback to source handle

NgoHarrison:
More codegen changes I needed to serialize node ports: main motivation is I needed to reference the ids of each node port somewhere

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -81,6 +81,8 @@ from ...nodes.second_node import SecondNode
 
 
 class SecondNodeDisplay(BaseNodeDisplay[SecondNode]):
+    node_input_ids_by_name = {}
+    port_displays = {}
     display_data = NodeDisplayData(position=NodeDisplayPosition(x=0, y=0), width=None, height=None)
 "
 `;
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -81,8 +81,6 @@ from ...nodes.second_node import SecondNode
 
 
 class SecondNodeDisplay(BaseNodeDisplay[SecondNode]):
-    node_input_ids_by_name = {}
-    port_displays = {}
     display_data = NodeDisplayData(position=NodeDisplayPosition(x=0, y=0), width=None, height=None)
 "
 `;
--- a/ee/codegen/src/__test__/nodes/__snapshots__/templating-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/templating-node.test.ts.snap
@@ -91,6 +91,43 @@ class TemplatingNode(BaseTemplatingNode[BaseState, Json]):
 "
 `;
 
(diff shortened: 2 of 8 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/templating-node.test.ts.snap
NgoHarrison, line 1: I chose templating node arbitrarily to show the change of prioritizing node port attribute if defined

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -68,6 +68,7 @@ from ...nodes.second_node import SecondNode
 
 class SecondNodeDisplay(BaseNodeDisplay[SecondNode]):
     node_input_ids_by_name = {}
+    port_displays = {}
```
dvargasfuertes: [nit] let's skip codegenning this guy if it's empty

dvargasfuertes: we recently merged a change that did the same for `node_input_ids_by_name`

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
```diff
@@ -52,6 +61,11 @@ from ...nodes.my_custom_node import MyCustomNode
 @BaseRetryNodeDisplay.wrap(node_id=UUID("adornment-3"))
 class MyCustomNodeDisplay(BaseNodeDisplay[MyCustomNode]):
     node_input_ids_by_name = {}
+    port_displays = {
+        MyCustomNode.Ports.default_port: PortDisplayOverrides(
```
dvargasfuertes, line 63: [nit] we should update our `nodePortFactory` to make sure the DEFAULT port always uses the `default` attribute name

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -213,6 +213,67 @@ export abstract class BaseNode<
...
+      return;
+    } else {
+      const portDisplayOverridesDict = new Map();
+      const portIds = new Set(this.nodeData.ports.map((port) => port.id));
+
+      Array.from(this.workflowContext.portContextById.entries()).forEach(
+        ([portId, context]) => {
+          const isPortInCurrentNode = portIds.has(portId);
+          console.log(portIds);
+          console.log(portId)
```
dvargasfuertes:
```suggestion
```

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -213,6 +213,67 @@ export abstract class BaseNode<
...
+          if (isPortInCurrentNode) {
+            const portDisplayOverrides = python.instantiateClass({
+              classReference: python.reference({
+                name: "PortDisplayOverrides",
```
dvargasfuertes, line 235: [oos] Oh damn, we should eventually consolidate these too to just `PortDisplay`

dvargas92495 (approved):

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
NgoHarrison: Updated

PR #1432, opened 2025-04-18
Fix chat history input

vincent0426:
Remove passed in chat history field


tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -113,13 +112,10 @@ def generate_prompt_events(*args, **kwargs) -> Iterator[ExecutePromptEvent]:
...
             PromptRequestStringInput(key="question", type="STRING", value="What's the weather like in San Francisco?"),
-            PromptRequestChatHistoryInput(
+            PromptRequestJsonInput(
                 key="chat_history",
-                type="CHAT_HISTORY",
-                value=[
-                    ChatMessage(role="USER", text="Hello, how are you?"),
-                    ChatMessage(role="ASSISTANT", text="I'm good, thank you!"),
-                ],
+                type="JSON",
```
vincent0426, line 117: `compile_prompt_inputs` treat empty array as `JSON` type instead of `CHAT_HISTORY`

dvargasfuertes: yea, alittle wonky, but all we actually can infer from a raw array is that it's just that

dvargas92495 (approved):

PR #1433, opened 2025-04-18
Serialize node ports attribute

NgoHarrison:
Wait for this PR to go in:
https://github.com/vellum-ai/vellum-python-sdks/pull/1431

Majority of these are just updating json's and assertions to expect a default node port attribute for nodes (conditional nodes will have their corresponding node port attributes like if, elif, else)


dvargas92495 (approved):
dope

PR #1435, opened 2025-04-18
Serialize function definition

vincent0426:
Initial serialization for function definitions, will add test serialization for the whole tool calling workflow

will get
`{'id': '78324739-ff89-47a5-902b-10da0cb95c6d', 'name': 'functions', 'value': {'type': 'FUNCTION_DEFINITIONS', 'functions': [{'id': '0328f399-1d74-4375-8e65-62ec31b7db57', 'block_type': 'FUNCTION_DEFINITION', 'properties': {'function_name': 'get_current_weather', 'function_description': None, 'function_parameters': {'type': 'object', 'properties': {'location': {'type': 'string'}, 'unit': {'type': 'string'}}, 'required': ['location', 'unit']}, 'function_forced': None, 'function_strict': None}}]}}`

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -1,4 +1,3 @@
-from collections.abc import Callable
 from functools import cached_property
 import inspect
 from uuid import UUID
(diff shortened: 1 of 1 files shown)


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -487,6 +490,35 @@ def serialize_value(self, display_context: "WorkflowDisplayContext", value: Any)
...
+                        "function_parameters": function.parameters,
+                        "function_forced": function.forced,
+                        "function_strict": function.strict,
+                    },
+                }
+                for i, function in enumerate(normalized_functions)
+            ]
+            return {
+                "type": "FUNCTION_DEFINITIONS",
+                "functions": functions,
```
vincent0426, line 518: any naming suggestions here?

dvargasfuertes: Yea I think the outcome of this TDD will be very relevant here: https://www.notion.so/vellum-ai/TDD-Serializing-Complex-Attributes-1d89035c57bd804a8e74c91af2451505

The most likely outcome is that the if check simply if it's a list, then recurses into `serialize_value` again for each item.

Then there's another case for just pydantic models where we could just `.dump()`

Happy to chat this one more about it tm, but in the meantime, I suggest we separately standup a test, see what that generates so far, and see how far into `vellum push` you can get for a workflow with a ToolCallingNode defined

PR #1436, opened 2025-04-18
Avoid Lazy Reference when second node references first in Loop

dvargas92495:
Noticed this in the demo we are preparing - in the case of a loop that goes 1 -> 2 -> 3 -> 4, we were ordering 2,3,4,1. This PR ensures that we now process 1 first.

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -24,6 +24,8 @@ exports[`WorkflowProjectGenerator > LazyReference > should not generate LazyRefe
 "from vellum.workflows.nodes import BaseNode
 from vellum.workflows.ports import Port
 
+from .start_node import StartNode
+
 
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -20,6 +20,20 @@ class APINode(BaseAPINode):
...
+exports[`WorkflowProjectGenerator > LazyReference > should not generate LazyReference when there is a simple loop 1`] = `
+"from vellum.workflows.nodes import BaseNode
+from vellum.workflows.ports import Port
+
+from .start_node import StartNode
+
+
+class RouterNode(BaseNode):
+    class Ports(BaseNode.Ports):
+        top = Port.on_if(StartNode.Outputs.result.equals("top"))
```
dvargasfuertes, line 32: This was previously a `LazyReference("")`

vincent0426 (commented):
✅

PR #1437, opened 2025-04-18
clean up default params to plain text block

dvargas92495:
Discovered while setting up a demo. We don't need to codegen these defaults, and now plain text prompt block works like the rest of them


marlenc3 (commented):
✅

PR #1439, opened 2025-04-18
Nuke BaseNodeVellumDisplay to consolidate around BaseNodeDisplay

dvargas92495:
This PR removes another redundant Vellum display class. As with the previous ones, this helps us centralize on BaseNodeDisplay, getting us incrementally closer to all legacy nodes being just base nodes by default


marlenc3:
✅

PR #1444, opened 2025-04-18
Ignore forward ref failures for now for pydantic v1 customers

dvargas92495:
Error was preventing users from running the SDK

Screenshot 2025-04-18 at 4 46 05 PM


dvargas92495:
The next release will need to include postprocess.ts changes to make sure we don't lose these changes

marlenc3 (approved):

PR #1445, opened 2025-04-19
Fix base inputs inheritance

dvargas92495:
Noticed this when trying to setup Map nodes during Freaky friday


src/vellum/workflows/inputs/base.py
```diff
@@ -15,10 +15,23 @@
 @dataclass_transform(kw_only_default=True)
 class _BaseInputsMeta(type):
     def __getattribute__(cls, name: str) -> Any:
```
dvargasfuertes, line 17: this method almost exactly matches `_BaseOutputsMeta(type).__getattribute__()`

PR #1446, opened 2025-04-19
Fix nested map node item reference

dvargas92495:
Discovered this pretty gnarly bug during FF. Essentially, the `item` input always resolved to the innermost map node's `item` value. So this PR adds handling that says that not just the names need to match, but the parent class does too. This essentially matches the UI which is matching by id.

### DIFF
--- a/src/vellum/workflows/inputs/base.py
+++ b/src/vellum/workflows/inputs/base.py
@@ -1,4 +1,4 @@
-from typing import Any, Iterator, Tuple, Type, Union, get_args, get_origin
+from typing import Any, Dict, Iterator, Set, Tuple, Type, Union, get_args, get_origin
 from typing_extensions import dataclass_transform
 
 from pydantic import GetCoreSchemaHandler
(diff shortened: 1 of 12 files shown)


src/vellum/workflows/inputs/base.py
```diff
@@ -42,14 +46,20 @@ def __getattribute__(cls, name: str) -> Any:
     def __iter__(cls) -> Iterator[InputReference]:
         # We iterate through the inheritance hierarchy to find all the WorkflowInputReference attached to this
         # Inputs class. __mro__ is the method resolution order, which is the order in which base classes are resolved.
+        yielded_attr_names: Set[str] = set()
```
dvargasfuertes, line 49: fixes another bug discovered, copying `_BaseOutputsMeta.__iter__`

tests/workflows/stream_subworkflow_node/workflow.py
```diff
@@ -8,12 +8,12 @@
 from vellum.workflows.state import BaseState
 
 
-class Inputs(BaseInputs):
+class InnerInputs(BaseInputs):
```
dvargasfuertes, line 11: We can no longer (and should never anyway) share inputs classes across workflows

PR #1450, opened 2025-04-20
Silence warning from ruff with a new Writer()

dvargas92495:
We keep seeing these guys locally

Screenshot 2025-04-20 at 4 42 31 PM

Figured out how to silence them by sporting an extension of our Writer class. Gonna check in with Fern to see if they'd be willing to main line this since there's no rush from us to implement this


dvargas92495:
Should be solved whenever we bump fern I think

PR #1451, opened 2025-04-21
Introduce node data factory builder for generic fields like ports and attributes

NgoHarrison:
Context: This came up from initial QAing of node ports and I noticed an error there. That was done on a code exec node (I chose this randomly for testing) and am now getting slightly annoyed of having to introduce these fields that are meant to be generic in the future on all node types (ports, attributes, etc.). This in my opinion also eliminates the inconsistency with our factories and how some legacy node types have ports as an arg and how some don't (This is because we chose whatever one we are fixing a bug for). This PR just introduces a simple builder pattern wrapper on our node data factories and I included a test I purposely skipped that will be implemented in a follow up PR where I do the actual fix needed for Ports code generation

### DIFF
--- a/ee/codegen/src/__test__/helpers/node-data-factory-builder.ts
+++ b/ee/codegen/src/__test__/helpers/node-data-factory-builder.ts
@@ -1,0 +1,29 @@
+import { NodeAttribute, NodePort } from "src/types/vellum";
+
+export class NodeDataFactoryBuilder<T> {
+  public nodeData: T;
+
+  constructor(nodeData: T) {
+    this.nodeData = nodeData;
+  }
+
+  withPorts(ports: NodePort[]): NodeDataFactoryBuilder<T> {
+    this.nodeData = {
+      ...this.nodeData,
+      ports: ports,
+    };
+    return this;
+  }
+
+  withAttributes(attributes: NodeAttribute[]): NodeDataFactoryBuilder<T> {
+    this.nodeData = {
+      ...this.nodeData,
+      attributes,
+    };
+    return this;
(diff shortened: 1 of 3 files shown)


ee/codegen/src/__test__/helpers/node-data-factory-builder.ts
```diff
@@ -0,0 +1,29 @@
+import { NodeAttribute, NodePort } from "src/types/vellum";
+
+export class NodeDataFactoryBuilder<T> {
```
dvargasfuertes, line 3: [nit] Ideally `T` is a `T extends WorkflowNodeType` or whatever the latter is called in our repo

ee/codegen/src/__test__/nodes/code-execution-node.test.ts
```diff
@@ -367,4 +372,40 @@ describe("CodeExecutionNode", () => {
...
+          id: ifPortId,
+        }),
+        nodePortFactory({
+          type: "ELSE",
+          id: elsePortId,
+        }),
+      ];
+
+      const nodeData = codeExecutionNodeFactory()
+        .withPorts(nodePortsData)
```
dvargasfuertes, line 392: [q] Why this over `codeExecutionNodeFactory({ports: nodePortsData})`? just curious, cool with new pattern too

dvargas92495 (approved):

vincent0426 (approved):

NgoHarrison:
I should note that I am gonna be slowly migrating the other factories to return this builder later on

PR #1453, opened 2025-04-21
Fix codegen of node ports on legacy nodes

NgoHarrison:
We are codegening this currently (Notice how we are referencing the current node's output in the ports):
`@TryNode.wrap()
class CodeExecutionNode(BaseCodeExecutionNode[BaseState, Union[float, int]]):
    filepath = "./script.py"
    code_inputs = {}
    runtime = "PYTHON_3_11_6"
    packages = []

    class Ports(BaseCodeExecutionNode.Ports):
        port_5036eae0_fd01_4afc_9825_a029b4db0bf4 = Port.on_if(CodeExecutionNode.Outputs.result.equals(1))
        port_1a5d5ca5_7abb_4a06_9472_8e8b88d6d19b = Port.on_elif(CodeExecutionNode.Outputs.result.does_not_equal(0))
        ace58a77_2dcd_428e_8d44_4b7167e356c6 = Port.on_else()`

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/code-execution-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/code-execution-node.test.ts.snap
@@ -112,6 +112,51 @@ class CodeExecutionNode(BaseCodeExecutionNode[BaseState, str]):
 "
 `;
 
+exports[`CodeExecutionNode > basic with ports > with ports referencing current node output > getNodeFile 1`] = `
+"from vellum.workflows.nodes.displayable import (
+    CodeExecutionNode as BaseCodeExecutionNode,
+)
+from vellum.workflows.ports import Port
+from vellum.workflows.references import LazyReference
+from vellum.workflows.state import BaseState
+
+
+class CodeExecutionNode(BaseCodeExecutionNode[BaseState, str]):
+    filepath = "./script.py"
+    code_inputs = {}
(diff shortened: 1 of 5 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/code-execution-node.test.ts.snap
```diff
@@ -112,6 +112,51 @@ class CodeExecutionNode(BaseCodeExecutionNode[BaseState, str]):
...
+
+class CodeExecutionNode(BaseCodeExecutionNode[BaseState, str]):
+    filepath = "./script.py"
+    code_inputs = {}
+    runtime = "PYTHON_3_11_6"
+    packages = None
+
+    class Ports(BaseCodeExecutionNode.Ports):
+        if_port = Port.on_if(
+            LazyReference("CodeExecutionNode.Outputs.result").equals("Hello, World!")
```
dvargasfuertes: [nit] we should use the lambda variant here:
```python
LazyReference(lambda: CodeExecutionNode.Outputs.result).equals("Hello, World!")
```

dvargas92495 (approved):

ee/codegen/src/__test__/nodes/__snapshots__/code-execution-node.test.ts.snap
NgoHarrison: Updated

PR #1461, opened 2025-04-22
[APO-132] Inline prompt node non streaming support

NgoHarrison:
Codegen PR will come after

### DIFF
--- a/src/vellum/workflows/nodes/displayable/bases/base_prompt_node/node.py
+++ b/src/vellum/workflows/nodes/displayable/bases/base_prompt_node/node.py
@@ -18,6 +18,8 @@ class BasePromptNode(BaseNode, Generic[StateType]):
 
     request_options: Optional[RequestOptions] = None
 
+    run_with_stream: bool = True
+
     class Trigger(BaseNode.Trigger):
         merge_behavior = MergeBehavior.AWAIT_ANY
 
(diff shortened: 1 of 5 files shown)


src/vellum/workflows/nodes/displayable/bases/base_prompt_node/node.py
```diff
@@ -18,6 +18,8 @@ class BasePromptNode(BaseNode, Generic[StateType]):
 
     request_options: Optional[RequestOptions] = None
 
+    run_with_stream: bool = True
```
dvargasfuertes: [blocking] the vellum side PR showed that you added this to `PromptSettings` already? I would expect `self.settings.stream_enabled`

src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
```diff
@@ -103,17 +103,36 @@ def _get_prompt_event_stream(self) -> Iterator[AdHocExecutePromptEvent]:
...
+                    parameters=self.parameters,
+                    blocks=self.blocks,
+                    settings=self.settings,
+                    functions=normalized_functions,
+                    expand_meta=self.expand_meta,
+                    request_options=request_options,
+                )
+                yield response
+
+            return single_prompt_event_generator()
```
dvargasfuertes: [nit] instead of defining a new function, does the following work?

```python
response = ...
return iter([response])
```

src/vellum/workflows/nodes/displayable/bases/base_prompt_node/node.py
NgoHarrison: Oops yea, this PR was before my front end changes so forgot to swap them. Updated now

src/vellum/workflows/nodes/displayable/bases/base_prompt_node/node.py
```diff
@@ -39,34 +39,9 @@ def run(self) -> Iterator[BaseOutput]:
                 code=WorkflowErrorCode.INTERNAL_ERROR,
             )
 
+    @abstractmethod
     def _process_prompt_event_stream(self) -> Generator[BaseOutput, None, Optional[List[PromptOutput]]]:
```
dvargasfuertes: [nit] I'd like to avoid making both `_get_prompt_event_stream` and `_process_prompt_event_stream` abstract. In fact, long term I'm hoping to unabstract `_get_prompt_event_stream` so that extensions of `BasePromptNode` have less that they are responsible for implementing

dvargas92495 (approved):

src/vellum/workflows/nodes/displayable/bases/base_prompt_node/node.py
NgoHarrison: Undid this change

PR #1462, opened 2025-04-22
Harrison/apo 132 codegen stream setting for inline prompt node

NgoHarrison:
Wait for this https://github.com/vellum-ai/vellum-python-sdks/pull/1461

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
@@ -73,7 +73,6 @@ class PromptNode(InlinePromptNode):
         logit_bias={},
         custom_parameters={},
     )
-    run_with_stream = True
 "
 `;
 
@@ -157,7 +156,6 @@ class PromptNode(InlinePromptNode):
         logit_bias={},
         custom_parameters={},
     )
-    run_with_stream = True
 "
 `;
 
@@ -234,6 +232,5 @@ class PromptNode(InlinePromptNode):
         logit_bias={},
         custom_parameters={},
     )
-    run_with_stream = True
 "
 `;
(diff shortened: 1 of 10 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
```diff
@@ -73,6 +73,7 @@ class PromptNode(InlinePromptNode):
         logit_bias={},
         custom_parameters={},
     )
+    run_with_stream = True
```
dvargasfuertes: [blocking] let's skip codgenning this field if it matches the default

Remember, we almost always want to minimize changes to as many existing snapshots as possible, or comment as to why we should make that change

ee/codegen/src/generators/nodes/inline-prompt-node.ts
```diff
@@ -142,6 +142,17 @@ export class InlinePromptNode extends BaseSingleFileNode<
       })
     );
 
+    if (this.nodeData.data.execConfig.settings) {
```
dvargasfuertes, line 147: [blocking] did we not already have logic codegen-ning `PromptSettings`? coupled with my feedback in the other PR, we should start to do so now

ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
NgoHarrison: Ah yes, updated

ee/codegen/src/generators/nodes/inline-prompt-node.ts
NgoHarrison: We did not, but updated it now to look for it and conditionally codegen if the fields are provided

ee/codegen/src/generators/nodes/inline-prompt-node.ts
```diff
@@ -142,6 +144,43 @@ export class InlinePromptNode extends BaseSingleFileNode<
...
+            name: "timeout",
+            value: python.TypeInstantiation.float(timeout),
+          })
+        );
+      }
+      if (!isNil(streamEnabled)) {
+        args.push(
+          python.methodArgument({
+            name: "stream_enabled",
+            value: python.TypeInstantiation.bool(streamEnabled),
```
dvargasfuertes, line 164: [nit] I think we should still skip if it's the default value (in this case, `true`), since the FE may be sending this value during this dual writing period

dvargas92495 (approved):

PR #1464, opened 2025-04-22
Fix try node adornment

vincent0426:
~~[blocking on error code]~~

Try node adornment is not generating correct arguments. 

Move the error output ID check after generating all adornments, and if the try node is not generated and the error output ID exists, then we generate it.

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -304,8 +304,8 @@ Summarize the following text: hello today is feb 7 bad weather
 
 exports[`WorkflowProjectGenerator > retry node adornment > should correctly generate default code for adornments 1`] = `
 "from vellum import PromptParameters
-from vellum.workflows.nodes.core import TryNode
 from vellum.workflows.nodes.core.retry_node.node import RetryNode
+from vellum.workflows.nodes.core.try_node.node import TryNode
 from vellum.workflows.nodes.displayable import InlinePromptNode
 
 
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -117,6 +117,31 @@ class FirstNode(BaseNode):
 "
 `;
 
+exports[`WorkflowProjectGenerator > Try adornment node > should generate try adornment node with correct attributes 1`] = `
+"from vellum import PromptParameters
+from vellum.workflows.nodes.core.try_node.node import TryNode
+from vellum.workflows.nodes.displayable import InlinePromptNode
+
+
+@TryNode.wrap(on_error_code="INVALID_WORKFLOW")
+class Prompt(InlinePromptNode):
+    ml_model = "gpt-4o-mini"
+    blocks = []
+    prompt_inputs = {}
+    parameters = PromptParameters(
+        stop=[],
+        temperature=None,
+        max_tokens=None,
+        top_p=None,
+        top_k=None,
+        frequency_penalty=None,
+        presence_penalty=None,
+        logit_bias=None,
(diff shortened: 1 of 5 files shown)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -279,8 +304,8 @@ Summarize the following text: hello today is feb 7 bad weather
 
 exports[`WorkflowProjectGenerator > retry node adornment > should correctly generate default code for adornments 1`] = `
 "from vellum import PromptParameters
-from vellum.workflows.nodes.core import TryNode
 from vellum.workflows.nodes.core.retry_node.node import RetryNode
+from vellum.workflows.nodes.core.try_node.node import TryNode
```
dvargasfuertes, line 308: [q] why did these imports update?

ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
```diff
@@ -135,8 +135,8 @@ from vellum.workflows.nodes.displayable import InlinePromptNode
 from ..inputs import Inputs
 
 
-@TryNode.wrap()
 @RetryNode.wrap(max_attempts=3, delay=2)
+@TryNode.wrap()
 class PromptNode(InlinePromptNode):
```
dvargasfuertes, line 140: [q] why did this order change? In any case, I would expect this new version to be "more" correct

dvargas92495 (approved):

ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
vincent0426: cause we now first generate adornments defined in adornments section then error output id if we didn't define try node adornment, before this we generate error output id one first

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
vincent0426: cause this try node is defined explicitly in adornments section which uses `adornment.base.module` as module path but the error output id one uses `this.workflowContext.sdkModulePathNames.CORE_NODES_MODULE_PATH` should we just use `this.workflowContext.sdkModulePathNames.CORE_NODES_MODULE_PATH` for all adornments?

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -117,6 +117,31 @@ class FirstNode(BaseNode):
 "
 `;
 
+exports[`WorkflowProjectGenerator > Try adornment node > should generate try adornment node with correct attributes 1`] = `
+"from vellum import PromptParameters
+from vellum.workflows.nodes.core.try_node.node import TryNode
+from vellum.workflows.nodes.displayable import InlinePromptNode
+
+
+@TryNode.wrap(on_error_code="INVALID_WORKFLOW")
```
NgoHarrison, line 126: I don't think this is valid. on_error_code takes `WorkflowErrorCode` which is an enum not a string

vincent0426: fixed here https://github.com/vellum-ai/vellum-python-sdks/pull/1465 but it might be better if we could generate enum type (not sure introduce an enum type or hardcode enum value)

NgoHarrison: https://github.com/vellum-ai/vellum-python-sdks/pull/1476/files#diff-3017a6fcfc7dd401ea7ff9e88c6d94e7e0d376ad80527cf5b8a85499b99f5e8dR413-R444

Take a look at how I did it here

PR #1465, opened 2025-04-23
Support try node str error

vincent0426:
This should unblock https://github.com/vellum-ai/vellum-python-sdks/pull/1464

We generate str as error in try adornment but we compare it with enum values

### DIFF
--- a/src/vellum/workflows/nodes/core/try_node/node.py
+++ b/src/vellum/workflows/nodes/core/try_node/node.py
@@ -21,7 +21,8 @@ class TryNode(BaseAdornmentNode[StateType], Generic[StateType]):
     subworkflow: Type["BaseWorkflow"] - The Subworkflow to execute
     """
 
-    on_error_code: Optional[WorkflowErrorCode] = None
+    # TODO: We could remove str after we support generating enum values.
+    on_error_code: Optional[WorkflowErrorCode | str] = None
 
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/core/try_node/node.py
```diff
@@ -65,7 +65,10 @@ def run(self) -> Iterator[BaseOutput]:
                     message="Subworkflow unexpectedly paused within Try Node",
                 )
             elif event.name == "workflow.execution.rejected":
-                if self.on_error_code and self.on_error_code != event.error.code:
+                # TODO: This is a hack to support the case where the on_error_code is a string.
+                # We should remove this once we support generating enum values.
```
dvargasfuertes: [oos] What if it wasn't a hack and we changed the type annotation of `on_error_code` to be `WorkflowErrorCode | str`?

dvargas92495 (approved):

src/vellum/workflows/nodes/core/try_node/node.py
vincent0426: I think we still need to do transform here for comparison if we are not passing in an enum type, but we could fix this after we support generating an enum

vincent0426: but i fixed the comment

PR #1467, opened 2025-04-23
Serialize null adornment value

vincent0426:
What I am pulling down for null adornment value is

```
'attributes': [{'id': 'b1bcf28e-a8d8-4746-9736-f96f7de73c73',
 'name': 'on_error_code',
 'value': None}]},
```

not sure if this would break something

### DIFF
--- a/ee/codegen_integration/fixtures/simple_node_with_try_wrap/display_data/simple_node_with_try_wrap.json
+++ b/ee/codegen_integration/fixtures/simple_node_with_try_wrap/display_data/simple_node_with_try_wrap.json
@@ -162,13 +162,7 @@
               {
                 "id": "55821c68-5207-4a5b-8345-e89dee65f71a",
                 "name": "on_error_code",
-                "value": {
-                  "type": "CONSTANT_VALUE",
-                  "value": {
-                    "type": "JSON",
-                    "value": null
-                  }
-                }
+                "value": null
               }
             ]
           }
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -157,7 +157,9 @@ class BaseNodeDisplay(Generic[NodeType], metaclass=BaseNodeDisplayMeta):
                     {
                         "id": id,
                         "name": attribute.name,
-                        "value": self.serialize_value(display_context, attribute.instance),
+                        "value": (
+                            self.serialize_value(display_context, attribute.instance) if attribute.instance else None
+                        ),
                     }
                 )
             except ValueError as e:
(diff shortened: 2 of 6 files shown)


ee/codegen_integration/fixtures/simple_node_with_try_wrap/display_data/simple_node_with_try_wrap.json
```diff
@@ -162,13 +162,7 @@
               {
                 "id": "55821c68-5207-4a5b-8345-e89dee65f71a",
                 "name": "on_error_code",
-                "value": {
-                  "type": "CONSTANT_VALUE",
-                  "value": {
-                    "type": "JSON",
-                    "value": null
```
dvargasfuertes, line 169: did our codegen not handle both? I think i'd prefern that approach

ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -157,7 +157,9 @@ def serialize(self, display_context: "WorkflowDisplayContext", **kwargs: Any) ->
                     {
                         "id": id,
                         "name": attribute.name,
-                        "value": self.serialize_value(display_context, attribute.instance),
+                        "value": (
+                            self.serialize_value(display_context, attribute.instance) if attribute.instance else None
+                        ),
```
dvargasfuertes, line 162: would love to keep each of these instances as they were - was there any negative downstream implications of serializaing `"JSON", null` over just `null`?

dvargas92495 (commented):
looking for some qualifying qs first

ee/vellum_ee/workflows/display/nodes/base_node_display.py
vincent0426: it may have two default values? 
- `null `
- `{"type": "CONSTANT_VALUE", "value": {"type": "JSON", "value": None}}`

default value may require additional logic to compare https://github.com/vellum-ai/vellum-python-sdks/blob/acb3a291d72d755091246abac2e94807dcee4719/ee/codegen/src/generators/nodes/bases/base.ts#L628-L644

https://github.com/vellum-ai/vellum-python-sdks/blob/acb3a291d72d755091246abac2e94807dcee4719/ee/codegen/src/generators/nodes/constants.ts#L5-L15

dvargasfuertes: ah ok I guess all I'm advocating for is that `serialize_value()` should be able to handle `value: None`, and return `None` there

ee/vellum_ee/workflows/display/nodes/vellum/try_node.py
```diff
@@ -37,7 +37,11 @@ def serialize(self, display_context: "WorkflowDisplayContext", **kwargs: Any) ->
                 {
                     "id": id,
                     "name": attribute.name,
-                    "value": self.serialize_value(display_context, cast(BaseDescriptor, attribute.instance)),
```
dvargasfuertes, line 40: [nit] we shouldn't need these `cast`s anymore, right?

PR #1468, opened 2025-04-23
Fix critical bug surrounding retry node adorned outputs

dvargas92495:
A user tried to move one of their workflows to the SDK, and saw one of their code nodes instantly regress. The root of the bug was pretty deep - all nodes that were adorned by a retry node was showing the same value. This bug was widespread, and is currently affecting any workflow with multiple retry nodes. This PR adds some tests catching the bug and puts out a fix

### DIFF
--- a/src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
+++ b/src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
@@ -59,8 +59,8 @@ def create_tool_router_node(
         function_name = function.__name__
         port_condition = LazyReference(
             lambda: (
-                ToolRouterNode.Outputs.results[0]["type"].equals("FUNCTION_CALL")
-                & ToolRouterNode.Outputs.results[0]["value"]["name"].equals(function_name)
+                node.Outputs.results[0]["type"].equals("FUNCTION_CALL")
+                & node.Outputs.results[0]["value"]["name"].equals(function_name)
             )
         )
         port = Port.on_if(port_condition)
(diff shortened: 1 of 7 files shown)


src/vellum/workflows/references/output.py
```diff
@@ -78,7 +78,7 @@ def _as_generator(self, node_output: Queue) -> Generator[_OutputType, None, Type
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, type(self)):
             return False
-        return super().__eq__(other) and self._outputs_class == other._outputs_class
+        return super().__eq__(other) and id(self._outputs_class) == id(other._outputs_class)
```
dvargasfuertes, line 81: Our Outputs class has custom equality method that we deferred editing and deferred investigating the root bug there, in case there were dependent consumers on how that equality evaluated today

marlenc3 (approved):

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -59,8 +59,8 @@ def create_tool_router_node(
         function_name = function.__name__
         port_condition = LazyReference(
             lambda: (
-                ToolRouterNode.Outputs.results[0]["type"].equals("FUNCTION_CALL")
-                & ToolRouterNode.Outputs.results[0]["value"]["name"].equals(function_name)
+                node.Outputs.results[0]["type"].equals("FUNCTION_CALL")
+                & node.Outputs.results[0]["value"]["name"].equals(function_name)
```
dvargasfuertes, line 63: cc: @vincent0426 for tool calling node changes. Note that we want to associate direct node output references instead of the base class

marlenc3:
✅  -- thought it was just tests on the first look, but hell yeah

PR #1469, opened 2025-04-23
Harrison/fix multiple port groups

NgoHarrison:
This came from qa'ing ports. Multiple port groups were not supported due to other validation logic


tests/workflows/basic_ports/elif_with_no_if/tests/test_workflow.py
```diff
@@ -5,8 +5,4 @@ def test_run_workflow():
     workflow = ElifWithNoIfWorkflow()
     terminal_event = workflow.run(inputs=Inputs(value="foo"))
     assert terminal_event.name == "workflow.execution.rejected"
-    base_module = __name__.split(".")[:-2]
-    assert (
-        terminal_event.error.message
-        == f"Class {'.'.join(base_module)}.workflow.ElifWithNoIfNode.Ports containing on_elif or on_else port conditions must have at least one on_if condition"  # noqa E501
-    )
+    assert terminal_event.error.message == "Port conditions must be in the following order: on_if, on_elif, on_else"
```
dvargasfuertes: [blocking] we should make sure the full class name is still included in the error message so that the user knows which node is malconfigured

tests/workflows/basic_ports/multiple_else/workflow.py
```diff
@@ -14,7 +14,6 @@ class Inputs(BaseInputs):
 class MultipleElseNode(BaseNode):
     class Ports(NodePorts):
         else_branch: Port = Port.on_else()
-        another_else_branch: Port = Port.on_else()
```
dvargasfuertes: [nit] why did we remove this one?

dvargas92495 (approved):
pending quick error message update

PR #1474, opened 2025-04-23
Support null value type for adornment

vincent0426:
We are getting both JSON null and pure None value in adornment value, this PR handles both scenarios

### DIFF
--- a/ee/codegen/src/generators/nodes/bases/base.ts
+++ b/ee/codegen/src/generators/nodes/bases/base.ts
@@ -640,6 +640,33 @@ export abstract class BaseNode<
       return true; // Include if no config exists for this attribute
     }
 
-    return attrConfig.defaultValue !== attributeValue; // Include if value differs from default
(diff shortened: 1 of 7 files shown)


ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -640,6 +640,33 @@ export abstract class BaseNode<
       return true; // Include if no config exists for this attribute
     }
 
-    return attrConfig.defaultValue !== attributeValue; // Include if value differs from default
+    const extractedValue = this.extractConstantValue(attributeValue);
+    return attrConfig.defaultValue !== extractedValue; // Include if value differs from default
+  }
+
+  private extractConstantValue(attributeValue: unknown): unknown {
```
dvargasfuertes, line 647: [oos] we should soon be able to type this as `attributeValue?: WorkflowValueDescriptor | null`, which will allow us to simplify the logic below

dvargas92495 (approved):

PR #1477, opened 2025-04-24
Align the workflow output name with the terminal output name

dvargas92495:
I noticed during the customer demo that we were serializing the wrong workflow output name. The full fix will require three PRs:
- This one, which edits terminal node name to be aligned with the workflow output name
- Next one, which will abstract out the BaseNodeDisplay.serialize_value() to its own util to be reused by workflows
- The last one, to start serializing workflow output values, which we have never done


ee/vellum_ee/workflows/display/workflows/tests/test_workflow_display.py
```diff
@@ -327,3 +328,43 @@ class Outputs(MyWorkflow.Outputs):
...
+    assert isinstance(terminal_node["data"], dict)
+    assert terminal_node["data"]["name"] == "answer"
+
+    # AND the output variable should have the correct name
+    assert isinstance(data["output_variables"], list)
+    assert isinstance(data["output_variables"][0], dict)
+    assert data["output_variables"][0]["key"] == "answer"
+    assert data["output_variables"][0]["type"] == "STRING"
+
+    # TODO: AND the output value should have the correct name
```
dvargasfuertes, line 365: Will be done with the third PR

vincent0426 (commented):
✅

PR #1478, opened 2025-04-24
Move serialize_value out of BaseNodeDisplay and into display utils

dvargas92495:
This PR simply moves the serialize methods out of BaseNodeDisplay and into the utils directory so that we could reuse it in the future for workflow outputs


marlenc3 (approved):

vincent0426 (commented):
✅

PR #1479, opened 2025-04-24
Fix try node display order

vincent0426:
Follow up display part for https://github.com/vellum-ai/vellum-python-sdks/pull/1464

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
@@ -93,11 +93,11 @@ from vellum_ee.workflows.display.nodes.types import (
 from ...nodes.prompt_node import PromptNode
 
 
+@BaseRetryNodeDisplay.wrap(node_id=UUID("e5de8d57-ae0d-4a4a-afb3-eb4cd6bdb0ac"))
 @BaseTryNodeDisplay.wrap(
     node_id=UUID("e7a1fbea-f5a7-4b31-a9ff-0d26c3de021f"),
     error_output_id=UUID("e7a1fbea-f5a7-4b31-a9ff-0d26c3de021f"),
 )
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
```diff
@@ -93,11 +93,11 @@ from vellum_ee.workflows.display.nodes.types import (
 from ...nodes.prompt_node import PromptNode
 
 
+@BaseRetryNodeDisplay.wrap(node_id=UUID("e5de8d57-ae0d-4a4a-afb3-eb4cd6bdb0ac"))
```
vincent0426, line 96: retry node is defined in front of try node in the adornments section

ee/codegen_integration/fixtures/simple_node_with_try_wrap/code/display/nodes/prompt_node.py
```diff
@@ -7,9 +7,7 @@
 from ...nodes.prompt_node import PromptNode
 
 
-@BaseTryNodeDisplay.wrap(
-    node_id=UUID("42823c15-2ba6-4c85-a0d7-74a4e0541a42"), error_output_id=UUID("42823c15-2ba6-4c85-a0d7-74a4e0541a42")
-)
+@BaseTryNodeDisplay.wrap(node_id=UUID("42823c15-2ba6-4c85-a0d7-74a4e0541a42"))
```
vincent0426, line 10: We explicitly defined base try node so it will be generated during adornments loop

ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node-retry.test.ts.snap
dvargasfuertes: oh I see in the adornments spec in the test file, retry node was always first, so this test was bugged. 👍

dvargas92495 (approved):
Again, recommend pairing with @NgoHarrison at some point to consolidate with the PR he was working on, seems like there was some overlap here

PR #1480, opened 2025-04-24
Fixing Subworkflow Deployment Inputs

marlenc3:
** related to https://linear.app/vellum/issue/APO-426/customer-workflow-from-evals-are-missing-input-events

adds in the prefix that is needed for lookup

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/subworkflow_deployment_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/subworkflow_deployment_node.py
@@ -29,7 +29,8 @@ class BaseSubworkflowDeploymentNodeDisplay(
                 input_name=variable_name,
                 value=variable_value,
                 display_context=display_context,
-                input_id=self.node_input_ids_by_name.get(variable_name),
+                input_id=self.node_input_ids_by_name.get(f"subworkflow_inputs.{variable_name}")
+                or self.node_input_ids_by_name.get(variable_name),
             )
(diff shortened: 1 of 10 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/subworkflow_deployment_node.py
```diff
@@ -29,7 +29,8 @@ def serialize(
                 input_name=variable_name,
                 value=variable_value,
                 display_context=display_context,
-                input_id=self.node_input_ids_by_name.get(variable_name),
+                input_id=self.node_input_ids_by_name.get(f"subworkflow_inputs.{variable_name}")
```
dvargasfuertes: [nit] 
```suggestion
                input_id=self.node_input_ids_by_name.get(f"{SubworkflowDeploymentNode.subworkflow_inputs.name}.{variable_name}")
```

This makes the dependencies on this attribute discrete instead of hardcoded

dvargas92495 (approved):
Dope! do we need to do the same for inline subworkflow node?

ee/vellum_ee/workflows/display/nodes/vellum/subworkflow_deployment_node.py
marlenc3: will update!

marlenc3:
> Dope! do we need to do the same for inline subworkflow node?

@dvargas92495  I double checked, and inline looked like it's working as intended!

PR #1481, opened 2025-04-24
Fixing Prompt Deployment Inputs

marlenc3:
** related to https://linear.app/vellum/issue/APO-426/customer-workflow-from-evals-are-missing-input-events

adding in the required prefix for prompt - inputs

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/prompt_deployment_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/prompt_deployment_node.py
@@ -28,7 +28,9 @@ class BasePromptDeploymentNodeDisplay(BaseNodeDisplay[_PromptDeploymentNodeType]
                     input_name=variable_name,
                     value=variable_value,
                     display_context=display_context,
-                    input_id=self.node_input_ids_by_name.get(f"prompt_inputs.{variable_name}")
+                    input_id=self.node_input_ids_by_name.get(
+                        f"{PromptDeploymentNode.prompt_inputs.name}.{variable_name}"
+                    )
                     or self.node_input_ids_by_name.get(variable_name),
                 )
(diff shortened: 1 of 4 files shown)


ee/codegen/src/generators/nodes/subworkflow-deployment-node.ts
```diff
@@ -11,10 +11,16 @@ import { BaseSingleFileNode } from "src/generators/nodes/bases/single-file-base"
 import { codegen } from "src/index";
 import { SubworkflowNode as SubworkflowNodeType } from "src/types/vellum";
 
+const INPUTS_PREFIX = "subworkflow_inputs";
```
dvargasfuertes: [nit] looks like the other PR crept into this one - would prefer to keep this one focused on prompt inputs

ee/vellum_ee/workflows/display/nodes/vellum/prompt_deployment_node.py
```diff
@@ -28,7 +28,8 @@ def serialize(
                     input_name=variable_name,
                     value=variable_value,
                     display_context=display_context,
-                    input_id=self.node_input_ids_by_name.get(variable_name),
+                    input_id=self.node_input_ids_by_name.get(f"prompt_inputs.{variable_name}")
```
dvargasfuertes: [nit] Same feedback here as the other PR to reference `PromptDeploymentNode.prompt_inputs.name` instead

dvargas92495 (approved):

ee/codegen/src/generators/nodes/subworkflow-deployment-node.ts
marlenc3: oh boo - I rebased main and everything. I will remove the subworkflow updates from here

PR #1483, opened 2025-04-24
Generate a warning when we get empty expressions on if and elif ports

NgoHarrison:
Moving some safety checks to here as stated in comments of this PR:
https://github.com/vellum-ai/vembda-service/pull/204

### DIFF
--- a/ee/codegen/src/generators/node-port.ts
+++ b/ee/codegen/src/generators/node-port.ts
@@ -4,6 +4,7 @@ import { Writer } from "@fern-api/python-ast/core/Writer";
 
 import { WorkflowContext } from "src/context";
 import { BaseNodeContext } from "src/context/node-context/base";
+import { NodeAttributeGenerationError } from "src/generators/errors";
 import { WorkflowValueDescriptor } from "src/generators/workflow-value-descriptor";
 import {
   NodePort as NodePortType,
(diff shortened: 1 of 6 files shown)


ee/codegen/src/generators/node-port.ts
```diff
@@ -110,19 +111,36 @@ export class NodePorts extends AstNode {
...
+        })
+      );
+    } else if (
+      isNilOrEmpty(descriptor) &&
+      (attribute === "on_if" || attribute === "on_elif")
+    ) {
+      this.workflowContext.addError(
+        new NodeAttributeGenerationError(
+          "Expected IF / ELIF Ports to contain an expression",
+          "WARNING"
```
dvargasfuertes, line 128: dope 👍

src/vellum/workflows/ports/port.py
```diff
@@ -73,11 +73,15 @@ def __rshift__(self, other: GraphTarget) -> Graph:
         return Graph.from_edge(edge)
 
     @staticmethod
-    def on_if(condition: BaseDescriptor, fork_state: bool = False) -> "Port":
+    def on_if(condition: Optional[BaseDescriptor] = None, fork_state: bool = False):
+        if condition is None:
+            raise WorkflowInitializationException("Please verify that your IF ports have defined expressions")
```
dvargasfuertes: [blocking] I would love to get the node class' name in this error message to help the user know exactly which node was broken. I think instead of raising here, we could define a `validate()` method on `Port`, and then call `port.validate()` after this line: https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/nodes/bases/base.py#L113

src/vellum/workflows/ports/port.py
```diff
@@ -73,11 +73,15 @@ def __rshift__(self, other: GraphTarget) -> Graph:
...
         return Port(condition=condition, condition_type=ConditionType.IF, fork_state=fork_state)
 
     @staticmethod
-    def on_elif(condition: BaseDescriptor, fork_state: bool = False) -> "Port":
+    def on_elif(condition: Optional[BaseDescriptor] = None, fork_state: bool = False) -> "Port":
+        if condition is None:
+            raise WorkflowInitializationException("Please verify that your ELIF ports have defined expressions")
```
dvargasfuertes: [oos] Part of me is also wondering if it's actually fine to accept null here and that always resolves to falsy, similar to python/javascript. am down to keep the exceptions for now though and just keep this noodling in the back of our mind

dvargas92495 (approved):
pending `validate` suggestion

PR #1484, opened 2025-04-24
Introduce workflow raw data output values serialization

dvargas92495:
Supports serializing the new `workflow_raw_data.output_values` field. Initially blocked on some tests that we should simplify first

EDIT: Now ready for review

### DIFF
--- a/ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_api_node_serialization.py
+++ b/ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_api_node_serialization.py
@@ -54,7 +54,6 @@ def test_serialize_workflow(vellum_client):
 
     # AND its raw data should be what we expect
     workflow_raw_data = serialized_workflow["workflow_raw_data"]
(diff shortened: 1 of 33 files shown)


ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_api_node_serialization.py
```diff
@@ -54,7 +54,6 @@ def test_serialize_workflow(vellum_client):
 
     # AND its raw data should be what we expect
     workflow_raw_data = serialized_workflow["workflow_raw_data"]
-    assert workflow_raw_data.keys() == {"edges", "nodes", "display_data", "definition"}
```
dvargasfuertes, line 57: Each of the following tests are ensuring serialization for a node - so this assertion isn't relevant. It's already covered by our workflow level tests.

The same could be said for a lot of the other assertions in these tests, but wanted to keep the PR focused

marlenc3:
✅

PR #1485, opened 2025-04-24
Enforce attribute ordering based on class definition

dvargas92495:
We want to enforce the ordering of attributes so that during serialization, we don't get false positives in changes. This improves our main helper for collecting class attributes to maintain order

### DIFF
--- a/ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_adornments_serialization.py
+++ b/ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_adornments_serialization.py
@@ -237,8 +237,12 @@ def test_serialize_node__stacked():
     exec_config = workflow_display.serialize()
 
     # THEN the workflow display is created successfully
+    assert isinstance(exec_config["workflow_raw_data"], dict)
+    assert isinstance(exec_config["workflow_raw_data"]["nodes"], list)
     inner_stacked_generic_node = [
-        node for node in exec_config["workflow_raw_data"]["nodes"] if node["type"] == "GENERIC"
+        node
+        for node in exec_config["workflow_raw_data"]["nodes"]
+        if isinstance(node, dict) and node["type"] == "GENERIC"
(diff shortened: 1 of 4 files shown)


ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_adornments_serialization.py
```diff
@@ -237,124 +237,82 @@ class StackedWorkflow(BaseWorkflow):
     exec_config = workflow_display.serialize()
 
     # THEN the workflow display is created successfully
+    inner_stacked_generic_node = [
+        node for node in exec_config["workflow_raw_data"]["nodes"] if node["type"] == "GENERIC"
+    ][0]
     assert not DeepDiff(
         {
-            "workflow_raw_data": {
```
dvargasfuertes, line 242: I was notified of this issue by noticing this test failing while I was doing: https://github.com/vellum-ai/vellum-python-sdks/pull/1484

This test should focus solely on the node we're serializing, since the rest of workflow raw data is unrelated. I'm going to follow up this PR with doing a similar cleanup refactor for the rest of the tests that were failing for unrelated reasons from the above PR

ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_adornments_serialization.py
```diff
@@ -237,124 +237,82 @@ class StackedWorkflow(BaseWorkflow):
...
-            },
-            "input_variables": [],
-            "state_variables": [],
-            "output_variables": [],
+            ],
+            "attributes": [],
+            "outputs": [],
         },
-        exec_config,
-        ignore_order=True,
```
dvargasfuertes, line 359: We should look to `ignore_order` entirely from this repo. We actually really care about order bc we need to ensure consistent snapshots/codegens

marlenc3:
✅

PR #1487, opened 2025-04-24
Remove str and use enums for error codes for adornments

NgoHarrison:


src/vellum/workflows/nodes/core/try_node/tests/test_node.py
```diff
@@ -59,7 +59,7 @@ def run(self) -> Outputs:
 
 def test_try_node__on_error_code__str():
```
vincent0426: I think we could remove this test

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -383,17 +383,54 @@ export abstract class BaseNode<
...
+                      constantValue.value.type === "NUMBER" &&
+                      constantValue.value.value !== undefined
+                    ) {
+                      const numValue = constantValue.value.value;
+                      return python.methodArgument({
+                        name: attr.name,
+                        value: Number.isInteger(numValue)
+                          ? python.TypeInstantiation.int(numValue)
+                          : python.TypeInstantiation.float(numValue),
+                      });
```
vincent0426: Could we just return as usual? I think it still works for both int and float
```python
else {
      return python.methodArgument({
        name: attr.name,
        value: new WorkflowValueDescriptor({
          workflowValueDescriptor: attr.value,
          nodeContext: this.nodeContext,
          iterableConfig: { endWithComma: false },
        }),
      });
    }
```

NgoHarrison: Yup agreed, just wanted to specify the actual difference to align with our type hints but can change

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -383,17 +383,54 @@ export abstract class BaseNode<
...
+              arguments_: filteredAttributes
+                .map((attr) => {
+                  if (attr.value) {
+                    const constantValue =
+                      attr.value as ConstantValueWorkflowReference;
+                    // This is for error codes on try and retry args
+                    if (
+                      constantValue.type === "CONSTANT_VALUE" &&
+                      constantValue.value &&
+                      constantValue.value.type === "STRING" &&
```
dvargasfuertes: [blocking] wait, doesn't this case _all_ string attributes as `WorkflowErrorCode`? Could we add a generic node test that is just `foo = "hello"` to confirm?

src/vellum/workflows/nodes/core/try_node/tests/test_node.py
NgoHarrison: Removed

ee/codegen/src/generators/nodes/bases/base.ts
NgoHarrison: Updated

ee/codegen/src/generators/nodes/bases/base.ts
dvargasfuertes: oh I see this is just for adornment attributes. Still, once we have some future string adornment attribute, this will break.

I believe we should:
- Keep `WorkflowValueDescriptor()`
- pass in some arg to `WorkflowValueDescriptor` that represents the "attribute config" of `TryNode.on_error_code`, namely that it resolves to an enum
- Eventually when we reach the constant descriptor, it should have that enum accessible to condition logic off of

ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
```diff
@@ -272,6 +272,14 @@ exports[`Workflow > graph > should define nested sets of nodes without compilati
 "
 `;
 
+exports[`Workflow > graph > should generate correct graph for a loop with ports 1`] = `
```
dvargasfuertes: whoops - wrong pr haha

ee/codegen/src/generators/nodes/bases/base.ts
NgoHarrison: Updated

ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
NgoHarrison: Lol yeah, been a long day 🫠 I cherrypicked that other one by accident but fixed the branches

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -391,6 +391,17 @@ export abstract class BaseNode<
                     nodeContext: this.nodeContext,
                     workflowContext: this.workflowContext,
                     iterableConfig: { endWithComma: false },
+                    attributeConfig: {
+                      lhs: python.reference({
+                        name: "WorkflowErrorCode",
```
dvargasfuertes: [blocking] this still has the same issue where all string adornment attributes will be WorkflowErrorCode. I'm thinking we add config here: https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/codegen/src/generators/nodes/constants.ts#L16 that we eventually prop through here

dvargas92495 (approved):
pending final blocking idea

ee/codegen/src/generators/nodes/bases/base.ts
NgoHarrison: Ahh gotcha! I added an extra test now to show this (should have done this earlier to test out 😓)

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -383,28 +383,36 @@ export abstract class BaseNode<
                 attribute: ["wrap"],
                 modulePath: adornment.base.module,
               }),
-              arguments_: filteredAttributes.map((attr) =>
-                python.methodArgument({
+              arguments_: filteredAttributes.map((attr) => {
+                const nodeConfig = NODE_DEFAULT_ATTRIBUTES[adornment.base.name];
+                const attrConfig = nodeConfig?.[attr.name];
```
dvargasfuertes, line 388: [oos] yea perfect. only thing left I could see is that we pass this `attrConfig` directly, so that `WorkflowValueDescriptor` eventually constructs `python.reference` internally. I like the idea of only raw data types going down, AstNodes building up.

but I'm looking to cut another release, so let's merge this in first

dvargas92495 (approved):
💥

PR #1488, opened 2025-04-24
Fix import paths based on the module path arg

dvargas92495:
This PR fixes all of the module path resolution issues we were having during codegen. After the release is cut, we'll need to update codegen service as well.

### DIFF
--- a/ee/codegen/src/__test__/project.test.ts
+++ b/ee/codegen/src/__test__/project.test.ts
@@ -22,15 +22,8 @@ import { WorkflowProjectGenerator } from "src/project";
 describe("WorkflowProjectGenerator", () => {
   let tempDir: string;
 
-  function expectProjectFileToMatchSnapshot(
-    project: WorkflowProjectGenerator,
-    filePath: string[]
-  ) {
-    const completeFilePath = join(
-      tempDir,
-      ...project.getModulePath(),
-      ...filePath
-    );
+  function expectProjectFileToMatchSnapshot(filePath: string[]) {
+    const completeFilePath = join(tempDir, ...filePath);
     expect(
       fs.existsSync(completeFilePath),
       `File does not exist: ${completeFilePath}. Files generated:\n${Object.keys(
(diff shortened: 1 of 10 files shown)


ee/codegen/src/__test__/project.test.ts
```diff
@@ -2975,4 +2986,89 @@ baz = foo + bar
...
+          ],
+        },
+        input_variables: [],
+        output_variables: [],
+      };
+
+      const project = new WorkflowProjectGenerator({
+        absolutePathToOutputDirectory: tempDir,
+        workflowVersionExecConfigData: displayData,
+        moduleName: "my.module",
```
dvargasfuertes, line 3043: main thing we were testing

ee/codegen/src/__test__/project.test.ts
```diff
@@ -22,36 +22,30 @@ import { WorkflowProjectGenerator } from "src/project";
 describe("WorkflowProjectGenerator", () => {
   let tempDir: string;
 
-  function expectProjectFileToMatchSnapshot(
-    project: WorkflowProjectGenerator,
```
dvargasfuertes, line 26: It's dangerous for implementation code to be used in test fixtures bc it allows the same bug in implementation to also be true in tests

ee/codegen/src/generators/workflow-sandbox-file.ts
```diff
@@ -26,7 +26,7 @@ export class WorkflowSandboxFile extends BasePersistedFile {
   }
 
   protected getModulePath(): string[] {
-    return [this.workflowContext.moduleName, "sandbox"];
+    return [...this.workflowContext.modulePath.slice(0, -1), "sandbox"];
```
dvargasfuertes, line 29: I think there's currently a conflation of "workflow context" and "project context" that _might_ be worth separating. but also tricky to do bc we pass projects into subworkflows. ideally, subworkflows are just workflow file/directory, and projects are top level. I plan to sketch out a quick proposal to post to the channel some time tn

ee/codegen/src/generators/workflow.ts
```diff
@@ -34,8 +34,6 @@ import { DictEntry, isDefined } from "src/utils/typing";
 
 export declare namespace Workflow {
   interface Args {
-    /* The name of the module that the workflow class belongs to */
-    moduleName: string;
```
dvargasfuertes, line 38: big part of this PR was getting rid of as many `moduleName` references as possible. We should always consider modules as paths.

PR #1489, opened 2025-04-24
Generate correct graph where there is a cycle between two nodes where the ports are pointing to each other

NgoHarrison:
For this https://app.vellum.ai/workflow-sandboxes/e46c2d7d-d974-4ded-9d7e-5cd52767c2d5

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
@@ -273,14 +273,10 @@ exports[`Workflow > graph > should define nested sets of nodes without compilati
 `;
 
 exports[`Workflow > graph > should generate correct graph for a loop with ports 1`] = `
-"(
-    {
-        FirstNode.Ports.if_port >> FinalNode,
-        FirstNode.Ports.else_port >> SecondNode,
-    }
-    >> SecondNode
-    >> FirstNode
-)
+"{
+    FirstNode.Ports.if_port >> FinalNode,
+    FirstNode.Ports.else_port >> SecondNode >> FirstNode,
+}
 "
 `;
 
--- a/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
@@ -272,6 +272,18 @@ exports[`Workflow > graph > should define nested sets of nodes without compilati
 "
 `;
 
+exports[`Workflow > graph > should generate correct graph for a loop with ports 1`] = `
+"(
+    {
+        FirstNode.Ports.if_port >> FinalNode,
+        FirstNode.Ports.else_port >> SecondNode,
+    }
+    >> SecondNode
+    >> FirstNode
+)
+"
+`;
+
 exports[`Workflow > graph > should handle a simple edge of generic nodes 1`] = `
 "StartNode >> EndNode
 "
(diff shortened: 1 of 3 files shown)


ee/codegen/src/__test__/graph-attribute.test.ts
```diff
@@ -47,6 +48,8 @@ describe("Workflow", () => {
           return;
         }
 
+        //console.log(node.label, node.id);
+        //console.log(node.data.label, node.id);
```
vincent0426:
```suggestion
```

ee/codegen/src/__test__/__snapshots__/graph-attribute.test.ts.snap
```diff
@@ -272,6 +272,18 @@ exports[`Workflow > graph > should define nested sets of nodes without compilati
 "
 `;
 
+exports[`Workflow > graph > should generate correct graph for a loop with ports 1`] = `
+"(
+    {
+        FirstNode.Ports.if_port >> FinalNode,
+        FirstNode.Ports.else_port >> SecondNode,
+    }
+    >> SecondNode
```
dvargasfuertes: [blocking] this still isn't correct - implies two invalid edges:
- FinalNode >> SecondNode
- SecondNode >> SecondNode

dvargasfuertes: I think this is what we want:
```
{
    FirstNode.Ports.if_port >> FinalNode,
    FirstNode.Ports.else_port >> SecondNode >> FirstNode,
}
```

NgoHarrison: Ooh got it. Updated

NgoHarrison: Ah I see. Updated

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -336,6 +336,36 @@ export class GraphAttribute extends AstNode {
       return;
     }
 
+    // Check if this is a cycle edge (target node already exists in the graph)
+    const isCycle = this.isNodeInBranch(targetNode, mutableAst);
+    const isSourceInGraph = this.isNodeInBranch(sourceNode, mutableAst);
+
+    if (
+      sourceNode &&
```
dvargasfuertes: [nit] this check seems redundant with `isSourceInGraph`

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -336,6 +336,36 @@ export class GraphAttribute extends AstNode {
...
+      isCycle &&
+      isSourceInGraph &&
+      mutableAst.type === "set" &&
```
dvargasfuertes: [nit] We have a condition below for `mutableAst.type === "set"`. in fact, addEdgeToGraph is actually the union of addEdgeToEmpty, addEdgeToSource, etc for each mutableAst type. we've added those helpers for 2 of the 5 already, plan is to have for each. so would prefer if we nest this bit of logic within it.

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -336,6 +336,36 @@ export class GraphAttribute extends AstNode {
...
+      this.isDirectCycle(sourceNode, targetNode, this.workflowContext)
```
dvargasfuertes: [nit] technically speaking, we already know that we have an edge from `sourceNode >> targetNode` at this point. so we only need to check if an edge from `targetNode >> sourceNode` exists here

dvargas92495 (approved):

PR #1490, opened 2025-04-25
[APO-421] Support value access for string in codeexec

vincent0426:


src/vellum/workflows/types/code_execution_node_wrappers.py
```diff
@@ -0,0 +1,64 @@
...
+            wrapped[key] = clean_for_dict_wrapper(wrapped[key])
+
+        return wrapped
+
+    elif isinstance(obj, list):
+        return ListWrapper(map(lambda item: clean_for_dict_wrapper(item), obj))
+    elif isinstance(obj, str):
+        return StringValueWrapper(obj)
+
+    return obj
```
vincent0426, line 64: Move to a new file as there is circular dependency

src/vellum/workflows/nodes/utils.py
```diff
@@ -176,6 +178,8 @@ def parse_type_from_str(result_as_str: str, output_type: Any) -> Any:
 
 
 def _get_type_name(obj: Any) -> str:
+    if inspect.isclass(obj) and issubclass(obj, StringValueWrapper):
+        return "str"
```
vincent0426, line 182: to avoid returning `StringValueWrapper`

dvargasfuertes: [oos] Should we do the same for `DictWrapper` and `ListWrapper`?

dvargas92495 (approved):

src/vellum/workflows/nodes/utils.py
vincent0426: yes! will do in follow PR

PR #1497, opened 2025-04-25
Ensure new URL is in the print log at the end

dvargas92495:
Make sure our print statement at the end uses the new VPC URL instead of the default SaaS one

### DIFF
--- a/ee/vellum_cli/push.py
+++ b/ee/vellum_cli/push.py
@@ -235,9 +235,11 @@ def push_command(
 """
         )  # type: ignore[attr-defined]
     else:
+        default_api_url = client._client_wrapper._environment.default
+        base_url = default_api_url.split("/v1")[0].replace("//api.", "//app.")
         logger.info(
             f"""Successfully pushed {workflow_config.module} to Vellum!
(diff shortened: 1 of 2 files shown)


ee/vellum_cli/push.py
```diff
@@ -235,9 +235,11 @@ def push_command(
 """
         )  # type: ignore[attr-defined]
     else:
+        default_api_url = client._client_wrapper._environment.default
+        base_url = default_api_url.split("/v1")[0].replace("//api.", "//app.")
```
dvargasfuertes, line 239: @Shaarson do you know if this is a safe assumption we could make long term about the webserver url, or should we just return the actual app URL from the API request?

Shaarson: Is the app url always guaranteed to be part of the request? If that's the case I would lean that direction. We can't always guarantee that the URLs are structured like this and this will not work with the way Fastweb has structured their URLs.

Shaarson:
✅

ee/vellum_cli/push.py
dvargasfuertes: ok I think we need to update the API then to return the webserver hostname

Shaarson: Will make a note of that

PR #1499, opened 2025-04-25
Thread parent display context through to node input serialization

dvargas92495:
Discovered this issue trying to push a workflow to a VPC environment. The client that we reinitialize during serialization is not the same as the one that we use to invoke the push API.

This PR focuses on just one use case, workspace secret serialization in legacy node inputs. The next PR will apply the same sort of threading to our other vellum client use cases

### DIFF
--- a/ee/vellum_cli/tests/test_push.py
+++ b/ee/vellum_cli/tests/test_push.py
@@ -30,19 +30,25 @@ def _extract_tar_gz(tar_gz_bytes: bytes) -> dict[str, str]:
     return files
 
 
-def _ensure_workflow_py(temp_dir: str, module: str) -> str:
-    base_dir = os.path.join(temp_dir, *module.split("."))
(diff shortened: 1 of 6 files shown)


ee/vellum_cli/tests/test_push.py
```diff
@@ -519,7 +525,39 @@ def test_push__strict_option_returns_diffs(mock_module, vellum_client):
...
+            "workflow.py": """\
+from vellum.workflows import BaseWorkflow
+from .nodes.start_node import StartNode
+
+class ExampleWorkflow(BaseWorkflow):
+    graph = StartNode
+""",
+        },
+    ],
+    ids=["base_case", "with_secret_reference"],
```
dvargasfuertes, line 558: this triggered the appropriate test failure bc we have an assertion in this test that ensures that we only initialize a client once

NgoHarrison:
✅

PR #1507, opened 2025-04-25
Serialize constant error inputs

dvargas92495:
This PR started as a resolution to a serialization problem I saw during the customer demo. The error node had a constant string value, but after running `vellum push`, it was missing. This revealed an issue with inferred node input serialization of constants in general.

This PR introduces `__serializable_inputs__` on BaseNodeDisplay. The goal is not just to solve this issue, but to allow us to continue generalizing all of our node input displays in the future

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/get_node_display_class.py
+++ b/ee/vellum_ee/workflows/display/nodes/get_node_display_class.py
@@ -2,7 +2,6 @@ import types
 from uuid import UUID
 from typing import TYPE_CHECKING, Any, Dict, Generic, Type, TypeVar
 
-from vellum.workflows.descriptors.base import BaseDescriptor
 from vellum.workflows.types.generics import NodeType
 from vellum.workflows.utils.uuids import uuid4_from_hash
 from vellum_ee.workflows.display.utils.registry import get_from_node_display_registry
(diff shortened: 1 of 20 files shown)


ee/vellum_ee/workflows/display/nodes/get_node_display_class.py
```diff
@@ -30,14 +29,14 @@ def _get_node_input_ids_by_ref(path: str, inst: Any):
...
-            return {path: uuid4_from_hash(f"{node_class.__id__}|{path}")}
-
-        return {}
+        return {path: uuid4_from_hash(f"{node_class.__id__}|{path}")}
 
     def exec_body(ns: Dict):
         node_input_ids_by_name: Dict[str, UUID] = {}
         for ref in node_class:
+            if ref not in base_node_display_class.__serializable_inputs__:
+                continue
```
dvargasfuertes, line 38: the key line to support the original bug

ee/vellum_ee/workflows/display/nodes/vellum/map_node.py
```diff
@@ -13,6 +13,8 @@
 
 
 class BaseMapNodeDisplay(BaseAdornmentNodeDisplay[_MapNodeType], Generic[_MapNodeType]):
+    __serializable_inputs__ = {MapNode.items}  # type: ignore[misc]
```
dvargasfuertes, line 16: mypy doesn't like class attribute references of generic types

ee/vellum_ee/workflows/display/nodes/vellum/tests/test_code_execution_node.py
```diff
@@ -29,7 +29,7 @@ class CodeExecutionNodeDisplay(BaseCodeExecutionNodeDisplay[Node]):  # type: ign
 @pytest.mark.parametrize(
     ["GetDisplayClass", "expected_input_id"],
     [
-        (_no_display_class, "e3cdb222-324e-4ad1-abb2-bdd7881b3a0e"),
+        (_no_display_class, "a5dbe403-0b00-4df6-b8f7-ed5f7794b003"),
```
dvargasfuertes, line 32: the default id generated is a bit different now since we are hashing `code_inputs.foo` instead of `foo`

ee/vellum_ee/workflows/display/nodes/vellum/tests/test_prompt_deployment_node.py
```diff
@@ -13,14 +14,14 @@ def _no_display_class(Node: Type[PromptDeploymentNode]):  # type: ignore
 
 
 def _display_class_with_node_input_ids_by_name(Node: Type[PromptDeploymentNode]):
-    class PromptDeploymentNodeDisplay(PromptDeploymentNode[Node]):  # type: ignore[valid-type]
+    class PromptDeploymentNodeDisplay(BasePromptDeploymentNodeDisplay[Node]):  # type: ignore[valid-type]
```
dvargasfuertes, line 17: busted tests here and for subworkflow display

PR #1512, opened 2025-04-25
[APO-465] Codegen ampersand instead of and on port conditions

NgoHarrison:


vincent0426 (approved):

ee/codegen/src/generators/extensions/protected-python-file.ts
```diff
@@ -308,16 +308,23 @@ export class PythonFile extends AstNode {
   }
 }
 
+export declare namespace PipeExpression {
```
dvargasfuertes, line 311: [oos] ah let's put these into their own files since they are a distinct concept from `protected-python-file.ts`

dvargas92495 (commented):
thanks for the quick fix! ✅

PR #1513, opened 2025-04-26
Support anotation node references of inherited nodes

dvargas92495:
A followup to https://github.com/vellum-ai/vellum-python-sdks/pull/1511, but now for inherited references


vincent0426 (commented):
✅

PR #1517, opened 2025-04-27
Catch unexpected errors during asset generation

dvargas92495:
Screenshot 2025-04-27 at 8 02 39 AM

### DIFF
--- a/ee/codegen/src/project.ts
+++ b/ee/codegen/src/project.ts
@@ -17,8 +17,10 @@ import { InputVariableContext } from "./context/input-variable-context";
 import { WorkflowOutputContext } from "./context/workflow-output-context";
 import { ErrorLogFile, InitFile, Inputs, Workflow } from "./generators";
 import {
+  BaseCodegenError,
   NodeDefinitionGenerationError,
   ProjectSerializationError,
+  WorkflowGenerationError,
 } from "./generators/errors";
 import { BaseNode } from "./generators/nodes/bases";
 import { GuardrailNode } from "./generators/nodes/guardrail-node";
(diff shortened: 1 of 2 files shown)


ee/codegen/src/project.ts
```diff
@@ -175,7 +178,17 @@ ${errors.slice(0, 3).map((err) => {
 
   public async generateCode(): Promise<void> {
     const assets = await this.generateAssets().catch((error) => {
-      this.workflowContext.addError(error);
+      if (error instanceof BaseCodegenError) {
+        this.workflowContext.addError(error);
+      } else {
+        console.error("Unexpected error generating assets", error);
```
dvargasfuertes, line 184: I looked into creating a test for this, but I couldn't figure out how to repro based on the sentry error. So am hoping to get more info from this console.error

ee/codegen/src/project.ts
```diff
@@ -476,60 +502,52 @@ ${errors.slice(0, 3).map((err) => {
     return orderedNodes;
   }
 
-  private async generateNodes(
+  private generateNodes(
```
dvargasfuertes, line 505: Unrelated cleanup opportunity I noticed

ee/codegen/src/project.ts
```diff
@@ -8,24 +8,13 @@ import { AstNode } from "@fern-api/python-ast/core/AstNode";
...
-} from "./generators/errors";
-import { BaseNode } from "./generators/nodes/bases";
-import { GuardrailNode } from "./generators/nodes/guardrail-node";
-import { InlineSubworkflowNode } from "./generators/nodes/inline-subworkflow-node";
-import { SearchNode } from "./generators/nodes/search-node";
-import { TemplatingNode } from "./generators/nodes/templating-node";
-
+} from "src/constants";
+import { createNodeContext, WorkflowContext } from "src/context";
+import { InputVariableContext } from "src/context/input-variable-context";
```
dvargasfuertes, line 17: vitest treats relative and absolut paths as different modules, so we should always be using absolute

PR #1528, opened 2025-04-27
Remove the inputs AST arg from the Workflo AST

dvargas92495:
Before adding support for Generics and Display for Workflow State, I wanted to first clean up how we thread through the workflow input AST. This allows us to not complicate the interface of a Workflow AST further when we introduce support for `State`

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
@@ -2,9 +2,12 @@
 
 exports[`Workflow > write > should generate correct code when there are input variables 1`] = `
 "from vellum.workflows import BaseWorkflow
+from vellum.workflows.state import BaseState
+
+from .inputs import Inputs
 
 
-class TestWorkflow(BaseWorkflow):
+class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
     pass
 "
 `;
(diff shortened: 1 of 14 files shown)


ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -2,9 +2,12 @@
 
 exports[`Workflow > write > should generate correct code when there are input variables 1`] = `
```
dvargasfuertes, line 3: This test seem to be doing the same thing as the no input variables one right after it, so made the snapshot more consistent with the labelled intent of the test

ee/codegen/src/generators/workflow.ts
```diff
@@ -36,57 +35,63 @@ export declare namespace Workflow {
   interface Args {
     /* The context for the workflow */
     workflowContext: WorkflowContext;
-    /* The inputs for the workflow */
-    inputs: Inputs;
```
dvargasfuertes, line 40: This removal was the goal of the PR

ee/codegen/src/generators/workflow.ts
```diff
@@ -236,68 +241,65 @@ export class Workflow {
...
     );
 
-    workflowDisplayClass.add(
-      python.field({
-        name: "inputs_display",
-        initializer: python.TypeInstantiation.dict(
-          Array.from(this.workflowContext.inputVariableContextsById)
-            .map(([_, inputVariableContext]) => {
-              const inputsClass = this.inputs.inputsClass;
-              if (!inputsClass) {
```
dvargasfuertes, line 246: replaced the check outside of the loop

vincent0426 (commented):
✅

PR #1529, opened 2025-04-27
Add support for StateValueDisplay generation

dvargas92495:
Adds support for state value displays that preserves ids and non-python-safe state names


vincent0426 (commented):
✅

PR #1530, opened 2025-04-27
Custom state class is imported into workflow.py

dvargas92495:
Adds support for our custom state to actually be configured with the workflow


vincent0426 (commented):
✅

PR #1531, opened 2025-04-28
Initial stab at codecop guidelines

dvargas92495:
This is to help out the codecop automation that has been appearing on our PR's lately. This markdown is not exhaustive, but is a decent starting point


vincent0426 (commented):
👮‍♂️

PR #1532, opened 2025-04-28
Support serialization of legacy nodes when they have custom ports

dvargas92495:
the source handle id helper was failing whenever we tried to push a legacy nodes with Ports attached. This PR addresses


vincent0426 (commented):
✅

PR #1535, opened 2025-04-28
Remove tests and mypy from pre-commit-config

m-abboud:
Talked about removing this earlier and I'm waiting around for a vembda test to finish. This removes notoriously flakey mypy and the slow/flakey tests from hooks.


dvargas92495 (approved):
now we can add thousands of more tests!

vincent0426 (commented):
✅

PR #1536, opened 2025-04-28
Code gen references to State

dvargas92495:
We need this to support Prompt node flows in our new function calling demos that reference State in chat history


marlenc3 (commented):
✅

NgoHarrison (commented):
✅

PR #1542, opened 2025-04-29
Support iter for list wrapper

vincent0426:
This will solve the first error encountered in APO-363

### DIFF
--- a/src/vellum/workflows/types/code_execution_node_wrappers.py
+++ b/src/vellum/workflows/types/code_execution_node_wrappers.py
@@ -23,6 +23,10 @@ class ListWrapper(list):
             return self
         raise AttributeError(f"'list' object has no attribute '{attr}'")
 
+    def __iter__(self):
+        for i in range(len(self)):
+            yield self[i]
+
 
 class DictWrapper(dict):
     """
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/types/code_execution_node_wrappers.py
```diff
@@ -23,6 +23,10 @@ def __getattr__(self, attr):
             return self
         raise AttributeError(f"'list' object has no attribute '{attr}'")
 
+    def __iter__(self):
+        for i in range(len(self)):
+            yield self[i]
```
dvargasfuertes, line 28: [q] woah. Why don't we get this for free from inheriting `list` ??

dvargasfuertes: [oos] do we need the same for `DictWrapper`?

dvargas92495 (approved):

src/vellum/workflows/types/code_execution_node_wrappers.py
vincent0426: > [q] woah. Why don't we get this for free from inheriting `list` ??

I think it's bc we override `__getitem__` and the default `__iter__` will bypass our `__getitem__` which then can't access value

dvargasfuertes: Wild

PR #1543, opened 2025-04-29
Migrate custom prompt node into SDK examples

dvargas92495:
Discussed with @awlevin earlier today, and we agreed on the SDK's `examples/workflows` repo being the new home for workflows 

Migrating the custom-prompt-node [example](https://github.com/vellum-ai/vellum-example-apps/tree/main/examples/custom-prompt-node) we spun up in the examples repo over to here. We can successfully keep the deps needed for these examples from polluting the sdk's dependencies since we have a dedicated pyproject.toml and poetry environment nested in the repo for these examples.

### DIFF
--- a/examples/workflows/custom_prompt_node/inputs.py
+++ b/examples/workflows/custom_prompt_node/inputs.py
@@ -1,0 +1,5 @@
+from vellum.workflows.inputs import BaseInputs
+
+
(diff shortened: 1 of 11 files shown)


noanflaherty (commented):
✅

PR #1544, opened 2025-04-29
push the custom prompt node workflow to vellum

dvargas92495:
https://www.loom.com/share/0a8d262fe503463294451191a5998b2f


awlevin:
I actually like the config idea, i could see it being an extension of the Environments project, maybe similar to `vercel env pull/push`. Once we can select models dynamically, why not be able to set them with env vars? 🤷

Also lol @ the end of that loom. Go Mets!

dvargas92495:
Hm yea interesting. I think bc the icp of this workflow will need a custom docker image regardless, going to support that route for now, but will give more thought on the alternative idea 🤔

PR #1547, opened 2025-04-29
Copy custom base node example into SDK

dvargas92495:
Cleaned up our custom base node example from the prototype repo that we will share with a prospect, and ported into the sdk repo. This workflow features a custom base node that requires custom docker image (emulated by the `utils` directory) and have three nodes that extend from it.


noanflaherty (commented):
✅

NgoHarrison (commented):
✅

PR #1548, opened 2025-04-29
support nullable state_variables field

dvargas92495:
We have workflows on the vellum side that actually have this field nullable. This change helps protect us in that situation


NgoHarrison (commented):
✅

PR #1549, opened 2025-04-29
Push Custom Base Node

dvargas92495:
This PR shows that we were able to push the custom base node example into Vellum, and successfully initializes with access to the custom functionality found in the docker image. Next step is to serialize prompt_inputs as expression inputs so that we could access state

Screenshot 2025-04-29 at 4 12 04 PM


NgoHarrison (commented):
✅

noanflaherty (commented):
💥

PR #1554, opened 2025-04-30
Improve the error message returned for code execution node errors

dvargas92495:
Previously, we were returning just the error message with no context (so a `KeyError: 0` was literally just `'0'`). Now, we forward a traceback to the user to help them debug their code nodes


joao-krauchenco-mottu (approved):

NgoHarrison (commented):
✅

vincent0426 (commented):
✅

PR #1559, opened 2025-04-30
[APO-340] Serialize tool calling node

vincent0426:


src/vellum/workflows/nodes/displayable/bases/utils.py
```diff
@@ -95,6 +96,8 @@ def primitive_to_vellum_value(value: Any) -> VellumValue:
         return value  # type: ignore
 
     try:
+        if callable(value):
+            value = compile_function_definition(value)
```
vincent0426: serialize to json if it's a callable type but needs to compile into function definition first

ee/vellum_ee/workflows/display/nodes/vellum/tool_calling_node.py
```diff
@@ -0,0 +1,14 @@
...
+from vellum_ee.workflows.display.nodes.base_node_display import BaseNodeDisplay
+from vellum_ee.workflows.display.types import WorkflowDisplayContext
+
+_ToolCallingNodeType = TypeVar("_ToolCallingNodeType", bound=ToolCallingNode)
+
+
+class BaseToolCallingNodeDisplay(BaseNodeDisplay[_ToolCallingNodeType], Generic[_ToolCallingNodeType]):
+    def serialize(self, display_context: WorkflowDisplayContext, **kwargs: Any) -> JsonObject:
+        serialized_node = super().serialize(display_context, **kwargs)
+        return serialized_node
```
dvargasfuertes: [nit] so here's the cool thing about where we're trying to go - since all this method does is called the base node's display serialize, we actually don't need this file at all!

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_serialization.py
```diff
@@ -0,0 +1,285 @@
+from deepdiff import DeepDiff
+
+from vellum_ee.workflows.display.workflows.get_vellum_workflow_display_class import get_workflow_display
+
+from tests.workflows.basic_tool_calling_node.workflow import BasicToolCallingNodeWorkflow
+
+
+def test_serialize_workflow():
```
dvargasfuertes, line 8: I would expect this test to pass with no changes even if we remove the first file

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_serialization.py
```diff
@@ -0,0 +1,285 @@
...
+    assert not DeepDiff(
+        [
+            {"id": "8e7c0147-930d-4b7f-b6b1-6d79641cd3eb", "key": "text", "type": "STRING"},
+            {"id": "01a07e6d-7269-4f45-8b44-ef0227a2e88d", "key": "chat_history", "type": "CHAT_HISTORY"},
+        ],
+        output_variables,
+        ignore_order=True,
+    )
+
+    # AND its raw data should be what we expect
```
dvargasfuertes, line 38: [nit] One thing I want to do with all of the tests in this directory is remove the redundant assertions that don't relate to the node we are testing. so that means getting rid of the next three lines, the entrypoint/terminal node assertions, etc

src/vellum/workflows/nodes/displayable/bases/utils.py
dvargasfuertes: [nit] what if we added this to `DefaultStateEncoder` instead?

dvargas92495 (approved):

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_serialization.py
vincent0426: @dvargas92495 Remove edges and definition too?

dvargasfuertes: yea I would

PR #1560, opened 2025-04-30
Use thread pool executor for handling concurrent threads on map node execution

NgoHarrison:


src/vellum/workflows/nodes/core/map_node/tests/test_node.py
```diff
@@ -172,3 +174,81 @@ class FruitMapNode(MapNode):
...
+        class Outputs(BaseOutputs):
+            output: str
+            thread_id: int
+
+        def run(self) -> Outputs:
+            current_thread_id = threading.get_ident()
+            thread_ids[self.item] = current_thread_id
+
+            # Simulate work
+            time.sleep(0.5)
```
dvargasfuertes: Is the bug reproducible (ie this test fails) if we do `0.01` instead of `0.5`? would love to keep each test on the order of ms

src/vellum/workflows/nodes/core/map_node/tests/test_node.py
```diff
@@ -172,3 +174,81 @@ class FruitMapNode(MapNode):
...
+            thread_id: int
+
+        def run(self) -> Outputs:
+            current_thread_id = threading.get_ident()
+
+            return self.Outputs(result=float(self.item), thread_id=current_thread_id)
+
+    # AND a workflow that connects these nodes
+    class TestWorkflow(BaseWorkflow[MapNode.SubworkflowInputs, BaseState]):
+        graph = BaseNode1 >> BaseNode2 >> FinalNode
```
dvargasfuertes: Is the bug reproducible without BaseNode2 and FinalNode? If so, let's get rid of them

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -75,95 +75,85 @@ def run(self) -> Iterator[BaseOutput]:
...
-                                extra={"output_reference_type": type(output_reference)},
+        fulfilled_iterations: List[bool] = [False] * len(self.items)
+
+        max_workers = self.max_concurrency if self.max_concurrency is not None else len(self.items)
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=max_workers) as executor:
+            futures = []
+            for index, item in enumerate(self.items):
+                current_execution_context = get_execution_context()
+                future = executor.submit(self._context_run_subworkflow, item, index, current_execution_context)
```
dvargasfuertes: [q] is there a way to submit via kwargs? if so would love to keep as kwargs

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -75,95 +75,85 @@ def run(self) -> Iterator[BaseOutput]:
...
-                                "Invalid key to map node's subworkflow event outputs",
-                                extra={"output_reference_type": type(output_reference)},
+        fulfilled_iterations: List[bool] = [False] * len(self.items)
+
+        max_workers = self.max_concurrency if self.max_concurrency is not None else len(self.items)
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=max_workers) as executor:
+            futures = []
+            for index, item in enumerate(self.items):
+                current_execution_context = get_execution_context()
```
dvargasfuertes: [nit] this should actually be the same for all iterations and could be lifted out of the loop

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -75,95 +75,85 @@ def run(self) -> Iterator[BaseOutput]:
...
-                    raise NodeException(
-                        f"Subworkflow failed on iteration {index} with error: {subworkflow_event.error.message}",
-                        code=subworkflow_event.error.code,
-                    )
-        except Empty:
-            pass
+                    if all_futures_done:
+                        if not all(fulfilled_iterations):
+                            if self._event_queue.empty():
+                                logger.warning("All threads completed but not all iterations fulfilled")
```
dvargasfuertes, line 152: [oos] we should likely add Rejected events for each iteration that is without one

dvargas92495 (approved):
sick!

PR #1576, opened 2025-05-01
Support serialization of prompt node inputs as attributes

dvargas92495:
This will allow the UI to have state references and nested value accessors


ee/vellum_ee/workflows/display/nodes/vellum/inline_prompt_node.py
```diff
@@ -19,6 +19,14 @@
 
 class BaseInlinePromptNodeDisplay(BaseNodeDisplay[_InlinePromptNodeType], Generic[_InlinePromptNodeType]):
     __serializable_inputs__ = {InlinePromptNode.prompt_inputs}
+    __unserializable_attributes__ = {
```
dvargasfuertes, line 22: The first four are supported by the UI, and therefore, we'll gradually reintroduce over the next week I suspect

PR #1578, opened 2025-05-01
support final output node streaming from prompt nodes

dvargas92495:
Original Workflows supported streaming through Final Output Nodes, mostly bc of special casing for those nodes. This PR explores how we could have first class support for streaming through terminal nodes.

I wanted to get this launched soon for the demo, but would be too high risk so am going to roll this out more slowly


dvargas92495:
closing due to complexity - we found a workaround here: https://github.com/vellum-ai/vellum-python-sdks/pull/1773

PR #1582, opened 2025-05-01
Update node context to look for try adornment and fallback to erroroutputid

NgoHarrison:


ee/codegen/src/context/node-context/base.ts
```diff
@@ -217,4 +218,20 @@ export abstract class BaseNodeContext<T extends WorkflowDataNode> {
...
+    });
+
+    const tryNode = tryNodes?.at(-1);
+
+    let fallbackErrorOutputId: string | undefined = undefined;
+    if (hasErrorOutput(this.nodeData)) {
+      fallbackErrorOutputId = this.nodeData.data.errorOutputId;
+    }
+
+    return tryNode ? tryNode.id : fallbackErrorOutputId;
```
dvargasfuertes, line 235: [nit] Looks like we could early return `tryNode` at around L229 instead of doing the second half if it's truthy

dvargas92495 (approved):

dvargas92495:
Merging through to include in 0.14.51

PR #1584, opened 2025-05-01
Add experimental path

vincent0426:


NgoHarrison (approved):

ee/codegen/src/context/node-context/base.ts
```diff
@@ -59,6 +59,12 @@ export abstract class BaseNodeContext<T extends WorkflowDataNode> {
           .startsWith(VELLUM_WORKFLOW_NODES_MODULE_PATH.join("."))
       ) {
         this.baseNodeClassModulePath = args.nodeData.base.module;
+      } else if (
+        args.nodeData.base &&
+        args.nodeData.base.name === "ToolCallingNode"
```
dvargasfuertes, line 64: [blocking] we actually don't want any special casing of node in the generic code generator going forward. The check in lines 57-59 are actually to protect against base node. I'll do a quick rev here

PR #1586, opened 2025-05-02
Exclude core, displayable, experimental in codemerge

vincent0426:


ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -102,7 +102,10 @@ export class GenericNode extends BaseSingleFileNode<
 
     const relativePath = `nodes/${fileName}`;
 
-    this.workflowContext.addPythonCodeMergeableNodeFile(relativePath);
+    // Dont add tool calling nodes to mergeable node files
+    if (this.nodeData.base.name !== "ToolCallingNode") {
+      this.workflowContext.addPythonCodeMergeableNodeFile(relativePath);
+    }
```
dvargasfuertes: [blocking] Similar to my feedback in [this PR](https://github.com/vellum-ai/vellum-python-sdks/pull/1590), we should avoid special casing by node names going forward. We should try to consider only base node's module path's, and only exclude from code merge if it's the `vellum.workflows.nodes.[displayable|core|experimental]` module's

dvargas92495 (approved):

PR #1587, opened 2025-05-02
[APO-341] Initial codegen for tool calling node

vincent0426:


ee/codegen/src/types/vellum.ts
```diff
@@ -676,6 +676,20 @@ export interface GenericNode extends BaseWorkflowNode {
   outputs: NodeOutput[];
 }
 
+export interface ToolCallingNode extends BaseWorkflowNode {
+  id: string;
+  label: string;
+  type: "GENERIC";
```
vincent0426, line 682: will FE pass back TOOL_CALLING type in the future?

vincent0426: Hmm or should we just treat tool calling node as generic node and remove all tool calling context or generators

dvargasfuertes: > Hmm or should we just treat tool calling node as generic node and remove all tool calling context or generators

This !!

To give some more context, we want to be able to run serialization and codegen for hundreds, potentially thousands, of node types, which won't scale with special casing

ee/codegen/src/__test__/nodes/__snapshots__/tool-calling-node.test.ts.snap
```diff
@@ -0,0 +1,125 @@
...
+"from vellum.workflows.nodes.experimental.tool_calling_node import (
+    ToolCallingNode as BaseToolCallingNode,
+)
+
+
+class ToolCallingNode(BaseToolCallingNode):
+    ml_model = "gpt-4o-mini"
+    blocks = [
+        {
+            "state": None,
```
dvargasfuertes, line 42: [oos] The next iteration we'll want here is to generate the prompt block dataclasses. This may require revisiting the last part of that doc we discussed last week

ee/codegen/src/__test__/nodes/__snapshots__/tool-calling-node.test.ts.snap
```diff
@@ -0,0 +1,125 @@
...
+            "chat_role": "USER",
+            "block_type": "CHAT_MESSAGE",
+            "chat_source": None,
+            "cache_config": None,
+            "chat_message_unterminated": None,
+        },
+    ]
+    functions = [
+        {
+            "name": "get_current_weather",
```
dvargasfuertes, line 90: [oos] this is the part that will get interesting. Similar to the above, we _won't_ want to generate JSON. but callables have a few more interesting characteristics that include the function source code that we'll need to think about how to store as data on the vellum side

ee/codegen/src/__test__/nodes/__snapshots__/tool-calling-node.test.ts.snap
```diff
@@ -0,0 +1,125 @@
...
+            "cache_config": None,
+        },
+    ]
+    prompt_inputs = {
+        "question": "What's the weather like in San Francisco?",
+    }
+    max_tool_calls = 1
+
+    class NodeTrigger(BaseToolCallingNode.Trigger):
+        merge_behavior = "AWAIT_ALL"
```
dvargasfuertes: [blocking] we want AWAIT_ATTRIBUTES going forward, which should be the default now and could remove this

ee/codegen/src/__test__/nodes/__snapshots__/tool-calling-node.test.ts.snap
```diff
@@ -0,0 +1,125 @@
...
+
+    class Outputs(BaseToolCallingNode.Outputs):
+        text = ""
+        chat_history = []
```
dvargasfuertes, line 120: [blocking] we shouldn't need to define these, and most importantly, I don't think they start out with default values for them both like this

ee/codegen/src/context/node-context/create-node-context.ts
```diff
@@ -171,6 +172,15 @@ function buildNodeContext(
...
       const genericNodeData = nodeData;
+      if (
+        genericNodeData.base &&
+        genericNodeData.base.name === "ToolCallingNode"
+      ) {
+        return new ToolCallingNodeContext({
+          ...args,
+          nodeData: genericNodeData,
+        });
+      }
```
dvargasfuertes: [nit] can we achieve what we need without this?

ee/codegen/src/generators/nodes/tool-calling-node.ts
```diff
@@ -0,0 +1,94 @@
...
+
+import { ToolCallingNodeContext } from "src/context/node-context/tool-calling-node";
+import { NodeOutputs } from "src/generators/node-outputs";
+import { NodeTrigger } from "src/generators/node-trigger";
+import { BaseSingleFileNode } from "src/generators/nodes/bases/single-file-base";
+import { WorkflowValueDescriptor } from "src/generators/workflow-value-descriptor";
+import { ToolCallingNode as ToolCallingNodeType } from "src/types/vellum";
+import { toPythonSafeSnakeCase } from "src/utils/casing";
+
+export class ToolCallingNode extends BaseSingleFileNode<
```
dvargasfuertes: [nit] similar to above, I'd like to relay on Base Node's default file implementation as much as possible

ee/codegen/src/__test__/nodes/__snapshots__/tool-calling-node.test.ts.snap
dvargasfuertes: However, I would _still_ expect the workflow to be runnable with the dict's instead. If they aren't we should tackle that first

dvargas92495 (approved):
looks great! I'm cutting 0.14.51 this morning so merging this through

dvargas92495:
Let's be sure to clean up the generated Trigger and Outputs class though from the blocking comments - idt it's as blocking as I originally thought, but we should try to clean up anyway

PR #1588, opened 2025-05-02
new codegen service action action is missing creds

dvargas92495:
We didn't need this before, so I'm confused why we need it [now](https://github.com/vellum-ai/vellum-python-sdks/actions/runs/14782445051). Hopefully should be the last of our codegen service problems


vincent0426 (commented):
✅

PR #1589, opened 2025-05-02
Revert "Add experimental path"

vincent0426:
Reverts vellum-ai/vellum-python-sdks#1584

@dvargas92495 should we keep the experimental path? I think we still need to write the path for tool calling node?


dvargas92495:
No need to revert: https://github.com/vellum-ai/vellum-python-sdks/pull/1590

PR #1594, opened 2025-05-02
Clean up default test output

vincent0426:


ee/codegen/src/__test__/helpers/node-data-factories.ts
```diff
@@ -1892,20 +1892,7 @@ export function toolCallingNodeFactory({
         },
       },
     ],
-    outputs: nodeOutputs ?? [
```
dvargasfuertes, line 1895: [oos] What will the frontend do? will it include these outputs?

If so, we should add a test that has the array defined, and come up with a way of handling that generically

dvargas92495 (approved):

PR #1596, opened 2025-05-02
Support dict blocks

vincent0426:


src/vellum/prompts/blocks/compilation.py
```diff
@@ -252,3 +259,88 @@ def _sanitize_inputs(inputs: Sequence[PromptInput]) -> list[PromptInput]:
...
+            )
+        elif block_type == "FUNCTION_CALL":
+            converted_blocks.append(
+                FunctionCallPromptBlock(
+                    name=block.get("name"),
+                    arguments=block.get("arguments"),
+                    cache_config=block.get("cache_config"),
+                )
+            )
+
```
vincent0426: do we need to include plain text here since PromptBlock doesn't have it

src/vellum/prompts/blocks/compilation.py
```diff
@@ -252,3 +259,88 @@ def _sanitize_inputs(inputs: Sequence[PromptInput]) -> list[PromptInput]:
             sanitized_inputs.append(input_)
 
     return sanitized_inputs
+
+
+def compile_blocks_to_prompt_blocks(blocks: list[Union[PromptBlock, Dict[str, Any]]]) -> list[PromptBlock]:
```
dvargasfuertes: [blocking] ah actually, let's keep our internal methods only dealing with structured data, and deserialize in the node. I'll show how below

src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
```diff
@@ -134,6 +135,14 @@ def _get_prompt_event_stream(self) -> Iterator[AdHocExecutePromptEvent]:
             )
 
     def _process_prompt_event_stream(self) -> Generator[BaseOutput, None, Optional[List[PromptOutput]]]:
+        try:
+            self.blocks = compile_blocks_to_prompt_blocks(self.blocks)  # type: ignore
```
dvargasfuertes: I think this could be as simple as:

`self.blocks = [PromptBlock.model_validate(block) block in blocks]`

dvargasfuertes: Actually, `PromptBlock` is a union, so you may need to do something like

```python
exec_config = PromptExecConfig.model_validate({
    "blocks": self.blocks,
    ..., # other fields here too
})

self.blocks = exec_config.blocks # or just feed `exec_config` directly into the eventual api request to vellum
```

src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
```diff
@@ -136,7 +136,16 @@ def _get_prompt_event_stream(self) -> Iterator[AdHocExecutePromptEvent]:
 
     def _process_prompt_event_stream(self) -> Generator[BaseOutput, None, Optional[List[PromptOutput]]]:
         try:
-            self.blocks = compile_blocks_to_prompt_blocks(self.blocks)  # type: ignore
+            # Compile dict blocks into PromptBlocks
+            exec_config = PromptExecConfig.model_validate(
+                {
+                    "ml_model": "",
+                    "input_variables": [],
+                    "parameters": {},
```
dvargasfuertes, line 144: in the future, we could do the same for parameters

dvargas92495 (approved):

PR #1597, opened 2025-05-02
[APO-486] Update guardrail node to hardcode score type to number

NgoHarrison:


ee/codegen/src/__test__/nodes/multiple-nodes.test.ts
```diff
@@ -657,3 +660,144 @@ describe("Inline Subworkflow Try adornment referenced by Conditional Node", () =
...
+    vi.spyOn(
+      MetricDefinitionsClient.prototype,
+      "metricDefinitionHistoryItemRetrieve"
+    ).mockResolvedValue({
+      id: "mocked-metric-output-id",
+      label: "mocked-metric-output-label",
+      name: "mocked-metric-output-name",
+      description: "mocked-metric-output-description",
+      outputVariables: [
+        { id: "score-output-id", key: "score", type: "STRING" },
```
NgoHarrison, line 682: Notice that this type is string

dvargasfuertes: [nit] We should put some mention of that in the test label

ee/codegen/src/context/node-context/guardrail-node.ts
```diff
@@ -35,10 +35,13 @@ export class GuardrailNodeContext extends BaseNodeContext<GuardrailNodeType> {
     }
 
     return Object.fromEntries(
-      this.metricDefinitionsHistoryItem.outputVariables.map((variable) => [
-        variable.id,
-        variable.type,
-      ])
+      this.metricDefinitionsHistoryItem.outputVariables.map((variable) => {
+        // If the variable key is "score", hardcode the type to "NUMBER". We assume float on sdk side anyway
+        if (variable.key === "score") {
```
dvargasfuertes, line 40: [nit] we might want to do the same for `"normalized_score"` and `"log"`

dvargas92495 (approved):

ee/codegen/src/context/node-context/guardrail-node.ts
NgoHarrison: Added

ee/codegen/src/__test__/nodes/multiple-nodes.test.ts
NgoHarrison: Added

PR #1608, opened 2025-05-03
Support code generation of custom non-generic base nodes

dvargas92495:
This is to support the custom prompt node demo


ee/codegen/src/__test__/nodes/__snapshots__/inline-prompt-node.test.ts.snap
```diff
@@ -1164,6 +1164,31 @@ class PromptNode(InlinePromptNode):
 "
 `;
 
+exports[`InlinePromptNode > basic with custom, same module, base > getNodeFile 1`] = `
+"from vellum import JinjaPromptBlock, PromptParameters
+
+from .my_inline_prompt import MyInlinePrompt
+
+
+class PromptNode(MyInlinePrompt):
```
dvargasfuertes, line 1173: This is the main goal of the PR

PR #1620, opened 2025-05-03
Resolve gnarly looping bug for nodes with AWAIT ATTRIBUTES merge behavior

dvargas92495:
We have an issue with AWAIT_ANY nodes where loops in the middle of a graph never execute. Turns out this problem _also_ existed for AWAIT_ATTRIBUTES node.

It should've never existed for the latter - those nodes only execute based on their attribute definitions. This PR moves the `queue_node_execution` into the `BaseNode.Trigger` class so that we can discriminate that step based on merge behavior. This solves this gnarly edge case for `AWAIT_ATTRIBUTES` nodes while deferring our need to fix it for AWAIT_ANY node in the future

### DIFF
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -2,12 +2,13 @@ from dataclasses import field
 from functools import cached_property, reduce
 import inspect
 from types import MappingProxyType
-from uuid import UUID
+from uuid import UUID, uuid4
 from typing import Any, Dict, Generic, Iterator, Optional, Set, Tuple, Type, TypeVar, Union, cast, get_args
 
 from vellum.workflows.constants import undefined
 from vellum.workflows.descriptors.base import BaseDescriptor
 from vellum.workflows.descriptors.utils import is_unresolved, resolve_value
+from vellum.workflows.edges.edge import Edge
 from vellum.workflows.errors.types import WorkflowErrorCode
 from vellum.workflows.exceptions import NodeException
 from vellum.workflows.graph import Graph
@@ -324,6 +325,36 @@ class BaseNode(Generic[StateType], metaclass=BaseNodeMeta):
                 message="Invalid Trigger Node Specification",
                 code=WorkflowErrorCode.INVALID_INPUTS,
             )
+
+        @classmethod
+        def _queue_node_execution(
+            cls, state: StateType, dependencies: Set["Type[BaseNode]"], invoked_by: Optional[Edge] = None
+        ) -> UUID:
+            """
+            Queues a future execution of a node, returning the span id of the execution.
+
+            We may combine this into the should_initiate method, but we'll keep it separate for now to avoid
+            breaking changes until the 0.15.0 release.
+            """
+
+            execution_id = uuid4()
+            if not invoked_by:
+                return execution_id
+
+            if cls.merge_behavior not in {MergeBehavior.AWAIT_ANY, MergeBehavior.AWAIT_ALL}:
+                return execution_id
+
+            source_node = invoked_by.from_port.node_class
+            for queued_node_execution_id in state.meta.node_execution_cache._node_executions_queued[cls.node_class]:
+                if source_node not in state.meta.node_execution_cache._dependencies_invoked[queued_node_execution_id]:
+                    state.meta.node_execution_cache._invoke_dependency(
+                        queued_node_execution_id, cls.node_class, source_node, dependencies
+                    )
+                    return queued_node_execution_id
+
+            state.meta.node_execution_cache._node_executions_queued[cls.node_class].append(execution_id)
+            state.meta.node_execution_cache._invoke_dependency(execution_id, cls.node_class, source_node, dependencies)
+            return execution_id
 
     class Execution(metaclass=_BaseNodeExecutionMeta):
         node_class: Type["BaseNode"]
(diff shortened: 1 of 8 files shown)


src/vellum/workflows/nodes/bases/base.py
```diff
@@ -325,6 +326,36 @@ def should_initiate(
...
+            We may combine this into the should_initiate method, but we'll keep it separate for now to avoid
+            breaking changes until the 0.15.0 release.
+            """
+
+            execution_id = uuid4()
+            if not invoked_by:
+                return execution_id
+
+            if cls.merge_behavior not in {MergeBehavior.AWAIT_ANY, MergeBehavior.AWAIT_ALL}:
+                return execution_id
```
dvargasfuertes, line 345: This is the only new piece of logic here. every thing else is copy pasted. The doc string for the method is also new

src/vellum/workflows/runner/runner.py
```diff
@@ -431,7 +431,7 @@ def _run_node_if_ready(
                     return
 
             all_deps = self._dependencies[node_class]
-            node_span_id = state.meta.node_execution_cache.queue_node_execution(node_class, all_deps, invoked_by)
+            node_span_id = node_class.Trigger._queue_node_execution(state, all_deps, invoked_by)
             if not node_class.Trigger.should_initiate(state, all_deps, node_span_id):
```
dvargasfuertes, line 435: I imagine combining these two in a future release

tests/workflows/basic_await_any/workflow.py
```diff
@@ -25,6 +26,9 @@ def run(self) -> Outputs:
 class AwaitAnyNode(BaseNode[BaseState]):
     top = TopNode.Outputs.total
 
+    class Trigger(BaseNode.Trigger):
+        merge_behavior = MergeBehavior.AWAIT_ANY
```
dvargasfuertes, line 30: This test was always busted. It claimed to be an AwaitAny node, but was still using the default AWAIT_ATTRIBUTES strategy

tests/workflows/basic_await_attributes/tests/test_workflow.py
```diff
@@ -14,4 +14,4 @@ def test_workflow__happy_path():
 
     # THEN the Workflow completes successfully
     assert final_event.name == "workflow.execution.fulfilled"
-    assert final_event.outputs.final_value == 2
+    assert final_event.outputs.final_value == 3
```
dvargasfuertes, line 17: This test was also busted. An AWAIT_ATTRIBUTES node should be invoked three times if there are three dependencies and the attributes are set for all three

tests/workflows/basic_emitter_workflow/tests/test_workflow.py
```diff
@@ -25,8 +25,9 @@ def test_run_workflow__happy_path(mock_uuid4_generator, mock_datetime_now):
     state_id_generator = mock_uuid4_generator("vellum.workflows.state.base.uuid4")
     state_id = state_id_generator()
     workflow_span_id = state_id_generator()
-    start_node_span_id = state_id_generator()
-    next_node_span_id = state_id_generator()
+    base_node_id_generator = mock_uuid4_generator("vellum.workflows.nodes.bases.base.uuid4")
```
dvargasfuertes, line 28: bc we moved the method, we needed to move the fixture

tests/workflows/basic_emitter_workflow/tests/test_workflow.py
```diff
@@ -120,12 +121,8 @@ def test_run_workflow__happy_path(mock_uuid4_generator, mock_datetime_now):
...
                     serialized_next_node_id: [str(next_node_span_id)],
                 },
-                "node_executions_queued": {
-                    serialized_next_node_id: [],
-                },
-                "dependencies_invoked": {
-                    str(next_node_span_id): [serialized_start_node_id],
-                },
+                "node_executions_queued": {},
+                "dependencies_invoked": {},
```
dvargasfuertes, line 125: we still consider the `node_executrion_cache` to be an unstable API, and we should feel free to mess with it

tests/workflows/circular_loop/tests/test_workflow.py
```diff
@@ -1,25 +1,56 @@
...
+
+    assert terminal_event.outputs.counter == 2
+    assert terminal_event.outputs.completed is True
+
+
 @pytest.mark.skip(
     reason="https://linear.app/vellum/issue/APO-390/workflow-erronously-terminates-in-looping-edge-case-with-prefixed-node"  # noqa: E501
 )
-def test_workflow__happy_path():
+def test_workflow__top_node_awaits_any():
```
dvargasfuertes, line 31: we still want to track and fix this issue for AWAIT ANY nodes one day

tests/workflows/multi_branch_merge_loop/workflow.py
```diff
@@ -49,6 +50,9 @@ class Ports(BaseNode.Ports):
     class Outputs(BaseNode.Outputs):
         value = TopNode.Outputs.value.coalesce(BottomNode.Outputs.value)
 
+    class Trigger(BaseNode.Trigger):
+        merge_behavior = MergeBehavior.AWAIT_ANY
```
dvargasfuertes, line 54: Another test that was assumed to be an AWAIT ANY node but wasn't under the hood

dvargas92495:
@vincent0426 with this, you should be able to iterate on ToolCalling Node

PR #1621, opened 2025-05-03
Make 400 push API errors cleaner to end users

dvargas92495:
Here's the error message I hit trying to push a workflow again to the base_custom_node example:

Screenshot 2025-05-03 at 3 05 56 PM

All user errors going forward _should not_ show the stack trace of the CLI. This PR introduces the pattern of how we could make these errors more user facing

### DIFF
--- a/ee/vellum_cli/push.py
+++ b/ee/vellum_cli/push.py
@@ -211,14 +211,14 @@ def push_command(
             )
 
             reported_diffs = f"""\
-    {e.body.get("detail")}
+{e.body.get("detail")}
 
-    {generated_only_str}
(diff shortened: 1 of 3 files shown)


ee/vellum_cli/push.py
```diff
@@ -178,51 +178,55 @@ def push_command(
             strict=strict,
         )
     except ApiError as e:
-        if e.status_code != 400 or not isinstance(e.body, dict) or "diffs" not in e.body:
```
dvargasfuertes, line 181: hide whitespace when reviewing below

PR #1625, opened 2025-05-04
[DO NOT MERGE] PoC of additional changes to route via API Key scheme

dvargas92495:


src/vellum/client/__init__.py
```diff
@@ -142,36 +142,57 @@ def __init__(
...
-        self.prompts = PromptsClient(client_wrapper=self._client_wrapper)
-        self.sandboxes = SandboxesClient(client_wrapper=self._client_wrapper)
-        self.test_suite_runs = TestSuiteRunsClient(client_wrapper=self._client_wrapper)
-        self.test_suites = TestSuitesClient(client_wrapper=self._client_wrapper)
-        self.workflow_deployments = WorkflowDeploymentsClient(client_wrapper=self._client_wrapper)
-        self.workflow_sandboxes = WorkflowSandboxesClient(client_wrapper=self._client_wrapper)
-        self.workflows = WorkflowsClient(client_wrapper=self._client_wrapper)
-        self.workspace_secrets = WorkspaceSecretsClient(client_wrapper=self._client_wrapper)
-        self.workspaces = WorkspacesClient(client_wrapper=self._client_wrapper)
+        self.ad_hoc = AdHocClient(client_wrapper=self._environment_client_wrapper)
```
dvargasfuertes, line 177: I wasn't diligent in the next few lines about which API is using which client. Just communicating that we will autogenerate the correct `*_client_wrapper` based on what's on the schema spec.

src/vellum/client/core/client_wrapper.py
```diff
@@ -22,14 +22,18 @@ def __init__(
...
             "X-Fern-Language": "Python",
             "X-Fern-SDK-Name": "vellum-ai",
             "X-Fern-SDK-Version": "0.14.53",
         }
-        headers["X-API-KEY"] = self._workspace_api_key
-        headers["X-API-KEY"] = self.environment_api_key
+        if scheme == "workspaceApiKeyAuth":
+            headers["X-API-KEY"] = self._workspace_api_key
+        elif scheme == "environmentApiKeyAuth":
+            headers["X-API-KEY"] = self.environment_api_key
```
dvargasfuertes, line 36: The other way we could avoid this entire PR is if environment api key can start being sent as a different header, then they both could be set without conflicting

dvargasfuertes: Oh but this change will require all users to change their current VellumClient setups to start adding a second kwarg. Maybe we could mitigate that by doing something like:

```python
self._environment_api_key = environment_api_key or workspace_api_key
```

PR #1632, opened 2025-05-05
Support display context gathering of workflow display classes with custom names

dvargas92495:
If a user pushes a workflow with a `CustomWorkflowNameLikeThis`, none of the display context mappings return correctly. This PR fixes that


vincent0426 (commented):
✅

PR #1633, opened 2025-05-05
Fix tool call default output

vincent0426:
found that other nodes do not specify default output?

### DIFF
--- a/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
+++ b/src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
@@ -81,10 +81,10 @@ class ToolCallingNode(BaseNode):
                     message="Subworkflow unexpectedly paused",
                 )
             elif terminal_event.name == "workflow.execution.fulfilled":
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -81,7 +81,7 @@ class Outputs(BaseWorkflow.Outputs):
                     message="Subworkflow unexpectedly paused",
                 )
             elif terminal_event.name == "workflow.execution.fulfilled":
-                node_outputs = self.Outputs()
+                node_outputs = self.Outputs(text="", chat_history=[])
 
                 for output_descriptor, output_value in terminal_event.outputs:
                     setattr(node_outputs, output_descriptor.name, output_value)
```
dvargasfuertes: [nit] realizing we should only set our _expected_ outputs, namely:

```python
node_outputs = self.Outputs(
    text=terminal_event.outputs.text,
    chat_history=terminal_event.outputs.chat_history,
)
```

dvargas92495 (approved):

PR #1636, opened 2025-05-05
Custom Base Node example works in prod

dvargas92495:
https://www.loom.com/share/c6c271c917da4b37bcd19abcccc13c7d

### DIFF
--- a/ee/vellum_cli/logger.py
+++ b/ee/vellum_cli/logger.py
@@ -44,5 +44,4 @@ def handle_cli_error(logger: logging.Logger, title: str, message: str, suggestio
         logger.error("\n\033[1;34mNext steps:\033[0m")  # Bold blue
         logger.error(suggestion)
 
-    logger.error(f"{title}: {message}")
     exit(1)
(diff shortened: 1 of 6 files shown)


ee/vellum_cli/logger.py
```diff
@@ -44,5 +44,4 @@ def handle_cli_error(logger: logging.Logger, title: str, message: str, suggestio
         logger.error("\n\033[1;34mNext steps:\033[0m")  # Bold blue
         logger.error(suggestion)
 
-    logger.error(f"{title}: {message}")
```
dvargasfuertes, line 47: redundant error message

examples/workflows/custom_base_node/nodes/mock_networking_client.py
```diff
@@ -33,6 +33,9 @@ class Outputs(BaseNode.Outputs):
         response: dict
 
     def run(self) -> BaseNode.Outputs:
+        if not self.state.chat_history:
+            self.state.chat_history = []
```
dvargasfuertes, line 37: this raises the idea of how we should explore default field factories, but punting that for now

PR #1646, opened 2025-05-07
Compile lambda func

vincent0426:
Not sure if this would happen, but I think we could still handle it?


dvargas92495 (approved):
I'll note that we are pretty far from serializing this as an attribute haha

PR #1651, opened 2025-05-08
[APO-503] Codegen function files

vincent0426:


ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -45,16 +58,54 @@ export class GenericNode extends BaseSingleFileNode<
...
-        python.field({
-          name: toPythonSafeSnakeCase(attribute.name),
-          initializer: new WorkflowValueDescriptor({
-            nodeContext: this.nodeContext,
-            workflowValueDescriptor: attribute.value,
-            workflowContext: this.workflowContext,
-          }),
-        })
-      );
+      if (attribute.name === "functions") {
```
dvargasfuertes: [nit] instead of hard coding here, let's use the attribute config that retry and try node uses to discriminate custom logic

dvargas92495 (approved):

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -44,17 +59,63 @@ export class GenericNode extends BaseSingleFileNode<
 
   getNodeClassBodyStatements(): AstNode[] {
     const statements: AstNode[] = [];
+    const nodeAttributes = NODE_DEFAULT_ATTRIBUTES["GenericNode"] as Record<
```
vincent0426: like to make sure if this is what we want? could update name to `NODE_ATTRIBUTES` and other files

vincent0426:
Move function files generation to `persist()` to avoid generating files during test

ee/codegen/src/generators/nodes/generic-node.ts
dvargasfuertes: Yup! but instead of `"GenericNode"`, we use `"ToolCallingNode"`. 

In the future, this config will be auto-generated from the SDK, similar to our generatedApi.ts in the monorepo

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -57,54 +59,62 @@ export class GenericNode extends BaseSingleFileNode<
...
-            }
-          });
-          const nodeName = this.nodeContext.getNodeLabel();
+    const nodeAttributes = NODE_DEFAULT_ATTRIBUTES["GenericNode"] as Record<
+      string,
+      AdornmentAttributeConfig
+    >;
 
+    this.nodeData.attributes.forEach((attribute) => {
+      switch (attribute.name) {
```
dvargasfuertes, line 68: we should switch on `type` instead of `name`, so that in the future we could support functions passed in as attributes to _any_ type

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -57,54 +59,62 @@ export class GenericNode extends BaseSingleFileNode<
...
+            functions.forEach((f) => {
+              if (f.definition && f.definition.name) {
+                functionNames.push(f.definition.name);
+              }
+            });
+            const nodeName = this.nodeContext.getNodeLabel();
+
+            statements.push(
+              python.field({
+                name: "functions",
```
dvargasfuertes: [nit] should be `attribute.name`

ee/codegen/src/generators/nodes/generic-node.ts
vincent0426: I think we are looping `NodeAttribute` which doesn't have `type`?

PR #1652, opened 2025-05-08
[APO-502] Fix port loop

vincent0426:
~~NOT 100% sure of the cause of this, may need to revisit it~~

the problem is expression referencing itself


ee/codegen/src/__test__/project.test.ts
```diff
@@ -3282,6 +3282,257 @@ baz = foo + bar
     });
   });
 
+  describe("port", () => {
+    it("should generate correct port references", async () => {
```
dvargasfuertes: [q] why here instead of graph-attribute.test.ts?

dvargas92495 (approved):

PR #1658, opened 2025-05-09
[APO-504] Support generic node prompt block

vincent0426:


ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -106,6 +109,39 @@ export class GenericNode extends BaseSingleFileNode<
           }
           break;
         }
+        case nodeAttributes.blocks?.name: {
```
dvargasfuertes, line 112: [oos] still bullish on switching this to attributeConfig.type as we mentioned

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -106,6 +109,39 @@ export class GenericNode extends BaseSingleFileNode<
...
+          const blocks = attribute.value;
+
+          if (
+            blocks &&
+            blocks.type === "CONSTANT_VALUE" &&
+            blocks.value?.type === "JSON"
+          ) {
+            const rawBlocks = blocks.value.value as PromptBlockSerializer.Raw[];
+            const deserializedBlocks: PromptBlockType[] = rawBlocks.map(
+              // @ts-ignore
```
dvargasfuertes: [nit] why do we need this? feels like a pretty important step here that we should try to get right

dvargas92495 (approved):

ee/codegen/src/generators/nodes/generic-node.ts
vincent0426: will fix in following PR

ee/codegen/src/generators/nodes/generic-node.ts
vincent0426: fixed, check if the result is `ok` first

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -119,8 +119,19 @@ export class GenericNode extends BaseSingleFileNode<
...
             const deserializedBlocks: PromptBlockType[] = rawBlocks.map(
-              // @ts-ignore
-              (block) => PromptBlockSerializer.parse(block).value
+              (block) => {
+                const parseResult = PromptBlockSerializer.parse(block);
+                if (parseResult.ok) {
+                  // TODO: Remove `as` once other types of blocks are supported
+                  return parseResult.value as PromptBlockType;
+                } else {
+                  throw new Error(
```
dvargasfuertes, line 128: [nit] Wherever possible, we should throw some extension of BaseCodegenError and catch them with workflow context so we don't kill the full code generation process

PR #1661, opened 2025-05-10
Fix attr config

vincent0426:
Will remove `name` in next PR

### DIFF
--- a/ee/codegen/src/generators/nodes/constants.ts
+++ b/ee/codegen/src/generators/nodes/constants.ts
@@ -26,7 +26,15 @@ export const NODE_ATTRIBUTES: Record<
     },
   },
   ToolCallingNode: {
-    functions: { name: "functions", defaultValue: null },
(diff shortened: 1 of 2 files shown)


ee/codegen/src/generators/nodes/constants.ts
```diff
@@ -26,7 +26,15 @@ export const NODE_ATTRIBUTES: Record<
     },
   },
   ToolCallingNode: {
-    functions: { name: "functions", defaultValue: null },
-    blocks: { name: "blocks", defaultValue: null },
+    blocks: {
+      name: "blocks",
+      defaultValue: null,
+      type: "blocks",
```
dvargasfuertes: [nit] I would have these somewhat match the type itself, like `PromptBlocks`. Also if we're going to reuse the value below, let's have the constants assigned to an exported variable

dvargas92495 (approved):

PR #1671, opened 2025-05-14
Auto include x api key into API node when invoking a Vellum url

dvargas92495:
This will relieve GTM from needing to create API Keys + Workspace Secrets on every target workspace they clone API Nodes that use Vellum APIs into. Longer term solutions look like:
- Vembda leases service tokens, we create a descriptor for those to make this look less "magical"
- We create OOTB nodes for other Vellum APIs (eg, GetDocument)


vincent0426 (commented):
✅

PR #1676, opened 2025-05-15
Serialize tool calling node inputs

vincent0426:
~~Serialize inputs for tool calling node to match inline prompt node~~
~~may not need this one as we should be able to select inputs for attribute~~
actually, we still need it

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/tool_calling_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/tool_calling_node.py
@@ -20,16 +20,17 @@ class BaseToolCallingNodeDisplay(BaseNodeDisplay[_ToolCallingNodeType], Generic[
 
         value = raise_if_descriptor(node.prompt_inputs)
         node_inputs: List[NodeInput] = []
-        for variable_name, variable_value in value.items():
-            node_input = create_node_input(
-                node_id=node_id,
-                input_name=variable_name,
-                value=variable_value,
-                display_context=display_context,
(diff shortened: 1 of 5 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/tool_calling_node.py
```diff
@@ -0,0 +1,37 @@
...
+                input_name=variable_name,
+                value=variable_value,
+                display_context=display_context,
+                input_id=self.node_input_ids_by_name.get(f"{node.prompt_inputs.name}.{variable_name}")
+                or self.node_input_ids_by_name.get(variable_name),
+            )
+            node_inputs.append(node_input)
+
+        return {
+            "inputs": [node_input.dict() for node_input in node_inputs],
```
vincent0426, line 36: need to pull out serialization for tool calling node to add `inputs` for it

tests/workflows/basic_tool_calling_node/tests/test_workflow.py
```diff
@@ -77,7 +77,11 @@ def generate_prompt_events(*args, **kwargs) -> Iterator[ExecutePromptEvent]:
     workflow = BasicToolCallingNodeWorkflow()
 
     # WHEN the workflow is executed
-    terminal_event = workflow.run()
+    terminal_event = workflow.run(
+        inputs=Inputs(
```
vincent0426, line 81: pass in inputs instead of hard coding

PR #1680, opened 2025-05-16
Handle lazy references of adorned nodes

NgoHarrison:
Context: Came up from a customer workflow issue. The problem was that if they did a lazy reference to a node which has an adornment, it would not find the correct reference since our lazy reference is reading from the wrapped node but the adornment (in this case retry) was the one that had the descriptor reference. Essentially lazy reference would reference `MyCustomNode.Outputs` but when you adorn it with retry, the output lives on `RetryNode.Outputs`.

### DIFF
--- a/src/vellum/workflows/nodes/core/retry_node/node.py
+++ b/src/vellum/workflows/nodes/core/retry_node/node.py
@@ -70,6 +70,12 @@ class RetryNode(BaseAdornmentNode[StateType], Generic[StateType]):
 
                     for output_descriptor, output_value in event.outputs:
                         setattr(node_outputs, output_descriptor.name, output_value)
+
+                        if self.__wrapped_node__:
+                            inner_desc = getattr(self.__wrapped_node__.Outputs, output_descriptor.name, None)
+                            if inner_desc:
+                                self.state.meta.node_outputs[inner_desc] = output_value
+
                 elif event.name == "workflow.execution.paused":
                     exception = NodeException(
                         code=WorkflowErrorCode.INVALID_OUTPUTS,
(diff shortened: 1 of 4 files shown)


tests/workflows/stream_retry_node_annotation/tests/test_workflow.py
```diff
@@ -142,7 +142,7 @@ def test_workflow_stream__happy_path():
 
     # workflow snapshotted events
     workflow_snapshotted_events = [e for e in events if e.name == "workflow.execution.snapshotted"]
-    assert len(workflow_snapshotted_events) == 2
+    assert len(workflow_snapshotted_events) == 3
 
     # AND the total number of events is correct
-    assert len(events) == 21
+    assert len(events) == 22
```
NgoHarrison, line 148: I had to bump the events by one since I am mutating the state for the workflow by adding an extra reference to the descriptor

src/vellum/workflows/nodes/core/retry_node/node.py
```diff
@@ -70,6 +70,12 @@ def run(self) -> BaseNode.Outputs:
 
                     for output_descriptor, output_value in event.outputs:
                         setattr(node_outputs, output_descriptor.name, output_value)
+
+                        if self.__wrapped_node__:
+                            inner_desc = getattr(self.__wrapped_node__.Outputs, output_descriptor.name, None)
+                            if inner_desc:
+                                self.state.meta.node_outputs[inner_desc] = output_value
+
```
vincent0426, line 78: will this happen on other types of adornment nodes (e.g., try) too?

vincent0426 (approved):

PR #1684, opened 2025-05-17
Harrison/add reason to guardrail node output

NgoHarrison:


ee/codegen/src/__test__/nodes/__snapshots__/guardrail-node.test.ts.snap
```diff
@@ -149,10 +151,10 @@ class GuardrailNodeDisplay(BaseGuardrailNodeDisplay[GuardrailNode]):
         "metric_inputs.actual": UUID("bed55ada-923e-46ef-8340-1a5b0b563dc1"),
     }
     output_display = {
-        GuardrailNode.Outputs.score1: NodeOutputDisplay(
+        LazyReference("GuardrailNode.Outputs.score1"): NodeOutputDisplay(
```
vincent0426, line 154: [q] Should `score1` be treated as a standard output key or only `score`?

NgoHarrison: Score is one of the outputs we expose from main vellum. Score1 was supposed to represent a random output variable from the metric execution

vincent0426 (commented):
✅

PR #1685, opened 2025-05-17
Add dependencies in await attribute

vincent0426:
We might need to know each `invoke_by` for each node to climb up the history to see if we are in the loop


NgoHarrison:
I am assuming the fix comes after this?

vincent0426:
> I am assuming the fix comes after this?

Yeah

PR #1686, opened 2025-05-19
[APO-390] Fix port looping issue on MERGE_ANY

vincent0426:
Error Workflow

```python
graph = (
    StartNode
    >> {
        MergeNode.Ports.group_1_if_port >> SomeNode >> AnotherNode >> StartNode,
        MergeNode.Ports.group_1_else_port >> FinalNode,
    }
)
```

We introduced `__node_execution_lookup__` in https://github.com/vellum-ai/vellum-python-sdks/pull/1673 to retrieve the full dependency graph.

In https://github.com/vellum-ai/vellum-python-sdks/pull/1685, we added support for the `AWAIT_ATTRIBUTE` dependency to ensure the graph continues tracing upward without stopping prematurely.

---

## The Problem

In a `MERGE_ANY` scenario, we want to **avoid triggering the merged node multiple times**.

### Example

```python
graph = (
    StartNode
    >> {
        TopNode,
        BottomNode,
    }
    >> EndNode
)
```

Here, `EndNode` is configured with `MERGE_ANY`. It should be triggered **only once**, by either `TopNode` or `BottomNode`, but **not both**.

---

### Problematic Case

```python
graph = (
    StartNode
    >> TopNode
    >> {
        MergeNode.Ports.group_1_if_port >> SomeNode >> AnotherNode >> TopNode,
        MergeNode.Ports.group_1_else_port >> FinalNode,
    }
)
```

In this case, `TopNode` is triggered once by `StartNode`, and then again via the loop after `AnotherNode` (which becomes the same scenario as the above one). Our current implementation skips the second trigger, causing the workflow to halt after `AnotherNode` and before `TopNode`.

---

This PR introduces two key changes:

1. **Loop Detection via Dependency Lookup**  
   We analyze the dependency graph to detect and correctly handle loops.

2. **`fork_id` Tracking**  
   We introduce a `fork_id` to identify fork/merge points. Once nodes merge, their `fork_id` is cleared. If a downstream node sees that the fork ID no longer exists, it avoids retriggering to ensure `MERGE_ANY` behavior.

---

## Known Issue

Due to loop detection, the test  
[`tests/workflows/multi_branch_merge_loop`](https://github.com/vellum-ai/vellum-python-sdks/tree/main/tests/workflows/multi_branch_merge_loop) now fails. When `TopNode` enters the loop a second time, the loop detection thinks a loop has started and retriggers the `LoopNode`.

### Failing Case

```python
graph = (
    StartNode
    >> {
        TopNode,
        BottomNode,
    }
    >> {
        LoopNode.Ports.loop >> StartNode,
        LoopNode.Ports.end >> EndNode,
    }
)
```

We added the `fork_id` logic in  
[this section of the code](https://github.com/vellum-ai/vellum-python-sdks/blob/e21799f31a0b4632862cf9204744a661b19ced79/src/vellum/workflows/runner/runner.py#L400-L411), where each forked node is assigned an ID. Once the nodes merge, this ID is cleared.

However, this does **not handle cases** where a forked path passes through intermediate nodes before merging. To handle that, we now **propagate the `fork_id`** forward so that the loop can continue correctly.

### Propagation Case

```python
graph = (
    StartNode
    >> {
        TopNode >> MiddleNode,
        BottomNode >> MiddleNode2,
    }
    >> {
        LoopNode.Ports.loop >> StartNode,
        LoopNode.Ports.end >> EndNode,
    }
)
```

### DIFF
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -369,23 +369,24 @@ class BaseNode(Generic[StateType], metaclass=BaseNodeMeta):
                 invoked_by_node = state.meta.node_execution_cache.__node_execution_lookup__.get(invoked_by)
 
                 # Check if invoked by node is a forked node
-                fork_id = state.meta.node_execution_cache.__node_to_fork_id__.get(invoked_by_node, None)
-                if fork_id:
-                    # If the invoked by node has a fork id and that fork id is in __all_fork_ids__
-                    #   We will
-                    #   1. remove the fork id from __all_fork_ids__
-                    #   2. remove the fork id from the __node_to_fork_id__ mapping
-                    # else (this mean the fork has already been triggered)
-                    #   remove the id from the node_to_fork_id mapping and not retrigger again
-                    all_fork_ids = state.meta.node_execution_cache.__all_fork_ids__
-                    if fork_id in all_fork_ids:
-                        # When the next forked node merge, it will not trigger the node again
-                        # We should consider adding a lock here to prevent race condition
-                        all_fork_ids.remove(fork_id)
-                        del state.meta.node_execution_cache.__node_to_fork_id__[invoked_by_node]
-                    else:
-                        should_retrigger = False
-                        del state.meta.node_execution_cache.__node_to_fork_id__[invoked_by_node]
+                if invoked_by_node is not None:
+                    fork_id = state.meta.node_execution_cache.__node_to_fork_id__.get(invoked_by_node, None)
+                    if fork_id:
+                        # If the invoked by node has a fork id and that fork id is in __all_fork_ids__
+                        #   We will
+                        #   1. remove the fork id from __all_fork_ids__
+                        #   2. remove the fork id from the __node_to_fork_id__ mapping
+                        # else (this mean the fork has already been triggered)
+                        #   remove the id from the node_to_fork_id mapping and not retrigger again
+                        all_fork_ids = state.meta.node_execution_cache.__all_fork_ids__
+                        if fork_id in all_fork_ids:
+                            # When the next forked node merge, it will not trigger the node again
+                            # We should consider adding a lock here to prevent race condition
+                            all_fork_ids.remove(fork_id)
+                            del state.meta.node_execution_cache.__node_to_fork_id__[invoked_by_node]
+                        else:
+                            should_retrigger = False
+                            del state.meta.node_execution_cache.__node_to_fork_id__[invoked_by_node]
 
                 # If should_retrigger is false, then we will not trigger the node already
                 # So we don't need to check loop behavior
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -356,8 +356,77 @@ class BaseNode(Generic[StateType], metaclass=BaseNodeMeta):
                 state.meta.node_execution_cache._dependencies_invoked[execution_id].add(invoked_by)
                 return execution_id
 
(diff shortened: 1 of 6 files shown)


tests/workflows/multi_branch_merge_loop_multi_length/workflow.py
```diff
@@ -0,0 +1,93 @@
...
+class MultiBranchMergeLoopMultiLengthWorkflow(BaseWorkflow[BaseInputs, State]):
+    """
+    This workflow is similar to the MultiBranchMergeLoopWorkflow, but the loop
+    has a different length.
+    """
+
+    graph = (
+        StartNode
+        >> {
+            TopNode >> MiddleNode,
```
vincent0426, line 82: Add a test for `fork_id` propagation, otherwise, all existing tests will pass even without propagation

tests/workflows/circular_loop/tests/test_workflow.py
```diff
@@ -24,9 +22,6 @@ def test_workflow__happy_path():
     assert terminal_event.outputs.completed is True
 
 
-@pytest.mark.skip(
-    reason="https://linear.app/vellum/issue/APO-390/workflow-erronously-terminates-in-looping-edge-case-with-prefixed-node"  # noqa: E501
-)
```
vincent0426, line 29: this PR's main goal

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -356,8 +356,78 @@ def _queue_node_execution(
...
+                        #   remove the id from the node_to_fork_id mapping and not retrigger again
+                        all_fork_ids = state.meta.node_execution_cache.__all_fork_ids__
+                        if fork_id in all_fork_ids:
+                            # When the next forked node merge, it will not trigger the node again
+                            # We should consider adding a lock here to prevent race condition
+                            all_fork_ids.remove(fork_id)
+                            del state.meta.node_execution_cache.__node_to_fork_id__[invoked_by_node]
+                        else:
+                            should_retrigger = False
+                            del state.meta.node_execution_cache.__node_to_fork_id__[invoked_by_node]
```
NgoHarrison, line 389: Because of the concern we have around race conditions would it be safer to do pop instead of hard `del`? To avoid key error?

noanflaherty (approved):
Nice! This is indeed tricky. Appreciate you leaving such helpful code comments to explain what's happening!

NgoHarrison (approved):

src/vellum/workflows/runner/runner.py
```diff
@@ -409,6 +411,21 @@ def _handle_invoked_ports(
...
+                            state.meta.node_execution_cache.__node_to_fork_id__[edge.to_node] = previous_node_fork_id
+                            # Remove previous node fork in __node_to_fork_id__
+                            del state.meta.node_execution_cache.__node_to_fork_id__[previous_node]
+
+            # If there are multiple edges, we need to create a fork ID
+            if len(nodes_to_fork) > 1:
+                fork_id = uuid4()
+                state.meta.node_execution_cache.__all_fork_ids__.add(fork_id)
+                for node in nodes_to_fork:
+                    state.meta.node_execution_cache.__node_to_fork_id__[node] = fork_id
```
dvargasfuertes, line 428: Let's sync up tm - would love to understand more about the need for this fork id

In general, I would love to find a way to keep logic outside of workflow runner and contained to the AWAIT_ANY case within the Node's Trigger

dvargas92495 (commented):
Love the extra test - would love to pair tm to see if we could clean up workflow runner

✅

PR #1690, opened 2025-05-20
Example: Multimodal Prompts - Client SDK

awlevin:
Just a simple example of using multimodal inputs to accompany [this help doc](https://docs.vellum.ai/product/prompts/multimodality)


examples/client/multimodal_inputs/requirements.txt
```diff
@@ -0,0 +1,2 @@
+python-dotenv==1.0.1
+vellum-ai==0.14.25
```
vincent0426: [q] do we want to keep this to the latest version? haha

vincent0426 (approved):

examples/client/multimodal_inputs/requirements.txt
awlevin: oh duh lol, good catch

PR #1698, opened 2025-05-21
Add ABC to BaseNode and BaseNodeAdornment

noanflaherty:
This lays some of the initial foundation to allow for Node Classes to be defined as Abstract.

With this, we can do stuff like:
```
class SearchNode(BaseSearchNode[StateType], ABC):
    ...
```

And use the presence of the `ABC` parent class to differentiate between node classes and their concrete usages.


NgoHarrison (approved):
Sweet didn't seem too bad so far although that compile error might signal a little bit more plumbing needed

PR #1712, opened 2025-05-27
[APO-563] Migrate code execution node

vincent0426:
TODO: 
- split `create_function_node` function node in `node.py` to separate workflow and function on a top level
- serialize/codegen runtime and package


vincent0426:
Will revisit. merge for serialization

PR #1714, opened 2025-05-28
[APO-559] Surface invalid inputs error for subworkflow deployments

NgoHarrison:
Context: This is related to prompt deployments not honoring optional inputs. This PR helps surface the actual invalid input error with message instead of internal server error


noanflaherty (approved):
Looks good, but would be great to handle the case when we encounter an unexpected/new error code and at least log a warning/error that we're hitting this case.

PR #1715, opened 2025-05-28
[APO-570] Cast primitives to vellum values for array inputs on code execution inputs

NgoHarrison:
Context: Customer was seeing internal server errors on their code execution inputs

### DIFF
--- a/src/vellum/workflows/nodes/displayable/code_execution_node/node.py
+++ b/src/vellum/workflows/nodes/displayable/code_execution_node/node.py
@@ -38,6 +38,7 @@ from vellum.workflows.errors.types import WorkflowErrorCode
 from vellum.workflows.exceptions import NodeException
 from vellum.workflows.nodes.bases import BaseNode
 from vellum.workflows.nodes.bases.base import BaseNodeMeta
+from vellum.workflows.nodes.displayable.bases.utils import primitive_to_vellum_value
 from vellum.workflows.nodes.displayable.code_execution_node.utils import read_file_from_path, run_code_inline
 from vellum.workflows.outputs.base import BaseOutputs
 from vellum.workflows.types.core import EntityInputsInterface, MergeBehavior, VellumSecret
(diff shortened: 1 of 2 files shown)


marlenc3 (approved):
Nice find!

src/vellum/workflows/nodes/displayable/code_execution_node/node.py
```diff
@@ -169,10 +180,41 @@ def _compile_code_inputs(self) -> List[CodeExecutorInput]:
                         )
                     )
                 else:
+                    # Convert primitive values to VellumValue objects
+                    vellum_values: List[VellumValue] = []
```
noanflaherty: I coulda swore we already had a helper util already somewhere that did this? If not, we probs should and should unit test it.

src/vellum/workflows/nodes/displayable/code_execution_node/node.py
```diff
@@ -169,10 +180,41 @@ def _compile_code_inputs(self) -> List[CodeExecutorInput]:
...
+                                ArrayVellumValue,
+                                ChatHistoryVellumValue,
+                                SearchResultsVellumValue,
+                            ),
+                        ):
+                            # If it's already a VellumValue (any type), keep it as is
+                            vellum_values.append(item)
+                        else:
+                            # For any other type, treat as JSON
+                            vellum_values.append(JsonVellumValue(value=item))
```
noanflaherty: This kinda spooks me as it could be a breaking change in the future when we introduce additional types (like boolean).

If we already have similar helpers, let's just double-check to make sure we do it the same way elsewhere so that we're at least consistent.

src/vellum/workflows/nodes/displayable/code_execution_node/node.py
noanflaherty: Either way, I'd extract this out into a helper method and write a parametrized unit test for it.

NgoHarrison: Ah yeah I just found it after digging around. Updated to use the existing one

noanflaherty (approved):

PR #1717, opened 2025-05-28
[APO-574] Fix dict iteration

vincent0426:


src/vellum/workflows/types/code_execution_node_wrappers.py
```diff
@@ -82,10 +82,6 @@ def __getattr__(self, attr):
     def __setattr__(self, name, value):
         self[name] = value
 
-    def __iter__(self):
```
vincent0426, line 85: iterate on dictionary should yield key not values

NgoHarrison (approved):

PR #1720, opened 2025-05-28
[APO-566] Do not include null values for prompt inputs

NgoHarrison:


src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
```diff
@@ -193,6 +193,8 @@ def _compile_prompt_inputs(self) -> Tuple[List[VellumVariable], List[PromptReque
             return input_variables, input_values
 
         for input_name, input_value in self.prompt_inputs.items():
+            if input_value is None:
+                continue
             if isinstance(input_value, str):
```
noanflaherty:
```suggestion
        for input_name, input_value in self.prompt_inputs.items():
            # Exclude inputs that resolved to be null. This ensure that we don't pass input values
            # to optional prompt inputs whose values were unresolved.
            if input_value is None:
                continue
            
            if isinstance(input_value, str):
```

src/vellum/workflows/nodes/displayable/bases/prompt_deployment_node.py
```diff
@@ -171,6 +171,8 @@ def _compile_prompt_inputs(self) -> List[PromptDeploymentInputRequest]:
             return compiled_inputs
 
         for input_name, input_value in self.prompt_inputs.items():
+            if input_value is None:
+                continue
             if isinstance(input_value, str):
```
noanflaherty:
```suggestion
        for input_name, input_value in self.prompt_inputs.items():
            # Exclude inputs that resolved to be null. This ensure that we don't pass input values
            # to optional prompt inputs whose values were unresolved.
            if input_value is None:
                continue
            
            if isinstance(input_value, str):
```

noanflaherty (approved):

PR #1726, opened 2025-05-29
[APO-577] Handle concurrent map output undefined

vincent0426:
We are setting `_outputs_post_init` on `self.Outputs`. In high concurrency, some threads may finish first and remove that attribute so the rest threads' outputs will not enter the following 

https://github.com/vellum-ai/vellum-python-sdks/blob/9e24ff58fe1fb883e18333936189841d71eb993d/src/vellum/workflows/outputs/base.py#L201-L202

### DIFF
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -490,7 +490,6 @@ class BaseNode(Generic[StateType], metaclass=BaseNodeMeta):
                         node_output_descriptor.name,
                         node_output_descriptor.instance.resolve(self.state),
                     )
-            delattr(self.Outputs, "_outputs_post_init")
 
         setattr(self.Outputs, "_outputs_post_init", _outputs_post_init)
 
--- a/tests/workflows/concurrent_map_execution/tests/test_workflow.py
+++ b/tests/workflows/concurrent_map_execution/tests/test_workflow.py
@@ -1,0 +1,17 @@
+from tests.workflows.concurrent_map_execution.workflow import ConcurrentMapExecutionWorkflow, Inputs
+
+
+def test_concurrent_map_execution():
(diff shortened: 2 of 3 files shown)


src/vellum/workflows/nodes/bases/base.py
```diff
@@ -490,7 +490,6 @@ def _outputs_post_init(outputs_self: "BaseNode.Outputs", **kwargs: Any) -> None:
                         node_output_descriptor.name,
                         node_output_descriptor.instance.resolve(self.state),
                     )
-            delattr(self.Outputs, "_outputs_post_init")
```
vincent0426, line 493: Chat w/ @NgoHarrison offline, need revisit, not sure if there is any side-effect as there is no test on it

tests/workflows/concurrent_map_execution/tests/test_workflow.py
```diff
@@ -0,0 +1,17 @@
...
+    workflow = ConcurrentMapExecutionWorkflow()
+
+    inputs = Inputs(arr=["a"] * 20)
+
+    # WHEN the workflow is run
+    terminal_event = workflow.run(inputs=inputs)
+
+    # THEN the workflow should be fulfilled
+    assert terminal_event.name == "workflow.execution.fulfilled"
+    assert all(output == 1 for output in terminal_event.outputs.final_output)
```
vincent0426, line 17: without the fix, output will have interleaving `undefined`

NgoHarrison (approved):

noanflaherty (approved):

src/vellum/workflows/nodes/bases/base.py
dvargasfuertes: no real reason this was here, the test provides us a reason to remove so good stuff

dvargas92495 (commented):
✅

PR #1728, opened 2025-05-30
Loosen Constraints of Dependencies

noanflaherty:
Make it easier to play nice with existing packages that are in use.


NgoHarrison (approved):

pyproject.toml
```diff
@@ -36,22 +36,21 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.9"
-Jinja2 = "^3.1.2"
-cdktf = "^0.20.5"
```
siddseethepalli, line 40: Is this removal intentional?

siddseethepalli (commented):
✅

pyproject.toml
noanflaherty: Yes, we don't use cdktf anymore. Thanks for caching though!

vincent0426 (commented):
✅

PR #1746, opened 2025-06-02
[APO-581] Fix lazy references for prompt inputs

NgoHarrison:
Context: Classic function calling demo was failing because of circular import. We were using node attributes now and need to pass in node context to various workflow descriptor references


dvargas92495 (approved):
wow - great find

vincent0426 (approved):

PR #1755, opened 2025-06-03
[APO-530] Serialize tool calling inline workflow

vincent0426:


ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_inline_workflow_serialization.py
```diff
@@ -0,0 +1,660 @@
...
+                                "type": "WORKFLOW_INPUT",
+                                "input_variable_id": "045942b7-e5b9-482c-b4d4-943309a20e05",
+                            },
+                        }
+                    ],
+                },
+            },
+            {
+                "id": "a4e3bc9f-7112-4d2f-94fb-7362a85db27a",
+                "name": "function_configs",
```
vincent0426, line 648: `function_configs` will be removed after workflow package

src/vellum/workflows/state/encoder.py
```diff
@@ -56,6 +57,16 @@ def default(self, obj: Any) -> Any:
             # in practice, we only ever pass the former
             return asdict(obj)  # type: ignore[call-overload]
 
+        if is_workflow_class(obj):
+            from vellum_ee.workflows.display.workflows.get_vellum_workflow_display_class import get_workflow_display
```
vincent0426: will cause cycle if import at the top

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_inline_workflow_serialization.py
```diff
@@ -0,0 +1,661 @@
...
+                                                "inputs": [],
+                                                "data": {
+                                                    "label": "Entrypoint Node",
+                                                    "source_handle_id": "c344fdee-282b-40c9-8c97-6dd08830948c",
+                                                },
+                                                "display_data": {"position": {"x": 0.0, "y": 0.0}},
+                                                "base": None,
+                                                "definition": None,
+                                            },
+                                            # type: ignore[misc]
```
vincent0426: not sure why `SUBWORKFLOW` causes `mypy` to hang indefinitely

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_inline_workflow_serialization.py
```diff
@@ -0,0 +1,661 @@
+# type: ignore  # subworkflow is causing mypy to hang indefinitely
```
vincent0426, line 1: not sure why SUBWORKFLOW node causes `mypy` to hang indefinitely

src/vellum/workflows/state/encoder.py
dvargasfuertes: [blocking] we should never import `vellum_ee` into `src/vellum` directory. Unfortunately, I think we'll need to do some recursive crawl on the display side

dvargas92495 (approved):

PR #1759, opened 2025-06-04
Remove fork for port loop

vincent0426:
Remove fork by comparing latest execution id in fulfilled queue


dvargas92495 (approved):
good stuff!

PR #1760, opened 2025-06-04
Move logic from base single file node to base node

vincent0426:


ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -52,6 +53,8 @@ export abstract class BaseNode<
 
   private readonly errorOutputId: string | undefined;
 
+  skipFormatting: boolean = false;
```
dvargasfuertes, line 56: @NgoHarrison has solved the need for this and we could remove

ee/codegen/src/generators/nodes/bases/single-file-base.ts
```diff
@@ -1,9 +1,6 @@
-import { AstNode } from "@fern-api/python-ast/core/AstNode";
-
 import { BaseNode } from "./base";
 
 import { BaseNodeContext } from "src/context/node-context/base";
-import { BasePersistedFile } from "src/generators/base-persisted-file";
 import { WorkflowDataNode } from "src/types/vellum";
 
 export abstract class BaseSingleFileNode<
```
dvargasfuertes, line 6: [oos] deleting this guy on the next one?

dvargas92495 (approved):

PR #1769, opened 2025-06-05
Generalize node attributes

vincent0426:


ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -36,10 +36,11 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
   }
 
   private generateNodeAttributes(): AstNode[] {
-    const nodeAttributes = NODE_ATTRIBUTES["ToolCallingNode"] as Record<
-      string,
-      AttributeConfig
-    >;
+    const nodeAttributes =
+      (NODE_ATTRIBUTES[this.nodeData.base.name] as Record<
```
dvargasfuertes: do we still need this `as` cast?

dvargas92495 (approved):

ee/codegen/src/generators/nodes/generic-node.ts
vincent0426: removed!

PR #1773, opened 2025-06-06
Adding a hacky workaround to resupport Workflow output deltas

dvargas92495:
Old workflow runner used to support the idea of streaming workflow outputs when a prompt node was connected to a terminal node. The goal of this PR was to bring back this behavior.

The first attempt was through node-to-node streaming, which could be found here: https://github.com/vellum-ai/vellum-python-sdks/pull/1578/files. This got really hairy, and didn't even get to a stable place. Given that Terminal nodes will be gone in some medium term future, I decided to stop investing in node to node streaming support and invest in an alternative approach.

This PR enables the idea that any node in the graph could define a custom method for when one of its streaming events could act as a workflow level streaming event. Long term, the hope is that this gets replaced with direct referencing of workflow outputs to node outputs without needing terminal nodes.

Going to sleep on this before trying to publish and test tm morning, but this feels overall far less risky than the original approach.

### DIFF
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -10,6 +10,7 @@ from vellum.workflows.constants import undefined
 from vellum.workflows.descriptors.base import BaseDescriptor
 from vellum.workflows.descriptors.utils import is_unresolved, resolve_value
 from vellum.workflows.errors.types import WorkflowErrorCode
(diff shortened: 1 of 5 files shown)


noanflaherty (approved):
This seems like a reasonable hack to me. I like the extra documentation and use of the extra-private-looking `__simulates_workflow_output__` attribute.

Between node-to-node streaming + connecting directly to workflow outputs being two viable longer-term solutions, this feels like a good 80/20 short term one.

PR #1774, opened 2025-06-06
[APO-616] Codegen nested tool calling node

vincent0426:
Screenshot 2025-06-05 at 5 33 38 PM
Screenshot 2025-06-05 at 5 33 42 PM


ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -25,6 +26,9 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
 
   private nodeAttributes: AstNode[] = [];
 
+  // True for node that has functions in attributes
```
dvargasfuertes:
```suggestion
  // True for node that has additional assets generated from attributes like functions or subworkflows
```

dvargas92495 (approved):

PR #1791, opened 2025-06-07
Fix tool calling inline workflow as tool

vincent0426:


ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_inline_workflow_serialization.py
```diff
@@ -1,4 +1,3 @@
-# type: ignore  # subworkflow is causing mypy to hang indefinitely
```
vincent0426, line 1: Removed, but this will take 3 min

dvargasfuertes: I'd remove it from the precommit if so. we should also separately investigate why this is the case, but not worth prioritizing now

dvargas92495 (approved):

PR #1792, opened 2025-06-07
Successfully pushed mcp demo in vellum

dvargas92495:
Now time to start running in the Vellum UI and debug issues there


vincent0426 (commented):
✅

PR #1800, opened 2025-06-09
Preserve context for tool calling node inline workflow

vincent0426:
While trying to use a code execution node in inline workflow in FE, found that it couldn't access generated files


dvargas92495 (approved):
good stuff

PR #1811, opened 2025-06-10
Fix output unique class

vincent0426:


ee/codegen/src/utils/paths.ts
```diff
@@ -76,6 +76,10 @@ export function getGeneratedNodeModuleInfo({
       nodeDefinition?.name ?? workflowContext.getUniqueClassName(nodeLabel);
   }
 
+  // Add the node class name to the used class names to ensure that the node class name that
+  // does not call getUniqueClassName still gets added to the used class names.
+  workflowContext.addUsedClassName(nodeClassName);
```
vincent0426: If `nodeDefinition.name` exists, it will be used directly and **NOT** added to the list of unique class names. The next time `workflowContext.getUniqueClassName(nodeLabel)` is called, it will be treated as if it's being called for the first time. As a result, it's possible for it to generate the same name as `nodeDefinition.name`.

```ts
from .final_output import FinalOutput
from .final_output_1 import FinalOutput as FinalOutput1FinalOutput
from .final_output_2 import FinalOutput1
```

vincent0426: We could also remove https://github.com/vellum-ai/vellum-python-sdks/blob/0bc7f14e10135f2d7ea1415b7c090b8155031195/ee/codegen/src/context/workflow-context/workflow-context.ts#L538
and https://github.com/vellum-ai/vellum-python-sdks/blob/0bc7f14e10135f2d7ea1415b7c090b8155031195/ee/codegen/src/context/workflow-context/workflow-context.ts#L547

as we now add for every class name but could also keep it to make the code more explicit

dvargasfuertes: I would expect this to be up a level within the node context's init instead of a `getGeneratedNodeModuleInfo` util whose naming implies no writes

dvargas92495 (approved):
nice catch

PR #1814, opened 2025-06-11
clean up flake ignores

dvargas92495:
cleans up some pyproject info


pyproject.toml
```diff
@@ -140,5 +139,4 @@ exclude = "(.git|venv|src/vellum/client|tests/client|scripts|examples/workflows)
...
+ignore = "E203, W503, E704, U100, U101"
```
dvargasfuertes, line 142: this experiment ended up being a disaster and I'm just ignoring `U100, U101` for now. Most of the offenses are due to methods that don't use the kwargs, but who's consumers expect them to exist, so renaming to `_` would break. Most cases involved redundant `_ = unused_kwarg` lines, at which point, we were just creating work for no reason, some of which I just left below since they are in tests

NgoHarrison (commented):
✅

PR #1817, opened 2025-06-12
[APO-617] Codegen inline workflow tool

vincent0426:
https://www.loom.com/share/3da0ba0cec914f43bb6313e142e4b35a?sid=7da51aed-4c4f-4d8e-93cd-d1eec0ec9e9e

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -400,6 +400,243 @@ class GetCurrentWeatherNode(ToolCallingNode):
 "
 `;
 
+exports[`WorkflowProjectGenerator > function > should generate inline workflow tool 1`] = `
+"from vellum.workflows.inputs import BaseInputs
+
+
+class Inputs(BaseInputs):
+    query: str
+    city: str
+    date: str
+"
+`;
+
+exports[`WorkflowProjectGenerator > function > should generate inline workflow tool 2`] = `
+"from .final_output import FinalOutput
+from .final_output_1 import FinalOutput1
+from .final_output_2 import FinalOutput2
+from .tool_call_get_current_weather_node import GetCurrentWeatherNode
+
+__all__ = [
+    "FinalOutput",
+    "FinalOutput1",
+    "FinalOutput2",
+    "GetCurrentWeatherNode",
+]
+"
+`;
+
+exports[`WorkflowProjectGenerator > function > should generate inline workflow tool 3`] = `
+"from vellum.workflows.nodes.displayable import FinalOutputNode
+from vellum.workflows.state import BaseState
+
+from ..inputs import Inputs
+
+
+class FinalOutput(FinalOutputNode[BaseState, str]):
+    class Outputs(FinalOutputNode.Outputs):
+        value = Inputs.city
+"
+`;
+
+exports[`WorkflowProjectGenerator > function > should generate inline workflow tool 4`] = `
+"from vellum.workflows.nodes.displayable import FinalOutputNode
+from vellum.workflows.state import BaseState
+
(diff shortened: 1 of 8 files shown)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -400,6 +400,243 @@ class GetCurrentWeatherNode(ToolCallingNode):
...
+    }
+    function_configs = None
+
+    class Outputs(ToolCallingNode.Outputs):
+        text: str
+        chat_history: List[ChatMessage]
+"
+`;
+
+exports[`WorkflowProjectGenerator > function > should generate inline workflow tool 7`] = `
```
dvargasfuertes, line 517: [q] is there a way for these snapshot labels to include _which_ file is being snapshotted here instead of `7` ?

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -108,6 +109,9 @@ export class WorkflowContext {
   // new nodes.
   private readonly nodeModuleNames: Set<string> = new Set();
 
+  // Track the custom workflow module name if it exists
+  public readonly nestedWorkflowModuleName?: string;
```
dvargasfuertes, line 113: [nit] what's the difference between this and `moduleName`? is it possible to consolidate?

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
dvargasfuertes: If there is, let's resolve as a followup since it'll likely affect all snapshots in this file

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -52,28 +66,105 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
...
+                  const rawExecConfig =
+                    f.exec_config as unknown as WorkflowVersionExecConfigSerializer.Raw;
+                  const workflowVersionExecConfigResult =
+                    WorkflowVersionExecConfigSerializer.parse(rawExecConfig);
+                  if (!workflowVersionExecConfigResult.ok) {
+                    throw new Error(
+                      `Failed to parse WorkflowVersionExecConfig: ${JSON.stringify(
+                        workflowVersionExecConfigResult.errors
+                      )}`
+                    );
```
dvargasfuertes, line 90: [blocking] let's make sure that:
- All errors we intentionally throw are extensions of `BaseCodegenError`
- We put the error in `this.workflowContext.addError`
- We mark the error as a `WARNING` unless we are _absolutely_ sure that the error is bad enough to kill the whole code generation process, which is rare if ever

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -52,28 +66,105 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
...
-                  functionNames.map((name) => {
-                    const snakeName = toPythonSafeSnakeCase(name);
-                    return python.reference({
-                      name: snakeName,
-                      modulePath: [`.${snakeName}`],
-                    });
-                  })
-                ),
+                initializer: python.TypeInstantiation.list([
+                  ...codeExecutionFunctions
```
dvargasfuertes, line 133: [blocking] this changes the order of how we codegen it. meaning if a user defines it in the UI as [subworkflow, code-exec], it will codegen as [code-exec, subworkflow]. We should try to preserve order

dvargas92495 (approved):
Great work 🫡 

Going to approve since the PR is meaty and we should try to get it into main. However, I would like the two blocking comments addressed as fast followup PRs before we cut a release

PR #1819, opened 2025-06-13
Handle serialization of prompt blocks as an attribute

dvargas92495:
This PR adds support for serializing the `blocks` attribute on inline prompt node _as_ an attribute. Most changes are snapshot changes, actual implementation is ~10 lines.

The first attempt was made here: https://github.com/vellum-ai/vellum-python-sdks/pull/1816, but prompt node serialization is so complex atm that Devin couldn't handle


NgoHarrison (commented):
✅

vincent0426 (commented):
✅

PR #1831, opened 2025-06-16
Translate from vellum variable to primitive type

vincent0426:
Parsing inputs returned from subworkflow for tool calling node subworkflow tool

### DIFF
--- a/src/vellum/workflows/utils/tests/test_vellum_variables.py
+++ b/src/vellum/workflows/utils/tests/test_vellum_variables.py
@@ -3,7 +3,10 @@ from typing import List, Optional
 
 from vellum import ChatMessage, SearchResult, VellumAudio, VellumDocument, VellumImage
 from vellum.workflows.types.core import Json
-from vellum.workflows.utils.vellum_variables import primitive_type_to_vellum_variable_type
+from vellum.workflows.utils.vellum_variables import (
+    primitive_type_to_vellum_variable_type,
+    vellum_variable_type_to_primitive_type,
+)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/utils/tests/test_vellum_variables.py
```diff
@@ -31,3 +34,24 @@
...
+        ("ARRAY", "array"),
+        ("FUNCTION_CALL", "object"),
+        ("IMAGE", "object"),
+        ("AUDIO", "object"),
+        ("DOCUMENT", "object"),
+        ("NULL", "null"),
+    ],
+)
+def test_vellum_variable_type_to_primitive_type(vellum_type, expected):
+    assert vellum_variable_type_to_primitive_type(vellum_type) == expected
```
dvargasfuertes: [nit] Instead of "primitive" type, this is the openapi type correct? We've been using the word "primitive" to mean python, we should make sure that this is different from that

src/vellum/workflows/utils/vellum_variables.py
```diff
@@ -78,6 +78,36 @@ def primitive_type_to_vellum_variable_type(type_: Union[Type, BaseDescriptor]) -
...
+    elif vellum_type == "ARRAY":
+        return "array"
+    elif vellum_type == "FUNCTION_CALL":
+        return "object"
+    elif vellum_type == "IMAGE":
+        return "object"
+    elif vellum_type == "AUDIO":
+        return "object"
+    elif vellum_type == "DOCUMENT":
+        return "object"
```
dvargasfuertes, line 104: [oos] technically, we'll want these to point to defined components. We can hold off on that for now

dvargas92495 (approved):

PR #1839, opened 2025-06-17
[APO-347] Support subworkflow deployment tool

vincent0426:
serialization shape would be: 

```
{
    "type": "DEPLOYMENT_WORKFLOW",
    "deployment": "",
    "release_tag": ""
}
```


src/vellum/workflows/nodes/experimental/tool_calling_node/node.py
```diff
@@ -106,11 +108,10 @@ def _build_graph(self) -> None:
 
         self._function_nodes = {}
         for function in self.functions:
-            function_name = snake_case(function.__name__)
-
+            function_name = get_function_name(function)
             # Get configuration for this function
             config = {}
-            if self.function_configs and function.__name__ in self.function_configs:
+            if callable(function) and self.function_configs and function.__name__ in self.function_configs:
```
vincent0426, line 115: added to avoid `Dict[str,str]` using __name__

src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
```diff
@@ -105,6 +109,13 @@ def _get_prompt_event_stream(self) -> Iterator[AdHocExecutePromptEvent]:
             for function in self.functions:
                 if isinstance(function, FunctionDefinition):
                     normalized_functions.append(function)
+                elif isinstance(function, Dict):
```
dvargasfuertes: [nit] Similar to `FunctionDefinition`, We should create a new pydantic model that we export, eg `DeploymentDefinition`, that gives us the structured data that we need here

dvargasfuertes: Actually going to [blocking] on this, I think we want to avoid dict parsing logic

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -68,14 +70,14 @@ def run(self) -> Iterator[BaseOutput]:
 def create_tool_router_node(
     ml_model: str,
     blocks: List[PromptBlock],
-    functions: List[Callable[..., Any]],
+    functions: List[Union[Callable[..., Any], DeploymentDefinition]],
```
dvargasfuertes, line 73: [oos] Why is inline workflow not in this list?

dvargas92495 (approved):

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
vincent0426: yeah i found that lol im going to include that in the following PR

PR #1840, opened 2025-06-17
Fix CI formatting

vincent0426:
I think we should avoid ignoring the whole file?


dvargas92495 (commented):
I ended up getting rid of this: https://github.com/vellum-ai/vellum-python-sdks/pull/1814

PR #1852, opened 2025-06-18
Tool function type

vincent0426:


src/vellum/workflows/types/core.py
```diff
@@ -39,3 +47,7 @@ class ConditionType(Enum):
     IF = "IF"
     ELIF = "ELIF"
     ELSE = "ELSE"
+
+
+# Type alias for functions that can be called in tool calling nodes
+ToolFunction = Union[Callable[..., Any], DeploymentDefinition, Type["BaseWorkflow"]]
```
NgoHarrison: [q] Do we anticipate other nodes using these functions? If so maybe choosing another name? Right now it seems coupled with Tool calling nodes (which makes sense as that is the only one using it)

vincent0426: @dvargas92495 thoughts on the naming? not sure if we would pass tool to other nodes

dvargasfuertes: I could see us both passing it to other nodes and the name making sense. Since we want to drive off of attribute types over node types going forward, and "Tool" is becoming a pretty first class concept in AI.

In fact, I could even see us calling this just `Tool`

PR #1858, opened 2025-06-18
[APO-669] Serialize deployment workflow tool

vincent0426:


ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_deployment_workflow_serialization.py
```diff
@@ -0,0 +1,168 @@
...
+            {"id": "7bd47b06-0481-48a7-a51e-6a998385ee9a", "key": "chat_history", "type": "CHAT_HISTORY"},
+        ],
+        output_variables,
+        ignore_order=True,
+    )
+
+    # AND its raw data should be what we expect
+    workflow_raw_data = serialized_workflow["workflow_raw_data"]
+    tool_calling_node = workflow_raw_data["nodes"][1]
+    assert tool_calling_node == {
```
dvargasfuertes: [nit] for a lot of these serialization tests, we should think about "what is the _new_ fact that I want to make sure works?" In this case, I think we could get away with just asserting the shape of the `tool_calling_node["attributes"]["functions"]` attribute

dvargas92495 (approved):

PR #1860, opened 2025-06-19
Skip codegen function configs

vincent0426:


ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -205,6 +205,15 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
           break;
         }
         default:
+          // Skip generating fields for explicitly null values
+          if (
+            attribute.value?.type === "CONSTANT_VALUE" &&
+            attribute.value.value?.type === "JSON" &&
+            attribute.value.value.value === null
```
dvargasfuertes: What provoked this?

In general, we don't want to skip generation of null, but skip generation of an attribute equaling its _default_ value, if its known

dvargasfuertes: In alot of cases, this is null, but not always

vincent0426: ah yeah, want to skip function configs for now

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -204,6 +204,13 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
           }
           break;
         }
+        case AttributeType.FunctionConfigs: {
```
dvargasfuertes, line 207: What's a FunctionConfig? Was this bc of packages for Code Exec Node?

I'd rather not continue spreading out usage of this as we'll roll back code exec nodes once we have workflow level packages

vincent0426: oh i see

PR #1874, opened 2025-06-21
Add test for default codegen release tag

vincent0426:
Not sure if we need to codegen `LATEST` if `release_tag` is `null`

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/tool-calling-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/tool-calling-node.test.ts.snap
@@ -71,6 +71,18 @@ class ToolCallingNode(BaseToolCallingNode):
 "
 `;
 
+exports[`ToolCallingNode > deployment workflow > should generate latest release tag if release_tag is null 1`] = `
+"from vellum.workflows.nodes.experimental.tool_calling_node import (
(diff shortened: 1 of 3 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/tool-calling-node.test.ts.snap
```diff
@@ -71,6 +71,18 @@ class ToolCallingNode(BaseToolCallingNode):
...
 
+exports[`ToolCallingNode > deployment workflow > should generate latest release tag if release_tag is null 1`] = `
+"from vellum.workflows.nodes.experimental.tool_calling_node import (
+    ToolCallingNode as BaseToolCallingNode,
+)
+from vellum.workflows.types.definition import DeploymentDefinition
+
+
+class ToolCallingNode(BaseToolCallingNode):
+    functions = [DeploymentDefinition(deployment="deployment_1", release_tag="LATEST")]
```
dvargasfuertes: [nit] I'd slightly prefer codegenning `DeploymentDefinition(deployment="deployment_1")`, and updating our SDK logic to know that no `release_tag` defaults to `"LATEST"` when left unspecified

dvargas92495 (approved):

dvargas92495 (approved):

PR #1879, opened 2025-06-23
[APO-342] Support max prompt iterations

vincent0426:


src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -41,6 +42,15 @@ class Trigger(InlinePromptNode.Trigger):
         merge_behavior = MergeBehavior.AWAIT_ATTRIBUTES
 
     def run(self) -> Iterator[BaseOutput]:
+        if hasattr(self, "max_prompt_iterations") and self.max_prompt_iterations is not None:
```
dvargasfuertes: [nit] Let's default the attribute to None so that we don't need this `hasattr` check

tests/workflows/basic_tool_calling_node_max_prompt_iterations/tests/test_workflow.py
```diff
@@ -0,0 +1,101 @@
...
+    uuid4_generator()
+
+    # GIVEN a limited tool calls workflow with max_tool_calls=2
+    workflow = BasicToolCallingNodeMaxPromptIterationsWorkflow()
+
+    # WHEN the workflow is executed
+    terminal_event = workflow.run()
+
+    # THEN the workflow is executed successfully
+    assert terminal_event.name == "workflow.execution.fulfilled"
```
dvargasfuertes: [blocking] I think in the [PRD](https://www.notion.so/vellum-ai/PRD-ToolCallingNode-1c39035c57bd8084b8a9defa1a767b31), we aligned on rejecting

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
dvargasfuertes: Actually, reading [Aaron's](https://www.notion.so/PRD-ToolCallingNode-1c39035c57bd8084b8a9defa1a767b31?d=1c39035c57bd807eab61001c6697f051&source=copy_link#1c39035c57bd8046a578d029cf5e5c0a) comment here, I'm seeing we should default to some concrete number, like 5.

tests/workflows/basic_tool_calling_node_max_prompt_iterations/tests/test_workflow.py
dvargasfuertes: To Aaron's point, we can [oos] another attribute called like `max_tool_behavior`, which is an enum between REJECT and PROMPT and maybe others in the future

dvargas92495 (approved):

PR #1880, opened 2025-06-23
Serialize inline workflow tool definition

vincent0426:


ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_inline_workflow_serialization.py
```diff
@@ -131,6 +131,10 @@ def test_serialize_workflow():
                         "value": [
                             {
                                 "type": "INLINE_WORKFLOW",
+                                "definition": {
+                                    "name": "BasicInlineSubworkflowWorkflow",
+                                    "description": "\n    A workflow that gets the weather for a given city and date.\n    ",  # noqa: E501
+                                },
```
dvargasfuertes: [blocking] I'd make this top level, so that all three tools have a consistent `name` and `description` to use.

I recognize that this means that tools like raw code tool will have redundant data, but think of it as the difference between the model's spec vs the code's spec, where we'll prefill the former with the latter

ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -329,14 +329,20 @@ def serialize_value(display_context: "WorkflowDisplayContext", value: Any) -> Js
         from vellum_ee.workflows.display.workflows.get_vellum_workflow_display_class import get_workflow_display
 
         workflow_display = get_workflow_display(workflow_class=value)
-        value = workflow_display.serialize()
+        serialized_value: dict = workflow_display.serialize()
+        name = serialized_value["workflow_raw_data"]["definition"]["name"]
+        description = value.__doc__ or ""
```
dvargasfuertes, line 334: [oos] we should split out `__doc__` in a separate PR that gets sourced and stored as part of `workflow_display.serialize()`. we'll want to have this stored in the exec_config somewhere, likely in display data consistent with how Nodes store their comments

dvargas92495 (approved):

PR #1881, opened 2025-06-23
[APO-695] Fix map node event concurrent processing

NgoHarrison:
[CURRENT STATUS] - The theory was that something is wrong with either state or the attributes within the state, specifically the descriptor instances. I believe the newest print below that is denoted with a PR comment and TODO debunks this theory since it seems like it's showing the same id across the subworkflow iterations leading me to believe that it is one descriptor instance shared throughout the runs.

Could use some more input for this. I think the fix might be to make our descriptors thread safe but that seems like a kinda big change that I didn't want to do unless agreed upon.

Context: User was reporting issues with map nodes reporting the wrong outputs. For example, if I had a list of [a, b, c, d, e, f] and I was using these map node outputs in a final output, the result could be something like [f, f, f, f, e, f]. My first commit was to address the first theory which was we were accessing/overwriting the queue index and values due to multiple threads. There was a comment about using slicing but not sure if that can also help since, in my head, that only reduces the number of access/reads due to batching everything but the problem can still be there. However, after debugging more my second theory is that we are passing a singular descriptor reference for each subworkflow iteration of the map node (`TemplatingNode.Outputs.processed_item`). I came to this new working theory because the following test setup passes but not the actual test I have in this PR (it uses a final output within the map node):

```    
    class TemplatingNode(BaseNode):
        item = MapNode.SubworkflowInputs.item

        class Outputs(BaseOutputs):
            processed_item: str

        def run(self) -> Outputs:
            processed_item = f"{self.item}!"
            return self.Outputs(processed_item=processed_item)

    # AND a workflow using those nodes
    class ProcessItemWorkflow(BaseWorkflow[MapNode.SubworkflowInputs, BaseState]):
        graph = TemplatingNode

            result = TemplatingNode.Outputs.processed_item

    # AND a map node with high concurrency
    class RaceConditionMapNode(MapNode):
        items = ["a", "b", "c", "d", "e", "f"]
        subworkflow = ProcessItemWorkflow
        max_concurrency = 50
```

### DIFF
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -457,21 +457,6 @@ class BaseNode(Generic[StateType], ABC, metaclass=BaseNodeMeta):
             resolved_value = resolve_value(descriptor.instance, self.state, path=descriptor.name, memo=inputs)
             setattr(self, descriptor.name, resolved_value)
 
-        # Resolve descriptors set as defaults to the outputs class
-        def _outputs_post_init(outputs_self: "BaseNode.Outputs", **kwargs: Any) -> None:
-            for node_output_descriptor in self.Outputs:
-                if node_output_descriptor.name in kwargs:
-                    continue
-
-                if isinstance(node_output_descriptor.instance, BaseDescriptor):
-                    setattr(
-                        outputs_self,
-                        node_output_descriptor.name,
-                        node_output_descriptor.instance.resolve(self.state),
-                    )
-
-        setattr(self.Outputs, "_outputs_post_init", _outputs_post_init)
-
         # We only want to store the attributes that were actually set as inputs, not every attribute that exists.
         all_inputs = {}
         for key, value in inputs.items():
--- a/src/vellum/workflows/nodes/core/map_node/node.py
+++ b/src/vellum/workflows/nodes/core/map_node/node.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 import concurrent.futures
+from copy import deepcopy
 import logging
 from queue import Empty, Queue
 from typing import (
@@ -168,7 +169,7 @@ class MapNode(BaseAdornmentNode[StateType], Generic[StateType, MapNodeItemType])
     def _run_subworkflow(self, *, item: MapNodeItemType, index: int) -> None:
         context = WorkflowContext.create_from(self._context)
         subworkflow = self.subworkflow(
-            parent_state=self.state,
+            parent_state=deepcopy(self.state),
             context=context,
         )
         SubworkflowInputsClass = self.subworkflow.get_inputs_class()
--- a/src/vellum/workflows/nodes/core/map_node/node.py
+++ b/src/vellum/workflows/nodes/core/map_node/node.py
@@ -2,7 +2,6 @@ from collections import defaultdict
 import concurrent.futures
 import logging
 from queue import Empty, Queue
-from threading import Lock
 from typing import (
     TYPE_CHECKING,
     Callable,
@@ -52,8 +51,6 @@ class MapNode(BaseAdornmentNode[StateType], Generic[StateType, MapNodeItemType])
 
     items: List[MapNodeItemType]
     max_concurrency: Optional[int] = None
-
-    _execution_lock = Lock()
 
     class Outputs(BaseAdornmentNode.Outputs):
         pass
@@ -181,10 +178,8 @@ class MapNode(BaseAdornmentNode[StateType], Generic[StateType, MapNodeItemType])
             event_filter=all_workflow_event_filter,
         )
 
-        # Only serialize the event queuing
-        with self._execution_lock:
-            for event in events:
-                self._event_queue.put((index, event))
+        for event in events:
+            self._event_queue.put((index, event))
 
     @overload
     @classmethod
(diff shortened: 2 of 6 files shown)


vincent0426:
Seems like deadlock.

Maybe we could slice those indices according to the number of workers before spawning threads? like a map-reduce style

NgoHarrison:
> Seems like deadlock.
> 
> Maybe we could slice those indices according to the number of workers before spawning threads? like a map-reduce style

Hmm by deadlock do you mean by this change possibly introducing that potential? If so I think we should be fine? We are using that one lock on the class for event queueing. Worst case is that high volume of events could cause this to be slow since they wait for the lock but I don't htink it should deadlock

vincent0426:
> > Seems like deadlock.
> > Maybe we could slice those indices according to the number of workers before spawning threads? like a map-reduce style
> 
> Hmm by deadlock do you mean by this change possibly introducing that potential? If so I think we should be fine? We are using that one lock on the class for event queueing. Worst case is that high volume of events could cause this to be slow since they wait for the lock but I don't htink it should deadlock

CI is still running the test haha not sure if it’s caused by that

NgoHarrison:
> CI is still running the test haha not sure if it’s caused by that

LOL nvm I think you're right. It's hanging on the nested map nodes test

dvargas92495:
> after debugging more my second theory is that we are passing a singular descriptor reference for each subworkflow iteration of the map node

I think this is a symptom of the issue. 

In theory, each subworkflow iteration should have its _own_ State, and that _nested_ state should be storing its own value for `TemplatingNode.Outputs.processed_item`. Is that what we're observing? Could we encode those properties in the test somehow?

NgoHarrison:
> In theory, each subworkflow iteration should have its _own_ State, and that _nested_ state should be storing its own value for `TemplatingNode.Outputs.processed_item`. Is that what we're observing? Could we encode those properties in the test somehow?

Currently they aren't unique I think but I tried this as part of one set of changes yesterday but let me try and update the test to show that.

EDIT:
https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/nodes/core/map_node/node.py#L171

I did a deepcopy here (as part of those local testing changes but not shown in this PR) but let me update the test anyway

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -168,7 +169,7 @@ def _context_run_subworkflow(
     def _run_subworkflow(self, *, item: MapNodeItemType, index: int) -> None:
         context = WorkflowContext.create_from(self._context)
         subworkflow = self.subworkflow(
-            parent_state=self.state,
+            parent_state=deepcopy(self.state),
```
dvargasfuertes: [blocking] Instead of a deepcopy, all of the child state should still point to this parent state. It's the child state's that should all be different copies. Do we know if we're generating new child copies for each iteration of the loop?

The deepcopy would prevent state updates of the parent in child workflows.

src/vellum/workflows/nodes/core/map_node/tests/test_node.py
```diff
@@ -221,3 +222,50 @@ class TestMapNode(MapNode):
     # AND each item should have run on a different thread
     thread_ids_list = list(thread_ids.values())
     assert len(set(thread_ids_list)) == 3
+
+
+def test_map_node__shared_state_race_condition():
+    state_ids: list[int] = []
```
dvargasfuertes: this is a great assertion to keep, and should still be true after my comment above

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -457,21 +457,6 @@ def __init__(
...
-                    continue
-
-                if isinstance(node_output_descriptor.instance, BaseDescriptor):
-                    setattr(
-                        outputs_self,
-                        node_output_descriptor.name,
-                        node_output_descriptor.instance.resolve(self.state),
-                    )
-
-        setattr(self.Outputs, "_outputs_post_init", _outputs_post_init)
```
dvargasfuertes, line 473: this was the root of the bug.

`_outputs_post_init` was referencing instances in its method (`self.state`), even though it was assigned to a class (`self.Outputs`). So parallel executions would keep overwriting `self.Outputs` in a non-thread-safe way.

Moral of the story here is do not do instance operations in class methods

src/vellum/workflows/nodes/core/map_node/tests/test_node.py
```diff
@@ -221,3 +222,56 @@ class TestMapNode(MapNode):
...
+            return self.Outputs(processed_item=processed_item)
+
+    # AND a final output node
+    class FinalOutput(FinalOutputNode[BaseState, str]):
+        class Outputs(FinalOutputNode.Outputs):
+            value = TemplatingNode.Outputs.processed_item
+
+        def run(self) -> Outputs:
+            outputs = super().run()
+            processed_items.append(outputs.value)
```
dvargasfuertes, line 248: this is what helped me find the bug. This, plus running the node a bunch of times increases the odds that we run into the concurrency issue

src/vellum/workflows/outputs/base.py
```diff
@@ -198,8 +200,27 @@ def __init__(self, **kwargs: Any) -> None:
         for name, value in kwargs.items():
             setattr(self, name, value)
 
-        if hasattr(self, "_outputs_post_init") and callable(self._outputs_post_init):
```
dvargasfuertes, line 201: I had been dying to get rid of this hack, and found out how. turns out any method in python has access to variables in its parent context 🤯

dvargas92495 (commented):
cc: @NgoHarrison found the issue!

dvargas92495 (approved):

PR #1887, opened 2025-06-24
Serialize definition for workflow deployment tool

vincent0426:


ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -348,6 +348,8 @@ def serialize_value(display_context: "WorkflowDisplayContext", value: Any) -> Js
                 "type": "JSON",
                 "value": {
                     "type": "WORKFLOW_DEPLOYMENT",
+                    "name": value.deployment,
+                    "description": f"Workflow deployment for {value.deployment}",
```
dvargasfuertes, line 352: [oos] oh man. We store description's on deployments I believe. Let's cut a ticket in linear with medium priority that uses our API to fetch the deployment description here for the default, I don't think it should block Thursday delivery

dvargas92495 (approved):

PR #1896, opened 2025-06-25
Fix tool calling node deployment type

vincent0426:
UUID deployment will fail without fix

https://github.com/vellum-ai/vellum-python-sdks/blob/bbe74826bc55034b3db0b0b723172f39edebbce7/src/vellum/workflows/nodes/displayable/subworkflow_deployment_node/node.py#L147-L148

### DIFF
--- a/src/vellum/workflows/nodes/experimental/tool_calling_node/tests/test_utils.py
+++ b/src/vellum/workflows/nodes/experimental/tool_calling_node/tests/test_utils.py
@@ -46,7 +46,7 @@ def test_get_function_name_subworkflow_deployment():
 
     result = get_function_name(deployment_config)
 
-    assert result == "my_test_deployment"
+    assert result == "mytestdeployment"
 
 
 def test_get_function_name_subworkflow_deployment_uuid():
@@ -55,4 +55,4 @@ def test_get_function_name_subworkflow_deployment_uuid():
 
     result = get_function_name(deployment_config)
 
-    assert result == "57_f_09_beb_b_463_40_e_0_bf_9_e_c_972_e_664352_f"
+    assert result == "57f09bebb46340e0bf9ec972e664352f"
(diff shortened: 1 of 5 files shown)


src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -314,6 +314,6 @@ def get_function_output_type() -> Type:
 
 def get_function_name(function: Tool) -> str:
     if isinstance(function, DeploymentDefinition):
-        return function.deployment
+        return str(function.deployment_id or function.deployment_name)
```
dvargasfuertes: [blocking] deployment id uuid will have dashes, breaking the class name here. likely need to replace dashes with underscores

dvargas92495 (approved):
Why do we need to support deployment id's at all?

src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
vincent0426: I thought backend support both, so we could also support both?

dvargasfuertes: this name is use to generate class names right? I think this class names would be invalid python without a replace or snake_case

vincent0426: Actually, it won't...not sure why but i think it makes sense to use snake case

src/vellum/workflows/nodes/experimental/tool_calling_node/tests/test_utils.py
```diff
@@ -46,4 +46,13 @@ def test_get_function_name_subworkflow_deployment():
...
+    assert result == "my_test_deployment"
+
+
+def test_get_function_name_subworkflow_deployment_uuid():
+    """Test subworkflow deployment with UUID."""
+    deployment_config = DeploymentDefinition(deployment="57f09beb-b463-40e0-bf9e-c972e664352f", release_tag="v1.0.0")
+
+    result = get_function_name(deployment_config)
+
+    assert result == "57_f_09_beb_b_463_40_e_0_bf_9_e_c_972_e_664352_f"
```
dvargasfuertes: yea this is a bit odd... for uuids, i'd either remove or replace the dashes

NgoHarrison (approved):

PR #1898, opened 2025-06-25
[APO-691] Fix tool calling blocks id key transformation

vincent0426:


NgoHarrison (approved):

ee/codegen/src/types/vellum.ts
```diff
@@ -810,6 +810,7 @@ export interface ExecutionCounterWorkflowReference {
 }
 
 export interface DictionaryWorkflowReferenceEntry {
+  id?: string;
```
dvargasfuertes: [nit] Let's leave a comment about how this is here temporarily to support prompt input_variables for now, and that in the future, we won't need to have prompt attributes reference each other like this once we clean up prompt inputs from the node

dvargas92495 (approved):

PR #1901, opened 2025-06-26
Fix inline workflow type

vincent0426:


ee/codegen/src/__test__/nodes/tool-calling-node.test.ts
```diff
@@ -327,6 +327,69 @@ describe("ToolCallingNode", () => {
       expect(await writer.toStringFormatted()).toMatchSnapshot();
     });
   });
+  describe("inline workflow", () => {
+    const createFunctionAttribute = <T>(
```
vincent0426, line 331: will refactor this in the following PR

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -321,16 +323,19 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
...
-        new NodeDefinitionGenerationError(
-          "Workflow definition name is required for inline workflows",
-          "WARNING"
-        )
-      );
-      return "inline_workflow_function";
+    if (definition?.name) {
+      return toPythonSafeSnakeCase(definition.name);
+    }
+    if (workflow.name) {
```
vincent0426, line 329: FE does not have definition at first so it will fallback to name

ee/codegen/src/__test__/nodes/tool-calling-node.test.ts
dvargasfuertes: [nit] yea we typically like to keep these in a factories file and call them a factory

dvargas92495 (approved):

NgoHarrison (approved):

PR #1911, opened 2025-06-26
Codegen generic node output display

vincent0426:
Screenshot 2025-06-26 at 9 50 10 AM

don't think it's possible to move all output display into `base.ts` now since we didn't group outputs?
https://github.com/vellum-ai/vellum-python-sdks/blob/19a13b75c21471f8349a83eaf716b10076ec64ba/ee/codegen/src/types/vellum.ts#L617-L620


dvargas92495 (approved):
> don't think it's possible to move all output display into base.ts now since we didn't group outputs?

We won't be able until we have base node generation figured out. I'd like to move all ids out of display files before we do that

PR #1919, opened 2025-06-27
[APO-558] Change pure function to use file

vincent0426:


src/vellum/workflows/nodes/experimental/tool_calling_node/utils.py
```diff
@@ -251,7 +251,12 @@ def execute_inline_workflow_function(self) -> BaseNode.Outputs:
         )
     else:
         # For regular functions, use CodeExecutionNode approach
-        function_source = inspect.getsource(function)
+        # function tool must be put in another file (no need to have the same name)
+        source_path = inspect.getmodule(function)
+        if source_path is not None:
+            function_source = inspect.getsource(source_path)
+        else:
+            function_source = f"<source code not available for {function.__name__}>"
```
dvargasfuertes, line 259: We should opt to add to workflow context errors instead of return uncompileable python, bc it can prevent the whole workflow from initializing. We'll fast follow here

dvargas92495 (approved):
Gonna merge and fast follow the feedback

PR #1938, opened 2025-06-30
Add better error message for push validation errors on release tags

NgoHarrison:
Screenshot 2025-06-30 at 2 19 33 PM

I am gonna do a fast follow to address the actual issue

PR #1939, opened 2025-06-30
[APO-725] Fix release tags click command typing

NgoHarrison:
https://github.com/vellum-ai/vellum-python-sdks/pull/1938

Related PR

Context: User was experiencing issues when passing in a release tag and I think having multiple=True and type=list was causing it to get into a weird validation error like:
`Input should be a valid string [type=string_type, input_value=['v', '0', '.', '0', '.',...'t', 'p', 'u', 't', 's'], input_type=list]`

### DIFF
--- a/ee/vellum_cli/tests/test_push.py
+++ b/ee/vellum_cli/tests/test_push.py
@@ -1004,6 +1004,7 @@ def test_push__deploy_with_malformed_release_tags_shows_friendly_validation_erro
     assert "--release-tag tag1 --release-tag tag2" in result.output
 
 
+@pytest.mark.usefixtures("info_log_level")
 def test_push__deploy_with_release_tags_success(mock_module, vellum_client):
     # GIVEN a single workflow configured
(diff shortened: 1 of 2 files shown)


marlenc3 (approved):

ee/vellum_cli/tests/test_push.py
```diff
@@ -996,9 +996,46 @@ def test_push__deploy_with_release_tag_shows_friendly_validation_error(mock_modu
 
     # WHEN calling `vellum workflows push` with --deploy and --release-tag
     runner = CliRunner()
-    result = runner.invoke(cli_main, ["workflows", "push", module, "--deploy", "--release-tag", "v1.0.0-testing"])
+    result = runner.invoke(cli_main, ["workflows", "push", module, "--deploy", "--release-tag", None])  # type: ignore
```
dvargasfuertes, line 999: [nit] Why are we using None here instead of a malformed release tag?

dvargas92495 (commented):
✅

PR #1945, opened 2025-06-30
Update codegen for new repository field

m-abboud:
Remind me do I need to cut a new release to get the codegen server to update?

### DIFF
--- a/ee/codegen/src/generators/nodes/code-execution-node.ts
+++ b/ee/codegen/src/generators/nodes/code-execution-node.ts
@@ -129,19 +129,21 @@ export class CodeExecutionNode extends BaseNode<
                     }),
                     ...(package_.repository
                       ? [
-                        python.methodArgument({
-                          name: 'repository',
-                          value: python.TypeInstantiation.str(package_.repository)
-                        })
-                      ]
-                      : [])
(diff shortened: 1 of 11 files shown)


noanflaherty (approved):
I'm pretty sure we need to cut a release

ee/codegen/src/generators/nodes/code-execution-node.ts
```diff
@@ -127,6 +127,16 @@ export class CodeExecutionNode extends BaseNode<
...
                     }),
+                    ...(package_.repository
+                      ? [
+                          python.methodArgument({
+                            name: "repository",
+                            value: python.TypeInstantiation.str(
+                              package_.repository
+                            ),
+                          }),
+                        ]
```
dvargasfuertes, line 138: [oos] do we have a test left that validates that we're not generating a kwarg when the repository is not passed in?

dvargas92495 (commented):
✅

ee/codegen/src/generators/nodes/code-execution-node.ts
m-abboud: I think there's one without a repository I didn't update

PR #1973, opened 2025-07-09
[APO-757] Create test for subworkflow event emission on tool calling nodes

NgoHarrison:
Actual implementation and unskipping / passing test will come in follow up PR


marlenc3 (approved):
✅

dvargas92495 (approved):

PR #1981, opened 2025-07-10
Allow workflows that have nodes in both used and unused graphs

NgoHarrison:


src/vellum/workflows/workflows/base.py
```diff
@@ -123,13 +124,66 @@ def collect_nodes(graph_item: Union[GraphAttribute, Set[GraphAttribute]]) -> Set
...
+                    filtered_graphs.add(item)
+
+            return filtered_graphs
+
         graph_nodes = collect_nodes(dct.get("graph", set()))
         unused_nodes = collect_nodes(dct.get("unused_graphs", set()))
 
         overlap = graph_nodes & unused_nodes
         if overlap:
+            logger = logging.getLogger(__name__)
```
dvargasfuertes: [nit] we usually initiate this from the top of the file

src/vellum/workflows/workflows/base.py
```diff
@@ -123,13 +124,66 @@ def collect_nodes(graph_item: Union[GraphAttribute, Set[GraphAttribute]]) -> Set
...
+                    if not overlapping_in_graph:
+                        filtered_graphs.add(item)
+                    else:
+                        non_overlapping_nodes = graph_nodes - overlapping_nodes
+                        for node in non_overlapping_nodes:
+                            filtered_graphs.add(node)
+
+                elif isinstance(item, set):
+                    filtered_set = set()
+                    for subitem in item:
```
dvargasfuertes: [nit] should be able to just recurse `filter_overlapping_nodes_from_unused_graphs(item, overlapping_nodes)` in this case

src/vellum/workflows/workflows/base.py
```diff
@@ -123,13 +124,66 @@ def collect_nodes(graph_item: Union[GraphAttribute, Set[GraphAttribute]]) -> Set
...
+                if isinstance(item, Graph):
+                    graph_nodes = set(item.nodes)
+                    overlapping_in_graph = graph_nodes & overlapping_nodes
+
+                    if not overlapping_in_graph:
+                        filtered_graphs.add(item)
+                    else:
+                        non_overlapping_nodes = graph_nodes - overlapping_nodes
+                        for node in non_overlapping_nodes:
+                            filtered_graphs.add(node)
```
dvargasfuertes, line 144: [oos] there's some wonky edge cases that we'll need to consider in the future and will likely want to test more granularly in `test_graph.py`. I think in the future, we should have:
- support a `Graph::remove_node` method that returns a set of new graphs
- support a `Graph::remove_nodes` method that does the same, taking in a set of nodes
- Iterate over `remove_nodes` iterates over the input and removes each node
- `filter_overlapping_nodes_from_unused_graphs` likely is defined within the graph module as `remove_nodes_from_graph_set` or something like that

dvargas92495 (approved):

PR #1984, opened 2025-07-10
Update checks on thinking mode types

NgoHarrison:
To account for new types in thinking vellum value, it is now a list:
`value: typing.List[StringVellumValue]`


dvargas92495 (commented):
Discussed offline - we are looking to revert

PR #1992, opened 2025-07-11
Introduce new test method for codegen and serializing workflows without display directories

NgoHarrison:
Context: Conditional nodes break when we pull a workflow with a conditional node and then push up the same workflow but WITHOUT a display directory (we delete before push). This PR just introduces the test so that we can do the round trip serialization test. I will follow this up with a PR to test conditional nodes specifically

### DIFF
--- a/ee/codegen/src/__test__/project.test.ts
+++ b/ee/codegen/src/__test__/project.test.ts
@@ -177,6 +177,95 @@ describe("WorkflowProjectGenerator", () => {
       }
     );
   });
+  describe.skip("generateCodeNodeDirectoryOnly", () => {
+    const excludeFilesAtPaths: RegExp[] = [/\.pyc$/];
+    const ignoreContentsOfFilesAtPaths: RegExp[] = [];
+
+    it.each(
+      getFixturesForProjectTest({
+        includeFixtures: [""],
+        fixtureMocks: {},
+      })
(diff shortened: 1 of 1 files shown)


ee/codegen/src/__test__/project.test.ts
```diff
@@ -177,6 +177,95 @@ describe("WorkflowProjectGenerator", () => {
...
+    const excludeFilesAtPaths: RegExp[] = [/\.pyc$/];
+    const ignoreContentsOfFilesAtPaths: RegExp[] = [];
+
+    it.each(
+      getFixturesForProjectTest({
+        includeFixtures: [""],
+        fixtureMocks: {},
+      })
+    )(
+      "should correctly generate code without display directory for fixture $fixtureName",
```
dvargasfuertes, line 190: I'm having a hard time grokking the goal here:
- doesn't `includeFixtures` mean that none are included?
- What's the codegen bug we are hoping to catch here? I thought it was a serialization bug

NgoHarrison: Discussed offline - will continue with this approach and after follow up PRs address the bug that kicked this off, will create a true test for symmetric codegen/serialization

dvargas92495 (commented):
✅

PR #1993, opened 2025-07-11
Fix serializtion of conditional nodes without display

NgoHarrison:


ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_conditional_node_serialization.py
```diff
@@ -285,7 +285,7 @@ def test_serialize_workflow():
                     {
                         "id": "de7b0b4e-7803-4d36-a275-2e7e3f60342b",
                         "type": "IF",
-                        "source_handle_id": "561b4e3a-8db3-448a-8933-1115937082ff",
```
dvargasfuertes, line 288: [oos] wow - these have been busted since launch.

piggy backs off of a convo we were having in the elevator. This assertion is stronger if we tested the fields we cared about instead of the full json data

ee/vellum_ee/workflows/display/nodes/vellum/conditional_node.py
```diff
@@ -166,7 +169,11 @@ def serialize_rule(
             rule_group_id = str(
                 condition_ids[idx].rule_group_id if condition_ids else uuid4_from_hash(f"{condition_id}|rule_group")
             )
-            source_handle_id = str(source_handle_ids.get(idx) or uuid4_from_hash(f"{node_id}|handles|{idx}"))
+            source_handle_id = (
+                str(source_handle_ids.get(idx))
+                if source_handle_ids.get(idx)
+                else cast(JsonObject, ports_by_index[idx])["id"]
```
dvargasfuertes: [nit] what if, for some reason, `ports_by_index[idx]` throws a KeyError? feels like we should do the three stage check just to be safe where we default to the previous default. May be simpler with three variants to just helper method this instead of continuing the ternary

dvargas92495 (approved):

PR #2021, opened 2025-07-14
Add parameters for tool calling node

vincent0426:


src/vellum/workflows/nodes/displayable/tool_calling_node/node.py
```diff
@@ -1,13 +1,15 @@
...
 from vellum import ChatMessage, PromptBlock
+from vellum.client.types.prompt_parameters import PromptParameters
 from vellum.workflows.context import execution_context, get_parent_context
 from vellum.workflows.errors.types import WorkflowErrorCode
 from vellum.workflows.events.workflow import is_workflow_event
 from vellum.workflows.exceptions import NodeException
 from vellum.workflows.graph.graph import Graph
 from vellum.workflows.inputs.base import BaseInputs
 from vellum.workflows.nodes.bases import BaseNode
+from vellum.workflows.nodes.displayable.bases.inline_prompt_node.constants import DEFAULT_PROMPT_PARAMETERS
```
dvargasfuertes, line 12: [nit] let's move this to some prompt-related constants file so that we're not doing cross-node imports

dvargasfuertes: eh, I guess we already have cross-node imports in `utils.py` bc toolcallingnode has some other nodes in dynamically extends under the hood. I guess the ask here is to just move this constant into some not node related file since it's prompt-specific not one workflow node class' specific

dvargas92495 (approved):

PR #2023, opened 2025-07-14
[APO-845] Fix serialization for optional input

vincent0426:
Will fix state variables too

### DIFF
--- a/ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
+++ b/ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
@@ -136,13 +136,19 @@ class BaseWorkflowDisplay(Generic[WorkflowType]):
                 if workflow_input_reference.instance
                 else None
             )
+
+            # An input is required if it has no default value AND is not optional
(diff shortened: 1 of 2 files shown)


ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -136,13 +136,19 @@ def serialize(self) -> JsonObject:
...
+            is_required = not has_default and not is_optional
+
             input_variables.append(
                 {
                     "id": str(workflow_input_display.id),
                     "key": workflow_input_display.name or workflow_input_reference.name,
                     "type": infer_vellum_variable_type(workflow_input_reference),
                     "default": default.dict() if default else None,
-                    "required": workflow_input_reference.instance is undefined,
+                    "required": is_required,
```
dvargasfuertes, line 151: [oos] Nice. I think we likely need to similarly update our logic for output and state variables too. Bonus points for consolidation there

dvargas92495 (approved):

PR #2037, opened 2025-07-15
[APO-485] Support timeout attribute on APINode

alex-nork:
The purpose of this PR is to add a `timeout` attribute to the `APINode`. This addresses a user-reported issue in which execution of their `APINode` was hanging.


src/vellum/workflows/nodes/displayable/bases/api_node/node.py
```diff
@@ -79,25 +85,36 @@ def _local_execute_api(self, data, headers, json, method, url):
...
         )
 
-    def _vellum_execute_api(self, bearer_token, data, headers, method, url):
+    def _vellum_execute_api(self, bearer_token, data, headers, method, url, timeout):
         client_vellum_secret = ClientVellumSecret(name=bearer_token.name) if bearer_token else None
+
+        # Create request_options if timeout is specified
+        request_options = None
+        if timeout is not None:
+            request_options = RequestOptions(timeout_in_seconds=timeout)
```
dvargasfuertes, line 108: oh sick - I was worried that we'd have to pipe this through to Vellum's backend but this is better

tests/workflows/basic_api_node_with_timeout/test_timeout.py
```diff
@@ -0,0 +1,70 @@
+from unittest.mock import Mock, patch
+
+from vellum.workflows.constants import APIRequestMethod
+from vellum.workflows.nodes.displayable.bases.api_node.node import BaseAPINode
+
+
+class TestAPINodeTimeout:
```
dvargasfuertes: [nit] let's move this test over to https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/nodes/displayable/api_node/tests/test_api_node.py

`tests/workflows` are intended for full workflows that we'd like to test at the workflow level, calling `workflow.run()` or `workflow.stream()

tests/workflows/basic_api_node_with_timeout/test_timeout.py
```diff
@@ -0,0 +1,70 @@
...
+        mock_response.json.return_value = {"result": "success"}
+        mock_response.headers = {"content-type": "application/json"}
+        mock_response.status_code = 200
+        mock_response.text = '{"result": "success"}'
+        mock_session_instance.send.return_value = mock_response
+
+        node: BaseAPINode = BaseAPINode()
+        node.url = "https://example.com"
+
+        node._local_execute_api(
```
dvargasfuertes: [nit] Instead of calling private methods in tests, could we run `node.run()`? testing private methods makes it more inflexible for us to make changes on non-user facing logic

tests/workflows/basic_api_node_with_timeout/test_timeout.py
```diff
@@ -0,0 +1,70 @@
...
+        node.timeout = 25
+
+        # Mock the _run method to verify timeout is passed
+        with patch.object(node, "_run") as mock_run:
+            mock_run.return_value = Mock()
+
+            node.run()
+
+            # Verify _run was called with timeout
+            mock_run.assert_called_once_with(
```
dvargasfuertes: [nit] if we do the above suggestion, then I think the more informative test in this case would be verifying the vellum API case, where we have an assertion validating the right RequestOptions was created and passed in

dvargas92495 (approved):
Implementation looks good! Have some testing suggestions that could be done now or as a fast follow

tests/workflows/basic_api_node_with_timeout/test_timeout.py
```diff
@@ -0,0 +1,70 @@
...
+    def _setup_vellum_mocks(self, node):
+        """Helper method to set up common vellum client mocks."""
+        mock_context = Mock()
+        mock_vellum_client = Mock()
+        mock_vellum_response = Mock()
+        mock_vellum_response.json_ = {"result": "success"}
+        mock_vellum_response.headers = {"content-type": "application/json"}
+        mock_vellum_response.status_code = 200
+        mock_vellum_response.text = '{"result": "success"}'
+        mock_vellum_client.execute_api.return_value = mock_vellum_response
```
NgoHarrison: [nit] We also already have a fixture for mocking the vellum client:
https://github.com/vellum-ai/vellum-python-sdks/blob/bbe74826bc55034b3db0b0b723172f39edebbce7/conftest.py

tests/workflows/basic_api_node_with_timeout/test_timeout.py
```diff
@@ -0,0 +1,70 @@
...
+        node._local_execute_api(
+            data=None,
+            headers={},
+            json={"test": "data"},
+            method=APIRequestMethod.POST,
+            url="https://example.com",
+            timeout=30,
+        )
+
+        # Verify timeout was passed to session.send
```
vincent0426: We would like to follow a pattern `GIVEN`, `WHEN`, `AND`, `THEN` as our first verb in our test comments!

vincent0426 (approved):

NgoHarrison (approved):

tests/workflows/basic_api_node_with_timeout/test_timeout.py
alex-nork: right 👍 should be focused on the public interface

PR #2040, opened 2025-07-15
[APO-485] Add timeout serialization support to APINode

alex-nork:
The purpose of this PR is to include `timeout` in the `APINode`'s serialized `attributes` field. This is the second instance of a node implementing custom attribute serialization, so the logic has been moved out of `BaseInlinePromptNodeDisplay` and into `BaseNodeDisplay` to avoid duplication.


ee/vellum_ee/workflows/display/nodes/vellum/api_node.py
```diff
@@ -31,6 +31,19 @@ class BaseAPINodeDisplay(BaseNodeDisplay[_APINodeType], Generic[_APINodeType]):
         APINode.authorization_type,
     }
 
+    # TODO: Only serialize attributes not inputs
```
dvargasfuertes, line 34: [oos] Yea for context, our long term goal is to eventually get all of these out of this array and as attributes. We've made good progress on this in the prompt nodes case

dvargas92495 (approved):
good stuff 🔥

PR #2059, opened 2025-07-16
[APO-485] Include attributes field in APINodeSerializer schema

alex-nork:
The purpose of this PR is to include the `attributes` field in the APINodeSerializer schema so that we actually use it when generating the node's python code.


NgoHarrison (approved):

alex-nork:
test fix incoming

dvargas92495 (approved):
good catch

NgoHarrison (approved):

PR #2067, opened 2025-07-18
[APO-834] Add input monitoring for tool router node

vincent0426:
Screenshot 2025-07-17 at 2 56 09 PM
Screenshot 2025-07-17 at 2 56 25 PM

### DIFF
--- a/src/vellum/workflows/expressions/concat.py
+++ b/src/vellum/workflows/expressions/concat.py
@@ -1,4 +1,4 @@
-from typing import Generic, Sequence, TypeVar
+from typing import Generic, Sequence, TypeVar, Union
 
 from vellum.workflows.descriptors.base import BaseDescriptor
 from vellum.workflows.descriptors.utils import resolve_value
@@ -12,8 +12,8 @@ class ConcatExpression(BaseDescriptor[list], Generic[LHS, RHS]):
     def __init__(
         self,
         *,
-        lhs: Sequence[LHS],
-        rhs: Sequence[RHS],
+        lhs: Union[BaseDescriptor[Sequence[LHS]], Sequence[LHS]],
+        rhs: Union[BaseDescriptor[Sequence[RHS]], Sequence[RHS]],
     ) -> None:
         super().__init__(name=f"{lhs} + {rhs}", types=(list,))
         self._lhs = lhs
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/expressions/concat.py
```diff
@@ -0,0 +1,25 @@
...
+        *,
+        lhs: Union[BaseDescriptor[Sequence[LHS]], Sequence[LHS]],
+        rhs: Union[BaseDescriptor[Sequence[RHS]], Sequence[RHS]],
+    ) -> None:
+        super().__init__(name=f"{lhs} + {rhs}", types=(list,))
+        self._lhs = lhs
+        self._rhs = rhs
+
+    def resolve(self, state: "BaseState") -> list:
+        lval = resolve_value(self._lhs, state) or []
```
dvargasfuertes: [nit] This supports `0.concat([...])` and `false.concat([...])`. I believe we should validate that each side is `isinstance(list)` instead of coalescing here

dvargas92495 (approved):
[oos] One improvement we should look into on the vellum side is to see how we could render the chat history's as the chat history ui instead of the raw JSON

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -234,6 +231,14 @@ def create_port_condition(fn_name):
             )
         )
 
+    node_prompt_inputs = {
+        **(prompt_inputs or {}),
+        CHAT_HISTORY_VARIABLE: ConcatExpression(
+            lhs=(prompt_inputs or {}).get(CHAT_HISTORY_VARIABLE, []),
+            rhs=ToolCallingState.chat_history,
+        ),
```
dvargasfuertes, line 239: If we make the suggested change from above, then we should use a ternary here instead of always concatting:
```python
        CHAT_HISTORY_VARIABLE: ConcatExpression(
            lhs=prompt_inputs[CHAT_HISTORY_VARIABLE],
            rhs=ToolCallingState.chat_history,
        ) if prompt_inputs.get(CHAT_HISTORY_VARIABLE) else ToolCallingState.chat_history,
```

vincent0426: I think we couldn't assign `ToolCallingState.chat_history` here without using a new list, otherwise it will lead to the deepcopy issue with inputs which will have same history for every events

vincent0426: `lhs=(prompt_inputs or {}).get(CHAT_HISTORY_VARIABLE, []),` should be safe to pass an empty list?

PR #2068, opened 2025-07-18
[APO-868] Introduce ComposioService

alex-nork:
The purpose of this PR is to introduce a basic `ComposioService` class that interacts with Composio's the `composio-client` python packages. This will be followed up to include their `composio-core` package, hence the facade pattern.

### DIFF
--- a/src/vellum/workflows/nodes/displayable/tool_calling_node/composio_service.py
+++ b/src/vellum/workflows/nodes/displayable/tool_calling_node/composio_service.py
@@ -28,11 +28,11 @@ class ComposioAccountService:
     """Manages user authorized connections using composio-client"""
 
     def __init__(self, api_key: str):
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/nodes/displayable/tool_calling_node/composio_service.py
```diff
@@ -0,0 +1,48 @@
+from dataclasses import dataclass
+from typing import List
+
+from composio_client import Composio as ComposioClient
```
dvargasfuertes, line 4: [nit] I can see this file living in a new `src/vellum/workflows/integrations` module - if the ToolCallingNode use case goes well, we've also talked about creating a first class composio node

dvargas92495 (approved):

PR #2070, opened 2025-07-19
Fix tool calling node display to match nodes folder structure

vincent0426:
### Display folder
Current vs After

Screenshot 2025-07-30 at 11 25 58 AM
Screenshot 2025-07-18 at 10 57 10 PM

<br/>

### Node folder

Screenshot 2025-07-18 at 10 57 19 PM

### DIFF
--- a/ee/codegen/src/generators/workflow.ts
+++ b/ee/codegen/src/generators/workflow.ts
@@ -754,10 +754,16 @@ class WorkflowDisplayFile extends BasePersistedFile {
 
   public getModulePath(): string[] {
     if (this.workflowContext.parentNode) {
-      return [
-        ...this.workflowContext.parentNode.getNodeDisplayModulePath(),
-        GENERATED_WORKFLOW_MODULE_NAME,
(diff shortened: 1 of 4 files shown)


ee/codegen/src/generators/workflow.ts
```diff
@@ -754,10 +754,16 @@ class WorkflowDisplayFile extends BasePersistedFile {
 
   public getModulePath(): string[] {
     if (this.workflowContext.parentNode) {
-      return [
-        ...this.workflowContext.parentNode.getNodeDisplayModulePath(),
-        GENERATED_WORKFLOW_MODULE_NAME,
-      ];
+      return this.workflowContext.nestedWorkflowModuleName
```
dvargasfuertes, line 757: [oos] I would expect a workflow context with a parent node would always have a nested workflow module name. not worth debugging now, but worth watching and seeing if we could cleanup in the future

dvargas92495 (approved):

PR #2073, opened 2025-07-20
[APO-890] Add ComposioToolDefinition type

alex-nork:
This PR introduces the foundational `ComposioToolDefinition` type that enables Composio tools to be used as first-class functions in Vellum workflows. This is the first step in a three-part implementation to fully support Composio tool integration.


src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -290,6 +290,21 @@ def create_function_node(
 
         return node
 
+    elif isinstance(function, ComposioToolDefinition):
+        # ComposioToolDefinition execution not yet implemented
+        def composio_not_implemented(**kwargs):
+            raise NotImplementedError("ComposioToolDefinition execution not yet implemented")
```
alex-nork, line 296: This will be implemented in the follow up to this PR.

dvargas92495 (approved):

PR #2075, opened 2025-07-21
[APO-865] Add runtime execution support for ComposioToolDefinition

alex-nork:
This PR builds on #2073 to add runtime execution support for `ComposioToolDefinition` objects. This will allow Composio tools to be executed with proper function call handling.


src/vellum/workflows/nodes/displayable/tool_calling_node/tests/test_composio_node.py
```diff
@@ -0,0 +1,123 @@
...
+        toolkit="GITHUB", action="GITHUB_CREATE_AN_ISSUE", description="Create a new issue in a GitHub repository"
+    )
+
+
+@pytest.fixture
+def mock_state():
+    """Fixture for a mock ToolCallingState."""
+    state = Mock(spec=ToolCallingState)
+    state.chat_history = Mock()
+    return state
```
dvargasfuertes: [nit] In general, we try to only mock external deps (eg api requests), and reuse our source code where possible.

so alternatives available, in order of increasing preference:
- Our test just instantiates `ComposioNode(state=ToolCallingState)` in the test.
- we test `create_function_node` instead, which captures what we're mocking
- we test a workflow that uses a ToolCallingNode with a Composio tool definition just as a user would define it

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -31,6 +33,30 @@
 CHAT_HISTORY_VARIABLE = "chat_history"
 
 
+class FunctionCallNodeMixin:
+    """Mixin providing common functionality for nodes that handle function calls."""
+
+    function_call_output: List[PromptOutput]
+    state: Any  # Will be properly typed by the inheriting node class
+
+    def extract_function_arguments(self) -> dict:
```
dvargasfuertes: [nit] let's make these private methods

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -31,6 +33,30 @@
 CHAT_HISTORY_VARIABLE = "chat_history"
 
 
+class FunctionCallNodeMixin:
+    """Mixin providing common functionality for nodes that handle function calls."""
+
+    function_call_output: List[PromptOutput]
+    state: Any  # Will be properly typed by the inheriting node class
```
dvargasfuertes: [blocking] I don't believe we need to define this? Am worried about some inference magic we do if this is defined as an `Any`

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -69,20 +95,11 @@ def run(self) -> Iterator[BaseOutput]:
             yield output
 
 
-class DynamicSubworkflowDeploymentNode(SubworkflowDeploymentNode[ToolCallingState]):
+class DynamicSubworkflowDeploymentNode(FunctionCallNodeMixin, SubworkflowDeploymentNode[ToolCallingState]):
```
dvargasfuertes: [nit] Let's make these mixins the second class passed in. We _might_ have logic in our serialization layer that is doing stuff using `__bases__[0]` just in case

dvargas92495 (commented):
Implementation looks great! Had a note on the tests and there may be a blocking issue with the Mixin. Though I vibe with the DRY approach the Mixin brings

PR #2077, opened 2025-07-21
Removed Thinking Output from Prompt Node Text View

meikemagdalena:


dvargas92495 (approved):
we'll need to update `src/vellum/workflows/nodes/displayable/prompt_deployment_node/node.py` too

PR #2083, opened 2025-07-21
[APO-891] Support literal type for tool calling node

vincent0426:


src/vellum/workflows/utils/functions.py
```diff
@@ -33,6 +33,14 @@ def compile_annotation(annotation: Optional[Any], defs: dict[str, Any]) -> dict:
     if get_origin(annotation) is Union:
         return {"anyOf": [compile_annotation(a, defs) for a in get_args(annotation)]}
 
+    if get_origin(annotation) is Literal:
+        values = list(get_args(annotation))
+        types = {type(value) for value in values}
+        if len(types) == 1:
+            return {"type": type_map[types.pop()], "enum": values}
```
dvargasfuertes: [blocking] we should ensure that `types.pop() in type_map` to avoid the KeyError here. I think this currently will internal server error with `Literal[SomeEnum.foo]`

dvargas92495 (approved):
quick blocker

PR #2084, opened 2025-07-22
[APO-866] Serialization support for Composio tool definition

alex-nork:
This PR was initially meant to add explicit serialization support for the `ComposioToolDefinition`, but we seem to get that for free [here](https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/vellum_ee/workflows/display/utils/expressions.py#L369). So now it only includes a test of the serialization.


dvargas92495 (approved):
yea let's do it

NgoHarrison (commented):
✅

PR #2090, opened 2025-07-23
New Wrapper Decorator for Execution Context + Tests

marlenc3:
this is first pass + mostly done 

started to try and play with some dynamic initialization in tests, but removable. 

wanted to get a yay/nay on pattern 

hopes are to remove all wtih execution_context() calls that are surrounding our sdk


src/vellum/workflows/monitoring/decorators.py
```diff
@@ -0,0 +1,214 @@
+"""Monitoring decorators for workflow tracing."""
```
dvargasfuertes: [nit] `src/vellum/workflows/events` is basically our monitoring module, would be nice to colocate there

src/vellum/workflows/monitoring/decorators.py
```diff
@@ -0,0 +1,214 @@
...
+    MonitoringExecutionContext,
+    _monitoring_context_store,
+    get_monitoring_execution_context,
+    set_monitoring_execution_context,
+)
+from vellum.workflows.types.definition import CodeResourceDefinition
+
+
+class MonitorDecorator:
+    """Self-sufficient monitoring decorator using wrapt."""
```
dvargasfuertes: [q] what does `wrapt` give us?

src/vellum/workflows/monitoring/tests/workflow.py
```diff
@@ -0,0 +1,137 @@
...
+        final_value = MapFruitsNode.Outputs.count
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        # Production-style: Top-level workflow initializes monitoring for all components
+        if not SimpleMapExample._monitoring_initialized:
+            self._initialize_monitoring_dynamic()
+            SimpleMapExample._monitoring_initialized = True
+
+    def _initialize_monitoring_dynamic(self):
```
dvargasfuertes: tests should show what the _user_ is defining. so this feels like it should be moved to `BaseWorkflow`

src/vellum/workflows/monitoring/tests/workflow.py
```diff
@@ -0,0 +1,137 @@
...
+        """Dynamically discover and wrap all execution methods with proper type classification. -- TESTING ONLY."""
+
+        # Classes to monitor with their proper types
+        workflow_classes = [SimpleMapExample, IterationSubworkflow]
+        node_classes = [MapFruitsNode, Iteration]
```
dvargasfuertes: these should be able to done dynamically and recursively based on the `graph` attribute. In fact, we have some get nodes helpers that already exist

src/vellum/workflows/monitoring/tests/workflow.py
```diff
@@ -0,0 +1,137 @@
...
+                        wrapped_methods.add(monitor_name)
+
+        # Successfully wrapped all discovered execution methods
+
+    def _get_execution_methods(self, cls: Type[Any]) -> List[str]:
+        """Get list of methods that should be monitored for a class."""
+        execution_methods = []
+
+        # Standard execution methods
+        standard_methods = ["run", "stream", "_run_subworkflow", "_context_run_subworkflow"]
```
dvargasfuertes: Thinking we may want to define these on `BaseNode` and `BaseWorkflow` instead of hardcoding within the helper.  We should also pick one subworkflow-running method to wrap and standardize our nodes around it

src/vellum/workflows/monitoring/tests/workflow.py
```diff
@@ -0,0 +1,137 @@
...
+    def _initialize_monitoring_dynamic(self):
+        """Dynamically discover and wrap all execution methods with proper type classification. -- TESTING ONLY."""
+
+        # Classes to monitor with their proper types
+        workflow_classes = [SimpleMapExample, IterationSubworkflow]
+        node_classes = [MapFruitsNode, Iteration]
+
+        wrapped_methods = set()
+
+        # Wrap workflow classes
```
dvargasfuertes: these next two blocks are the good stuff and seems like the main goal of this setup work we want to do before introducing otel emissions. It's my impression that we could do this and still use our existing execution context class's, then iterate to add otel logic to it

src/vellum/workflows/monitoring/tests/workflow.py
```diff
@@ -0,0 +1,137 @@
...
+    graph = MapFruitsNode
+    _monitoring_initialized = False  # Class-level flag for the entire workflow system
+
+    class Outputs(BaseOutputs):
+        final_value = MapFruitsNode.Outputs.count
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        # Production-style: Top-level workflow initializes monitoring for all components
+        if not SimpleMapExample._monitoring_initialized:
```
dvargasfuertes: [nit] `_monitoring_initialized = True` skips monitoring for my workflow which is a little unintuitive. Would expect opt-in

dvargasfuertes: making it opt-in also allows us to add it to BaseWorkflow

dvargas92495 (commented):
Available to sync this morning if helpful!

src/vellum/workflows/monitoring/decorators.py
marlenc3: it's a cleaner way to implement decorators for the most part -- and it keeps the call signature intact
saw it in the otel packages and thought to give it a whirl

src/vellum/workflows/monitoring/tests/workflow.py
marlenc3: ya -- I was playing around with different ways to initialize the wrappers and settled on the below

I'll refactor to use ExecutionContext 👍

src/vellum/workflows/workflows/base.py
```diff
@@ -211,6 +212,8 @@ class BaseWorkflow(Generic[InputsType, StateType], metaclass=_BaseWorkflowMeta):
     unused_graphs: ClassVar[Set[GraphAttribute]]  # nodes or graphs that are defined but not used in the graph
     emitters: List[BaseWorkflowEmitter]
     resolvers: List[BaseWorkflowResolver]
+    _enable_monitoring: ClassVar[bool] = True
```
marlenc3: need to turn this off before merging @marlenc3

src/vellum/workflows/events/decorators.py
```diff
@@ -0,0 +1,338 @@
...
+    ) -> ParentContext:
+        """Create new monitoring parent context for this call."""
+        # Use current monitoring context as parent if it has one
+        if current_monitoring.parent_context:
+            # Convert current monitoring's parent context into the parent for this new context
+            # This creates proper nesting: new_method -> current_method -> parent_method
+            current_parent = current_monitoring.parent_context
+        else:
+            # If we have a workflow_class and no current parent, create workflow context as root
+            current_parent = None
```
marlenc3, line 122: I actualy dont know if we really need this guy tbh

src/vellum/workflows/events/decorators.py
```diff
@@ -0,0 +1,338 @@
...
+                    name=f"{self.workflow_class.__name__}.run",  # Include the run method name
+                    module=(
+                        [self.workflow_class.__module__]
+                        if hasattr(self.workflow_class, "__module__")
+                        else ["workflows"]
+                    ),
+                )
+                current_parent = WorkflowParentContext(
+                    span_id=uuid4(),
+                    type="WORKFLOW",
```
dvargasfuertes, line 135: [nit] redundant

src/vellum/workflows/events/decorators.py
```diff
@@ -0,0 +1,338 @@
...
+            # Convert current monitoring's parent context into the parent for this new context
+            # This creates proper nesting: new_method -> current_method -> parent_method
+            current_parent = current_monitoring.parent_context
+        else:
+            # If we have a workflow_class and no current parent, create workflow context as root
+            current_parent = None
+            if hasattr(self, "workflow_class") and self.workflow_class:
+                workflow_resource_def = CodeResourceDefinition(
+                    id=uuid4(),
+                    name=f"{self.workflow_class.__name__}.run",  # Include the run method name
```
dvargasfuertes: [blocking] Wait why are we including the method name in the parent context's workflow definition?

src/vellum/workflows/events/decorators.py
dvargasfuertes: [nit] can we remove until we have a test that proves that we need it?

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -266,6 +267,8 @@ class BaseNode(Generic[StateType], ABC, metaclass=BaseNodeMeta):
     state: StateType
     _context: WorkflowContext
     _inputs: MappingProxyType[NodeReference, Any]
+    _enable_monitoring: ClassVar[bool] = False
+    _monitoring_initialized: ClassVar[bool] = False
```
dvargasfuertes, line 270: [blocking] do we need these on nodes? I thought our monitoring logic would handle all of these by attaching themselves to the method automatically?

src/vellum/workflows/workflows/base.py
```diff
@@ -211,6 +212,8 @@ class BaseWorkflow(Generic[InputsType, StateType], metaclass=_BaseWorkflowMeta):
     unused_graphs: ClassVar[Set[GraphAttribute]]  # nodes or graphs that are defined but not used in the graph
     emitters: List[BaseWorkflowEmitter]
     resolvers: List[BaseWorkflowResolver]
+    _enable_monitoring: ClassVar[bool] = False
+    _monitoring_initialized: ClassVar[bool] = False
```
dvargasfuertes, line 216: [blocking] do we need both?

dvargas92495 (commented):
Gonna leave some top level comments on https://github.com/vellum-ai/vellum-python-sdks/pull/2091

src/vellum/workflows/workflows/base.py
marlenc3: my thought is if a user re-uses a node/workflow, we shouldn't rewrap it — but we can also just check that from the decorator helpers. can move 👍

dvargas92495:
Gonna wait to re-review this one until the base one is merged into main and once you notify that it's ready, which can be done with the button below!

Screenshot 2025-07-25 at 8 40 13 AM

PR #2091, opened 2025-07-23
New Multi-Threaded Context Handler

marlenc3:
*** allows for us to store context for threads without having to use with context manager shenanigans

goal -- any big concerns with this pattern? I have one currently, and its for large workflows, and if it'll take up more memory to keep track of everything

I may introduce some cleanups at the end of spans


src/vellum/workflows/monitoring/context.py
```diff
@@ -0,0 +1,206 @@
...
+                return self._contexts[trace_key]
+
+            # Global current trace_id fallback for worker threads
+            if trace_id == default_uuid and self._current_trace_id:
+                global_trace_key = str(self._current_trace_id)
+                if global_trace_key in self._contexts:
+                    return self._contexts[global_trace_key]
+
+            # Look for any recent context as last resort
+            if self._thread_contexts:
```
marlenc3: gunna remove this

src/vellum/workflows/monitoring/context.py
```diff
@@ -0,0 +1,206 @@
...
+    try:
+        context = getattr(_monitoring_execution_context, "context", None)
+        if context is not None:
+            return context
+    except AttributeError:
+        pass
+
+    # Try to auto-create from execution context
+    try:
+        exec_ctx = get_execution_context()
```
marlenc3: hope is to not need execution_context here

src/vellum/workflows/monitoring/context.py
```diff
@@ -0,0 +1,206 @@
...
+from vellum.workflows.context import get_execution_context
+from vellum.workflows.events.types import ParentContext
+
+
+class MonitoringExecutionContext(UniversalBaseModel):
+    """Monitoring-specific execution context that runs in parallel with ExecutionContext."""
+
+    trace_id: UUID = field(default_factory=lambda: UUID("00000000-0000-0000-0000-000000000000"))
+    parent_context: Optional[ParentContext] = None
+    monitoring_enabled: bool = True
```
dvargasfuertes: do we need to introduce _another_ context object in order to do this forked path and maintain both? especially since `monitoring_enabled` should never be false here correct?

dvargasfuertes: Gonna do the rest of the review on the other PR, I think we could save needing to introduce this one

marlenc3: 👍  I'm like 1/5 for a new one, it's just easier to test/compare, and was thinking it'd be easier to roll out in peices
I can refactor out MonitoringExecutionContext -- but we'll still need _MonitoringContextStore for threading

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+    trace_id = _monitoring_context_store.get_current_trace_id()
+    if trace_id:
+        default_uuid = UUID("00000000-0000-0000-0000-000000000000")
+        if trace_id != default_uuid:
+            parent_context = _monitoring_context_store.retrieve_context(trace_id, None)
+            if parent_context:
+                return ExecutionContext(
+                    trace_id=trace_id,
+                    parent_context=parent_context,
+                    monitoring_enabled=True,
```
dvargasfuertes: [nit] in neither of the two PRs, I'm not seeing any reads of this field, hinting that we may not need it

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+            _monitoring_context_store.store_context(context.trace_id, context.parent_context)
+
+
+@contextmanager
+def monitoring_execution_context(
+    parent_context: Optional[ParentContext] = None, trace_id: Optional[UUID] = None, monitoring_enabled: bool = True
+) -> Iterator[None]:
+    """Context manager for handling monitoring execution context - self-contained monitoring system."""
+    prev_context = get_monitoring_execution_context()
+    default_uuid = UUID("00000000-0000-0000-0000-000000000000")
```
dvargasfuertes: [nit] we should have this as a constant defined somewhere globally

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+    # This prevents overwriting deeper contexts when restoring shallower ones
+    if context.trace_id and context.parent_context:
+        default_uuid = UUID("00000000-0000-0000-0000-000000000000")
+        if context.trace_id != default_uuid:
+            _monitoring_context_store.store_context(context.trace_id, context.parent_context)
+
+
+@contextmanager
+def monitoring_execution_context(
+    parent_context: Optional[ParentContext] = None, trace_id: Optional[UUID] = None, monitoring_enabled: bool = True
```
dvargasfuertes: [oos] I'd be interested to see if there was a way to avoid the kwargs and go off of just `get_monitoring_execution_context` (or vice versa). Optimization for later

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+
+class _MonitoringContextStore:
+    """Thread-safe storage for monitoring contexts."""
+
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._contexts: Dict[str, ParentContext] = {}
+        self._thread_contexts: Dict[int, ParentContext] = {}
+        self._current_trace_id: Optional[UUID] = None
+        # self._current_span_id: Optional[UUID] = None # Removed as per edit hint
```
dvargasfuertes: [nit] do we need this?

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+        with self._lock:
+            return self._contexts.get(span_key)
+
+    def retrieve_context(
+        self, trace_id: UUID, current_parent_context: Optional[ParentContext] = None
+    ) -> Optional[ParentContext]:
+        """Retrieve monitoring parent context with multiple fallback strategies."""
+        thread_id = threading.get_ident()
+        if current_parent_context:
+            pass  # Removed debug output
```
dvargasfuertes: [nit] this is a no-op?

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+def set_monitoring_execution_context(context: ExecutionContext) -> None:
+    """Set the current monitoring execution context and persist it for cross-boundary access."""
+    _monitoring_execution_context.context = context
+
+    # Only store in global store if this context has a parent_context (i.e., it's not a default/restored context)
+    # This prevents overwriting deeper contexts when restoring shallower ones
+    if context.trace_id and context.parent_context:
+        default_uuid = UUID("00000000-0000-0000-0000-000000000000")
+        if context.trace_id != default_uuid:
+            _monitoring_context_store.store_context(context.trace_id, context.parent_context)
```
dvargasfuertes: [nit] why not just ` _monitoring_context_store.store_context(context)`

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+    def retrieve_context(
+        self, trace_id: UUID, current_parent_context: Optional[ParentContext] = None
+    ) -> Optional[ParentContext]:
+        """Retrieve monitoring parent context with multiple fallback strategies."""
+        thread_id = threading.get_ident()
+        if current_parent_context:
+            pass  # Removed debug output
+
+        with self._lock:
+            # Removed debug output
```
dvargasfuertes: [nit] remove?

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+    def retrieve_context_by_span_id(self, span_id: UUID) -> Optional[ParentContext]:
+        """Retrieve monitoring parent context directly by span_id."""
+        span_key = f"span:{str(span_id)}"
+        with self._lock:
+            return self._contexts.get(span_key)
+
+    def retrieve_context(
+        self, trace_id: UUID, current_parent_context: Optional[ParentContext] = None
+    ) -> Optional[ParentContext]:
+        """Retrieve monitoring parent context with multiple fallback strategies."""
```
dvargasfuertes, line 67: [q] why do we need multiple fallback strategies?

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+
+        with self._lock:
+            # Removed debug output
+
+            # Try span_id if we have current parent context (most specific)
+            if current_parent_context and hasattr(current_parent_context, "span_id"):
+                span_key = f"span:{str(current_parent_context.span_id)}"
+                if span_key in self._contexts:
+                    result = self._contexts[span_key]
+                    return result
```
dvargasfuertes: [nit] shouldn't this _always_ be itself? makes me additionally confused as to why `current_parent_context` is a kwarg

src/vellum/workflows/events/context.py
```diff
@@ -0,0 +1,199 @@
...
+            # Use trace:span:thread for unique context storage
+            trace_span_thread_key = f"trace:{str(trace_id)}:span:{str(parent_context.span_id)}:thread:{thread_id}"
+
+            # Store with all keys
+            self._contexts[span_key] = parent_context
+            self._contexts[trace_thread_key] = parent_context
+            self._contexts[trace_span_thread_key] = parent_context
+
+            # Thread-based fallback
+            self._thread_contexts[thread_id] = parent_context
```
dvargasfuertes: [q] why does the thread lookup have its own map, but the others are combined into one map?

dvargas92495 (commented):
Ok so I timebox'd the reviews to around 45m for both. The part I'm struggling the most with is the sheer number of design decisions grouped in both PRs.

I think we could start to simplify the implementation tremendously, incrementally, and more quickly if we start with a base case then worked our way up to the Map Node example. I think a good place to start would be [Trivial Workflow](https://github.com/vellum-ai/vellum-python-sdks/blob/main/tests/workflows/trivial/tests/test_workflow.py).

We can start with a simple test that just sets the `_enable_monitoring = True`. we then run the workflow, and assert that the parent context on the streaming events are what we expect, all the same trace id, etc. We may want to include one more assertion that proves we are using this new logic instead of the existing execution context logic. We can even then start introducing otel _here_, and assert that we are generating the emitted file for this use case, ensuring that we've gone end-to-end with a Trivial workflow.

Once we nail the experience for a trivial workflow, we can then start broadening our capabilities to all of the intricate subworkflow cases. It's likely that we will end up changing how some of the nodes _themselves_ work in order to accomodate this functionality.

I blocked off 330-6p est tm in case you think it'd be helpful to pair on some of this stuff

dvargas92495 (approved):
Looks simple enough to iterate on and don't want to further slow down, especially with another full calendar today :(

PR #2093, opened 2025-07-24
[APO-933] Support method as string for api node

vincent0426:


src/vellum/workflows/nodes/displayable/bases/api_node/node.py
```diff
@@ -31,7 +31,7 @@ class Trigger(BaseNode.Trigger):
         merge_behavior = MergeBehavior.AWAIT_ANY
 
     url: str = ""
-    method: Optional[APIRequestMethod] = APIRequestMethod.GET
+    method: Optional[Union[APIRequestMethod, str]] = APIRequestMethod.GET
```
dvargasfuertes: [nit] I'd prefer `src/vellum/client/types/method_enum.py` instead of `str`

src/vellum/workflows/nodes/displayable/bases/api_node/node.py
```diff
@@ -43,6 +43,34 @@ class Outputs(BaseOutputs):
         status_code: int
         text: str
 
+    def _normalize_http_method(self, method: Union[APIRequestMethod, str]) -> APIRequestMethod:
```
dvargasfuertes: [nit] Both `_vellum_execute_api` and `_local_execute_api` eventually just convert back to a string with `method.value` so I could actually see normalizing the opposite direction

dvargas92495 (approved):

PR #2097, opened 2025-07-24
[APO-942] Remove composio dependencies

alex-nork:
🧹 🧹 🧹 this PR cleans up the `ComposioService` class by swapping out composio-related dependencies for good old fashioned API requests, allowing us to remove that dependency entirely. This was also an opportunity to remove some unnecessary complexity in that and related classes.


src/vellum/workflows/nodes/displayable/tool_calling_node/composio_service.py
```diff
@@ -13,65 +15,58 @@ class ConnectionInfo:
...
+                response = requests.get(url, headers=headers, params=params or {}, timeout=30)
+            elif method == "POST":
+                response = requests.post(url, headers=headers, json=json_data or {}, timeout=30)
+            else:
+                raise ValueError(f"Unsupported HTTP method: {method}")
+
+            response.raise_for_status()
+            return response.json()
+        except Exception as e:
+            logger.error(f"Composio API request failed: {e}")
```
vincent0426: [q] should we raise node exception here instead of logger error?

vincent0426 (approved):

dvargas92495 (commented):
✅

src/vellum/workflows/nodes/displayable/tool_calling_node/composio_service.py
alex-nork: Good call - swapping this out for NodeException makes the errors within the node visible

PR #2098, opened 2025-07-25
[APO-945] Validate code execution node path during serialization

vincent0426:
before the fix if the filepath is not found it will be none

  ```
 'inputs': [{'id': '59c49667-189f-40f5-a86a-ea0d9c0acb29',

     'key': 'code',

     'value': {'rules': [{'data': {'type': 'JSON', 'value': None},

        'type': 'CONSTANT_VALUE'}],

      'combinator': 'OR'}},
```

correct file content will be
   ```
'inputs': [{'id': '59c49667-189f-40f5-a86a-ea0d9c0acb29',

     'key': 'code',

     'value': {'rules': [{'data': {'type': 'STRING', 'value': ''},

        'type': 'CONSTANT_VALUE'}],

      'combinator': 'OR'}},
```

we now check if the filepath exists and throw an error if not
Screenshot 2025-07-24 at 6 06 44 PM

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/code_execution_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/code_execution_node.py
@@ -2,6 +2,7 @@ import inspect
 from uuid import UUID
 from typing import ClassVar, Generic, Optional, TypeVar
 
+from vellum.workflows.exceptions import NodeException, WorkflowErrorCode
 from vellum.workflows.nodes.displayable.code_execution_node import CodeExecutionNode
 from vellum.workflows.nodes.displayable.code_execution_node.utils import read_file_from_path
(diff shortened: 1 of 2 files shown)


ee/vellum_ee/workflows/display/nodes/vellum/code_execution_node.py
```diff
@@ -40,6 +41,11 @@ def serialize(
                 node_filepath=node_file_path,
                 script_filepath=filepath,
             )
+            if not file_code:
+                raise NodeException(
```
dvargasfuertes: [nit] `NodeException` is meant for runtime, let's just use ordinary exception's for now

dvargas92495 (approved):

PR #2099, opened 2025-07-25
[APO-927] Fix composio codegen bug

alex-nork:
This fixes a codegen bug that was throwing a TypeError when deserializing Composio tools in the TCN. Now the code is generated as follows:
```python
functions = [
        ComposioToolDefinition(
            toolkit="gmail",
            action="GMAIL_CREATE_EMAIL_DRAFT",
            description="Creates a gmail email draft, supporting to/cc/bcc, subject, plain/html body (ensure `is html=true` for html), attachments, and threading.",
            display_name="Create email draft",
        )
    ]
```

Now that we've switched from the Composio python packages to straight API requests the responses have changed, hence the growing list of optional fields on the ComposioToolDefinition. Cleanup is tracked by APO-951.


dvargas92495 (commented):
✅

PR #2101, opened 2025-07-25
[APO-930] Hydrate ComposioToolDefinition

alex-nork:
The purpose of this PR is to retrieve from Composio details about the selected tool, and then use that response to fill in the parameters attribute of the `ComposioToolDefinition` class. This makes it go!


dvargas92495 (commented):
✅

PR #2107, opened 2025-07-25
Support dict type for tool calling node block

vincent0426:
Support blocks as dictionary

### DIFF
--- a/src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
+++ b/src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
@@ -1,6 +1,6 @@
 import json
 import logging
-from typing import Any, Callable, Iterator, List, Optional, Type, Union, cast
+from typing import Any, Callable, Dict, Iterator, List, Optional, Type, Union, cast
 
 from pydash import snake_case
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -287,7 +287,12 @@ def create_port_condition(fn_name):
 
     # Add a chat history block to blocks only if one doesn't already exist
     has_chat_history_block = any(
-        block.block_type == "VARIABLE" and block.input_variable == CHAT_HISTORY_VARIABLE for block in blocks
+        (
+            (block.get("block_type") if isinstance(block, dict) else getattr(block, "block_type", None)) == "VARIABLE"  # type: ignore  # noqa: E501
+            and (block.get("input_variable") if isinstance(block, dict) else getattr(block, "input_variable", None))  # type: ignore  # noqa: E501
```
dvargasfuertes: [nit] should be able to just return `block.block_type` instead of `getattr`. Why the type ignore?

dvargas92495 (approved):

PR #2112, opened 2025-07-26
Non-Decorator Global Execution Context Store [WIP]

marlenc3:
there's some flakiness I'm seeing when testing tho :/ 

main changes:
- context store now only accepts thread_id as a key
- introduction of relational threads -> storing parent thread id to retrieve their context
- adding global store support in current execution context manager
- removing run_item_with_context ( or w/ it was called) we can now directly access thread context via global store

### DIFF
--- a/src/vellum/workflows/context.py
+++ b/src/vellum/workflows/context.py
@@ -5,6 +5,7 @@ from uuid import UUID
 from typing import Iterator, Optional, cast
 
 from vellum.client.core import UniversalBaseModel
+from vellum.workflows.events.context import DEFAULT_TRACE_ID, MonitoringContextStore
 from vellum.workflows.events.types import ParentContext
 
 
@@ -17,15 +18,38 @@ _CONTEXT_KEY = "_execution_context"
 
 local = threading.local()
 
+monitoring_context_store = MonitoringContextStore()
+
 
 def get_execution_context() -> ExecutionContext:
-    """Retrieve the current execution context."""
-    return getattr(local, _CONTEXT_KEY, ExecutionContext())
+    """Get the current monitoring execution context, with intelligent fallback."""
+    context = getattr(local, _CONTEXT_KEY, ExecutionContext())
+    if context.trace_id != DEFAULT_TRACE_ID or context.parent_context:
+        return context
+
+    # If no thread-local context, try to restore from global store using current trace_id
+    trace_id = monitoring_context_store.get_current_trace_id()
+    span_id = monitoring_context_store.get_current_span_id()
+    if trace_id and trace_id != DEFAULT_TRACE_ID:
+        context = monitoring_context_store.retrieve_context(trace_id, span_id)
+        if context and (context.parent_context or context.trace_id != DEFAULT_TRACE_ID):
+            set_execution_context(context)
+            return context
+    return ExecutionContext()
 
 
 def set_execution_context(context: ExecutionContext) -> None:
-    """Set the current execution context."""
+    """Set the current monitoring execution context and persist it for cross-boundary access."""
     setattr(local, _CONTEXT_KEY, context)
+
+    # Always store in global store for cross-thread access
+    monitoring_context_store.store_context(context)
+    
+    # Set current trace and span IDs for this thread
+    if context.trace_id and context.trace_id != DEFAULT_TRACE_ID:
+        monitoring_context_store.set_current_trace_id(context.trace_id)
+    if context.parent_context:
+        monitoring_context_store.set_current_span_id(context.parent_context.span_id)
 
 
 def get_parent_context() -> ParentContext:
--- a/src/vellum/workflows/context.py
+++ b/src/vellum/workflows/context.py
@@ -5,7 +5,7 @@ from uuid import UUID
 from typing import Iterator, Optional, cast
 
 from vellum.client.core import UniversalBaseModel
-from vellum.workflows.events.context import DEFAULT_TRACE_ID, MonitoringContextStore
+from vellum.workflows.events.context import MonitoringContextStore
 from vellum.workflows.events.types import ParentContext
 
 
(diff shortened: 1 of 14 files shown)


src/vellum/workflows/events/relational_threads.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def __init__(self, *args, **kwargs):
+        self._collect_parent_thread()
+        threading.Thread.__init__(self, *args, **kwargs)
+
+    def _collect_parent_thread(self) -> None:
+        self._parent_thread = threading.get_ident()
+
+    def get_parent_thread(self) -> Optional[int]:
+        return self._parent_thread
```
marlenc3, line 16: TODO: 
on start -> pull in the execution context automatically -- should reduce the need to set up context

src/vellum/workflows/context.py
```diff
@@ -17,16 +18,37 @@ class ExecutionContext(UniversalBaseModel):
...
+monitoring_context_store = MonitoringContextStore()
+
 
 def get_execution_context() -> ExecutionContext:
-    """Retrieve the current execution context."""
-    return getattr(local, _CONTEXT_KEY, ExecutionContext())
+    """Get the current monitoring execution context, with intelligent fallback."""
+    context = getattr(local, _CONTEXT_KEY, ExecutionContext())
+    if context.parent_context:
+        return context
```
dvargasfuertes, line 28: [nit] I would default to `None` here now and just return `context`. Reason being is that we have valid cases of local thread context's not having a parent which is when you run the workflow locally

src/vellum/workflows/context.py
```diff
@@ -17,16 +18,37 @@ class ExecutionContext(UniversalBaseModel):
...
+
+    # If no thread-local context, try to restore from global store using current trace_id
+    context = monitoring_context_store.retrieve_context()
+    if context and context.parent_context:
+        set_execution_context(context)
```
dvargasfuertes, line 33: [q] why do we need to `set` here?

src/vellum/workflows/events/relational_threads.py
```diff
@@ -0,0 +1,16 @@
+import threading
+from typing import Optional
+
+
+class RelationalThread(threading.Thread):
+    _parent_thread: Optional[int] = None
+
+    def __init__(self, *args, **kwargs):
+        self._collect_parent_thread()
+        threading.Thread.__init__(self, *args, **kwargs)
```
dvargasfuertes, line 10: [blocking] make sure to use `super()`

src/vellum/workflows/events/relational_threads.py
```diff
@@ -0,0 +1,16 @@
+import threading
+from typing import Optional
+
+
+class RelationalThread(threading.Thread):
+    _parent_thread: Optional[int] = None
+
+    def __init__(self, *args, **kwargs):
+        self._collect_parent_thread()
```
dvargasfuertes, line 9: [nit] I'd just inline the parent tracking if it's just one line

src/vellum/workflows/events/relational_threads.py
```diff
@@ -0,0 +1,16 @@
+import threading
+from typing import Optional
+
+
+class RelationalThread(threading.Thread):
```
dvargasfuertes, line 5: This is neat

dvargasfuertes: [oos] One thing we'll have to think about is when users create their own nodes that invoke subworkflows. We'll likely want to support a BaseNode api for it. Can tackle that in the future

src/vellum/workflows/events/relational_threads.py
dvargasfuertes: [q] TODO in this PR or a future one? Seems like could be its own task

src/vellum/workflows/events/tests/test_basic_workflow.py
```diff
@@ -0,0 +1,41 @@
+import logging
+from uuid import uuid4
+
+from vellum.workflows.context import execution_context
+from vellum.workflows.workflows.event_filters import all_workflow_event_filter
+
+from tests.workflows.trivial.workflow import TrivialWorkflow
```
dvargasfuertes, line 7: [nit] Am trying to avoid imports from `tests.` -> `src.`. So Id prefer if we either move this test over into the trivial module or inline a new simple workflow below

src/vellum/workflows/runner/runner.py
```diff
@@ -188,15 +189,16 @@ def __init__(
         ]
 
     def _snapshot_state(self, state: StateType, deltas: List[StateDelta]) -> StateType:
+        execution = get_execution_context()
```
dvargasfuertes, line 192: [nit] would like to call or suffic these var's with `execution_context` since we might track actual execution's in the future

src/vellum/workflows/runner/runner.py
```diff
@@ -868,5 +861,13 @@ def _generate_events(self) -> Generator[WorkflowEvent, None, None]:
         self._background_thread_queue.put(None)
         cancel_thread_kill_switch.set()
 
+        # Clear MonitoringContextStore at the end of root workflow execution
+        # Only clear if this is a root workflow (no parent context)
+        if not self._execution_context.parent_context or self._execution_context.parent_context.type in [
+            "WORKFLOW_RELEASE_TAG",
+            "WORKFLOW_SANDBOX",
+        ]:
+            clear_execution_context()
```
dvargasfuertes, line 886: [blocking] why do we need this step? is it for memory consumption? If so, I think I'd prefer some root workflow tracking instead of these same checks we do in django, since these are not the only two root contexts in the no-django

dvargas92495 (commented):
Dig it! most feedback is minor, two blocking notes I think are quick to resolve

marlenc3:
eek - was too old, broke on merge/rebase. re-creating locally

PR #2116, opened 2025-07-28
Add mcp tool definition

vincent0426:


src/vellum/workflows/types/core.py
```diff
@@ -50,4 +50,4 @@ class ConditionType(Enum):
 
 
 # Type alias for functions that can be called in tool calling nodes
-Tool = Union[Callable[..., Any], DeploymentDefinition, Type["BaseWorkflow"], ComposioToolDefinition]
+Tool = Union[MCPToolDefinition, Callable[..., Any], DeploymentDefinition, Type["BaseWorkflow"], ComposioToolDefinition]
```
dvargasfuertes: [nit] add new entries towards the end to keep the callable first

src/vellum/workflows/types/definition.py
```diff
@@ -118,3 +121,21 @@ class ComposioToolDefinition(UniversalBaseModel):
...
+    api_key_header_value: Optional[Union[str, VellumSecret]] = None
+
+    model_config = {"arbitrary_types_allowed": True}
+
+
+class MCPToolDefinition(UniversalBaseModel):
+    name: str
+    server: MCPServer
+    description: Optional[str] = None
+    parameters: Dict[str, str] = {}
```
dvargasfuertes, line 141: [q] Is this one tool going to define all of the MCP tool? If so, parameters is odd to track since each MCP tool within the server will have its own schema

src/vellum/workflows/types/tests/test_definition.py
```diff
@@ -44,3 +47,70 @@ def test_composio_tool_definition_creation():
...
+        parameters={"repository_name": "string", "description": "string"},
+    )
+
+    assert mcp_tool.name == "create_repository"
+    assert mcp_tool.server.name == "github"
+    assert mcp_tool.server.url == "https://api.githubcopilot.com/mcp/"
+    assert mcp_tool.server.authorization_type == authorization_type
+    assert mcp_tool.parameters == {"repository_name": "string", "description": "string"}
+
+    if authorization_type == AuthorizationType.BEARER_TOKEN:
```
dvargasfuertes: [nit] instead of if statements in tests, we should either use parametrized tests or separate tests to avoid logic in the tests themselves

dvargas92495 (approved):

src/vellum/workflows/types/definition.py
vincent0426: each mcp server will have multiple tools so `MCPToolDefinitioin` should hold each single tool only

PR #2118, opened 2025-07-28
fixing push blocker

0xPCDefenders:
Switched from raising a ValueError to raising a logger warning. Pushes shouldn't be blocked with this refactor


dvargas92495 (approved):
nice! it helps for reviewers to have the linear ticket in the description so that they have context

vincent0426 (commented):
✅

PR #2120, opened 2025-07-28
Add MCP server attribute to tool calling node

vincent0426:
add mcp_servers


src/vellum/workflows/nodes/displayable/tool_calling_node/node.py
```diff
@@ -38,6 +40,7 @@ class ToolCallingNode(BaseNode):
     ml_model: ClassVar[str] = "gpt-4o-mini"
     blocks: ClassVar[List[Union[PromptBlock, Dict[str, Any]]]] = []
     functions: ClassVar[List[Tool]] = []
+    mcp_servers: ClassVar[List[MCPServer]] = []
```
dvargasfuertes, line 43: [blocking] this is the sort of API change that is good to hash out and align on over the TDD before implementation

I'd love to support multiple "tool sources", of which `MCPServer` is only one option. In the future, we should support OpenAPI, Composio, and possibly first class integrations ourselves. 

so I'd like to advocate for `tool_sources` or `tool_servers` or something more generic that hints at the fetching of multiple tools

PR #2126, opened 2025-07-29
Mark inline prompt nodes with callable functions as code mergable

NgoHarrison:
https://linear.app/vellum/issue/APO-995/mark-inline-prompt-nodes-as-code-mergable

### DIFF
--- a/ee/codegen/src/generators/nodes/inline-prompt-node.ts
+++ b/ee/codegen/src/generators/nodes/inline-prompt-node.ts
@@ -17,7 +17,7 @@ import {
   InlinePromptNode as InlinePromptNodeType,
   PlainTextPromptTemplateBlock,
   FunctionArgs,
-  Functions as FunctionsType,
+  ToolArgs as FunctionsType,
 } from "src/types/vellum";
 import { isNilOrEmpty } from "src/utils/typing";
 
@@ -364,9 +364,5 @@ export class InlinePromptNode extends BaseNode<
 
   protected getErrorOutputId(): string | undefined {
     return this.nodeData.data.errorOutputId;
-  }
-
-  public async persist(): Promise<void> {
-    await super.persist();
   }
 }
--- a/ee/codegen/src/generators/nodes/inline-prompt-node.ts
+++ b/ee/codegen/src/generators/nodes/inline-prompt-node.ts
@@ -17,9 +17,7 @@ import {
   InlinePromptNode as InlinePromptNodeType,
   PlainTextPromptTemplateBlock,
   FunctionArgs,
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/nodes/inline-prompt-node.ts
```diff
@@ -341,4 +345,39 @@ export class InlinePromptNode extends BaseNode<
   protected getErrorOutputId(): string | undefined {
     return this.nodeData.data.errorOutputId;
   }
+
+  public async persist(): Promise<void> {
```
dvargasfuertes: [blocking] ideally we keep our persist logic simple and do all of our work where we are already doing this sort of attribute parsing above. Is it possible to move?

ee/codegen/src/generators/nodes/inline-prompt-node.ts
```diff
@@ -341,4 +345,39 @@ export class InlinePromptNode extends BaseNode<
...
+      functionsAttribute.value?.type === "CONSTANT_VALUE" &&
+      functionsAttribute.value.value?.type === "JSON" &&
+      Array.isArray(functionsAttribute.value.value.value)
+    ) {
+      const functions: Array<
+        | FunctionArgs
+        | InlineWorkflowFunctionArgs
+        | DeploymentWorkflowFunctionArgs
+        | ComposioToolFunctionArgs
+      > = functionsAttribute.value.value.value;
```
dvargasfuertes: [nit] we should likely save this type union to a constant somewhere, likely that we reuse

ee/codegen/src/generators/nodes/inline-prompt-node.ts
NgoHarrison: Updated

dvargasfuertes: [nit] looks like we could delete this now

ee/codegen/src/types/vellum.ts
```diff
@@ -938,3 +938,10 @@ export type ComposioToolFunctionArgs = {
   connection_id?: string;
   integration_name?: string;
 } & NameDescription;
+
+export type Functions = (
```
dvargasfuertes: [nit] `ToolArgs`, consistent with what we call it on ToolCallingNode (`Tool`)

dvargas92495 (approved):
As a heads up, the customer has moved to loading JSON's instead of callables.

So while I think we should still include this, I'm think e2e support is no longer as urgent

vincent0426 (commented):
✅

PR #2130, opened 2025-07-30
[APO-983] Codegen node trigger

NgoHarrison:


vincent0426 (approved):

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -405,6 +420,9 @@ class MyToolCallingNode(ToolCallingNode):
     class Outputs(ToolCallingNode.Outputs):
         text: str
         chat_history: List[ChatMessage]
+
+    class NodeTrigger(ToolCallingNode.Trigger):
+        merge_behavior = MergeBehavior.AWAIT_ATTRIBUTES
```
dvargasfuertes, line 425: [blocking] we actually don't want to codegen outputs that are already expected to be default behavior. The ToolCallingNode already has this merge behavior for this merge behavior

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -529,6 +530,17 @@ export abstract class BaseNode<
     }
   }
 
+  protected getNodeTrigger(): AstNode | undefined {
+    if (this.nodeData.trigger) {
```
dvargasfuertes, line 534: to follow up with my comment above, we likely need a comparison here with the default trigger

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
NgoHarrison: In other words, are you saying don't codegen when getting `MergeBehavior.AWAIT_ATTRIBUTES`?

dvargasfuertes: Each node can have a different default trigger. all base nodes going forward will be `MergeBehavior.AWAIT_ATTRIBUTES`, so we shouldn't codegen in that case. Other nodes will may have AWAIT_ANY as the default (for now) so it'd be great to avoid codegenning in those cases too for those nodes

PR #2131, opened 2025-07-30
Fix node trigger codegen to enum

NgoHarrison:


ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
```diff
@@ -47,7 +47,7 @@ class MyCustomNode(BaseNode):
     default_attribute_2 = Inputs.count
 
     class NodeTrigger(BaseNode.Trigger):
-        merge_behavior = "AWAIT_ALL"
+        merge_behavior = MergeBehavior.AWAIT_ALL
```
dvargasfuertes, line 50: [oos] can we support both?

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
```diff
@@ -127,7 +127,7 @@ class MyCustomNode(BaseNode):
...
```
dvargasfuertes, line 130: [blocking] wait, won't this start causing runtime errors because we're not importing `MergeBehavior`?

dvargas92495 (commented):
One potentially blocking observation

ee/codegen/src/__test__/nodes/generic-nodes/__snapshots__/generic-node.test.ts.snap
NgoHarrison: https://github.com/vellum-ai/vellum-python-sdks/pull/2132

Addressed here

vincent0426 (commented):
✅

PR #2132, opened 2025-07-30
fix merge behavior import

NgoHarrison:
Accidentally forgot to inherit references for node trigger

### DIFF
--- a/ee/codegen/src/serializers/vellum.ts
+++ b/ee/codegen/src/serializers/vellum.ts
@@ -1024,7 +1024,7 @@ export declare namespace NodePortSerializer {
 export declare namespace NodeTriggerSerializer {
   interface Raw {
     id: string;
-    merge_behavior: "AWAIT_ATTRIBUTES" | "AWAIT_ALL" | "AWAIT_ANY" | "CUSTOM";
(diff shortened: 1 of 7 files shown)


ee/codegen/src/serializers/vellum.ts
```diff
@@ -1024,7 +1024,7 @@ export declare namespace NodePortSerializer {
 export declare namespace NodeTriggerSerializer {
   interface Raw {
     id: string;
-    merge_behavior: "AWAIT_ATTRIBUTES" | "AWAIT_ALL" | "AWAIT_ANY" | "CUSTOM";
+    merge_behavior: "AWAIT_ATTRIBUTES" | "AWAIT_ALL" | "AWAIT_ANY";
```
dvargasfuertes, line 1027: [blocking] why did we remove `| "CUSTOM"`?

dvargas92495 (commented):
✅

ee/codegen/src/serializers/vellum.ts
NgoHarrison: Was gonna do it in a follow up PR, backend does not have that currently

dvargasfuertes: ah if django is missing it, we could likely defer until we invest in custom triggers from the UI

vincent0426 (commented):
✅

PR #2136, opened 2025-07-30
Skip codegen of default behavior

NgoHarrison:


vincent0426 (approved):

dvargas92495 (approved):
lovely.

In general, we should have our alarms raised whenever we are _adding_ LoC to multiple snapshots, since that means we are complicating several end user cases

PR #2146, opened 2025-07-31
[APO-987] Implement VellumEmitter to directly emit events to Vellum

0xPCDefenders:


src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        if not self._enabled:
+            return
+
+        try:
+            state_data = self._serialize_state(state)
+
+            # Send the state snapshot as an event
+            event_data = {
+                "type": "workflow.state.snapshotted",
```
NgoHarrison: [nit] Could we extract these values into variables?

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+            event_data: The serialized event data to send.
+        """
+        last_exception = None
+
+        for attempt in range(self._max_retries + 1):
+            try:
+                # Use the Vellum client's underlying HTTP client to make the request
+                # For proper authentication headers and configuration
+                base_url = self._client._client_wrapper.get_environment().default
+                response = self._client._client_wrapper.httpx_client.request(
```
NgoHarrison: [nit/oos] I am guessing we don't have the vellum endpoint yet so maybe we can come back to this when it's up but would think we use something like `self._client.emit_event()` where emit_event is the endpoint generated by fern for our client sdk (can name the endpoint whatever)

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+
+        Args:
+            event: The event to serialize.
+
+        Returns:
+            Serialized event data.
+        """
+        # Use the same serialization approach as the rest of the system
+        return {
+            "type": "workflow.event",
```
NgoHarrison: [nit] Same as the comment above about extracting to variable

tests/workflows/basic_vellum_emitter_workflow/tests/test_vellum_emitter.py
```diff
@@ -0,0 +1,144 @@
...
+    mock_event = Mock()
+    mock_event.name = "test.event"
+
+    # These should not raise any exceptions or make any calls
+    emitter.emit_event(mock_event)
+    emitter.snapshot_state(TestState())
+
+
+@patch("vellum.workflows.emitters.vellum_emitter.create_vellum_client")
+@patch("vellum.workflows.emitters.vellum_emitter.default_serializer")
```
NgoHarrison: [nit] Make these fixtures

tests/workflows/basic_vellum_emitter_workflow/tests/test_vellum_emitter.py
```diff
@@ -0,0 +1,144 @@
...
+
+@patch("vellum.workflows.emitters.vellum_emitter.create_vellum_client")
+def test_vellum_emitter_error_handling(mock_create_client):
+    """Test that emitter handles errors gracefully."""
+    mock_client = Mock()
+    mock_create_client.return_value = mock_client
+    mock_client._client_wrapper.get_environment.return_value.default = "https://api.vellum.ai"
+    mock_client._client_wrapper.get_headers.return_value = {"Authorization": "Bearer test"}
+
+    from httpx import HTTPStatusError
```
NgoHarrison: [nit] Move local imports to top of file

tests/workflows/basic_vellum_emitter_workflow/tests/test_vellum_emitter.py
NgoHarrison, line 1: [nit] For our tests, we like to comment the flow of it in a GIVEN, WHEN, THEN structure. You can look at other tests for examples

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    Emitter that sends workflow events to Vellum's infrastructure for monitoring
+    externally hosted SDK-powered workflows.
+
+    Usage:
+        class MyWorkflow(BaseWorkflow):
+            emitters = [VellumEmitter]
+
+    Configuration:
+        - VELLUM_API_KEY: Environment variable for authentication
```
dvargasfuertes: [nit] I don't see this defined. I also think we can avoid requiring api_key and api_url from being passed in which I'll highlight below.

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+            timeout: Request timeout in seconds.
+            max_retries: Maximum number of retry attempts for failed requests.
+            enabled: Whether emitting is enabled. Useful for development/testing.
+        """
+        self._enabled = enabled
+        if not self._enabled:
+            return
+
+        try:
+            self._client = create_vellum_client(api_key=api_key, api_url=api_url)
```
dvargasfuertes: [oos] the [BaseWorkflow](https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/workflows/base.py#L245) already has access to a vellum client that we should reuse here to make sure we are using the same creds. We could do this by:
- Supporting a new `BaseWorkflowEmitter.register_context(context: WorkflowContext)`, that simply sets `self._context = context`
- instead of `self._client`, we could then use `self._context.vellum_client` ([source](https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/state/context.py#L33))

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+    ):
+        """
+        Initialize the VellumEmitter.
+
+        Args:
+            api_key: Vellum API key. If None, uses VELLUM_API_KEY environment variable.
+            api_url: Custom API URL. If None, uses default Vellum environment.
+            timeout: Request timeout in seconds.
+            max_retries: Maximum number of retry attempts for failed requests.
+            enabled: Whether emitting is enabled. Useful for development/testing.
```
dvargasfuertes: [q] why allow this as a init kwarg? if it's not enabled, I imagine the user would just skip entirely

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+
+        try:
+            event_data = self._serialize_event(event)
+
+            self._send_event(event_data)
+
+        except Exception as e:
+            logger.exception(f"Failed to emit event {event.name}: {e}")
+
+    def snapshot_state(self, state: BaseState) -> None:
```
dvargasfuertes, line 64: [blocking] let's just ignore this with a `pass` for now.

We currently send state snapshots as its own event so we may not need this method anymore

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+        Args:
+            event: The event to serialize.
+
+        Returns:
+            Serialized event data.
+        """
+        # Use the same serialization approach as the rest of the system
+        return {
+            "type": "workflow.event",
+            "data": default_serializer(event),
```
dvargasfuertes: [nit] Once we remove snapshot_state, instead of wrapping this payload we could just send the event directly

src/vellum/workflows/emitters/vellum_emitter.py
dvargasfuertes: Yea we won't be generating this though fern in the near term while team Sigma decides where to put this endpoint, so for now, this current approach looks good

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+                    request_options={"timeout_in_seconds": self._timeout},
+                )
+
+                response.raise_for_status()
+
+                if attempt > 0:
+                    logger.info(f"Event sent successfully after {attempt + 1} attempts")
+                return
+
+            except httpx.HTTPStatusError as e:
```
dvargasfuertes, line 105: I dig this error handling

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,197 @@
...
+                        continue
+                    else:
+                        logger.error(
+                            f"Server error emitting event after {self._max_retries + 1} attempts: "
+                            f"{e.response.status_code} {e.response.text}"
+                        )
+                else:
+                    # Client errors (4xx) are not retriable
+                    logger.error(f"Client error emitting event: {e.response.status_code} {e.response.text}")
+                    raise
```
dvargasfuertes: [blocking] let's not raise any errors for now. This emitter will be running in its own background Thread and therefore if we raise an exception, we could crash the thread and hang the process. the logger.error's are perfect. Actually, we should use `logger.exception` over `logger.error` whenever we are logging within an `except` block

tests/workflows/basic_vellum_emitter_workflow/tests/test_vellum_emitter.py
```diff
@@ -0,0 +1,144 @@
...
+    emitter = VellumEmitter(api_key="test-key", max_retries=0)
+
+    mock_event = Mock()
+    mock_event.name = "test.event"
+
+    # This should not raise an exception (errors are logged but not propagated)
+    emitter.emit_event(mock_event)
+
+
+def test_vellum_emitter_usage_pattern():
```
dvargasfuertes: [nit] this test doesn't do much for us since we already have general emitter tests

tests/workflows/basic_vellum_emitter_workflow/tests/test_vellum_emitter.py
```diff
@@ -0,0 +1,144 @@
...
+    # This demonstrates how users would use the VellumEmitter
+    class MyWorkflow(BaseWorkflow):
+        emitters = [VellumEmitter()]
+
+    # Verify an instance of VellumEmitter is in the emitters list
+    assert any(isinstance(e, VellumEmitter) for e in MyWorkflow.emitters)
+
+
+if __name__ == "__main__":
+    pytest.main([__file__])
```
dvargasfuertes: [q] what does this do? run the file directly? In general, we're unlikely to add it to all of our files so let's omit for now

dvargas92495 (commented):
looking good! only a couple of blockers. The comment(s) marked `[oos]` could be done in a followup

src/vellum/workflows/emitters/vellum_emitter.py
0xPCDefenders: This was a holdover from my previous commits on the old test file for the vellum emitter. The next push will omit this field.

src/vellum/workflows/emitters/vellum_emitter.py
0xPCDefenders: Sure thing

src/vellum/workflows/emitters/vellum_emitter.py
0xPCDefenders: Thanks!

tests/workflows/basic_vellum_emitter_workflow/tests/test_vellum_emitter.py
0xPCDefenders: Okay sounds good. This is function is omitted in my recent pushes to comply with the 1 test policy.

tests/workflows/basic_vellum_emitter_workflow/tests/test_vellum_emitter.py
0xPCDefenders: Didn't catch this initially, this will also be removed

src/vellum/workflows/emitters/vellum_emitter.py
dvargasfuertes: looks like we missed this one. let's remove from here and L156 below

src/vellum/workflows/emitters/vellum_emitter.py
```diff
@@ -0,0 +1,156 @@
...
+            timeout: Request timeout in seconds.
+            max_retries: Maximum number of retry attempts for failed requests.
+        """
+        self._context: Optional["WorkflowContext"] = None
+        self._timeout = timeout
+        self._max_retries = max_retries
+        self._events_endpoint = "events"  # TODO: make this configurable with the correct url
+
+    # For now the user will need to register workflow context with the emitter via dependency injection
+    def register_context(self, context: "WorkflowContext") -> None:
```
dvargasfuertes: [nit] let's actually put this on `BaseWorkflowEmitter`, could see this being useful for the others too. let's also set the `self._context: Optional["WorkflowContext"] = None` on the base class

tests/workflows/basic_vellum_emitter_workflow/workflow.py
```diff
@@ -0,0 +1,91 @@
...
+            timeout=60.0,  # Custom timeout for API requests
+            max_retries=5,  # More aggressive retry policy
+        )
+    ]
+
+    class Outputs(BaseOutputs):
+        final_value = NextNode.Outputs.final_value
+
+
+class DisabledVellumEmitterWorkflow(BaseWorkflow[BaseInputs, State]):
```
dvargasfuertes: [nit] looks like this is now the same as the Basic one and can be removed

dvargas92495 (approved):
last blocking comment should be quick, nice work!

dvargas92495 (approved):
🚢

PR #2154, opened 2025-08-01
[APO-995] Generate callable functions in top level of inline prompt node if present

NgoHarrison:
Context: This is for user who have callable functions defined in the functions attr of an inline prompt node

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -1,5 +1,69 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`WorkflowProjectGenerator > Get inline prompt node files > should generate the function at top level of node file > inline_prompt_test/nodes/inline_prompt_node.py 1`] = `
+"from vellum import JinjaPromptBlock, PromptParameters
+from vellum.workflows.nodes.displayable import InlinePromptNode as BaseInlinePromptNode
+from vellum.workflows.types.core import MergeBehavior
+
+
+def my_test_function(arg1: str, arg2: str) -> str:
+    """Processes input data and returns formatted output"""
+    return f"arg1: {arg1}, arg2: {arg2}"
+
+
+class InlinePromptNode(BaseInlinePromptNode):
+    ml_model = "gpt-4"
+    blocks = [
+        JinjaPromptBlock(template="""Hello world"""),
+    ]
+    functions = [
+        {
+            "type": "CODE_EXECUTION",
+            "name": "my_test_function",
+            "description": "Some sample test function",
+            "src": 'def my_test_function(arg1: str, arg2: str) -> str:\\n    """Processes input data and returns formatted output"""\\n    return f"arg1: {arg1}, arg2: {arg2}"',
+            "definition": {
+                "name": "my_test_function",
+                "description": "Some sample test function",
+                "parameters": {
+                    "type": "object",
+                    "required": [
+                        "arg1",
+                        "arg2",
+                    ],
+                    "properties": {
+                        "arg1": {
+                            "type": "string",
+                        },
+                        "arg2": {
+                            "type": "string",
+                        },
+                    },
+                },
+                "state": None,
+                "cache_config": None,
+                "forced": None,
+                "strict": None,
+            },
+        },
+    ]
+    parameters = PromptParameters(
+        stop=[],
+        temperature=0.7,
+        max_tokens=None,
+        top_p=None,
+        top_k=None,
+        frequency_penalty=None,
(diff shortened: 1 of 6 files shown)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1,5 +1,133 @@
...
+    return f"arg1: {arg1}, arg2: {arg2}"
+
+
+class InlinePromptNode(BaseInlinePromptNode):
+    ml_model = "gpt-4"
+    blocks = [
+        JinjaPromptBlock(template="""Hello world"""),
+    ]
+    functions = [
+        {
```
dvargasfuertes: [blocking] I would've expected that we just reference `my_test_function` here?

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1,5 +1,133 @@
...
+        top_p=None,
+        top_k=None,
+        frequency_penalty=None,
+        presence_penalty=None,
+        logit_bias=None,
+        custom_parameters=None,
+    )
+
+    class NodeTrigger(BaseInlinePromptNode.Trigger):
+        merge_behavior = MergeBehavior.AWAIT_ATTRIBUTES
```
dvargasfuertes: [nit] for codegen tests, we should try to zero in and just generate the piece of logic that we care about. If this is mostly because of the factory, then we could [oos] updates to the factory such that we codegen less unrelated logic overall

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -829,7 +832,19 @@ class NodeImplementationFile<
   }
 
   protected getFileStatements(): AstNode[] {
-    return [this.node.generateNodeClass()];
+    const statements: AstNode[] = [];
+    // This is to generate functions in the same file as inline prompt node if present
+    if (isInlinePromptNode(this.node.nodeData)) {
```
dvargasfuertes: [blocking] logic that is Inline Prompt Node-specific should live in the inline prompt node generator. We want to have this base file free of all _node_ level special casing, and instead, know how to generate _attributes_, in case other nodes have a `functions` attribute to generate.

So we should either:
- just move this to inline-prompt-node generator
- add handling to our Node Attributes logic

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
NgoHarrison: This is what I get for coding at 2am lol. Good catch and updated

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
NgoHarrison: Agreed. This is related to the factory and definitely would love to start incrementally refactoring it to a true builder pattern where just tack on what we want and not have it default to random stuff

ee/codegen/src/generators/nodes/bases/base.ts
NgoHarrison: Updated

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1,5 +1,70 @@
...
+    return f"arg1: {arg1}, arg2: {arg2}"
+
+
+class InlinePromptNode(BaseInlinePromptNode):
+    ml_model = "gpt-4"
+    blocks = [
+        JinjaPromptBlock(template="""Hello world"""),
+    ]
+    functions = [my_test_function]
+    functions = [
```
dvargasfuertes: [blocking] I'd expect only the former

ee/codegen/src/generators/nodes/bases/base.ts
```diff
@@ -99,6 +99,12 @@ export abstract class BaseNode<
   // added to the node class.
   protected abstract getErrorOutputId(): string | undefined;
 
+  // Override to provide additional statements that should be included in the node file
+  // before the node class definition.
+  public getAdditionalFileStatements(): AstNode[] {
```
dvargasfuertes: [nit] we should name this something that includes the word "Prefix" or "Before" or something like that. something to imply that it will come before the node class

ee/codegen/src/utils/typing.ts
```diff
@@ -53,3 +53,13 @@ export function hasErrorOutput(
     "errorOutputId" in node.data
   );
 }
+
+export function isInlinePromptNode(
```
dvargasfuertes: [nit] looks like this could now be removed

dvargas92495 (approved):
one more hopefully quick blocker

PR #2155, opened 2025-08-01
Fix node attribute factory

vincent0426:
Following

https://github.com/vellum-ai/vellum-python-sdks/blob/02487508d05a0f0402c715d18d2707f9bcd301b0/ee/codegen/src/types/vellum.ts#L858-L861


dvargas92495 (commented):
✅

PR #2175, opened 2025-08-04
Skip codegen empty string description for workflow

vincent0426:
Skip generating description if its empty
Screenshot 2025-08-04 at 5 48 44 PM

### DIFF
--- a/ee/codegen/src/generators/workflow.ts
+++ b/ee/codegen/src/generators/workflow.ts
@@ -149,7 +149,11 @@ export class Workflow {
     const workflowClass = python.class_({
       name: workflowClassName,
       extends_: [baseWorkflowClassRef],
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/workflow.ts
```diff
@@ -149,7 +149,11 @@ export class Workflow {
     const workflowClass = python.class_({
       name: workflowClassName,
       extends_: [baseWorkflowClassRef],
-      docs: this.workflowContext.workflowClassDescription,
+      docs:
+        this.workflowContext.workflowClassDescription &&
+        this.workflowContext.workflowClassDescription.length > 0
```
dvargasfuertes: [nit] these two conditions are redundant of each other

dvargas92495 (approved):

PR #2181, opened 2025-08-05
Add type for mcp server

vincent0426:


dvargas92495 (approved):
this will also help with FE discrimination as well

PR #2186, opened 2025-08-05
Raise invalid path with node validation error

vincent0426:
Validation on filepath will fail when we introduce serialization for graph in workflow init event.

Move validation to `serialize_module` so it will only validate during `push`


ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -898,6 +904,28 @@ def serialize_module(
...
+    @staticmethod
+    def _validate_code_execution_files(workflow_class: Type[BaseWorkflow]) -> None:
+        """Validate that all code execution node files exist."""
+
+        for node_class in workflow_class.get_all_nodes():
+            if issubclass(node_class, CodeExecutionNode):
+                BaseWorkflowDisplay._raise_if_invalid_code_execution_filepath(node_class)
+
+    @staticmethod
+    def _raise_if_invalid_code_execution_filepath(node_class: Type[CodeExecutionNode]) -> None:
```
dvargasfuertes: [blocking] oh I don't think we should have special handling for node at the workflow level. See my comment above

ee/vellum_ee/workflows/display/nodes/vellum/code_execution_node.py
```diff
@@ -40,8 +40,6 @@ def serialize(
                 node_filepath=node_file_path,
                 script_filepath=filepath,
             )
-            if not file_code:
-                raise Exception(f"Filepath '{filepath}' of node {node.__name__} does not exist")
```
dvargasfuertes: [nit] I think we define a new custom exception similar to what we have for codegen attribute errors, and we update our base workflow display to catch those exceptions to eventually add to `add_error(...)`. 

as a follow-on, node displays should have access to its parent workflow display and add_error to it directly

ee/vellum_ee/workflows/display/nodes/vellum/tests/test_code_execution_node.py
```diff
@@ -166,6 +167,56 @@ class Workflow(BaseWorkflow):
...
+            ],
+            "edges": [
+                {
+                    "id": "ab6ef06e-df2c-4877-9c3e-9d7261b39748",
+                    "source_node_id": "9b9e2a5d-01a4-46b2-80a3-d9484b2c0e08",
+                    "source_handle_id": "3e2a3f52-5047-4e2e-9a21-37bd43c63250",
+                    "target_node_id": "ac90c0ce-f393-438c-a24f-e5e9a9286182",
+                    "target_handle_id": "3a39ea63-9f86-4891-a902-0216a7190720",
+                    "type": "DEFAULT",
+                }
```
dvargasfuertes, line 210: [oos] We may want to prune any edges that map to nodes that fail

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -197,6 +198,9 @@ def serialize(self) -> JsonObject:
             except NotImplementedError as e:
                 self.add_error(e)
                 continue
+            except NodeValidationError as e:
```
dvargasfuertes: [nit] can group above `except (NotImplementedError, NodeValidationError) as e:`

dvargas92495 (approved):

ee/vellum_ee/workflows/display/nodes/vellum/tests/test_code_execution_node.py
vincent0426: ticketed

PR #2189, opened 2025-08-05
Fix mcp server auth type

vincent0426:


ee/codegen/src/types/vellum.ts
```diff
@@ -942,7 +942,7 @@ export type MCPServerFunctionArgs = {
   type: "MCP_SERVER";
   name: WorkflowValueDescriptor;
   url: WorkflowValueDescriptor;
-  authorization_type: "BEARER_TOKEN" | "API_KEY";
+  authorization_type: WorkflowValueDescriptor;
```
dvargasfuertes, line 945: [q] why do we need to workflow value descriptor this guy too? it's unlikely that we make this dynamic

vincent0426: cause we don't handle serialization for mcp server and most string attrs become constant value

```python
if not isinstance(value, BaseDescriptor):
    vellum_value = primitive_to_vellum_value(value)
    return {
        "type": "CONSTANT_VALUE",
        "value": vellum_value.dict(),
    }
```

PR #2190, opened 2025-08-06
[APO-1032][Part 1] Add new router node for tool calling node graph

NgoHarrison:
This is going to be part 1 of a multi PR stream to introduce a new graph structure for tool calling node. The plan is to split up the existing `ToolRouterNode` into a prompt and router node. What you are seeing here is just the router node. The next PR will be converting the `ToolRouterNode` into a prompt node without its port conditions and then subsequent clean up PRs. The final PR will include new connections and tests

### DIFF
--- a/src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
+++ b/src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
@@ -112,6 +112,18 @@ class ToolRouterNode(InlinePromptNode[ToolCallingState]):
             yield output
 
 
+class RouterNode(BaseNode[ToolCallingState]):
+    """Router node that handles routing to function nodes based on outputs."""
+
+    class Trigger(BaseNode.Trigger):
+        merge_behavior = MergeBehavior.AWAIT_ATTRIBUTES
+
+    def run(self) -> Iterator[BaseOutput]:
+        # Router node doesn't process outputs or create chat messages
+        # It just handles the routing logic via its ports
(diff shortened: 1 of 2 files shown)


vincent0426 (approved):

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -112,6 +112,18 @@ def run(self) -> Iterator[BaseOutput]:
             yield output
 
 
+class RouterNode(BaseNode[ToolCallingState]):
+    """Router node that handles routing to function nodes based on outputs."""
+
+    class Trigger(BaseNode.Trigger):
+        merge_behavior = MergeBehavior.AWAIT_ATTRIBUTES
```
dvargasfuertes, line 119: [nit] this is redundant and can be removed

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -112,6 +112,18 @@ def run(self) -> Iterator[BaseOutput]:
...
+
+    def run(self) -> Iterator[BaseOutput]:
+        # Router node doesn't process outputs or create chat messages
+        # It just handles the routing logic via its ports
+        yield from []
```
dvargasfuertes, line 124: [nit] this too

dvargas92495 (commented):
✅

PR #2233, opened 2025-08-07
[APO-1010] Fix inline subworkflow tool import levels

NgoHarrison:
Screenshot 2025-08-07 at 10 16 20 AM

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -1373,6 +1373,13 @@ class TemplatingNode(BaseTemplatingNode[BaseState, str]):
 "
 `;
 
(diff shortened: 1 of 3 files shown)


dvargas92495 (approved):
Would be great to back this change up with a test so we could ensure no regressions

NgoHarrison:
> Would be great to back this change up with a test so we could ensure no regressions

Yup marked it as draft to work on it, test added now

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1373,6 +1373,13 @@ class TemplatingNode(BaseTemplatingNode[BaseState, str]):
 "
 `;
```
NgoHarrison, line 1375: The error was from `from .....display import *`
( 5 levels up, but now we do 4)

dvargas92495 (commented):
✅

PR #2251, opened 2025-08-08
[APO-1041] Serialize graph in workflow init event

vincent0426:


src/vellum/workflows/events/tests/test_event.py
```diff
@@ -70,6 +70,151 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
                 body=WorkflowExecutionInitiatedBody(
                     workflow_definition=MockWorkflow,
                     inputs=MockInputs(foo="bar"),
+                    serialized_workflow_data={
+                        "workflow_raw_data": {
+                            "nodes": [
+                                {
```
vincent0426, line 78: should we make this just an empty arr?

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -674,6 +674,16 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
         workflow_class = get_args(cls.__orig_bases__[0])[0]  # type: ignore [attr-defined]
         register_workflow_display_class(workflow_class=workflow_class, workflow_display_class=cls)
 
+        # Override the workflow class's __serialize__ method to use this display class
+        def create_serialize_method():
```
dvargasfuertes: [nit] `create_serialize_method` is technically redundant, should be able to set `workflow_class.__serialize__ = __serialize__` directly

src/vellum/workflows/events/tests/test_event.py
```diff
@@ -70,6 +70,151 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
                 body=WorkflowExecutionInitiatedBody(
                     workflow_definition=MockWorkflow,
                     inputs=MockInputs(foo="bar"),
+                    serialized_workflow_data={
```
dvargasfuertes: [nit] we use the term `workflow_version_exec_config` everywhere, would be nice to keep consistent

src/vellum/workflows/events/tests/test_event.py
dvargasfuertes: [q] why would it be an empty arr?

src/vellum/workflows/workflows/base.py
```diff
@@ -636,6 +636,11 @@ def deserialize_state(
 
         return state_class(**state)
 
+    @classmethod
+    def __serialize__(cls):
```
dvargasfuertes: [nit] we could actually use `serialize(cls)` here. The reason why we use dunder's in BaseNode and Ports is because those classes allow user defined attributes

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -674,6 +674,16 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
...
+            def __serialize__():
+                workflow_display = cls(dry_run=True)
+                return workflow_display.serialize()
```
dvargasfuertes: [blocking] we should try/except this just in case. It seems like we are going to be returning this on _every_ workflow execution, so we don't want our existing ones to fail

src/vellum/workflows/runner/runner.py
```diff
@@ -611,13 +611,15 @@ def _handle_work_item_event(self, event: WorkflowEvent) -> Optional[WorkflowErro
         return None
 
     def _initiate_workflow_event(self) -> WorkflowExecutionInitiatedEvent:
+        serialized_workflow_data = self.workflow.__class__.__serialize__()
```
dvargasfuertes: [nit] I think I'm still partial to making this opt-in for now. At least until we prove it out for the ToolCallingNode and External workflow use cases. We could make that opt-in quite explicit, passed in by ToolCallingNode

dvargas92495 (approved):
one blocking comment which should have a quick resolve

src/vellum/workflows/events/tests/test_event.py
vincent0426: just to avoid long test cases but could keep it

src/vellum/workflows/events/tests/test_event.py
```diff
@@ -448,3 +730,20 @@ class MyData(UniversalBaseModel):
...
+        def serialize(self) -> dict:
+            raise Exception("Serialization failed")
+
+    register_workflow_display_class(MockWorkflow, TestWorkflowDisplay)
+
+    # WHEN we serialize the workflow that throws an exception
+    result = MockWorkflow.serialize()
+
+    # THEN the serialization should handle the error gracefully
+    assert result is None
```
vincent0426, line 749: test for graceful error handling

src/vellum/workflows/workflows/tests/test_base_workflow.py
```diff
@@ -684,3 +698,228 @@ class TestWorkflow(BaseWorkflow[BaseInputs, State]):
...
+
+    # AND we should find workflow execution initiated events
+    initiated_events = [event for event in events if event.name == "workflow.execution.initiated"]
+    assert len(initiated_events) == 3  # Main workflow + tool calling internal + inline workflow
+
+    assert initiated_events[0].body.workflow_version_exec_config is None  # Main workflow
+    assert initiated_events[1].body.workflow_version_exec_config is None  # Tool calling internal
+    assert initiated_events[2].body.workflow_version_exec_config == {  # Inline workflow tool
+        "config": "inline workflow"
+    }
```
vincent0426, line 840: test that only subworkflow will include `workflow_version_exec_config`

PR #2274, opened 2025-08-11
[APO-1007] Add parent context for subworkflow tool

vincent0426:
This PR fix that when a subworkflow tool reference parent workflow, it would fail during serialization


dvargas92495 (approved):
nice

PR #2288, opened 2025-08-12
Fix nested input subworkflow

vincent0426:


dvargas92495 (approved):
pretty dope that no other tests failed, a sign of this being in the correct direction

PR #2290, opened 2025-08-12
Add active workflow display registry

vincent0426:
vembda serialization for workflow init event needs parent context

### DIFF
--- a/ee/vellum_ee/workflows/display/utils/registry.py
+++ b/ee/vellum_ee/workflows/display/utils/registry.py
@@ -1,3 +1,4 @@
+from uuid import UUID
 from typing import TYPE_CHECKING, Dict, Optional, Type
 
 from vellum.workflows.events.types import BaseEvent
@@ -18,7 +19,7 @@ _workflow_display_registry: Dict[Type[BaseWorkflow], Type["BaseWorkflowDisplay"]
 _node_display_registry: Dict[Type[BaseNode], Type["BaseNodeDisplay"]] = {}
 
 # Registry to store active workflow display contexts by span ID for nested workflow inheritance
-_active_workflow_display_contexts: Dict[str, "WorkflowDisplayContext"] = {}
+_active_workflow_display_contexts: Dict[UUID, "WorkflowDisplayContext"] = {}
 
 
 def get_from_workflow_display_registry(workflow_class: Type[BaseWorkflow]) -> Optional[Type["BaseWorkflowDisplay"]]:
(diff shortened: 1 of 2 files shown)


ee/vellum_ee/workflows/tests/test_registry.py
```diff
@@ -0,0 +1,164 @@
...
+    result = get_parent_display_context_from_event(event)
+
+    # THEN it should return None (no workflow parent found)
+    assert result is None
+
+
+def test_get_parent_display_context_from_event__nested_workflow_parents():
+    """Test event with nested workflow parents traverses correctly"""
+    # GIVEN a chain of nested contexts:
+    # Event -> NodeParent -> WorkflowParent -> WorkflowParent
```
NgoHarrison: [q] Are these events in reverse order? I.e should it be workflowparent -> workflowparent -> nodeparent?

NgoHarrison (approved):

ee/vellum_ee/workflows/display/utils/registry.py
```diff
@@ -35,3 +41,42 @@ def get_from_node_display_registry(node_class: Type[BaseNode]) -> Optional[Type[
...
+
+    Returns:
+        The parent workflow display context if found, None otherwise
+    """
+    if not event.parent:
+        return None
+
+    current_parent: Optional["ParentContext"] = event.parent
+    while current_parent:
+        if hasattr(current_parent, "type") and current_parent.type == "WORKFLOW":
```
dvargasfuertes: [nit] shouldn't need `hasattr` once we know `BaseEvent`

ee/vellum_ee/workflows/display/utils/registry.py
```diff
@@ -14,6 +17,9 @@
 # Used to store the mapping between node types and their display classes
 _node_display_registry: Dict[Type[BaseNode], Type["BaseNodeDisplay"]] = {}
 
+# Registry to store active workflow display contexts by span ID for nested workflow inheritance
+_active_workflow_display_contexts: Dict[str, "WorkflowDisplayContext"] = {}
```
dvargasfuertes: [nit] we could avoid casts by just making the key `UUID` instead of `str`

ee/vellum_ee/workflows/display/utils/registry.py
```diff
@@ -35,3 +41,42 @@ def get_from_node_display_registry(node_class: Type[BaseNode]) -> Optional[Type[
...
 def register_node_display_class(node_class: Type[BaseNode], node_display_class: Type["BaseNodeDisplay"]) -> None:
     _node_display_registry[node_class] = node_display_class
+
+
+def register_workflow_display_context(span_id: str, display_context: "WorkflowDisplayContext") -> None:
+    """Register a workflow display context by span ID for nested workflow inheritance."""
+    _active_workflow_display_contexts[span_id] = display_context
+
+
+def get_parent_display_context_for_span(span_id: str) -> Optional["WorkflowDisplayContext"]:
```
dvargasfuertes: [nit] can make this private if we don't expect external use

ee/vellum_ee/workflows/display/utils/registry.py
```diff
@@ -35,3 +41,42 @@ def get_from_node_display_registry(node_class: Type[BaseNode]) -> Optional[Type[
...
+    current_parent: Optional["ParentContext"] = event.parent
+    while current_parent:
+        if hasattr(current_parent, "type") and current_parent.type == "WORKFLOW":
+            # Found a parent workflow, try to get its display context
+            parent_span_id = str(current_parent.span_id)
+            parent_display_context = get_parent_display_context_for_span(parent_span_id)
+            if parent_display_context:
+                return parent_display_context
+        # Move up the parent chain
+        current_parent = getattr(current_parent, "parent", None)
```
dvargasfuertes: [nit] should be able to just do `current_parent.parent`

ee/vellum_ee/workflows/tests/test_registry.py
dvargasfuertes: yea would read better from the Event

Also, in practice, a workflow should never have another workflow as a direct parent, always a node parent context in between

dvargas92495 (approved):

PR #2305, opened 2025-08-13
Add `is_dynamic` attribute to workflow

vincent0426:


src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
```diff
@@ -30,6 +30,8 @@ def __new__(cls, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any
         if not subworkflow_attribute:
             return node_class
 
+        subworkflow_attribute.is_dynamic = True
```
vincent0426: this will also set the subworkflow tool that we create in tool calling node

src/vellum/workflows/events/tests/test_event.py
```diff
@@ -70,6 +70,7 @@ class MockWorkflow(BaseWorkflow[MockInputs, BaseState]):
                 body=WorkflowExecutionInitiatedBody(
                     workflow_definition=MockWorkflow,
                     inputs=MockInputs(foo="bar"),
+                    is_dynamic=False,
```
dvargasfuertes: [blocking] why do we need it on the event?

src/vellum/workflows/nodes/core/inline_subworkflow_node/node.py
dvargasfuertes: [blocking] can we do this for just the Subworkflow Tool and not for other Inline Subworkflow Nodes for now?

dvargas92495 (approved):

PR #2310, opened 2025-08-14
Add default composio name

vincent0426:


src/vellum/workflows/types/definition.py
```diff
@@ -111,11 +111,11 @@ class ComposioToolDefinition(UniversalBaseModel):
     action: str  # Specific action like "GITHUB_CREATE_AN_ISSUE"
     description: str
     user_id: Optional[str] = None
+    name: Optional[str] = None  # Optional name, defaults to action if not provided
```
dvargasfuertes: [nit] I wouldn't make this `Optional` - just `name: str = ""`

dvargas92495 (approved):

PR #2314, opened 2025-08-15
Multithreaded Execution Context Update

marlenc3:
New Relational Threads - unsued for now so should be safe to merge in asap

changes: 
relational threads - keeps track of span, trace id, and parent thread associated with a threads
updates to global context store - using new span and trace_id in threads
updates to execution context -> ability to use new threads if local excution context does not exist

remove the runner changes, to unblock, iterate

### DIFF
--- a/src/vellum/workflows/context.py
+++ b/src/vellum/workflows/context.py
@@ -1,10 +1,11 @@
 from contextlib import contextmanager
 from dataclasses import field
 import threading
-from uuid import UUID
+from uuid import UUID, uuid4
 from typing import Iterator, Optional, cast
 
 from vellum.client.core import UniversalBaseModel
+from vellum.workflows.events.context import MonitoringContextStore
 from vellum.workflows.events.types import ParentContext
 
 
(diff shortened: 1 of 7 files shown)


src/vellum/workflows/context.py
```diff
@@ -17,16 +18,37 @@ class ExecutionContext(UniversalBaseModel):
...
 def get_execution_context() -> ExecutionContext:
-    """Retrieve the current execution context."""
-    return getattr(local, _CONTEXT_KEY, ExecutionContext())
+    """Get the current monitoring execution context, with intelligent fallback."""
+    context = getattr(local, _CONTEXT_KEY, ExecutionContext())
+    if context.parent_context:
+        return context
+
+    # If no thread-local context, try to restore from global store using current trace_id
+    context = monitoring_context_store.retrieve_context()
```
marlenc3, line 31: we should always have thread-local for now -- nothing is using our new threads yet

src/vellum/workflows/events/context.py
```diff
@@ -2,110 +2,85 @@
...
-
-            span_key = f"trace:{str(trace_id)}:span:{str(span_id)}:thread:{thread_id}"
-            if span_key in self._contexts:
-                result = self._contexts[span_key]
-                return result
+            # Get span_id from RelationalThread first
+            span_id = None
 
-        return None
+            if context.parent_context and hasattr(context.parent_context, "span_id"):
```
marlenc3, line 39: default to context span + trace, then fall back to thread span and trace

src/vellum/workflows/state/context.py
```diff
@@ -35,6 +35,8 @@ def __init__(
 
         if self._execution_context.parent_context is None:
             self._execution_context.parent_context = ExternalParentContext(span_id=uuid4())
+            # Propagate the updated context back to the global execution context
+            set_execution_context(self._execution_context)
```
marlenc3, line 39: this was needed for global context to see the set context that's passed through -- might move this out to the end, once execution_contexdt is set

dvargas92495 (approved):

PR #2319, opened 2025-08-15
[codegen] Support prompt blocks for file input

clopen-set:
Intended use case is to have a `CHAT_MESSAGE` block with some static file input blocks.

I can also extend support for the normal `PromptBlock` generator used in Generic Nodes if needed: https://github.com/vellum-ai/vellum-python-sdks/blob/bf3f82e31dd5f5ba9f7319dc4f087068cfd850f4/ee/codegen/src/generators/prompt-block.ts#L19


dvargas92495 (approved):

vincent0426 (commented):
✅

PR #2332, opened 2025-08-20
[APO-1025] Fix Dict Contains Dict TypeError In Workflows

VJagiasi:
Fix ticket [APO-1025 ](https://linear.app/vellum/issue/APO-1025/fix-issue-with-some-valcontains-causing-errors-when-rhs-is-an-object )where `some_val.contains({...})` operations crashed with `TypeError: unhashable type: 'dict'` when checking if one dictionary contains another dictionary.

  ## Details
  - Updated `ContainsExpression.resolve()` to handle dict-contains-dict comparisons using JSON serialization
  - Added support for string-contains-dict comparisons
  - Preserves all existing functionality for lists, tuples, sets, and strings
  - Added comprehensive test suite with 12 test cases covering all scenarios and edge cases

  ## Changes Made
  - **Modified**: `src/vellum/workflows/expressions/contains.py` - Added JSON-based comparison logic
  - **Added**: `src/vellum/workflows/expressions/tests/test_contains.py` - Complete test coverage

 ## Testing
  - ✅ All existing tests still pass
  - ✅ New comprehensive test suite covers all edge cases
  - ✅ Validated fix resolves the exact error from Linear ticket

### DIFF
--- a/src/vellum/workflows/expressions/contains.py
+++ b/src/vellum/workflows/expressions/contains.py
@@ -1,4 +1,3 @@
-import json
 from typing import Generic, TypeVar, Union
 
 from vellum.workflows.constants import undefined
@@ -37,13 +36,14 @@ class ContainsExpression(BaseDescriptor[bool], Generic[LHS, RHS]):
 
         rhs = resolve_value(self._rhs, state)
 
-        if isinstance(lhs, str) and isinstance(rhs, dict):
-            rhs_json = json.dumps(rhs, sort_keys=True)
-            return rhs_json in lhs
-
         if isinstance(lhs, dict) and isinstance(rhs, dict):
-            lhs_json = json.dumps(lhs, sort_keys=True)
-            rhs_json = json.dumps(rhs, sort_keys=True)
-            return rhs_json in lhs_json
+            raise InvalidExpressionException(
+                "Cannot check if dict contains dict. Use dict keys/values or convert to strings for comparison."
+            )
+
+        if isinstance(lhs, str) and isinstance(rhs, dict):
+            raise InvalidExpressionException(
+                "Cannot check if string contains dict. Convert dict to string first or check for specific keys."
+            )
 
         return rhs in lhs
--- a/src/vellum/workflows/expressions/contains.py
+++ b/src/vellum/workflows/expressions/contains.py
@@ -1,3 +1,4 @@
+import json
 from typing import Generic, TypeVar, Union
 
 from vellum.workflows.constants import undefined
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/expressions/tests/test_contains.py
```diff
@@ -0,0 +1,178 @@
...
+    Tests that ContainsExpression handles string-contains-dict scenarios.
+    """
+    state = TestState()
+    lhs_string = 'Response: {"status": "success"} was returned'
+    rhs_dict = {"status": "success"}
+
+    expression = ContainsExpression(lhs=lhs_string, rhs=rhs_dict)
+    result = expression.resolve(state)
+
+    assert result is True
```
dvargasfuertes: [blocking] this is a bit unintuitive both in python and typescript this is either false or error.

We should be fine continuing to raise an error here, as long as it makes it more clear to the end user what's happening instead of the current "Internal Server Error".

If `rhs` is a dict (or in general, any incompatible type for the left side of the ` in ` expression), I think we should raise an `InvalidExpressionException`. Our node logic properly catches and propagates this error so that the end user could see their error more easily and remedy accordingly

dvargas92495 (commented):
Love the tests! I think we could simplify our implementation by just raising a user facing error in this situation

src/vellum/workflows/expressions/tests/test_contains.py
VJagiasi: added

NgoHarrison (approved):

src/vellum/workflows/expressions/tests/test_contains.py
```diff
@@ -0,0 +1,178 @@
...
+    dict_value: dict = {"key": "value"}
+    list_value: list = [1, 2, 3]
+    string_value: str = "hello world"
+
+
+def test_dict_contains_dict_raises_error():
+    """
+    Tests that ContainsExpression raises clear error for dict-contains-dict scenarios.
+
+    This addresses Linear ticket APO-1025 where dict.contains(dict)
```
vincent0426: @dvargas92495 I don't think we need to keep this in the code?

vincent0426 (approved):

src/vellum/workflows/expressions/tests/test_contains.py
dvargasfuertes: [nit] correct, we should keep linear references for completed work out

src/vellum/workflows/expressions/contains.py
```diff
@@ -35,4 +35,15 @@ def resolve(self, state: "BaseState") -> bool:
             )
 
         rhs = resolve_value(self._rhs, state)
+
+        if isinstance(lhs, dict) and isinstance(rhs, dict):
+            raise InvalidExpressionException(
+                "Cannot check if dict contains dict. Use dict keys/values or convert to strings for comparison."
+            )
+
+        if isinstance(lhs, str) and isinstance(rhs, dict):
```
dvargasfuertes: [nit] looks like these two checks can simplify to just checking the `rhs`

dvargas92495 (approved):

src/vellum/workflows/expressions/contains.py
VJagiasi: fixed

PR #2341, opened 2025-08-20
[APO-1072] Add Backward Compatibility For TemplatingNode Value/Type Access

VJagiasi:
### Issue

[APO-1072](https://linear.app/vellum/issue/APO-1072/templating-nodes-should-handle-legacy-value-and-type-references-like)

TemplatingNode broke after SDK-enabled workflows - can't access .value and .type properties on inputs anymore. Breaks workflows using Map node outputs like array[0].value.

### Fix

Add backward compatibility wrappers to TemplatingNode, same as CodeExecutionNode already has.

### Changes

  - Extract shared wrapper function to eliminate duplication between nodes
  - Wrap TemplatingNode inputs before template rendering
  - Refactor CodeExecutionNode to use shared function
  - Add test coverage for wrapper functionality

### DIFF
--- a/src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
+++ b/src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
@@ -319,102 +319,69 @@ def test_api_error_templating_node():
     assert outputs.result == ""
 
 
-def test_templating_node__string_value_access():
-    # GIVEN a templating node that accesses string value
-    class TemplateNode(TemplatingNode[BaseState, str]):
-        template = "{{ text_input.value }}"
-        inputs = {"text_input": "hello world"}
+@pytest.mark.parametrize(
+    "template,inputs,expected_result,output_type",
+    [
+        # String value access
+        ("{{ text_input.value }}", {"text_input": "hello world"}, "hello world", str),
+        # Function call value access
+        (
+            "{{ func.value.name }}",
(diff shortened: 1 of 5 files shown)


src/vellum/workflows/types/code_execution_node_wrappers.py
NgoHarrison, line 1: [nit] Could we move this to somewhere that isn't specific to code execution nodes since you are now using this in templating nodes?

src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
NgoHarrison, line 1: [nit] Because these tests are so similar in assertions and setup, could we parameterize these tests? One for value access and one for type access?

dvargas92495 (approved):
Nice! 🔥

NgoHarrison (approved):

vincent0426 (approved):

src/vellum/workflows/types/code_execution_node_wrappers.py
VJagiasi: fixed

src/vellum/workflows/nodes/core/templating_node/tests/test_templating_node.py
VJagiasi: fixed

PR #2344, opened 2025-08-21
Add resolver

vincent0426:


src/vellum/workflows/resolvers/resolver.py
```diff
@@ -0,0 +1,36 @@
...
+        client = self._context.vellum_client
+        response = client.workflow_executions.retrieve_workflow_execution_detail(
+            execution_id=str(previous_execution_id),
+        )
+
+        if response.state is None:
+            return None
+
+        meta = StateMeta.model_validate(response.state.pop("meta"))
+        return BaseState(**response.state, meta=meta)
```
dvargasfuertes: [blocking] we want to return an instance of the state class of the _workflow_, rather than an instance of `BaseState`. Somehow, we need to make that available to this resolver

src/vellum/workflows/resolvers/tests/test_resolver.py
```diff
@@ -0,0 +1,50 @@
...
+from uuid import uuid4
+
+from vellum.client.types.workflow_execution_detail import WorkflowExecutionDetail
+from vellum.workflows.inputs.base import BaseInputs
+from vellum.workflows.resolvers.resolver import VellumResolver
+from vellum.workflows.state.base import BaseState, NodeExecutionCache
+from vellum.workflows.state.context import WorkflowContext
+
+
+def test_load_state_with_context_success():
```
dvargasfuertes, line 13: [oos] eventually want to see a test at the workflow API level, like our other `tests/workflows` tests

dvargas92495 (approved):
quick blocker

src/vellum/workflows/resolvers/base.py
```diff
@@ -1,21 +1,26 @@
...
 
 class BaseWorkflowResolver(ABC):
     def __init__(self):
         self._context: Optional["WorkflowContext"] = None
+        self._workflow_class: Optional[Type["BaseWorkflow"]] = None
 
     def register_context(self, context: "WorkflowContext") -> None:
         self._context = context
 
+    def register_workflow(self, workflow_class: Type["BaseWorkflow"]) -> None:
```
dvargasfuertes: [nit] if we pass in the instance here, we could save both the class and the context with one register method

dvargas92495 (approved):

dvargas92495 (approved):

PR #2345, opened 2025-08-21
Add mcp tool description field

vincent0426:
frontend tool type uses both name and description, though description is nullable, I think we could set it as a placeholder?


NgoHarrison (approved):
A question I have then is why is the UI expecting this field then? It sounds like we can change the schema

vincent0426:
> A question I have then is why is the UI expecting this field then? It sounds like we can change the schema

the initial frontend design combines name and description as common fields and using it as schema

vincent0426:
ah but i think maybe we might support mcp server description in the future

dvargas92495 (approved):
Yea seems useful to keep

PR #2356, opened 2025-08-22
[APO-1144] [APO-1147] Add previous execution id

vincent0426:


src/vellum/workflows/runner/runner.py
```diff
@@ -627,6 +645,37 @@ def _handle_work_item_event(self, event: WorkflowEvent) -> Optional[WorkflowErro
...
+        if self._span_link_info:
+            from vellum.client.types.vellum_code_resource_definition import (
+                VellumCodeResourceDefinition as ClientVellumCodeResourceDefinition,
+            )
+            from vellum.client.types.workflow_parent_context import WorkflowParentContext as ClientWorkflowParentContext
+            from vellum.workflows.types.definition import serialize_type_encoder_with_id
+
+            serialized_data = serialize_type_encoder_with_id(self.workflow.__class__)
+            serialized_data["id"] = str(serialized_data["id"])
+            workflow_definition = ClientVellumCodeResourceDefinition(**serialized_data)
```
vincent0426: this looks wrong but `SpanLink` is from client types

src/vellum/workflows/resolvers/base.py
```diff
@@ -28,5 +28,5 @@ def get_state_snapshot_history(self) -> Iterator[BaseState]:
         pass
 
     @abstractmethod
-    def load_state(self, previous_execution_id: Optional[UUID] = None) -> Optional[BaseState]:
+    def load_state(self, previous_execution_id: Optional[str] = None) -> LoadStateResult:
```
dvargasfuertes: [nit] why did we change from uuid to str?

src/vellum/workflows/resolvers/resolver.py
```diff
@@ -16,27 +18,66 @@ def get_latest_execution_events(self) -> Iterator[WorkflowEvent]:
...
-    def load_state(self, previous_execution_id: Optional[UUID] = None) -> Optional[BaseState]:
+    def _find_execution_events(
+        self, execution_id: str, spans: list
+    ) -> Tuple[Optional[WorkflowExecutionInitiatedEvent], Optional[WorkflowExecutionInitiatedEvent]]:
+        previous_invocation: Optional[WorkflowExecutionInitiatedEvent] = None
+        root_invocation: Optional[WorkflowExecutionInitiatedEvent] = None
+
+        for span in spans:
+            # Look for workflow execution spans
+            if isinstance(span, WorkflowExecutionSpan):
```
dvargasfuertes: [nit] we typically use `.name` and `.type` attribute references to type discriminate

src/vellum/workflows/resolvers/resolver.py
```diff
@@ -16,27 +18,66 @@ def get_latest_execution_events(self) -> Iterator[WorkflowEvent]:
...
 
-    def load_state(self, previous_execution_id: Optional[UUID] = None) -> Optional[BaseState]:
+    def _find_execution_events(
+        self, execution_id: str, spans: list
+    ) -> Tuple[Optional[WorkflowExecutionInitiatedEvent], Optional[WorkflowExecutionInitiatedEvent]]:
+        previous_invocation: Optional[WorkflowExecutionInitiatedEvent] = None
+        root_invocation: Optional[WorkflowExecutionInitiatedEvent] = None
+
+        for span in spans:
+            # Look for workflow execution spans
```
dvargasfuertes: can do a `span.span_id == execution_id` here to start ruling out spans

src/vellum/workflows/resolvers/resolver.py
```diff
@@ -16,27 +18,66 @@ def get_latest_execution_events(self) -> Iterator[WorkflowEvent]:
...
+
+    def load_state(self, previous_execution_id: Optional[str] = None) -> LoadStateResult:
         if previous_execution_id is None:
-            return None
+            return LoadStateResult(state=None, span_link_info=None)
 
         if not self._context:
             logger.warning("Cannot load state: No workflow context registered")
-            return None
+            return LoadStateResult(state=None, span_link_info=None)
```
dvargasfuertes: [nit] not sure of the benefit of a second nested type (span_link_info) and could see just including it all in the root

src/vellum/workflows/resolvers/resolver.py
```diff
@@ -16,27 +18,66 @@ def get_latest_execution_events(self) -> Iterator[WorkflowEvent]:
...
+            if isinstance(span, WorkflowExecutionSpan):
+                # Find the WorkflowExecutionInitiatedEvent in the span's events
+                for event in span.events:
+                    if isinstance(event, WorkflowExecutionInitiatedEvent):
+                        # Check if this is the previous execution matches the execution_id
+                        if event.span_id == execution_id:
+                            previous_invocation = event
+
+                        if not event.links:
+                            root_invocation = event
```
dvargasfuertes: [blocking] this is not currently correct, since this is treating every span as a possible root invocation. Once we find the previous span, we check _its_ links and see it's root invocation to reuse

src/vellum/workflows/runner/runner.py
```diff
@@ -140,6 +145,19 @@ def __init__(
                 if issubclass(ei.inputs_class.__parent_class__, BaseNode)
             ]
             self._is_resuming = True
+        elif previous_execution_id:
+            self.previous_execution_id = str(previous_execution_id)
+            for resolver in self.workflow.resolvers:
+                try:
+                    load_state_result = resolver.load_state(self.previous_execution_id)
+                    if load_state_result.state is not None:
+                        self._initial_state = cast(StateType, load_state_result.state)
```
dvargasfuertes: [nit] we should almost never cast. In this case, we should ensure that `isinstance(load_state_result.state, StateType)`

src/vellum/workflows/runner/runner.py
dvargasfuertes: [nit] When we first introduced it, I had a sense we would need our own SDK version first. Would like to take the opportunity to do so now

Long term, we'll have to think about how to consolidate the two, but our SDK versions are almost more feature rich than our client versions

src/vellum/workflows/resolvers/base.py
vincent0426: I do the transformation at the upper layer but yea doing it here would be better

src/vellum/workflows/runner/runner.py
marlenc3: \[oos\] for what it's worth — I think from one of the fires of the weekend before last, we were talking about removing a lot of the drf serializers internally, and using our sdk to run the validation. if we keep finding reasons to consolidate, it might be worth doing! (plus it'll help with latency on our side)

src/vellum/workflows/resolvers/resolver.py
```diff
@@ -16,7 +19,38 @@ def get_latest_execution_events(self) -> Iterator[WorkflowEvent]:
     def get_state_snapshot_history(self) -> Iterator[BaseState]:
         return iter([])
 
-    def load_state(self, previous_execution_id: Optional[UUID] = None) -> Optional[BaseState]:
+    def _find_previous_and_root_span(
```
marlenc3, line 22: [nit] this might be a util we'll want in the core.events directory 👍  we have a ton of other span/root utils there I think you might be able to re-use one here 🤔

src/vellum/workflows/resolvers/resolver.py
```diff
@@ -26,17 +60,33 @@ def load_state(self, previous_execution_id: Optional[UUID] = None) -> Optional[B
 
         client = self._context.vellum_client
         response = client.workflow_executions.retrieve_workflow_execution_detail(
-            execution_id=str(previous_execution_id),
+            execution_id=previous_execution_id,
```
marlenc3, line 67: [nit] there's a new query parameter we can use now -- if you want to avoid loading up spans in this request

src/vellum/workflows/resolvers/resolver.py
```diff
@@ -26,17 +60,33 @@ def load_state(self, previous_execution_id: Optional[UUID] = None) -> Optional[B
...
-            return BaseState(**response.state, meta=meta)
+            state = BaseState(**response.state, meta=meta)
+
+        return LoadStateResult(
+            state=state,
+            previous_trace_id=previous_trace_id,
+            previous_span_id=previous_span_id,
+            root_trace_id=root_trace_id,
+            root_span_id=root_span_id,
+        )
```
marlenc3, line 96: [aside] ah - so this is where we're pulling previous state for an execution? don't worry about the api stuff for now then, this will get udpated once we have the RTA for Executions

src/vellum/workflows/events/types.py
```diff
@@ -86,6 +85,12 @@ class ExternalParentContext(BaseParentContext):
     type: Literal["EXTERNAL"] = "EXTERNAL"
 
 
+class SpanLink(UniversalBaseModel):
+    trace_id: str
+    type: Literal["PREVIOUS_SPAN", "ROOT_SPAN"]
```
vincent0426: do we add triggered by here?

dvargasfuertes: yea we want it to be compatible - eventually, usage tracking might take place on the vembda side

dvargas92495 (approved):

PR #2364, opened 2025-08-24
[APO-1219] Codegen boilerplate for barebone generic nodes

NgoHarrison:


dvargas92495 (changes requested):
I added some extra thoughts to the ticket on how we could solve this app-side instead of codegen side. I'd love to maintain the invariant that codegen on the SDK. Let's sync on this to align if need be

PR #2365, opened 2025-08-25
[codegen] add file types to `WorkflowSandboxInput` union type

clopen-set:
1. Bump `vellum-ai` dependency to gain access to the new file input types
2. Add file types to the `WorkflowSandboxInput` union type. AFAICT, this is mostly a cosmetic change since our logic already uses the polymorphic `VellumValue` and just passes the raw value through to codegen

### DIFF
--- a/ee/codegen/src/types/vellum.ts
+++ b/ee/codegen/src/types/vellum.ts
@@ -5,10 +5,13 @@ import {
   LogicalOperator,
 } from "vellum-ai/api";
 import {
+  AudioInputRequest,
   ChatHistoryInput,
   ChatMessageRequest,
   ChatMessageRole,
+  DocumentInputRequest,
   FunctionDefinition,
+  ImageInputRequest,
   JsonInput,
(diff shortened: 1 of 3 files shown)


ee/codegen/src/types/vellum.ts
```diff
@@ -800,7 +804,11 @@ type WorkflowSandboxInput =
   | StringInput
   | JsonInput
   | ChatHistoryInput
-  | NumberInput;
+  | NumberInput
+  | AudioInputRequest
+  | VideoInputRequest
+  | ImageInputRequest
+  | DocumentInputRequest;
```
clopen-set, line 811: These types technically come from prompt API definitions.

If we want to use the (otherwise identical) types generated from the actual workflow APIs, we'll have to wait for the next SDK release

NgoHarrison (approved):

dvargas92495 (commented):
✅

PR #2366, opened 2025-08-26
Fix default trace id

vincent0426:
# Fix external workflow trace ID defaulting to zeros

## Summary
Fixes the issue where external workflow trace IDs were defaulting to "00000000-0000-0000-0000-000000000000" instead of proper UUIDs.

**Root Cause**: The `execution_context()` function was using `int(prev_context.trace_id)` to determine when to preserve vs. generate trace IDs, but this logic didn't properly handle the relationship between nested contexts and fresh contexts.

**Solution**: Updated the trace ID logic to:
1. Use `parent_context` presence as the primary indicator for nested contexts
2. Preserve existing trace_id when in nested contexts (and trace_id is not zero)
3. Allow explicit trace_id when starting fresh contexts
4. Generate new UUID as fallback

## Review & Testing Checklist for Human
- [ ] **Test external workflow execution end-to-end** - Verify that external workflows now generate proper UUID trace IDs instead of zeros
- [ ] **Verify nested context behavior** - Test that trace IDs are preserved correctly when contexts are nested (e.g., subworkflows, node executions)
- [ ] **Check conditional logic edge cases** - Review the 3-condition logic in `execution_context()` for potential edge cases or unintended behavior
- [ ] **Run the originally failing tests** - Confirm that the 4 tests mentioned in CI (test_context_trace_and_parent, inline_prompt_node tests, etc.) now pass consistently

### Notes
- The fix addresses both the symptom (zero trace IDs) and ensures proper trace ID propagation in nested contexts
- Only minimal test changes were made - main validation comes from existing test suite passing
- CI tests are now passing, but end-to-end external workflow testing is recommended

---
**Link to Devin run**: https://app.devin.ai/sessions/0d8fa169a5774d218abd6599a4fcfd39  
**Requested by**: @vincent0426 (vincent@vellum.ai)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -12,7 +12,7 @@ from pytest_mock import MockerFixture
 import requests_mock
 
 from vellum.client.environment import VellumEnvironment
-from vellum.workflows.context import monitoring_context_store
+from vellum.workflows.context import clear_execution_context
 from vellum.workflows.logging import load_logger
 
 
@@ -20,12 +20,12 @@ from vellum.workflows.logging import load_logger
 def clear_monitoring_context():
     """Clear the global monitoring context store between tests to prevent collisions."""
     # Clear before test
-    monitoring_context_store.clear_context()
+    clear_execution_context()
 
     yield
 
     # Clear after test
-    monitoring_context_store.clear_context()
+    clear_execution_context()
 
 
 def pytest_collection_modifyitems(session, config, items):
--- a/src/vellum/workflows/state/context.py
+++ b/src/vellum/workflows/state/context.py
@@ -41,6 +41,7 @@ class WorkflowContext:
 
         if self._execution_context.parent_context is None:
             self._execution_context.parent_context = ExternalParentContext(span_id=uuid4())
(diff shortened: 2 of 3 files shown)


src/vellum/workflows/context.py
```diff
@@ -60,7 +60,14 @@ def execution_context(
...
     prev_context = get_execution_context()
-    set_trace_id = prev_context.trace_id if int(prev_context.trace_id) else trace_id or uuid4()
+    if prev_context.parent_context is not None and prev_context.trace_id != UUID(
+        "00000000-0000-0000-0000-000000000000"
+    ):
+        set_trace_id = prev_context.trace_id
+    elif trace_id is not None:
+        set_trace_id = trace_id
+    else:
+        set_trace_id = uuid4()
```
dvargasfuertes: [blocking] I would solve this at the `WorkflowContext` level so that it's specific to when we add an `ExternalParentContext`

marlenc3 (approved):

dvargas92495 (approved):

PR #2369, opened 2025-08-26
Fix state meta

vincent0426:


src/vellum/workflows/resolvers/tests/test_resolver.py
```diff
@@ -33,17 +33,6 @@ class TestWorkflow(BaseWorkflow[BaseInputs, TestState]):
     # GIVEN a state dictionary that matches what the resolver expects
     state_dict = {
         "test_key": "test_value",
-        "meta": {
```
dvargasfuertes, line 36: [nit] I think we can keep this in the test and assert that the new state as a different id / span_id

dvargas92495 (approved):

dvargas92495 (approved):

PR #2370, opened 2025-08-26
Remove lazy reference for tool calling node

vincent0426:


dvargas92495 (approved):
yea wait why did we have this at all haha

PR #2372, opened 2025-08-27
[APO-1263] WebSearchNode Foundation

VJagiasi:
[Ticket](https://linear.app/vellum/issue/APO-1263/create-websearchnode-foundation-and-data-models)

foundational structure for WebSearchNode to enable web search capabilities in Vellum workflows. This is the first phase of implementing web search functionality that will allow users to perform Google searches via SerpAPI integration.

## Changes
  - Created `WebSearchNode` class extending `BaseNode[StateType]`
  - Defined configuration attributes: `query`, `api_key`, `num_results`, `location`
  - Added structured `Outputs` class with `text`, `urls`, and `results` fields
  - Added proper module exports to make the node available in the SDK
  - Includes comprehensive type hints and docstrings following SDK patterns

 ## Implementation Status for Phase 1 (WebSearchNode fully executable locally)
  - ✅ Foundation structure complete
  - 🔄 SerpAPI integration (next PR)
  - 🔄 Local execution support 
  - 🔄 Comprehensive testing

### DIFF
--- a/src/vellum/workflows/nodes/displayable/web_search_node/node.py
+++ b/src/vellum/workflows/nodes/displayable/web_search_node/node.py
@@ -1,0 +1,52 @@
+from typing import Any, ClassVar, Dict, List, Optional
+
+from vellum.workflows.nodes.bases import BaseNode
+from vellum.workflows.outputs import BaseOutputs
+from vellum.workflows.types.core import MergeBehavior, VellumSecret
+from vellum.workflows.types.generics import StateType
+
+
+class WebSearchNode(BaseNode[StateType]):
+    """
+    Used to perform web search using SerpAPI.
+
+    query: str - The search query to execute
+    api_key: VellumSecret - SerpAPI authentication key
+    num_results: int - Number of search results to return (default: 10)
+    location: Optional[str] - Geographic location filter for search
+    """
+
+    query: ClassVar[str] = ""
+    api_key: ClassVar[Optional[VellumSecret]] = None
+    num_results: ClassVar[int] = 10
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/nodes/displayable/web_search_node/node.py
```diff
@@ -0,0 +1,52 @@
...
+    Used to perform web search using SerpAPI.
+
+    query: str - The search query to execute
+    api_key: VellumSecret - SerpAPI authentication key
+    num_results: int - Number of search results to return (default: 10)
+    location: Optional[str] - Geographic location filter for search
+    """
+
+    query: ClassVar[str] = ""
+    api_key: ClassVar[Optional[VellumSecret]] = None
```
dvargasfuertes: [blocking] we want to start moving away from `VellumSecret` and use Environment Variable references instead. So in practice, I think that means we could type it as just `str` for now

src/vellum/workflows/nodes/displayable/web_search_node/node.py
```diff
@@ -0,0 +1,52 @@
...
+    num_results: ClassVar[int] = 10
+    location: ClassVar[Optional[str]] = None
+
+    class Trigger(BaseNode.Trigger):
+        merge_behavior = MergeBehavior.AWAIT_ANY
```
dvargasfuertes: [blocking] This is an older await strategy that we also want to move away from. We can simply remove the class here

dvargas92495 (approved):
nice - blocking comments should be quick

PR #2392, opened 2025-08-29
[APO-1318] Loosen dictwrapper getattr check used in templating nodes when attr does not exist

NgoHarrison:


src/vellum/workflows/types/code_execution_node_wrappers.py
```diff
@@ -71,7 +71,7 @@ def __getattr__(self, attr):
                 # several values as VellumValue objects, we use the "value" key to return itself
                 return self
 
-            raise AttributeError(f"dict has no key: '{attr}'")
+            return None
```
dvargasfuertes: [q] does old jinja return None or their own empty singleton? (the analog to our `undefined`)

dvargas92495 (approved):

src/vellum/workflows/types/code_execution_node_wrappers.py
NgoHarrison: Ah good call out, I should have checked. They seem to return an undefined instance (so yeah similar to ours):
```   def getitem(
        self, obj: t.Any, argument: t.Union[str, t.Any]
    ) -> t.Union[t.Any, Undefined]:
        """Get an item or attribute of an object but prefer the item."""
        try:
            return obj[argument]
        except (AttributeError, TypeError, LookupError):
            if isinstance(argument, str):
                try:
                    attr = str(argument)
                except Exception:
                    pass
                else:
                    try:
                        return getattr(obj, attr)
                    except AttributeError:
                        pass
            return self.undefined(obj=obj, name=argument)
```
and that is essentially resolving to an empty "" or evalues to a False if used in an conditional based on their dunder str and bool methods in the undefined class

NgoHarrison: Updated our undefined metaclass for this and returned undefined

tests/workflows/missing_node_output/tests/test_workflow.py
```diff
@@ -10,4 +10,4 @@ def test_workflow__happy_path():
 
     # THEN we should expect the workflow to succeed with the expected output
     assert terminal_event.name == "workflow.execution.fulfilled"
-    assert terminal_event.outputs == {"final_value": "hello undefined"}
+    assert terminal_event.outputs == {"final_value": "hello "}
```
dvargasfuertes, line 13: [oos] this is the only piece I feel alittle iffy on, but can be sold that empty string is actually better

dvargas92495 (approved):
love it

PR #2401, opened 2025-08-30
[APO-1357] Guard against deepcopy of dictwrapper

NgoHarrison:


src/vellum/workflows/state/tests/test_state.py
```diff
@@ -229,3 +231,15 @@ def side_effect(*args, **kwargs):
...
+def test_state_deepcopy_handles_undefined_values():
+    # GIVEN a state with undefined values in node outputs
+    state = MockState(foo="bar")
+    state.meta.node_outputs[MockNode.Outputs.baz] = DictWrapper({"foo": undefined})
+
+    # WHEN we deepcopy the state (this previously crashed with "undefined() takes no arguments")
+    deepcopied_state = deepcopy(state)
+
+    # THEN the undefined values are preserved
+    assert deepcopied_state.meta.node_outputs[MockNode.Outputs.baz] == {"foo": undefined}
```
vincent0426, line 245: how does this test related to the change?

NgoHarrison: I wanted to show that state still goes through. Without the changes I did, this test would crash from line 242 above where we can't deepcopy

NgoHarrison: I also mentioned this in a message with @dvargas92495 but alternatively, we could give the dictwrapper dunder deepcopy's which might be more correct but this fix was just to align with existing behavior

src/vellum/workflows/types/code_execution_node_wrappers.py
```diff
@@ -72,6 +72,12 @@ def __getattr__(self, attr):
                 # several values as VellumValue objects, we use the "value" key to return itself
                 return self
 
+            if attr.startswith("__") and attr.endswith("__"):
+                raise AttributeError(
+                    f"'{type(self).__name__}' does not expose Python protocol attribute '{attr}'."
+                    f"Only non-dunder attributes fall back to `undefined`."
+                )
```
dvargasfuertes: I like this change. I would actually just return `super().__getattr__(name)`

dvargas92495 (approved):

vincent0426 (approved):

PR #2402, opened 2025-08-30
[APO-1189] Add WebSearchNode Code Generation Support

VJagiasi:
[Ticket](https://linear.app/vellum/issue/APO-1189/add-support-for-code-generation-for-web-search-node)

## Rationale
  Implements code generation support for WebSearchNode to enable users to pull and generate Python workflow code containing web search functionality, completing the WebSearchNode implementation chain.

 ## Changes
  - **Core Configuration**: Added WebSearchNode to NODE_ATTRIBUTES in `constants.ts` with attributes: `query`, `api_key`, `num_results`, `location`
  - **Test Infrastructure**: Created `webSearchNodeFactory` for generating test node data
  - **Comprehensive Test Suite**: 8 focused test cases covering:
    - Basic node and display file generation
    - Workflow input integration
    - Different value types (NODE_OUTPUT, DICTIONARY_REFERENCE)
    - Parameter validation and edge cases
   
## Details
  - WebSearchNode uses simple attribute configuration (no complex AttributeType enums like ToolCallingNode)
  - All attributes processed through standard WorkflowValueDescriptor handling
  - Generated code imports from `vellum.workflows.nodes.displayable.web_search_node`
  - VELLUM_SECRET attributes use `vellumSecretName` field for API key references

### DIFF
--- a/ee/codegen/src/__test__/nodes/__snapshots__/web-search-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/web-search-node.test.ts.snap
@@ -34,71 +34,14 @@ exports[`WebSearchNode > basic > getNodeFile 1`] = `
 "from vellum.workflows.nodes.displayable.web_search_node import (
     WebSearchNode as BaseWebSearchNode,
 )
+from vellum.workflows.references import EnvironmentVariableReference
 
 
 class WebSearchNode(BaseWebSearchNode):
     query = "latest AI developments"
-    api_key = None
+    api_key = EnvironmentVariableReference(name="SERP_API_KEY")
     num_results = 10
     location = "United States"
-"
-`;
-
-exports[`WebSearchNode > different secret configurations > should handle different VELLUM_SECRET key names 1`] = `
-"from vellum.workflows.nodes.displayable.web_search_node import (
-    WebSearchNode as BaseWebSearchNode,
-)
-
-
-class WebSearchNode(BaseWebSearchNode):
-    query = "search with custom key"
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/nodes/__snapshots__/web-search-node.test.ts.snap
```diff
@@ -0,0 +1,129 @@
...
+)
+from vellum.workflows.references import EnvironmentVariableReference
+
+from ..inputs import Inputs
+
+
+class WebSearchNode(BaseWebSearchNode):
+    query = {
+        "search_term": Inputs.query,
+    }
```
dvargasfuertes: [q] Is this a valid shape that our Node / SerpAPI supports?

dvargas92495 (approved):

vincent0426 (approved):

ee/codegen/src/__test__/nodes/__snapshots__/web-search-node.test.ts.snap
VJagiasi: you're right - that dictionary structure isn't valid for our WebSearchNode

dug into this and confirmed:
  1. The WebSearchNode expects query to be a string (validates with isinstance(self.query, str))
  2. SerpAPI's q parameter only accepts strings, not objects
  3. We have no dict-to-string processing in the node implementation

That test case was generating invalid Python code that would fail at runtime. removed the DICTIONARY_REFERENCE test case since it represents an invalid scenario

PR #2415, opened 2025-09-03
[APO-864] Add use tool inputs decorator for code tool

vincent0426:


tests/workflows/basic_tool_calling_node_parent_input/tests/test_workflow.py
```diff
@@ -0,0 +1,125 @@
...
+    def generate_prompt_events(*_args, **_kwargs) -> Iterator[ExecutePromptEvent]:
+        execution_id = str(uuid4())
+
+        call_count = vellum_adhoc_prompt_client.adhoc_execute_prompt_stream.call_count
+        expected_outputs: List[PromptOutput]
+        if call_count == 1:
+            expected_outputs = [
+                FunctionCallVellumValue(
+                    value=FunctionCall(
+                        arguments={"populated_input": "Hello from user"},
```
vincent0426, line 37: root change

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -520,12 +520,25 @@ def create_function_node(
             },
         )
     else:
-        # For regular functions, use FunctionNode
+
+        def create_function_wrapper(func):
+            def wrapper(self, **kwargs):
+                merged_kwargs = kwargs.copy()
+                inputs = getattr(func, "_inputs", {})
```
dvargasfuertes: [nit] let's use dunder's and namespace it (`__vellum_inputs__`). It's a more standard python convention for hacking onto classes, and the namespace allow us to play nicely with other libs

dvargas92495 (approved):
very dope. We'll need to still support serialization and codegen of this new annotation. We should likely cut a `1.4.0` release when this is ready since it's a new public export (`@use_tool_inputs`)

dvargas92495 (approved):

PR #2416, opened 2025-09-03
Move virtual open function

vincent0426:


ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -58,6 +59,22 @@
     from vellum_ee.workflows.display.types import WorkflowDisplayContext
 
 
+def virtual_open(file_path: str, mode: str = "r"):
+    """
+    Open a file, checking VirtualFileFinder instances first before falling back to regular open().
+    """
+    for finder in sys.meta_path:
+        if hasattr(finder, "loader") and hasattr(finder.loader, "_get_code"):
```
dvargasfuertes: [nit] now that we're in ee, we could just do an isinstance check for `finder.loader`

dvargas92495 (approved):

NgoHarrison (approved):

PR #2419, opened 2025-09-03
Use virtual open for code execution file open

NgoHarrison:
This is needed for the serialize route

Originally surfaced from issues with saving edited code execution node files

### DIFF
--- a/ee/vellum_ee/workflows/display/nodes/vellum/code_execution_node.py
+++ b/ee/vellum_ee/workflows/display/nodes/vellum/code_execution_node.py
@@ -1,9 +1,10 @@
 import inspect
+import os
 from uuid import UUID
 from typing import ClassVar, Generic, Optional, TypeVar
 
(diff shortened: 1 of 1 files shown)


ee/vellum_ee/workflows/display/types.py
```diff
@@ -54,6 +54,7 @@ class WorkflowDisplayContext:
     _errors: List[Exception] = field(default_factory=list)
     _invalid_nodes: List[Type[BaseNode]] = field(default_factory=list)
     _dry_run: bool = False
+    generated_files: Dict[str, str] = field(default_factory=dict)
```
dvargasfuertes: [nit] instead of threading through the original files, could we use a similar approach that `virtual_open` is using here? https://github.com/vellum-ai/vellum-python-sdks/pull/2416/files#diff-57899bd571a4ab4e7aab18a151b0b0390644e4607b3fc95f4b115299704c52f8R68-R74

dvargas92495 (approved):

PR #2422, opened 2025-09-03
Serialize use tool inputs

vincent0426:


ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -422,6 +423,17 @@ def serialize_value(display_context: "WorkflowDisplayContext", value: Any) -> Js
...
+        inputs = getattr(value, "__vellum_inputs__", {})
+
+        if inputs:
+            serialized_inputs = {}
+            for param_name, input_ref in inputs.items():
+                serialized_inputs[param_name] = serialize_value(display_context, input_ref)
+
+            model_data = function_definition.model_dump()
+            model_data["inputs"] = serialized_inputs
+            function_definition = FunctionDefinition(**model_data)
```
dvargasfuertes: [blocking] why do we need to convert back to the pydantic model before we redump below?

If we must, any raw -> Pydantic model should always use `.model_validate`

dvargas92495 (approved):

ee/vellum_ee/workflows/display/utils/expressions.py
vincent0426: we dump above to add inputs type since FunctionDefinition is read only

dvargasfuertes: I suppose my point was that we could just keep it as the dump when we assign it to the dump below

vincent0426: oh then we need to handle both dict and model type below `model.` or `dict[""]`, will this one be faster to avoid serialize/redump?

dvargasfuertes: in general, always faster if it can be avoided. can extract the name/description field before doing this dump so that we no longer need the model type below

PR #2430, opened 2025-09-03
[APO-1354] Fix codegen to generate proper graphs for single-node workflows

alex-nork:
## Problem
Single-node workflows were incorrectly generating empty graphs, breaking subworkflow input generation and preventing proper execution.

## Solution
- Added `tryGetEntrypointNode()` method for safe entrypoint detection
- Modified graph generation to properly handle workflows without edges from entrypoint
- Empty workflows now generate `Graph.empty()` instead of no graph attribute
- Added proper node usage tracking to prevent incorrect unused marking

## Key Changes
- `WorkflowContext`: Added null-safe `tryGetEntrypointNode()` method
- `GraphAttribute`: Simplified graph generation logic using proper null checks
- `Graph`: Added static `empty()` method for explicit empty graph creation
- Updated codegen to only add graph attribute when non-empty

## Backward Compatibility
✅ No breaking changes - existing workflows unchanged, previously broken single-node workflows now work correctly.

---


ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -97,10 +98,31 @@ export class GraphAttribute extends AstNode {
       return graphMutableAst;
     }
 
-    const edgesQueue = this.workflowContext.getEntrypointNodeEdges();
+    let edgesQueue: WorkflowEdge[] = [];
+    try {
+      edgesQueue = this.workflowContext.getEntrypointNodeEdges();
+    } catch {
+      // No entrypoint node exists - handle case where there are standalone nodes
```
vincent0426: I think we could check by using `this.workflowContext.getEntrypointNode().id;` instead of try catch?

ee/codegen/src/generators/graph-attribute.ts
```diff
@@ -97,10 +98,31 @@ export class GraphAttribute extends AstNode {
...
+    }
+
     const edgesByPortId = this.workflowContext.getEdgesByPortId();
     const processedEdges = new Set<WorkflowEdge>();
 
+    // Handle case where there are no edges from entrypoint but entrypoint exists
+    if (edgesQueue.length === 0) {
+      const singleNodeResult = this.getSingleNodeReference();
+      if (singleNodeResult.type !== "empty") {
+        return singleNodeResult;
```
vincent0426: q: what will this happen? will it start from the node or entrypoint?

ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -68,7 +79,7 @@ from .nodes.final_output import FinalOutput
 
 
 class TestWorkflow(BaseWorkflow):
-    unused_graphs = {FinalOutput}
+    graph = FinalOutput
```
dvargasfuertes: [blocking] this looks like a regression

ee/codegen/src/__test__/__snapshots__/workflow.test.ts.snap
```diff
@@ -2,60 +2,71 @@
...
 "from vellum.workflows import BaseWorkflow
+from vellum.workflows.graph import Graph
 from vellum.workflows.state import BaseState
 
 from .inputs import Inputs
 
 
 class TestWorkflow(BaseWorkflow[Inputs, BaseState]):
-    pass
+    graph = Graph.empty()
```
dvargasfuertes: [blocking] do we need to pass this in explicitly, or could we initiate this as part of `BaseWorkflow`'s metaclass? that way, we don't add logic to simple codegen cases

ee/codegen/src/generators/graph-attribute.ts
dvargasfuertes: [nit] would +1 ^

ee/codegen/src/generators/graph-attribute.ts
dvargasfuertes: There's a codegen test above that regressed I think because of this change. 

If there are no edges, then I think wecould just return without fetching a single node

dvargas92495 (approved):

PR #2433, opened 2025-09-04
Codegen use tool decorators

vincent0426:


ee/codegen/src/types/vellum.ts
```diff
@@ -967,7 +967,7 @@ interface NameDescription {
 export type FunctionArgs = {
   type: "CODE_EXECUTION";
   src: string;
-  definition: FunctionDefinition;
+  definition: FunctionDefinition; // legacy frontend does not send definition field
```
dvargasfuertes, line 970: [q] why is this considered legacy? is this not where the new `inputs` would come in?

dvargas92495 (approved):

ee/codegen/src/types/vellum.ts
vincent0426: ahh i mean "legacy frontend". would make changes to make it optional

PR #2437, opened 2025-09-04
[APO-1372] Fix function name casing issue in Python imports

alex-nork:
The purpose of this PR is to fix a function name casing issue in Python imports that was causing conflicts. User reported camelcase tool names being converted to snakecase in generated import statements, raising an exception on import.

## Changes Made
- Use original function name instead of snake_case name in Python reference
- Maintain snake_case module import path to preserve existing functionality
- Add explanatory comments referencing APO-1372

## Alternatives Considered
- Converting all function names to snake_case consistently: Would break existing workflow UI expectations
- Using camelCase throughout: Would be inconsistent with Python conventions
- Creating name mapping: Added unnecessary complexity

This approach felt like the fix resulting in the least surprise when switching between workflow UI and SDK, as it preserves the original function names users expect while maintaining proper Python module import conventions.

## Testing
- Unit tests verified that the change resolves the casing conflict issue
- Existing tests should continue to pass

---

### DIFF
--- a/ee/codegen/src/__test__/nodes/tool-calling-node.test.ts
+++ b/ee/codegen/src/__test__/nodes/tool-calling-node.test.ts
@@ -787,80 +787,18 @@ describe("ToolCallingNode", () => {
 
   describe("function name casing (APO-1372)", () => {
     const testCases = [
-      // Core functionality - camelCase preservation
-      {
-        name: "getCWD",
-        expectedRef: "getCWD",
-        expectedImport: "get_cwd",
-        description: "camelCase function names",
-      },
-      {
-        name: "parseJSON",
-        expectedRef: "parseJSON",
-        expectedImport: "parse_json",
-        description: "mixed case function names",
-      },
-      {
-        name: "XMLHttpRequest",
-        expectedRef: "XMLHttpRequest",
-        expectedImport: "xmlhttp_request",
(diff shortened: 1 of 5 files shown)


ee/codegen/src/__test__/nodes/tool-calling-node.test.ts
```diff
@@ -784,4 +784,92 @@ describe("ToolCallingNode", () => {
       );
     });
   });
+
+  describe("function name casing (APO-1372)", () => {
```
vincent0426: we could remove ticket number

ee/codegen/src/__test__/nodes/tool-calling-node.test.ts
```diff
@@ -784,4 +784,92 @@ describe("ToolCallingNode", () => {
...
+        const nodeContext = (await createNodeContext({
+          workflowContext,
+          nodeData,
+        })) as GenericNodeContext;
+        const node = new GenericNode({ workflowContext, nodeContext });
+
+        node.getNodeFile().write(writer);
+        const output = await writer.toStringFormatted();
+
+        expect(output).toContain(expectedRef);
```
vincent0426: could we generate the snapshot directly

dvargas92495 (commented):
✅

PR #2439, opened 2025-09-04
[APO-1404] Fix undefined value serialization error

alex-nork:
The purpose of this PR is to fix undefined value serialization error in workflow display expressions.

## Changes Made
- Add handling for undefined values in expression serialization utilities
- Add comprehensive test coverage for undefined value serialization scenarios
- Ensure workflow displays properly handle cases where expression values are undefined

## Testing
- Added unit tests covering various undefined value scenarios in test_expressions.py
- Tests verify proper handling of undefined values during expression serialization

## Notes
This fix prevents serialization errors when workflow expressions contain undefined values, improving the stability of the display system.

---


ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -245,6 +246,15 @@ def serialize_key(key: Any) -> str:
...
 
 def serialize_value(display_context: "WorkflowDisplayContext", value: Any) -> JsonObject:
+    if value is undefined:
+        return {
+            "type": "CONSTANT_VALUE",
+            "value": {
+                "type": "JSON",
+                "value": None,
+            },
+        }
```
alex-nork, line 256: is this a reasonable default to keep us from erroring out?

ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -245,6 +246,15 @@ def serialize_key(key: Any) -> str:
 
 
 def serialize_value(display_context: "WorkflowDisplayContext", value: Any) -> JsonObject:
+    if value is undefined:
```
dvargasfuertes, line 253: The difference between `undefined` and `null` is that `undefined` should always be skipped from the JSON wherever it's resolved. For example:
```
{"foo": "bar", "baz": null}
```

vs

```
{"foo": "bar"}
```

alex-nork: The latest changes should address this

ee/vellum_ee/workflows/display/utils/tests/test_expressions.py
```diff
@@ -0,0 +1,27 @@
+"""Tests for expressions serialization utilities"""
+
+from vellum.workflows.constants import undefined
+from vellum_ee.workflows.display.types import WorkflowDisplayContext
+from vellum_ee.workflows.display.utils.expressions import serialize_value
+
+
+class TestSerializeValue:
```
vincent0426: I think we should use function-based testing?

ee/vellum_ee/workflows/display/utils/tests/test_expressions.py
```diff
@@ -0,0 +1,27 @@
+"""Tests for expressions serialization utilities"""
```
vincent0426: could remove this

ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -244,7 +245,14 @@ def serialize_key(key: Any) -> str:
         return str(key)
 
 
+# Sentinel value to indicate a value should be omitted from serialization
+_UNDEFINED_SENTINEL: JsonObject = {"__undefined__": True}
```
dvargasfuertes, line 249: [nit] Do we need a new undefined sentinel or could we just reuse our existing one: `undefined` ?

dvargas92495 (approved):

ee/vellum_ee/workflows/display/utils/expressions.py
alex-nork: I preferred this approach because it keeps the `serialize_value` function return type a clean `JsonObject` rather than `undefined` or `JsonObject`

dvargasfuertes: it does mean that the user now can't generate this object in their workflow (however unlikely that is)

PR #2440, opened 2025-09-04
[APO-1443] Add custom process parameters method

vincent0426:


src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -398,6 +399,11 @@ def create_tool_prompt_node(
                 "prompt_inputs": node_prompt_inputs,
                 "parameters": parameters,
                 "max_prompt_iterations": max_prompt_iterations,
+                "process_parameters": (
+                    process_parameters_method
+                    if process_parameters_method is not None
+                    else ToolPromptNode.process_parameters
+                ),
```
dvargasfuertes: [blocking] I don't think we want to add `ToolPromptNode.process_parameters` to the class' dict's if the method is None. We should make sure it's not in the dict at all

dvargas92495 (approved):

PR #2441, opened 2025-09-05
[APO-1444] Support custom process blocks method

vincent0426:


src/vellum/workflows/nodes/displayable/tool_calling_node/node.py
```diff
@@ -140,6 +140,7 @@ class Outputs(BaseWorkflow.Outputs):
     def _build_graph(self) -> None:
         # Get the process_parameters method if it exists on this class
         process_parameters_method = getattr(self.__class__, "process_parameters", None)
```
vincent0426, line 142: maybe we could pass down class and parsed in build graph instead of here?

dvargasfuertes: 0/5 preference on my end

dvargas92495 (approved):

PR #2454, opened 2025-09-07
Resolve leftover TODO's from Map Node work

dvargas92495:
Figured out the issues that were tormenting me here: https://github.com/vellum-ai/vellum-python-sdks/pull/2444

### DIFF
--- a/src/vellum/workflows/executable.py
+++ b/src/vellum/workflows/executable.py
@@ -1,0 +1,9 @@
+from uuid import UUID
+from typing import Dict
+
+from vellum.workflows.utils.uuids import uuid4_from_hash
+
+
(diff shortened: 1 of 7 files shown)


src/vellum/workflows/executable.py
```diff
@@ -0,0 +1,9 @@
+from uuid import UUID
+from typing import Dict
+
+from vellum.workflows.utils.uuids import uuid4_from_hash
+
+
+class BaseExecutable:
```
dvargasfuertes, line 7: This is interesting, as it means we could start hoisting up logic common to nodes and workflows

vincent0426 (commented):
✅

PR #2458, opened 2025-09-08
Fix default parent class outputs

vincent0426:
an issue while trying to start python workflow server

```bash
self_outputs_class.__parent_class__.__qualname__ == other_outputs_class.__parent_class__.__qualname__
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Field' object has no attribute '__qualname__'. Did you mean: '__set_name__'?
```

### DIFF
--- a/src/vellum/workflows/outputs/base.py
+++ b/src/vellum/workflows/outputs/base.py
@@ -1,5 +1,5 @@
 import inspect
-from typing import Any, Dict, Generic, Iterator, Optional, Set, Tuple, Type, TypeVar, Union, cast
+from typing import Any, Dict, Generic, Iterator, Set, Tuple, Type, TypeVar, Union, cast
 from typing_extensions import dataclass_transform
 
 from pydantic import GetCoreSchemaHandler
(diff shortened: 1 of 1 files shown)


src/vellum/workflows/outputs/base.py
```diff
@@ -100,6 +100,10 @@ def __hash__(self) -> int:
 
 @dataclass_transform(kw_only_default=True)
 class _BaseOutputsMeta(type):
+    def __new__(cls, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any:
+        dct["__parent_class__"] = type(None)
```
dvargasfuertes, line 104: [nit] L124 we do an `is None` check so would be cool with either or. Could also see us defaulting to `BaseExecutable()` on the class itself instead `field` but that might get weird

dvargas92495 (approved):
thanks for catching!

dvargas92495 (approved):

PR #2472, opened 2025-09-09
[APO-1440] Serialize generic block

vincent0426:


ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -157,12 +161,18 @@ def serialize(self, display_context: "WorkflowDisplayContext", **kwargs: Any) ->
                 if self.attribute_ids_by_name.get(attribute.name)
                 else str(uuid4_from_hash(f"{node_id}|{attribute.name}"))
             )
+
+            value = attribute.instance
+            if attribute.name == "blocks":
```
dvargasfuertes, line 166: [blocking] we don't ever want to discriminate based on the attribute's name, only their types

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_basic_tool_calling_node_inline_workflow_serialization.py
```diff
@@ -114,7 +114,7 @@ def test_serialize_workflow():
                                                 "block_type": "VARIABLE",
                                                 "state": None,
                                                 "cache_config": None,
-                                                "input_variable": "question",
+                                                "input_variable": "d5c4d578-6ef1-4786-88f6-1ab0892d0798",
```
dvargasfuertes, line 117: [blocking] I would love to try to solve this problem without having to store `id`'s here. Happy to sync on this

PR #2476, opened 2025-09-11
Fix dictionary id serialization

vincent0426:


alex-nork (approved):
looks good

ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -378,7 +378,7 @@ def serialize_value(executable_id: UUID, display_context: "WorkflowDisplayContex
             if serialized_val != _UNDEFINED_SENTINEL:
                 serialized_entries.append(
                     {
-                        "id": str(uuid4_from_hash(f"{key}|{val}")),
+                        "id": str(uuid4_from_hash(f"{executable_id}|{key}")),
```
dvargasfuertes, line 381: [oos] technically, would be great to get the attribute path in here (in this case, `prompt_inputs`)

If that's too annoying to also pipe through, disregard and let's just roll with this. I'm optimistic of a future where we could eventually deprecate the dict ref id's

dvargas92495 (approved):

PR #2481, opened 2025-09-12
Move function definition serialization

vincent0426:


src/vellum/workflows/types/definition.py
```diff
@@ -100,6 +107,33 @@ def deployment_name(self) -> Optional[str]:
...
         return None
 
+    @model_serializer(mode="wrap")
+    def _serialize(self, handler, info: SerializationInfo):
+        """Allow Pydantic to serialize directly given a `client` in context.
+
+        Falls back to the default serialization when client is not provided.
+        """
+        context = info.context if info and hasattr(info, "context") else {}
+        client = cast(Optional[Vellum], context.get("client") if context else None)
```
vincent0426: casting to make it Vellum | None explicitly otherwise it can be `Any`

dvargasfuertes: [nit] Instead of `cast`, always prefer `isinstance` checks

dvargas92495 (approved):

PR #2504, opened 2025-09-13
Add uuid utils for serializing variable prompt blocks

vincent0426:


tests/client/utils/test_utils.py
```diff
@@ -0,0 +1,34 @@
+import uuid
+from unittest.mock import Mock
+from vellum.client.utils import convert_input_variable_to_uuid
+
+
+def test_convert_input_variable_to_uuid_with_valid_uuid():
```
dvargasfuertes, line 6: [nit] I think we'll want to add this file too to the fern ignore

dvargas92495 (approved):

tests/client/utils/test_utils.py
vincent0426: will fern override this path?

dvargasfuertes: I think they have control of `tests/client`

PR #2506, opened 2025-09-13
Fix flaky test by increasing sleep time

vincent0426:


tests/workflows/basic_vellum_emitter_workflow/tests/test_vellum_emitter_debounce.py
```diff
@@ -97,7 +97,7 @@ def test_vellum_emitter_debounce_timer_reset():
     """
     Test that debounce timer resets when new events arrive.
     """
-    emitter = VellumEmitter(debounce_timeout=0.2)
+    emitter = VellumEmitter(debounce_timeout=0.5)
```
dvargasfuertes: I think we want the sleep to be longer than the debounce here for this to work consistently

vincent0426: shouldn't the first two be < and all be > ?

vincent0426: or last sleep > 0.5?

dvargasfuertes: not sure what you're referring to by `first two` and `all` and `last`, but the spirit of this test is that the vellum emitter will flush the events after `debounce_timeout` seconds, at which point, we then expect `mock_client.events.create` to be called

PR #2508, opened 2025-09-14
Fix compile workflow deployment function

vincent0426:


ee/vellum_ee/workflows/display/nodes/vellum/inline_prompt_node.py
```diff
@@ -163,9 +163,7 @@ def _generate_function_tools(
         elif callable(function):
             normalized_functions = compile_function_definition(function)
         elif isinstance(function, DeploymentDefinition):
-            normalized_functions = compile_workflow_deployment_function_definition(
-                function.model_dump(), display_context.client
-            )
+            normalized_functions = compile_workflow_deployment_function_definition(function, display_context.client)
```
vincent0426, line 166: not dumping here

src/vellum/workflows/utils/functions.py
```diff
@@ -238,25 +225,21 @@ def compile_inline_workflow_function_definition(workflow_class: Type["BaseWorkfl
...
-    release_tag = deployment_config["release_tag"]
+    release_info = deployment_definition.get_release_info(vellum_client)
 
-    workflow_deployment_release = vellum_client.workflow_deployments.retrieve_workflow_deployment_release(
-        deployment, release_tag
-    )
-
-    input_variables = workflow_deployment_release.workflow_version.input_variables
-    description = workflow_deployment_release.description
+    name = release_info["name"]
```
vincent0426, line 241: use the name returned from the api which i think would be better than using `deployment` since it could be id

dvargas92495 (approved):

PR #2524, opened 2025-09-16
Fix Codegen of Generic Node Display Data

noanflaherty:
Generic Nodes have a different JSON schema for their display data as compared to legacy node types. As such, they need their own display data codegen class (unless I'm missing something).

The root bug I'm trying to fix is that codegen isn't including color/icon for Custom Nodes.

### DIFF
--- a/ee/codegen/src/types/vellum.ts
+++ b/ee/codegen/src/types/vellum.ts
@@ -183,8 +183,8 @@ export interface NodeDisplayData {
   width?: number;
   height?: number;
   comment?: NodeDisplayComment;
-  icon?: string;
-  color?: string;
+  icon?: string | null;
+  color?: string | null;
 }
 
 export interface WorkflowEdgeDisplayData {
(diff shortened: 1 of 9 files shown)


ee/codegen/src/types/vellum.ts
```diff
@@ -717,6 +717,8 @@ export interface ErrorNode extends BaseDisplayableWorkflowNode {
 export interface GenericNodeDisplayData {
   position?: { x: number; y: number };
   comment?: NodeDisplayComment;
+  color?: string | null;
+  icon?: string | null;
```
dvargasfuertes, line 722: the two seem pretty similar - was it just that we didn't have any generic node display data support?

dvargas92495 (approved):

ee/codegen/src/types/vellum.ts
noanflaherty: Yeah exactly

vincent0426 (approved):

NgoHarrison (approved):

PR #2533, opened 2025-09-16
[APO-1367] Validate mismatced output types for final output nodes

NgoHarrison:


src/vellum/workflows/nodes/displayable/final_output_node/node.py
```diff
@@ -27,6 +28,9 @@ def __new__(mcs, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any
             **annotations,
             "value": parent.get_output_type(),
         }
+
+        mcs._validate_output_type_consistency(parent)
```
dvargasfuertes: [blocking] I thought we landed on these methods being called during serialize instead of on import? On import prevents workflows from even running at all, even if its unused

src/vellum/workflows/nodes/displayable/final_output_node/node.py
```diff
@@ -38,6 +42,60 @@ def get_output_type(cls) -> Type:
         else:
             return all_args[1]
 
+    @classmethod
+    def _validate_output_type_consistency(mcs, cls: Type) -> None:
```
dvargasfuertes, line 47: [nit] do we still plan on each node defining a `__validate__()` method or something like that? the final output node's validate method could then call this guy

src/vellum/workflows/nodes/displayable/final_output_node/node.py
NgoHarrison: Updated

src/vellum/workflows/nodes/displayable/final_output_node/node.py
NgoHarrison: Yeah this was more so a result of going the on import route instead of serialize route. Added

ee/vellum_ee/workflows/display/nodes/vellum/tests/test_final_output_node.py
```diff
@@ -0,0 +1,76 @@
...
+
+        class Outputs(BaseWorkflow.Outputs):
+            final_result = BadOutput.Outputs.value
+
+    # WHEN we attempt to serialize the workflow
+    workflow_display = get_workflow_display(workflow_class=MyWorkflow)
+
+    # THEN a ValueError should be raised during serialization due to validation
+    with pytest.raises(ValueError) as exc_info:
+        workflow_display.serialize()
```
dvargasfuertes: [blocking] let's ensure we still create the json data, but that the error is found in `workflow_display.errors` after we serialize

ee/vellum_ee/workflows/display/nodes/vellum/final_output_node.py
```diff
@@ -23,6 +23,9 @@ def serialize(self, display_context: WorkflowDisplayContext, **_kwargs) -> JsonO
         node = self._node
         node_id = self.node_id
 
+        # Validate node configuration before serialization
+        node.__validate__()
```
dvargasfuertes: [nit] I think we could `validate` all of the nodes at the workflow or base node level so that this is generic. In general, we want to stray away and resist adding additional logic to specific node display's (and codegen's) from this point forward

dvargas92495 (approved):

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -196,6 +196,12 @@ def serialize(self) -> JsonObject:
             node_display = self.display_context.node_displays[node]
 
             try:
+                try:
+                    node.__validate__()
+                except ValueError as validation_error:
```
dvargasfuertes, line 201: [oos] I think we should create our own exception instead of `ValueError` to catch here, since `ValueError` can come from all sorts of underlying libs/impls

dvargas92495 (commented):
✅

PR #2534, opened 2025-09-16
Add dynamic under project

vincent0426:
https://python-poetry.org/docs/pyproject/#version

running into some issues while trying to use local sdk in AB

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [project]
 name = "vellum-ai"
+dynamic = ["version"]
 
 [tool.poetry]
 name = "vellum-ai"


dvargas92495:
does this affect our publishing at all? what's the error in AB?

vincent0426:
```
error: Failed to generate package metadata for `vellum-ai @ directory+../vellum-python-sdks`
  Caused by: Failed to parse: `vellum-python-sdks/pyproject.toml`
  Caused by: TOML parse error at line 1, column 1
  |
1 | [project]
  | ^^^^^^^^^
`pyproject.toml` is using the `[project]` table, but the required `project.version` field is neither set nor present in the `project.dynamic` list
make: *** [run-chat] Error 2 
```

dvargas92495 (approved):
I have a sneaky suspicion that `tool.poetry.[name|version]` is redundant with `project.[name|version]`, and we may be better off with both living in just the `project` table instead. am equally worried about regressions in that approach.

Feel free to do whatever of the two seems the most promising, and we could adjust on the next release

PR #2536, opened 2025-09-17
Codegen empty inputs file

vincent0426:


dvargas92495 (changes requested):
We don't want to codegen it if it's empty. Merged the fix for the sandbox issue you flagged here: https://github.com/vellum-ai/vellum-python-sdks/pull/2535

PR #2541, opened 2025-09-18
Support streaming settings for Agent Node

vincent0426:


dvargas92495 (approved):
Nice. Let's be sure this defaults to false in the UI as well

PR #2549, opened 2025-09-19
[APO-1634] Add VellumIntegrationToolDefinition with placeholders

alex-nork:
The purpose of this PR is to add VellumIntegrationToolDefinition type with placeholder implementations for future Vellum integration support.

## Changes Made
- Add VellumIntegrationToolDefinition class with provider, integration, and tool_name fields
- Add VellumIntegrationProviderType enum to constants
- Add NotImplementedError placeholders in inline_prompt_node and tool_calling_node for [APO-1636](https://linear.app/vellum/issue/APO-1636/implement-vellumintegrationservice)
- Create compile_vellum_integration_tool_definition function stub for future implementation

## Notes
- Full implementation will be completed in [APO-1636](https://linear.app/vellum/issue/APO-1636/implement-vellumintegrationservice)

---

### DIFF
--- a/src/vellum/workflows/constants.py
+++ b/src/vellum/workflows/constants.py
@@ -1,4 +1,4 @@
-from enum import Enum, StrEnum
+from enum import Enum
 from typing import Any, cast
 
 
@@ -60,5 +60,5 @@ class AuthorizationType(Enum):
     API_KEY = "API_KEY"
 
 
-class VellumIntegrationProviderType(StrEnum):
+class VellumIntegrationProviderType(Enum):
     COMPOSIO = "COMPOSIO"
--- a/src/vellum/workflows/constants.py
+++ b/src/vellum/workflows/constants.py
@@ -1,4 +1,4 @@
-from enum import Enum
+from enum import Enum, StrEnum
 from typing import Any, cast
 
 
@@ -58,3 +58,7 @@ class APIRequestMethod(Enum):
 class AuthorizationType(Enum):
     BEARER_TOKEN = "BEARER_TOKEN"
     API_KEY = "API_KEY"
+
+
+class VellumIntegrationProviderType(StrEnum):
+    COMPOSIO = "COMPOSIO"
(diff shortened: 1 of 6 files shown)


src/vellum/workflows/types/definition.py
```diff
@@ -165,6 +165,24 @@ def model_post_init(self, __context: Any):
...
+    type: Literal["INTEGRATION"] = "INTEGRATION"
+
+    # Core identification
+    provider: VellumIntegrationProviderType
+    integration: str  # "GITHUB", "SLACK", etc.
+    tool_name: str  # Specific action like "GITHUB_CREATE_AN_ISSUE"
+
+    # Fields for tool base consistency
+    description: str
+    name: str = ""
```
noanflaherty: Do we really need these? If we do, the we should remove `tool_name` in favor of `name`. And ideally `description` would be loaded at runtime when we call `get_tool_by_name` and would not be statically defined like this.

noanflaherty (approved):

src/vellum/workflows/types/definition.py
alex-nork: I checked with @vincent0426 earlier and he confirmed they're necessary. I can make the change to `name`. Agreed on loading description at runtime being ideal but will leave as is for now

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -533,5 +549,7 @@ def get_function_name(function: ToolBase) -> str:
     elif isinstance(function, ComposioToolDefinition):
         # model post init sets the name to the action if it's not set
         return function.name  # type: ignore[return-value]
+    elif isinstance(function, VellumIntegrationToolDefinition):
```
vincent0426, line 552: do we need that type ignore here?

alex-nork: good catch. shouldn't be necessary anymore after changing the field name on the definition class

src/vellum/workflows/constants.py
```diff
@@ -58,3 +58,7 @@ class APIRequestMethod(Enum):
 class AuthorizationType(Enum):
     BEARER_TOKEN = "BEARER_TOKEN"
     API_KEY = "API_KEY"
+
+
+class VellumIntegrationProviderType(Enum):
+    COMPOSIO = "COMPOSIO"
```
dvargasfuertes, line 64: [oos] I would expect us to be able to replace this with the client sdk imports once we publish those API's

dvargas92495 (commented):
✅

PR #2556, opened 2025-09-20
Support string type in type map

vincent0426:
https://github.com/vellum-ai/vellum-copilot/pull/292

### DIFF
--- a/src/vellum/workflows/utils/functions.py
+++ b/src/vellum/workflows/utils/functions.py
@@ -22,7 +22,7 @@ from vellum.workflows.utils.vellum_variables import vellum_variable_type_to_open
 if TYPE_CHECKING:
     from vellum.workflows.workflows.base import BaseWorkflow
 
-type_map = {
+type_map: dict[Any, str] = {
     str: "string",
     int: "integer",
     float: "number",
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/utils/functions.py
```diff
@@ -32,6 +32,14 @@
     None: "null",
     type(None): "null",
     inspect._empty: "null",
+    # Add string versions for __future__ import annotations support
```
dvargasfuertes: [q] instead of outlining all of these explicitly, is there a way to load them and index the existing type_map, like `type("str")` or something like that?

dvargas92495 (approved):

src/vellum/workflows/utils/functions.py
```diff
@@ -34,6 +34,12 @@
     inspect._empty: "null",
 }
 
+for k, v in list(type_map.items()):
+    if isinstance(k, type):
+        type_map[k.__name__] = v
+    # None is not a type, but we want to map it to "null"
+    type_map["None"] = "null"
```
dvargasfuertes: we can just set this directly on the map since it's unrelated to the forloop

dvargas92495 (approved):

PR #2590, opened 2025-09-24
[APO-1639] Add VELLUM_INTEGRATION tool type support and serialization

alex-nork:
The purpose of this PR is to add comprehensive support for VELLUM_INTEGRATION tool types across codegen, runtime, and serialization.

## Demo
Please ignore the function details since the integrations microservice isn't set up to return them yet...but it does codegen!
Screenshot 2025-09-24 at 3 15 28 PM

## Changes Made
- Add VELLUM_INTEGRATION tool type support to TypeScript codegen templates
- Implement runtime execution support for VellumIntegrationToolDefinition 
- Add serialization tests for tool calling nodes with Vellum integration tools
- Update utility functions to handle Vellum integration tool name generation
- Add comprehensive test coverage for new functionality

## Testing
- Added unit tests for VellumIntegrationToolDefinition creation and validation
- Added serialization tests to verify proper workflow display serialization
- Updated tool calling node utility tests with Vellum integration scenarios
- All existing tests continue to pass

---

### DIFF
--- a/ee/codegen/src/generators/nodes/generic-node.ts
+++ b/ee/codegen/src/generators/nodes/generic-node.ts
@@ -390,7 +390,9 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
                     }),
                     python.methodArgument({
                       name: "name",
-                      value: python.TypeInstantiation.str(integrationTool.name || "UNKNOWN"),
+                      value: python.TypeInstantiation.str(
+                        integrationTool.name || "UNKNOWN"
+                      ),
                     }),
                     python.methodArgument({
                       name: "description",
--- a/ee/codegen/src/generators/nodes/generic-node.ts
+++ b/ee/codegen/src/generators/nodes/generic-node.ts
@@ -382,7 +382,7 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
                   const args = [
                     python.methodArgument({
                       name: "provider",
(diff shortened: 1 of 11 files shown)


vincent0426 (approved):

ee/codegen/src/generators/nodes/generic-node.ts
```diff
@@ -369,13 +371,55 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
...
                 }
+                case "VELLUM_INTEGRATION": {
+                  const integrationTool =
+                    f as VellumIntegrationToolFunctionArgs;
+
+                  const args = [
+                    python.methodArgument({
+                      name: "provider",
+                      value: python.TypeInstantiation.str(
+                        integrationTool.provider || "COMPOSIO"
```
noanflaherty, line 382: Do we really want all of these default values? It shouldn't be null in the first place, yeah? Also, if we do, we probs want `??` instead of `||` so as to handle empty strings.

noanflaherty (approved):

ee/codegen/src/generators/nodes/generic-node.ts
alex-nork: will follow up to remove these. I was using them to help bridge the gap between what the UI was receiving integrations-wise and what we want to store on the node itself while manually testing these changes.

dvargas92495:
✅

PR #2596, opened 2025-09-24
Add results output for tool calling node from prompt node

vincent0426:
later use for streaming results from prompt node


dvargas92495 (changes requested):
Wait, do we need a new `results` output? I thought we aligned on streaming from `chat_history`?

vincent0426:
oh correct we should only need results for internal use from prompt node

dvargas92495:
Wait why did we merge this when it was still in [blocking] ?

PR #2602, opened 2025-09-26
WIP State Locking and local copies

marlenc3:
from thread from suggestic - not going to merge until I get some testing on this done and get some extra eyes

### DIFF
--- a/src/vellum/workflows/state/base.py
+++ b/src/vellum/workflows/state/base.py
@@ -87,7 +87,9 @@ class _SnapshottableDict(dict, _Snapshottable):
     def __deepcopy__(self, memo: Any) -> "_SnapshottableDict":
         y: dict = {}
         memo[id(self)] = y
-        for key, value in self.items():
+        # Create a snapshot to avoid "dictionary changed size during iteration"
+        items_snapshot = list(self.items())
+        for key, value in items_snapshot:
             y[deepcopy(key, memo)] = deepcopy(value, memo)
 
         y = _SnapshottableDict(y)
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/state/base.py
```diff
@@ -87,7 +87,9 @@ def __setitem__(self, key: Any, value: Any) -> None:
     def __deepcopy__(self, memo: Any) -> "_SnapshottableDict":
         y: dict = {}
         memo[id(self)] = y
-        for key, value in self.items():
+        # Create a snapshot to avoid "dictionary changed size during iteration"
+        items_snapshot = list(self.items())
```
dvargasfuertes, line 91: we should fetch a list of keys for each of these instead of snapshotting a list of items

src/vellum/workflows/state/base.py
```diff
@@ -536,14 +542,16 @@ def __iter__(self) -> Iterator[Tuple[Any, Any]]:
         on an instance of this class.
         """
 
-        # If the user sets a default value on state (e.g. something = "foo"), it's not on `instance_attributes` below.
-        # So we need to include class_attributes here just in case
-        class_attributes = {key: value for key, value in self.__class__.__dict__.items() if not key.startswith("_")}
-        instance_attributes = {key: value for key, value in self.__dict__.items() if not key.startswith("__")}
+        # Use lock to create a consistent snapshot of the state
+        with self.__lock__:
```
dvargasfuertes, line 546: would love to avoid locks if we can since I'm not sure what the deadlocking implications could be if not careful.

src/vellum/workflows/types/utils.py
```diff
@@ -150,14 +150,30 @@ def deepcopy_with_exclusions(
...
 
     exclusions = exclusions or {}
 
-    for key, value in obj.__dict__.items():
-        if key in exclusions:
-            continue
-        new_instance.__dict__[key] = deepcopy(value, memo)
+    # Check if object has a lock for thread-safe copying
+    obj_lock = getattr(obj, "__lock__", None)
+    if obj_lock is not None:
```
dvargasfuertes, line 158: would love to start with the `keys()` change mentioned above to see if it helps. This logic is all try/except'd so low urgency on more sweeping changes here involving lock's

dvargas92495:
Merged a slim down version of this PR here to further focus on Triggers this week: https://github.com/vellum-ai/vellum-python-sdks/pull/2663

PR #2619, opened 2025-10-01
Rename VellumIntegrationToolDefinition.integration to integration_name

alex-nork:
The purpose of this PR is to fix a field name mismatch between the Python SDK and frontend JSON format that was causing integration names to appear as 'UNKNOWN' in API requests.

## Changes Made
- Rename `integration` field to `integration_name` in `VellumIntegrationToolDefinition` model
- Update service layer and node utils to use the new field name
- Update TypeScript codegen types and generator to output correct field name
- Update all tests and fixtures to use `integration_name`

## Testing
- Updated unit tests and serialization tests
- Verified field naming consistency across Python SDK and TypeScript codegen

## Notes
The field was incorrectly named 'integration' which caused deserialization to fail when receiving 'integration_name' from the UI. When the field was missing, empty values were being interpreted as 'UNKNOWN' by the backend API.

---

### DIFF
--- a/ee/codegen/src/generators/nodes/generic-node.ts
+++ b/ee/codegen/src/generators/nodes/generic-node.ts
@@ -383,9 +383,9 @@ export class GenericNode extends BaseNode<GenericNodeType, GenericNodeContext> {
                       ),
                     }),
                     python.methodArgument({
-                      name: "integration",
+                      name: "integration_name",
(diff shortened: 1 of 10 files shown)


emmiekehoe (commented):
🙏

dvargas92495:
✅

PR #2664, opened 2025-10-06
[APO-1832] Add BaseTrigger and ManualTrigger for workflow triggering

alex-nork:
The purpose of this PR is to implement a foundational trigger system that enables workflows to be initiated from multiple sources beyond manual invocation.

## Changes Made
- Add `BaseTrigger` abstract class with `>>` operator support for graph integration
- Add `BaseTriggerMeta` metaclass to enable class-level syntax (consistent with nodes)
- Implement `ManualTrigger` as the default trigger representing explicit workflow invocation
- Create `TriggerEdge` class to represent Trigger >> Node connections in the workflow graph
- Extend `Graph` class with trigger edge tracking, new factory methods (`from_trigger_edge`, `from_trigger_edges`), and trigger iteration properties
- Add validation in `Port` and `Graph.__rshift__` to prevent invalid `Node >> Trigger` patterns
- Add comprehensive test coverage with 5 new tests covering trigger-to-node, trigger-to-set, trigger-to-graph, chaining, and error cases

## Architecture
The design follows the >> operator pattern established in the graph system, with triggers working just like nodes (no instantiation required for simple cases):

```python
# Class-level usage (no parentheses, like nodes)
ManualTrigger >> MyNode
ManualTrigger >> {NodeA, NodeB}
ManualTrigger >> NodeA >> NodeB

# Instance-level usage (for configured triggers)
ScheduledTrigger(cron="0 * * * *") >> MyNode
SlackTrigger(channel="alerts") >> MyNode
```

Type-safe implementation with full mypy validation and proper error messages for invalid usage.

## Testing
- All 30 graph tests pass (25 existing + 5 new trigger tests)

## Future Extensions
This foundation supports planned extensions:
- `IntegrationTrigger` base class for external services (Slack, GitHub, webhooks)
- `ScheduledTrigger` for cron/time-based execution
- `TriggerReference` system for accessing trigger attributes (e.g., `SlackTrigger.channel`)

---


src/vellum/workflows/triggers/base.py
```diff
@@ -0,0 +1,147 @@
...
+                    edges.append(TriggerEdge(self, item))
+                elif isinstance(item, Graph):
+                    # Trigger >> {Graph1, Graph2}
+                    for entrypoint in item.entrypoints:
+                        edges.append(TriggerEdge(self, entrypoint))
+                else:
+                    raise TypeError(
+                        f"Cannot connect trigger to {type(item).__name__}. " f"Expected BaseNode or Graph in set."
+                    )
+            return Graph.from_trigger_edges(edges)
```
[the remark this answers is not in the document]
alex-nork: @codex address this feedback

alex-nork:
@codex review

src/vellum/workflows/triggers/base.py
```diff
@@ -0,0 +1,195 @@
...
```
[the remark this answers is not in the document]
alex-nork: @codex address this feedback

src/vellum/workflows/edges/__init__.py
```diff
@@ -1,5 +1,7 @@
 from .edge import Edge
+from .trigger_edge import TriggerEdge
```
dvargasfuertes, line 2: [blocking] This might make sense, but I'd like to rule out that our current Edge can't be used first. What do we currently perceive as the blocker preventing `Edge` from being able to handle `Trigger`'s as a source?

src/vellum/workflows/graph/graph.py
```diff
@@ -38,6 +40,7 @@ def __rshift__(self, other: "GraphTarget") -> "Graph":
     "Port",
     "Graph",
     "NoPortsNode",
+    "BaseTrigger",
```
dvargasfuertes: [blocking] I wouldn't expect BaseTrigger to be a valid `GraphTarget`, since we don't want to allow `Node >> Trigger`

src/vellum/workflows/ports/port.py
```diff
@@ -61,6 +61,25 @@ def edges(self) -> Iterator[Edge]:
...
+                f"Triggers must be at the start of a graph path, not as targets. "
+                f"Did you mean: {other.__name__} >> {self.node_class.__name__}?"
+            )
+
+        # Check if other is a trigger instance
+        if isinstance(other, BaseTrigger):
+            raise TypeError(
+                f"Cannot create edge targeting trigger {other.__class__.__name__}. "
+                f"Triggers must be at the start of a graph path, not as targets. "
+                f"Did you mean: {other.__class__.__name__}() >> {self.node_class.__name__}?"
```
dvargasfuertes: [blocking] can we keep working with just BaseTrigger classes instead of instances? just to keep consistent with how our nodes work

src/vellum/workflows/triggers/base.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Notes:
+            - ManualTrigger returns empty dict (no trigger-provided inputs)
+            - IntegrationTrigger subclasses return parsed payload data
+            - ScheduledTrigger might return timestamp/schedule metadata
+        """
+        return {}
+
+    @property
+    def trigger_type(self) -> str:
```
dvargasfuertes: [q] what do we imagine using this for?

src/vellum/workflows/triggers/base.py
```diff
@@ -0,0 +1,207 @@
...
+    def __eq__(self, other: object) -> bool:
+        """Triggers are equal if they're the same class."""
+        if not isinstance(other, BaseTrigger):
+            return False
+        return self.__class__ == other.__class__
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}()"
+
+    def get_trigger_inputs(self) -> Dict[str, Any]:
```
dvargasfuertes: [nit] would be nice to say `get_trigger_attributes`, would be nice to keep consistent

dvargasfuertes: oh actually, [blocking] we don't want any top level attributes here in the base as it could conflict with the trigger's own attributes. What would we need this for? Downstream nodes will use Trigger data by doing things like:
```
class MyPrompt(BaseNode):
     blocks = [SlackTrigger.message]
```

dvargas92495 (commented):
got some thoughts, happy to discuss synchronously too!

src/vellum/workflows/edges/__init__.py
alex-nork: my thought with the separation is that Triggers aren't Nodes and don't have Ports, so creating a dedicated edge makes it easier to separate the concepts.

alex-nork: ~~but happy to keep it simple and and start by reusing `Edge`~~ actually I think it might be a pain to try using Edge for both. for instance on `Edge` we now have to change `from_port` to something like `from_source` whose type is a union of `Port` and `BaseTrigger`.

dvargasfuertes: Ok I think I buy it. This will likely have serialization and codegen implications down the road, but nothing that's unmanageable

src/vellum/workflows/graph/graph.py
```diff
@@ -96,12 +101,79 @@ def from_set(targets: GraphTargetOfSets) -> "Graph":
...
+    @staticmethod
+    def from_trigger_edges(edges: List[TriggerEdge]) -> "Graph":
+        """
+        Create a graph from multiple TriggerEdges (e.g., Trigger >> {NodeA, NodeB}).
+
+        Args:
+            edges: List of TriggerEdges
+
+        Returns:
+            Graph with all trigger edges and target nodes' ports as entrypoints/terminals
```
dvargasfuertes, line 140: [oos] we may need to redefine what we mean by `entrypoints` in this world with trigger (or maybe even nuke it). Entrypoints in the old world are just TriggerEdge's from `ManualTrigger` to `Node` in the new world.

dvargas92495 (approved):
we're off to the races 🏃

PR #2677, opened 2025-10-06
Fix router node port ordering for multiple MCP tools

alex-nork:
The purpose of this PR is to fix port validation errors when ToolCallingNode uses MCP servers that expose multiple tools.

## Problem
When an MCP server exposes multiple tools, the router node was creating all tool ports as `Port.on_if()` conditions. This violates port validation rules which require the pattern: one `on_if`, zero or more `on_elif`, and optionally one `on_else`.

The error manifested as:
```
Class vellum.workflows.nodes.displayable.tool_calling_node.utils.RouterNode.Ports 
must have ports in the following order: on_if, on_elif, on_else
```

## Root Cause
The `create_router_node` function in `utils.py` performs tool discovery on MCP servers, creating one port per discovered tool. Previously, all tool ports were created with `Port.on_if()`, leading to multiple `on_if` ports in sequence.

## Changes Made
- Modified `create_port_condition` to accept an `is_first` parameter
- First tool port now uses `Port.on_if()`, subsequent ports use `Port.on_elif()`
- Applied this logic consistently across all tool types (Composio, VellumIntegration, MCP, and regular functions)
- Added `test_router_node_port_ordering_with_multiple_tools` to verify port ordering and validation

## Testing
- ✅ New test validates port ordering: IF → ELIF → ELIF → ELSE
- ✅ Test fails without the fix (assertion: 'IF' == 'ELIF' on second port)
- ✅ All 41 related tests pass (tool calling, multi-tool, ports validation)
- ✅ Reproduces and fixes the reported user issue

---

### DIFF
--- a/src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
+++ b/src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
@@ -409,8 +409,8 @@ def create_router_node(
         # Avoid using lambda to capture function_name
         # lambda will capture the function_name by reference,
         # and if the function_name is changed, the port_condition will also change.
-        def create_port_condition(fn_name):
-            return Port.on_if(
+        def create_port_condition(fn_name, is_first):
+            condition = (
                 ToolCallingState.current_prompt_output_index.less_than(tool_prompt_node.Outputs.results.length())
                 & tool_prompt_node.Outputs.results[ToolCallingState.current_prompt_output_index]["type"].equals(
                     "FUNCTION_CALL"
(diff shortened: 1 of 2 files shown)


alex-nork:
@codex review

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -419,26 +419,33 @@ def create_port_condition(fn_name):
                     "name"
                 ].equals(fn_name)
             )
+            # First port should be on_if, subsequent ports should be on_elif
+            return Port.on_if(condition) if is_first else Port.on_elif(condition)
 
+        is_first_port = True
```
dvargasfuertes, line 425: [nit] we can avoid this boolean tracking by maping functions -> conditions, then conditions -> ports

dvargas92495 (commented):
✅

PR #2718, opened 2025-10-09
[APO-1834] Add IntegrationTrigger base class

alex-nork:
## Summary
Introduce `IntegrationTrigger` base class for integration-based workflow triggers (Slack, Email, GitHub, etc.).

This provides the foundation for all integration triggers and will be extended by APO-1833 (SlackTrigger implementation).

## Changes
- Add `IntegrationTrigger` abstract base class extending `BaseTrigger`
  - Support structured outputs that downstream nodes can reference
  - Add abstract `process_event()` method for parsing external webhooks
  - Include optional `config` ClassVar for runtime configuration
- Add 6 unit tests covering:
  - Abstract nature (cannot instantiate directly)
  - Subclassing capability
  - Graph syntax support (`IntegrationTrigger >> Node`)
  - Multiple entry points
  - Config attribute

## Test Plan
- [x] All 6 unit tests passing
- [x] Pre-commit hooks passing
- [x] Type checking passing

## Dependencies
None - this is a foundation PR

## Follow-up PRs
- APO-1833: SlackTrigger implementation (depends on this)
- Future: EmailTrigger, GitHubTrigger, etc.

### DIFF
--- a/src/vellum/workflows/triggers/integration.py
+++ b/src/vellum/workflows/triggers/integration.py
@@ -1,0 +1,62 @@
+from abc import ABC
+from typing import ClassVar, Optional
+
+from vellum.workflows.outputs.base import BaseOutputs
+from vellum.workflows.triggers.base import BaseTrigger
+
+
+class IntegrationTrigger(BaseTrigger, ABC):
+    """
+    Base class for integration-based triggers (Slack, Email, etc.).
+
+    Integration triggers:
+    - Are initiated by external events (webhooks, API calls)
+    - Produce outputs that downstream nodes can reference
+    - Require configuration (auth, webhooks, etc.)
+
+    Examples:
+        # Define an integration trigger
+        class MyIntegrationTrigger(IntegrationTrigger):
+            class Outputs(IntegrationTrigger.Outputs):
+                data: str
+
+            @classmethod
+            def process_event(cls, event_data: dict):
+                return cls.Outputs(data=event_data.get("data", ""))
+
+        # Use in workflow
+        class MyWorkflow(BaseWorkflow):
+            graph = MyIntegrationTrigger >> ProcessNode
+
+    Note:
+        Unlike ManualTrigger, integration triggers provide structured outputs
+        that downstream nodes can reference directly via Outputs.
+    """
+
+    class Outputs(BaseOutputs):
+        """Base outputs for integration triggers."""
+
+        pass
+
+    # Configuration that can be set at runtime
+    config: ClassVar[Optional[dict]] = None
+
+    @classmethod
+    def process_event(cls, event_data: dict) -> "IntegrationTrigger.Outputs":
+        """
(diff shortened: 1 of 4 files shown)


alex-nork:
@codex review

src/vellum/workflows/triggers/integration.py
```diff
@@ -0,0 +1,62 @@
...
+    class Outputs(BaseOutputs):
+        """Base outputs for integration triggers."""
+
+        pass
+
+    # Configuration that can be set at runtime
+    config: ClassVar[Optional[dict]] = None
+
+    @classmethod
+    def process_event(cls, event_data: dict) -> "IntegrationTrigger.Outputs":
```
dvargasfuertes, line 45: [q] can we just use `__init__`? again, would like to be sensitive to non-dunder top level attributes

src/vellum/workflows/triggers/integration.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Note:
+        Unlike ManualTrigger, integration triggers provide structured outputs
+        that downstream nodes can reference directly via Outputs.
+    """
+
+    class Outputs(BaseOutputs):
+        """Base outputs for integration triggers."""
+
+        pass
```
dvargasfuertes, line 39: [blocking] based on what we are serializing to/from on the django side, would like to keep this to just top level attributes for now. The current thought is for downstream nodes to reference attributes as:
```
class MyPrompt(BaseNode):
    blocks = [SlackTrigger.message]
```

Steering away from `Outputs` will also help in making Triggers feel distinct from nodes

src/vellum/workflows/triggers/integration.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Configuration that can be set at runtime
+    config: ClassVar[Optional[dict]] = None
```
dvargasfuertes, line 42: [q] what do we imagine this looking like?

alex-nork: sort of hand-wavy at this point but what I had in mind is storing config for a slack app, like signing secret and bot token

src/vellum/workflows/triggers/integration.py
alex-nork: will address in #2719

src/vellum/workflows/triggers/integration.py
dvargasfuertes: > like signing secret and bot token

could be good to align with @emmiekehoe today in our sync with her later. My impression is that creds will be taken care for us

emmiekehoe: i've done some initial research, but yeah if we go through composio then creds will be taken care of for us

src/vellum/workflows/triggers/integration.py
alex-nork: do we want triggers to be instantiable?

PR #2725, opened 2025-10-09
[APO-1860] Fix MapNode output references lost in serialization

alex-nork:
The purpose of this PR is to fix MapNode output references being serialized as None instead of proper NODE_OUTPUT references when used by downstream nodes in the Agent Builder.

## Changes Made
- Added `setattr(outputs_class, reference.name, reference)` to `MapNode.__annotate_outputs_class__()` to match the behavior of BaseAdornmentNode and InlineSubworkflowNode
- This ensures `output.instance` is properly set to an OutputReference instead of undefined, allowing correct serialization

## Testing
- All existing MapNode serialization tests pass
- MapNode reference serialization test passes
- Basic map node workflow tests pass
- Verified fix with comprehensive test matching the exact APO-1860 scenario

## Notes
The root cause was that MapNode was missing a single line that other adornment nodes had, which prevented the OutputReference from being set as an attribute on the outputs class. This caused the serialization code in `base_node_display.py` to produce None values instead of NODE_OUTPUT pointers.

---


alex-nork:
@codex review

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -218,6 +218,9 @@ def __annotate_outputs_class__(cls, outputs_class: Type[BaseOutputs], reference:
         #     value: List[str]
         outputs_class.__annotations__ = {**previous_annotations, reference.name: annotation}
 
+        # Set the reference as an attribute so it can be serialized properly
+        setattr(outputs_class, reference.name, reference)
```
dvargasfuertes: Ok so the desired endstate we want is:
```python
        # class Outputs(BaseOutputs):
        #     value: List[str]
```

we don't actually want to set it to the workflow reference directly, bc it will be a list of that output, not exact equality. This line you added produces the following endstate:

```python
        # class Outputs(BaseOutputs):
        #     value: List[str] = Subworkflow.Outputs.value
```

which can have undesired (or at the least) redundant data when the map node itself serializes.

---

The issue is that the map node _itself_ is not referenceable, correct? I would assume doing `for output in Outputs` would hit `value` in our example above, but maybe the but is that it's not?

alex-nork: yeah you're right mapnode outputs aren't getting discovered during iteration

src/vellum/workflows/nodes/core/map_node/node.py
```diff
@@ -218,6 +219,15 @@ def __annotate_outputs_class__(cls, outputs_class: Type[BaseOutputs], reference:
...
         outputs_class.__annotations__ = {**previous_annotations, reference.name: annotation}
 
+        # Create a NEW OutputReference with the List-wrapped type for discoverability during iteration
+        map_output_reference = OutputReference(
+            name=reference.name,
+            types=(annotation,),
+            instance=undefined,
+            outputs_class=outputs_class,
+        )
+        setattr(outputs_class, reference.name, map_output_reference)
```
dvargasfuertes, line 229: [oos] I would expect this method to catch `reference.name` without including L229: https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/outputs/base.py#L151C22-L151C42

If that proves to be too challenging, cool with merging this as is and iterating on this clean up later

dvargas92495 (approved):

src/vellum/workflows/nodes/core/map_node/node.py
alex-nork: merging and will do cleanup later as necessary

PR #2745, opened 2025-10-13
Add quota error

vincent0426:
how do we differentiate between workflow error / vellum error / workflow event error?


dvargas92495 (approved):
> how do we differentiate between workflow error / vellum error / workflow event error

One of the two goes away once we delete legacy workflow runner

The second will go away once we consolidate vellum value schemas. So let's keep in sync for now

PR #2755, opened 2025-10-14
fix the disappearance of map nodes

dvargas92495:
Nested map nodes can throw an error, leading to its removal from the graph

### DIFF
--- a/ee/codegen/src/generators/workflow.ts
+++ b/ee/codegen/src/generators/workflow.ts
@@ -216,46 +216,50 @@ export class Workflow {
                   }),
                 ]
               : []),
-            python.methodArgument({
-              name: "display_data",
-              value: python.instantiateClass({
-                classReference: python.reference({
-                  name: "WorkflowDisplayData",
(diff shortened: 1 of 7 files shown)


ee/codegen/src/generators/workflow.ts
```diff
@@ -216,46 +216,50 @@ export class Workflow {
...
               : []),
-            python.methodArgument({
-              name: "display_data",
-              value: python.instantiateClass({
-                classReference: python.reference({
-                  name: "WorkflowDisplayData",
-                  modulePath: VELLUM_WORKFLOWS_DISPLAY_BASE_PATH,
-                }),
-                arguments_: [
+            ...(this.displayData
```
dvargasfuertes, line 219: file was unrelated to the bug, but let me simplify the fixture a tad

alex-nork:
was this causing the orphaned edges we've been seeing from AB-built workflows?

dvargas92495:
potentially - there may be other exceptions throughout codegen that are not being caught by our error handler

PR #2764, opened 2025-10-14
Execution configuration types for integration triggers

alex-nork:
## Summary
Add execution configuration types for integration triggers to support serialization/deserialization of trigger configurations.

This PR introduces:
- `BaseIntegrationTriggerExecConfig` - Base class for all integration trigger exec configs
- `ComposioIntegrationTriggerExecConfig` - Specific implementation for Composio provider triggers

## Part of PR Series
This is **PR 1 of 2** in the APO-1833 integration triggers feature:
1. **This PR**: Trigger execution config types (foundation)
2. Coming next: VellumIntegrationTrigger implementation

## Details
These types define the structure of execution configuration data sent to/from the backend for integration triggers. They enable proper communication with integration providers like Composio.

The types are:
- Self-contained pure data structures (Pydantic models)
- Provider-extensible through inheritance
- Ready for use in trigger serialization/deserialization

## Dependencies
None - this PR can be reviewed and merged independently.

## Testing
- Types are validated through Pydantic
- Will be exercised by integration trigger implementation in next PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2765, opened 2025-10-14
VellumIntegrationTrigger factory with dynamic attributes

alex-nork:
## Summary
Add VellumIntegrationTrigger with factory pattern for dynamic integration trigger support.

Enables workflows to be triggered by external events from white-labeled integrations (Slack, GitHub, Google, etc.) without requiring SDK updates for each new integration.

## Problem & Solution
**Problem**: Vellum workflows need to respond to external events (e.g., Slack messages, GitHub PRs). Users want to set up triggers without custom code and use the same trigger across multiple workflows.

**Solution**: Factory-based trigger system that dynamically creates trigger classes for any integration provider (Composio, etc.), matching the extensibility pattern of `VellumIntegrationToolDefinition`.

## Key Features
- **Factory pattern** via `for_trigger()` for dynamic trigger class creation
- **Custom metaclass** enabling dynamic attribute discovery from event payloads
- **Automatic caching** of generated trigger classes for consistency

## Example Usage
```python
SlackNewMessage = VellumIntegrationTrigger.for_trigger(
    integration_name="SLACK",
    trigger_name="SLACK_NEW_MESSAGE"
)

class MyWorkflow(BaseWorkflow):
    graph = SlackNewMessage >> ProcessNode
    
        message_text = SlackNewMessage.message
        channel = SlackNewMessage.channel
```

## Part of PR Series
**PR 2 of 2** in APO-1833 integration triggers:
1. PR #2764: Trigger execution config types ✅
2. **This PR**: VellumIntegrationTrigger implementation


src/vellum/workflows/triggers/vellum_integration.py
```diff
@@ -0,0 +1,250 @@
...
+        # Create cache key
+        cache_key = (provider_enum.value, integration_name, trigger_name)
+
+        # Return cached class if it exists
+        if cache_key in cls._trigger_class_cache:
+            return cls._trigger_class_cache[cache_key]
+
+        # Generate unique class name
+        class_name = f"VellumIntegrationTrigger_{integration_name}_{trigger_name}"
+
```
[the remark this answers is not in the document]
alex-nork: @codex address this feedback

alex-nork:
@codex review

src/vellum/workflows/triggers/vellum_integration.py
```diff
@@ -0,0 +1,336 @@
...
+        # Validate and normalize provider
+        provider_enum = VellumIntegrationProviderType(provider)
+
+        # Normalize attributes
+        attrs = attributes or {}
+
+        # Create cache key - include all identifying parameters including attributes
+        # Convert attributes dict to hashable tuple for caching
+        attrs_tuple = tuple(sorted(attrs.items())) if attrs else ()
+        cache_key = (provider_enum.value, integration_name, slug, trigger_nano_id, attrs_tuple)
```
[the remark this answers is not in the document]
alex-nork: @codex address this feedback

src/vellum/workflows/triggers/vellum_integration.py
```diff
@@ -0,0 +1,383 @@
...
+    This design ensures parity with VellumIntegrationToolDefinition and allows users to
+    work with any integration trigger without requiring SDK updates for new integrations.
+
+    Examples:
+        Create triggers dynamically for different integrations:
+            >>> SlackNewMessage = VellumIntegrationTrigger.for_trigger(
+            ...     integration_name="SLACK",
+            ...     slug="slack_new_message",
+            ...     trigger_nano_id="abc123def456"
+            ... )
```
dvargasfuertes, line 119: [nit] thoughts on:
```python
class SlackNewMessageTrigger(VellumIntegrationTrigger):
    integration_name="SLACK"
    slug="slack_new_message"
    trigger_nano_id="abc123def456"
```
?

src/vellum/workflows/triggers/vellum_integration.py
```diff
@@ -0,0 +1,383 @@
...
+            >>>
+            >>> GithubPush = VellumIntegrationTrigger.for_trigger(
+            ...     integration_name="GITHUB",
+            ...     slug="github_push_event",
+            ...     trigger_nano_id="xyz789ghi012"
```
dvargasfuertes, line 124: [oos] This id is specific to composio correct? Do we plan to expose this in the UI?

I wonder if there's a way long term to obfuscate this from the user. My sense is we only need this on the vellum side and that once we are in workflows sdk we don't use this value?

src/vellum/workflows/triggers/vellum_integration.py
```diff
@@ -0,0 +1,383 @@
...
+        except AttributeError:
+            # For VellumIntegrationTrigger factory-generated classes, create dynamic references
+            # Only enable dynamic attribute creation for factory-generated classes, not the base
+            # VellumIntegrationTrigger class itself. We check the internal __name__ attribute
+            # (e.g., "VellumIntegrationTrigger_COMPOSIO_SLACK_slack_new_message"), not the user-facing
+            # variable name (e.g., "SlackNewMessage").
+            try:
+                is_factory_class = super().__getattribute__("__name__").startswith(
+                    "VellumIntegrationTrigger_"
+                ) and not name.startswith("_")
```
dvargasfuertes, line 66: [nit] often cleaner to just `if not name.startswith("_")` towards the start of the method

src/vellum/workflows/triggers/vellum_integration.py
```diff
@@ -0,0 +1,383 @@
...
+            return super().__getattribute__(name)
+        except AttributeError:
+            # For VellumIntegrationTrigger factory-generated classes, create dynamic references
+            # Only enable dynamic attribute creation for factory-generated classes, not the base
+            # VellumIntegrationTrigger class itself. We check the internal __name__ attribute
+            # (e.g., "VellumIntegrationTrigger_COMPOSIO_SLACK_slack_new_message"), not the user-facing
+            # variable name (e.g., "SlackNewMessage").
+            try:
+                is_factory_class = super().__getattribute__("__name__").startswith(
+                    "VellumIntegrationTrigger_"
```
dvargasfuertes, line 65: [nit] this name startswith is somewhat brittle. You likely can do:
`if issubclass(cls, VellumIntegrationTrigger) and cls is not VellumIntegrationTrigger:`
since we are in a method already. Actually that check only becomes possible with my suggestion below

src/vellum/workflows/triggers/vellum_integration.py
```diff
@@ -0,0 +1,383 @@
...
+                ) and not name.startswith("_")
+            except AttributeError:
+                is_factory_class = False
+
+            if is_factory_class:
```
dvargasfuertes, line 70: [nit] slight pref for most of this logic being in the "expected" case instead of the `exception` case

src/vellum/workflows/triggers/vellum_integration.py
```diff
@@ -0,0 +1,383 @@
...
+                trigger_cls = cast(Type["VellumIntegrationTrigger"], cls)
+
+                # Check cache first
+                cache = super().__getattribute__("__trigger_attribute_cache__")
+                if name in cache:
+                    return cache[name]
```
dvargasfuertes, line 76: [q] why do we need a cache?

src/vellum/workflows/triggers/vellum_integration.py
```diff
@@ -0,0 +1,383 @@
...
+                if name not in attribute_ids:
+                    # Generate stable ID from trigger semantics, not class naming
+                    provider = super().__getattribute__("provider")
+                    integration_name = super().__getattribute__("integration_name")
+                    slug = super().__getattribute__("slug")
+                    trigger_identity = f"{provider.value}|{integration_name}|{slug}"
+                    attribute_ids[name] = uuid4_from_hash(f"{trigger_identity}|{name}")
+
+                # Create a new dynamic reference for this attribute
+                types = (object,)
```
dvargasfuertes, line 92: [oos] long term, would be cool to have this reflect the actual types of the various trigger attribute fields

dvargas92495 (commented):
✅

src/vellum/workflows/triggers/vellum_integration.py
alex-nork: > My sense is we only need this on the vellum side and that once we are in workflows sdk we don't use this value?

yeah that sounds right...
cc @emmiekehoe

src/vellum/workflows/triggers/vellum_integration.py
alex-nork: I was thinking it'd be necessary for consistency when instantiating a given trigger class in multiple places, but on closer inspection that doesn't seem to be the case. I'll get rid of it.

src/vellum/workflows/triggers/vellum_integration.py
emmiekehoe: we won't have this `trigger_nano_id` until we create the record in the deployer service and 100% aligned we should obfuscate this from the user

PR #2822, opened 2025-10-21
[APO-1919] Emit initiated event of resolved inline subworkflow

NgoHarrison:


tests/workflows/basic_subworkflow_deployment/tests/test_workflow.py
```diff
@@ -402,3 +402,77 @@ def generate_subworkflow_events(*args: Any, **kwargs: Any) -> Iterator[WorkflowS
...
+        "temperature": 70,
+        "reasoning": "I went to weather.com and looked at today's forecast.",
+    }
+
+    # AND the node events should have the correct parent context
+    assert node_events[0].name == "node.execution.initiated"
+    assert isinstance(node_events[0].parent, WorkflowParentContext)
+    assert (
+        node_events[0].parent.workflow_definition.model_dump()
+        == WorkflowParentContext(
```
marlenc3, line 479: I think the check we need is that we get Workflow Deployment Parent -> WorkflowParent

which might be hard to test with  just generate stream

tests/workflows/basic_subworkflow_deployment/tests/test_workflow.py
```diff
@@ -402,3 +402,84 @@ def generate_subworkflow_events(*args: Any, **kwargs: Any) -> Iterator[WorkflowS
...
+                ts=datetime.now(),
+                outputs=expected_outputs,
+            ),
+        ),
+    ]
+
+    def generate_subworkflow_events(*args: Any, **kwargs: Any) -> Iterator[WorkflowStreamEvent]:
+        yield from expected_events
+
+    vellum_client.execute_workflow_stream.side_effect = generate_subworkflow_events
```
dvargasfuertes: During subworkflow deployment inlining, we don't call the execute workflow stream method, but rather, vellum pull gets called so that the workflow is run locally (or is already there from the outset of the method). I would take a look at the impl of `_run_resolved_workflow`

NgoHarrison: Ah yeah I see. I updated the test and included a source code change

src/vellum/workflows/nodes/displayable/subworkflow_deployment_node/node.py
```diff
@@ -165,7 +165,8 @@ def _run_resolved_workflow(self, resolved_workflow: "BaseWorkflow") -> Iterator[
             )
 
         try:
-            next(subworkflow_stream)
+            first_event = next(subworkflow_stream)
+            self._context._emit_subworkflow_event(first_event)
```
dvargasfuertes, line 169: Woah!

tests/workflows/basic_subworkflow_deployment/tests/test_workflow.py
dvargasfuertes: Yea marlen's right, we want to ensure:
- `.parent.type` == "WORKFLOW_DEPLOYMENT" (what's missing today)
- `.parent.parent.type` == "WORKFLOW_NODE" (the subworkflow deployment node)
- `.parent.parent.parent.type` == "WORKFLOW" (the root workflow)

I've also been thinking that we extend the `WorkflowDeploymentParentContext` to include a new optional flag or enum on it, stating that it was inlined or not, to differentiate this resolution from vellum-side resolution from predict

dvargas92495 (approved):
the `next()` fix is big enough to merge on its own. We will want to follow up for the parent context fix

PR #2823, opened 2025-10-22
Add extend expression

vincent0426:


src/vellum/workflows/expressions/tests/test_extend.py
```diff
@@ -0,0 +1,95 @@
...
+class TestState(BaseState):
+    pass
+
+
+def test_extend_expression_happy_path():
+    # GIVEN a list and a single item
+    state = TestState()
+    lhs_ref = ConstantValueReference([1, 2, 3])
+    rhs_ref = ConstantValueReference(4)
+    extend_expr = lhs_ref.extend(rhs_ref)
```
dvargasfuertes, line 18: doesn't this make `extend` redundant with `concat`?

dvargasfuertes: oh concat doesn't support individual. I would lean towards concat supporting individual args, consistent with javascript's implementation

PR #2825, opened 2025-10-22
[APO-1836] Support workflow.run and stream for integration triggers

alex-nork:
Adds support for passing `IntegrationTrigger` instances to `BaseWorkflow.run()` and `.stream()`, enabling workflows with IntegrationTriggers to be executed with runtime event data.

## Changes Made
- Added `trigger: Optional[IntegrationTrigger]` parameter to `BaseWorkflow.run()` and `.stream()`
- Added `_process_trigger()` to WorkflowRunner that:
  - Validates trigger type matches workflow's declared triggers
  - Binds trigger attributes to state
  - Routes to the specific trigger subclass path (e.g., SlackMessageTrigger vs GmailTrigger vs ManualTrigger)
  - Respects caller-provided entrypoints
- Updated TypeScript codegen tests to reflect triggers as entrypoints
- Added 10 integration tests covering successful execution, streaming, validation errors, and multi-trigger routing

## Key Features
- Type-safe trigger instances instead of untyped dicts
- Automatic routing in multi-trigger workflows
- Clear validation errors when trigger requirements aren't met

---


alex-nork:
Note: The TypeScript codegen fixes for trigger-to-entrypoint connections have been split into PR #2835.

alex-nork:
@codex review

src/vellum/workflows/runner/runner.py
```diff
@@ -250,6 +258,107 @@ def __init__(
...
+
+            # Validate that the trigger instance is compatible with one of the workflow's trigger types
+            # Allow subclasses: if workflow declares BaseSlackTrigger, accept SpecificSlackTrigger instances
+            trigger_class = type(trigger)
+            workflow_trigger_types = [
+                t
+                for subgraph in self.workflow.get_subgraphs()
+                for t in subgraph.triggers
+                if issubclass(t, IntegrationTrigger)
+            ]
```
alex-nork: me > as an llm is it easier for you to reason about a nested list comprehension
or nested for loops? wondering about this
```
# Get ALL IntegrationTrigger types once
      workflow_integration_triggers = [
          t
          for subgraph in self.workflow.get_subgraphs()
          for t in subgraph.triggers
          if issubclass(t, IntegrationTrigger)
      ]
```

claude > 
⏺ Great question! As an LLM, I actually find nested for loops easier to reason
   about when the logic gets complex, especially when explaining code to
  humans or tracking data flow. Let me show you why:

⏺ Honest LLM Perspective on Code Readability

  Nested For Loops (Easier for me to reason about):

  # Get ALL IntegrationTrigger types once
  workflow_integration_triggers = []
  for subgraph in self.workflow.get_subgraphs():
      for t in subgraph.triggers:
          if issubclass(t, IntegrationTrigger):
              workflow_integration_triggers.append(t)

  Why this is easier for me:
  1. Linear flow: I can trace execution step-by-step
  2. Clear scope: Each variable's scope is obvious
  3. Debugging points: I can "mentally insert" print statements or checks at
  each level
  4. Explicit operations: The append makes it clear we're building a list

  List Comprehension (More compact but requires "unpacking"):

  workflow_integration_triggers = [
      t
      for subgraph in self.workflow.get_subgraphs()
      for t in subgraph.triggers
      if issubclass(t, IntegrationTrigger)
  ]

  Why this is harder for me:
  1. Right-to-left then top-to-bottom parsing: I have to mentally reorder:
  "collect t, but first iterate subgraphs, then iterate triggers, but only
  if..."
  2. Implicit state: The list building is implicit
  3. Nested scope ambiguity: With multiple for clauses, I need to carefully
  track which loop is outer vs inner

src/vellum/workflows/runner/runner.py
```diff
@@ -109,6 +111,7 @@ def __init__(
         max_concurrency: Optional[int] = None,
         timeout: Optional[float] = None,
         init_execution_context: Optional[ExecutionContext] = None,
+        trigger: Optional[IntegrationTrigger] = None,
```
dvargasfuertes: [blocker] We want to use this same param for scheduled, manual, and any other trigger in the future. We should make this `Optional[BaseTrigger]`

src/vellum/workflows/runner/runner.py
```diff
@@ -250,6 +258,105 @@ def __init__(
...
+
+        # Case 2: No trigger but workflow has IntegrationTrigger(s)
+        elif workflow_integration_triggers:
+            if not self._has_manual_trigger():
+                # Workflow has ONLY IntegrationTrigger - this is an error
+                raise WorkflowInitializationException(
+                    message="Workflow has IntegrationTrigger which requires trigger parameter",
+                    workflow_definition=self.workflow.__class__,
+                    code=WorkflowErrorCode.INVALID_INPUTS,
+                )
```
dvargasfuertes, line 390: [oos] This case IIUC can be:
```
graph = {
    SlackTrigger >> TopNode,
    BottomNode,
}
```

In this kind of graph, I believe we still want to consider `BottomNode` as being "Manually Triggered".

src/vellum/workflows/runner/runner.py
```diff
@@ -250,6 +258,105 @@ def __init__(
...
+
+            # Validate that the trigger instance is compatible with one of the workflow's trigger types
+            # Allow subclasses: if workflow declares BaseSlackTrigger, accept SpecificSlackTrigger instances
+            trigger_class = type(trigger)
+
+            # Check if the provided trigger_class is a subclass of any declared trigger type
+            if not any(
+                issubclass(trigger_class, declared_trigger) for declared_trigger in workflow_integration_triggers
+            ):
```
dvargasfuertes: [nit] feels like we could use the initial loop in L301-L304 to just search or the trigger we want to invoke, instead of maintain an array of them through here

src/vellum/workflows/runner/runner.py
```diff
@@ -117,6 +120,8 @@ def __init__(
         self._is_resuming = False
         self._should_emit_initial_state = True
         self._span_link_info: Optional[Tuple[str, str, str, str]] = None
+        # Track if caller explicitly provided entrypoints
+        self._caller_provided_entrypoints = bool(entrypoint_nodes or external_inputs)
```
dvargasfuertes: [nit] I think we can avoid tracking this flag if we just add an `elif trigger` case in this if/else tree below

src/vellum/workflows/runner/runner.py
```diff
@@ -250,6 +258,105 @@ def __init__(
...
+
+            # Bind trigger to state
+            trigger.bind_to_state(self._initial_state)
+
+            # Filter entrypoints to the SPECIFIC trigger subclass path only
+            # UNLESS caller explicitly provided entrypoints (e.g., via entrypoint_nodes or external_inputs)
+            if not self._caller_provided_entrypoints:
+                # Use the specific trigger subclass, not the parent IntegrationTrigger
+                specific_entrypoints = self._get_entrypoints_for_trigger_type(trigger_class)
+                if specific_entrypoints:
```
dvargasfuertes: [oos] are we already rejecting if the workflow has no entrypoints? we should maybe consider doing so to catch triggers that point to nothing

src/vellum/workflows/workflows/base.py
```diff
@@ -507,6 +516,7 @@ def stream(
         node_output_mocks: Optional[MockNodeExecutionArg] = None,
         max_concurrency: Optional[int] = None,
         timeout: Optional[float] = None,
+        trigger: Optional[IntegrationTrigger] = None,
```
dvargasfuertes: cc: @NgoHarrison - this means it will be up to python workflow server to take the raw dict and deserialize into an Integration Trigger. We already do this for the Inputs class, and we have line of sight on how to re-home this into the SDK one day in the near future

tests/workflows/integration_trigger_execution/tests/test_integration_trigger_execution.py
```diff
@@ -0,0 +1,146 @@
...
+    # GIVEN a workflow without IntegrationTrigger
+    workflow = NoTriggerWorkflow()
+
+    # AND a trigger instance
+    trigger = SlackMessageTrigger(
+        event_data={
+            "message": "This should fail",
+            "channel": "C123456",
+            "user": "U789012",
+        }
```
dvargasfuertes, line 89: [oos] We should consider allowing the `__init__` just accepting the dict or a spread of kwargs directly

src/vellum/workflows/runner/runner.py
```diff
@@ -250,6 +258,105 @@ def __init__(
...
+        if trigger:
+            if not workflow_integration_triggers:
+                raise WorkflowInitializationException(
+                    message="trigger provided but workflow does not have an IntegrationTrigger",
+                    workflow_definition=self.workflow.__class__,
+                    code=WorkflowErrorCode.INVALID_INPUTS,
+                )
+
+            # Validate that the trigger instance is compatible with one of the workflow's trigger types
+            # Allow subclasses: if workflow declares BaseSlackTrigger, accept SpecificSlackTrigger instances
```
dvargasfuertes: [oos] If I have two triggers that extend `BaseSlackTrigger`, how will the consumer know which one to invoke?

Mid comment realizing that this is likely the responsibility of PWS (cc: @NgoHarrison )

dvargas92495 (commented):
main blocker is the kwarg type of the trigger

src/vellum/workflows/runner/runner.py
alex-nork: 🤦 my bad

dvargas92495 (approved):

PR #2832, opened 2025-10-23
[APO-1836] Implement trigger-to-entrypoint serialization for TypeScript codegen

alex-nork:
The purpose of this PR is to implement trigger-to-entrypoint serialization for TypeScript codegen using edges with trigger as source (Option 3).

### TypeScript Codegen - Option 3 Implementation
- Use **edges with trigger IDs as sourceNodeId** instead of metadata field
- Update `getEntrypointNodeEdges()` to filter edges where `sourceNodeId` matches trigger IDs
- Update tests to include edges from trigger to firstNode (e.g., `{sourceNodeId: triggerId, targetNodeId: firstNode.id}`)
- Simplified implementation - no synthetic edge generation needed
- **Result**: Tests now pass with output `ManualTrigger >> FirstNode >> SecondNode`

### Why Option 3 (Edges) vs Option 2 (Metadata)
- ✅ **Simpler**: No new metadata field needed
- ✅ **More explicit**: Edges exist in serialized data, not computed
- ✅ **First-class**: Trigger is a real graph node via edges
- ✅ **Natural**: Uses existing edge model

### Python Runtime Refactor (from earlier commits)
- Refactor trigger API from dict to `IntegrationTrigger` instance in `workflow.run(trigger=...)`
- Add `_process_trigger()` method to WorkflowRunner for trigger validation and routing
- Implement smart entrypoint filtering: routes to specific trigger subclass path when trigger provided
- Add comprehensive integration tests for trigger execution scenarios

## Testing

- ✅ All TypeScript codegen tests pass (601/602 tests passing)
- ✅ Both previously failing trigger tests now pass
- ✅ Added comprehensive Python integration tests for trigger execution
- ✅ Verified backward compatibility with ENTRYPOINT nodes

## Implementation Details

**Serialized Data Structure:**
```typescript
{
  nodes: [firstNode, secondNode],
  edges: [
    {sourceNodeId: "trigger-1", targetNodeId: "first-node"},  // Trigger edge
    {sourceNodeId: "first-node", targetNodeId: "second-node"}
  ],
  triggers: [{id: "trigger-1", type: "MANUAL"}]
}
```

**Codegen Logic:**
- Filter edges where `sourceNodeId` matches any trigger ID
- Use those edges as entrypoint edges for graph construction
- Fall back to ENTRYPOINT nodes for backward compatibility

## Notes

- Option 3 is cleaner than Option 2 - no metadata field, just natural edge relationships
[...]

PR #2835, opened 2025-10-23
[APO-1836] Add sourceHandleId to triggers and fix entrypoint display codegen

alex-nork:
## Summary

Implements trigger edge support in TypeScript codegen where IntegrationTriggers use trigger IDs directly in edges without ENTRYPOINT nodes, and adds sourceHandleId to trigger type definitions.

## Changes

- **Added `sourceHandleId` to IntegrationTrigger type** (required field)
  - IntegrationTrigger workflows use trigger ID as sourceHandleId
  - ManualTrigger type does NOT have sourceHandleId (not needed)
- **Updated test to use `triggerId` as `sourceHandleId`** in both edges and trigger objects
- **Fixed entrypoint display UUID generation bug** - use `edge.sourceNodeId` when `entrypointNode` is null to prevent `UUID('')` ValueError
- **Updated `getEntrypointNodeEdges()`** to check for edges from triggers first, then fall back to ENTRYPOINT nodes
- **Updated graph generation logic** to treat trigger sources like ENTRYPOINT nodes
- **Updated workflow edge filtering** to exclude edges from triggers

### Edge Structure
```typescript
{
  sourceNodeId: "trigger-1",       // Trigger ID
  sourceHandleId: "trigger-1",     // Same as trigger ID
  targetNodeId: "first-node"
}
```

### Trigger Type Structure
```typescript
export type WorkflowTrigger =
  | {
      id: string;
      type: WorkflowTriggerType.MANUAL;
      attributes: NodeAttribute[];
      // No sourceHandleId - uses ENTRYPOINT node
    }
  | {
      id: string;
      type: WorkflowTriggerType.INTEGRATION;
      attributes: NodeAttribute[];
      className: string;
      modulePath: string[];
      sourceHandleId: string;  // Required - acts as edge source
    };
```

### Architecture
- **IntegrationTrigger**: No ENTRYPOINT node, edges come directly from trigger, requires sourceHandleId
- **ManualTrigger**: ENTRYPOINT node created, edges come from ENTRYPOINT node, no sourceHandleId needed
- **Type safety**: sourceHandleId is required for IntegrationTrigger, absent from ManualTrigger

## Bug Fixes

Fixed Codex Review feedback: When IntegrationTrigger workflows don't have an ENTRYPOINT node, the entrypoint display instantiation was using `entrypointNode?.id ?? ""` which created invalid `UUID('')` causing ValueError at runtime. Now uses `edge.sourceNodeId` (the trigger ID) as fallback.

## Tests

- ✅ All trigger codegen tests pass
- ✅ TypeScript type checking passes
- ✅ Prettier formatting passes
- ✅ Expected graph output: `IntegrationTrigger >> FirstNode >> SecondNode`

## Related

- Works with PR #2836 (Python serialization changes)
- Split from original PR #2832

### DIFF
--- a/ee/codegen/src/__test__/graph-attribute.test.ts
+++ b/ee/codegen/src/__test__/graph-attribute.test.ts
@@ -1054,9 +1054,6 @@ describe("Workflow", () => {
 
       const triggerId = "trigger-1";
 
-      // With triggers, entrypoint node exists with ID matching the trigger ID
-      const entrypointNode = entrypointNodeDataFactory(triggerId);
-
       const firstNode = genericNodeFactory({
         id: "first-node",
         label: "FirstNode",
@@ -1119,9 +1116,6 @@ describe("Workflow", () => {
       const writer = new Writer();
 
       const triggerId = "trigger-1";
-
-      // With triggers, entrypoint node exists with ID matching the trigger ID
-      const entrypointNode = entrypointNodeDataFactory(triggerId);
 
       const firstNode = genericNodeFactory({
         id: "first-node",
--- a/ee/codegen/src/__test__/graph-attribute.test.ts
+++ b/ee/codegen/src/__test__/graph-attribute.test.ts
@@ -1055,9 +1055,7 @@ describe("Workflow", () => {
       const triggerId = "trigger-1";
 
       // Manual triggers have an associated entrypoint node with ID matching the trigger
-      const entrypointNodeWithTriggerId = entrypointNodeDataFactory({
-        id: triggerId,
-      });
+      const entrypointNodeWithTriggerId = entrypointNodeDataFactory(triggerId);
 
       const firstNode = genericNodeFactory({
         id: "first-node",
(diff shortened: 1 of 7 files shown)


alex-nork:
Note: This PR focuses on the TypeScript codegen changes only. The complementary Python runtime changes (trigger instance support) are already in PR #2825.

alex-nork:
## Addressed Codex Feedback

Both concerns were **valid** ✅. The original investigation included these fixes in commit f5554c03, but they weren't carried forward when cherry-picking the Option 3 refactor.

### Fixes Applied

**Issue 1: Skip trigger-based entrypoints when generating graph**
- ❌ Problem: Code checked for `entrypointNode` before calling `getEntrypointNodeEdges()`
- ✅ Fix: Always call `getEntrypointNodeEdges()` since it handles trigger edges internally

**Issue 2: Discard trigger edges due to missing source node contexts**
- ❌ Problem: `addEdgeToGraph` tried to resolve trigger IDs as node IDs, which fails
- ✅ Fix: Detect trigger IDs and treat them like entrypoints (sourceNode = null)

**Additional:** Updated `workflow.ts` to exclude trigger edges from edge displays, consistent with ENTRYPOINT handling.

Changes pushed in commit 80d7672a.

alex-nork:
@codex review

alex-nork:
@codex review

dvargas92495 (commented):
✅

PR #2845, opened 2025-10-24
[APO-1919] Construct workflow deployment parent context for inlined subworkflow deployments

NgoHarrison:


src/vellum/workflows/state/context.py
```diff
@@ -148,7 +161,7 @@ def _get_all_node_output_mocks(self) -> List[MockNodeExecution]:
 
     def resolve_workflow_deployment(
         self, deployment_name: str, release_tag: str, state: "BaseState"
-    ) -> Optional["BaseWorkflow"]:
+    ) -> Optional[Tuple[Type["BaseWorkflow"], "BaseState", Optional[WorkflowDeploymentMetadata]]]:
```
dvargasfuertes: why do we need to return state if it's the same one we're passing in?

src/vellum/workflows/state/context.py
```diff
@@ -158,20 +171,22 @@ def resolve_workflow_deployment(
...
+        deployment_metadata = self._fetch_deployment_metadata(deployment_name, release_tag)
+
         try:
             from vellum.workflows.workflows.base import BaseWorkflow
 
             WorkflowClass = BaseWorkflow.load_from_module(f"{self.namespace}.{expected_prefix}")
             WorkflowClass.is_dynamic = True
-            workflow_instance = WorkflowClass(context=WorkflowContext.create_from(self), parent_state=state)
-            return workflow_instance
+            # Return the class, not an instance, so caller can instantiate within proper execution context
```
dvargasfuertes, line 188: oh wow. Yea makes sense

src/vellum/workflows/state/context.py
```diff
@@ -196,18 +211,59 @@ def resolve_workflow_deployment(
                 prefixed_file_name = f"{expected_prefix}/{file_name}"
                 self._generated_files[prefixed_file_name] = content
 
+            # Re-fetch deployment metadata if it wasn't available before
+            if deployment_metadata is None:
+                deployment_metadata = self._fetch_deployment_metadata(deployment_name, release_tag)
```
dvargasfuertes: I wouldn't expect to need to do this twice

src/vellum/workflows/state/context.py
```diff
@@ -196,18 +211,59 @@ def resolve_workflow_deployment(
...
+        try:
+            # Fetch deployment details
+            deployment = self.vellum_client.workflow_deployments.retrieve(deployment_name)
+
+            deployment_id = UUID(deployment.id)
+
+            # Fetch release tag details
+            release_tag_info = self.vellum_client.workflow_deployments.retrieve_workflow_release_tag(
+                deployment.id, release_tag
+            )
```
dvargasfuertes, line 248: [oos] this works for now. Ideally our pull api and our initial artifact bundle makes it so that we don't need to do two separate api calls in the future

dvargas92495 (approved):

PR #2869, opened 2025-10-29
Add workflow version id to workflow deployment metadata

NgoHarrison:
After doing two api calls here, I think we should definitely modify the pull call from client to return all this info as mentioned in previous PR comment. For now we can use this and fast follow


dvargas92495 (approved):
oh damn I missed this. But yea, I'll follow up with updating the `pull` metadata to include

PR #2874, opened 2025-10-29
Fix root cause: skip serializing base class outputs

alex-nork:
The purpose of this PR is to fix the root cause of redundant output serialization by preventing Python from serializing outputs that are identical to the base class.

## Changes Made

- Python: Skip serializing outputs when they match the base class (return empty array)
- TypeScript: Remove `hasRedundantOutputs()` special case logic (no longer needed)
- Fixtures: Regenerate with ToolCallingNode having empty `outputs: []`
- Tests: Update to verify empty outputs behavior

## Context

PR #2872 fixed the symptom in TypeScript codegen. This PR fixes the root cause in Python serialization, eliminating the need for the TypeScript bandaid.

## Testing

✅ All TypeScript codegen tests passing (602 tests)
✅ Python serialization tests passing
✅ Snapshots updated

---


alex-nork:
@codex address the ci test failures

PR #2880, opened 2025-10-29
[APO-2007] Add BaseNode.Display serialization support

alex-nork:
## What Changed

Added serialization support for nested `Display` classes on workflow nodes, enabling Agent Builder to define icon and color metadata inline without requiring separate `./display` module files.

## Why

Agent Builder generates workflows with nodes that use nested `Display` classes for visual customization:

```python
class MyNode(BaseNode):
    class Display:
        icon = "vellum:icon:star"
        color = "purple"
```

Previously, serialization would fail because it couldn't find the icon/color metadata, resulting in "Failed to serialize workflow" errors.

## Implementation

- `BaseNodeDisplay.get_display_data()` now detects and extracts `icon` and `color` from nested `Display` classes
- When both a `Display` class and explicit `display_data` exist, they are merged with `Display` class taking precedence for icon/color
- Fully backward compatible - nodes without `Display` classes continue to work as before

This replaces the need for separate `./display` module files, simplifying the workflow structure for Agent Builder-generated code.

---

### DIFF
--- a/ee/codegen/src/generators/generic-node-display-data.ts
+++ b/ee/codegen/src/generators/generic-node-display-data.ts
@@ -67,13 +67,14 @@ export class GenericNodeDisplayData extends AstNode {
       );
     }
 
-    if (!isNil(this.sourceNodeDisplayData?.z_index)) {
+    // Include z_index if the field exists in source data (even if null)
+    if (this.sourceNodeDisplayData && "z_index" in this.sourceNodeDisplayData) {
       args.push(
         python.methodArgument({
           name: "z_index",
-          value: python.TypeInstantiation.int(
-            this.sourceNodeDisplayData.z_index
-          ),
+          value: !isNil(this.sourceNodeDisplayData.z_index)
+            ? python.TypeInstantiation.int(this.sourceNodeDisplayData.z_index)
+            : python.TypeInstantiation.none(),
         })
       );
     }
--- a/ee/codegen/src/generators/generic-node-display-data.ts
+++ b/ee/codegen/src/generators/generic-node-display-data.ts
@@ -67,14 +67,13 @@ export class GenericNodeDisplayData extends AstNode {
       );
     }
 
-    // Include z_index if the field exists in source data (even if null)
-    if (this.sourceNodeDisplayData && "z_index" in this.sourceNodeDisplayData) {
+    if (!isNil(this.sourceNodeDisplayData?.z_index)) {
       args.push(
         python.methodArgument({
           name: "z_index",
-          value: !isNil(this.sourceNodeDisplayData.z_index)
-            ? python.TypeInstantiation.int(this.sourceNodeDisplayData.z_index)
-            : python.TypeInstantiation.none(),
+          value: python.TypeInstantiation.int(
+            this.sourceNodeDisplayData.z_index
+          ),
         })
       );
     }
--- a/ee/vellum_ee/workflows/display/nodes/base_node_display.py
+++ b/ee/vellum_ee/workflows/display/nodes/base_node_display.py
@@ -535,14 +535,14 @@ class BaseNodeDisplay(Generic[NodeType], metaclass=BaseNodeDisplayMeta):
 
         # No explicit_value - build from Display class and/or docstring
         if display_class_icon is not None or display_class_color is not None or docstring:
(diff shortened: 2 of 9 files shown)


ee/vellum_ee/workflows/display/nodes/tests/test_base_node_display.py
```diff
@@ -181,3 +181,44 @@ class MyNodeDisplay(BaseNodeDisplay[MyNode]):
...
+            icon="vellum:icon:times",
+            color="navy",
+        )
+
+    data = MyNodeDisplay().serialize(WorkflowDisplayContext())
+
+    display_data = data["display_data"]
+    assert isinstance(display_data, dict)
+    assert display_data["icon"] == "vellum:icon:check"
+    assert display_data["color"] == "gold"
```
dvargasfuertes: [blocking] Our current posture is that `BaseNodeDisplay` class always annotates/overwrites the execution class (`BaseNode.Display`)

dvargasfuertes: So once we have serialization support, we'll want to update codegen to start generating this way too

ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -463,38 +463,86 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
 
         register_node_display_class(node_class=node_class, node_display_class=cls)
 
+    def _get_display_class_overrides(self) -> tuple[Optional[str], Optional[str]]:
+        """Extract icon and color from nested Display class if present."""
+        if not hasattr(self._node, "Display"):
+            return None, None
+
+        display_cls = getattr(self._node, "Display")
+        return getattr(display_cls, "icon", None), getattr(display_cls, "color", None)
```
dvargasfuertes: [nit] If we add `Display` class to `BaseNode`, we could clean up these hasattr/getattr's and have direct references

ee/vellum_ee/workflows/display/nodes/base_node_display.py
```diff
@@ -463,38 +463,86 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
...
+
+    def _merge_display_attributes(
+        self,
+        base_kwargs: Dict[str, Any],
+        explicit_icon: Optional[str],
+        explicit_color: Optional[str],
+        display_class_icon: Optional[str],
+        display_class_color: Optional[str],
```
dvargasfuertes: [nit] this will scale poorly as we eventually migrate the rest of `BaseNodeDisplay` -> `BaseNode.Display`. 

I believe we can simplify by just fully serializing with `BaseNode.Display`, then overwriting with `BaseNodeDisplay`

alex-nork:
## PR Updates Complete ✅

I've addressed all review comments:

### 1. ✅ **[BLOCKING] Fixed precedence** 
   - BaseNodeDisplay now correctly overrides BaseNode.Display attributes
   - Updated tests to verify correct precedence behavior

### 2. ✅ **Cleaned up hasattr/getattr checks**
   - Added Display class with default None values to BaseNode
   - Now using direct references instead of runtime attribute checks

### 3. ✅ **Simplified merge logic**
   - Refactored to serialize BaseNode.Display first, then override with BaseNodeDisplay
   - Much cleaner and more scalable approach

### 4. 📝 **Follow-up Task: Codegen Updates**
   - Once this PR is merged, we should update the codegen to start generating Display classes inline with nodes
   - This will allow removing the need for separate ./display module files
   - The codegen should generate nodes with the pattern:
     ```python
     class MyNode(BaseNode):
         class Display:
             icon = "vellum:icon:star"
             color = "purple"
     ```

ee/codegen/src/generators/generic-node-display-data.ts
```diff
@@ -67,13 +67,14 @@ export class GenericNodeDisplayData extends AstNode {
       );
     }
 
-    if (!isNil(this.sourceNodeDisplayData?.z_index)) {
+    // Include z_index if the field exists in source data (even if null)
```
dvargasfuertes: [nit] looks unrelated. Why do we want this? we typically lean towards not codegenning data that take their default values

dvargas92495 (approved):

PR #2882, opened 2025-10-29
Add scheduled trigger and integration contexts as parent context types

NgoHarrison:
Forgot to add here

### DIFF
--- a/src/vellum/workflows/events/types.py
+++ b/src/vellum/workflows/events/types.py
@@ -158,6 +158,8 @@ ParentContext = Annotated[
         WorkflowSandboxParentContext,
         APIRequestParentContext,
         ExternalParentContext,
+        WorkflowDeploymentScheduledTriggerContext,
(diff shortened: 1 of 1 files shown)


src/vellum/workflows/events/types.py
```diff
@@ -158,6 +158,8 @@ def __get_pydantic_core_schema__(self, source_type: Any, handler: GetCoreSchemaH
         WorkflowSandboxParentContext,
         APIRequestParentContext,
         ExternalParentContext,
+        WorkflowDeploymentScheduledTriggerContext,
+        WorkflowDeploymentIntegrationTriggerContext,
```
dvargasfuertes, line 162: Oh I missed the names - `WorkflowDeployment` will be a bit redundant bc we will already have a Workflow Deployment Parent Context pointing to this

dvargas92495 (commented):
✅ - changing the class names while technically breaking, shouldn't affect the execution work and can wait for the next release

PR #2887, opened 2025-10-29
[APO-1965] SetState Node run

vincent0426:


src/vellum/workflows/nodes/displayable/set_state_node/node.py
```diff
@@ -0,0 +1,65 @@
...
+
+            # Resolve the value if it's a descriptor
+            if isinstance(value, BaseDescriptor):
+                resolved_value = resolve_value(value, self.state)
+            else:
+                resolved_value = value
+
+            # Set the state value
+            setattr(self.state, path, resolved_value)
+            result[path] = resolved_value
```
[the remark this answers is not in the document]
vincent0426: fixing

vincent0426:
@codex review

vincent0426:
@codex review

src/vellum/workflows/nodes/displayable/set_state_node/tests/test_node.py
```diff
@@ -0,0 +1,239 @@
...
+
+    # WHEN we run the node
+    state = TestState()
+    node = SetState(state=state)
+
+    # THEN it should raise a NodeException
+    with pytest.raises(NodeException) as exc_info:
+        node.run()
+
+    # AND the error message should mention the non-existent attribute
```
dvargasfuertes: [nit] we should assert the error code here too

src/vellum/workflows/nodes/displayable/set_state_node/node.py
```diff
@@ -0,0 +1,80 @@
...
+            # Resolve the value if it's a descriptor against the current (unmodified) state
+            if isinstance(value, BaseDescriptor):
+                resolved_value = resolve_value(value, self.state)
+            else:
+                resolved_value = value
+
+            resolved_updates[path] = resolved_value
+
+        # Second pass: apply on a deepcopy first to ensure all setters succeed,
+        # then swap onto the real state atomically to avoid partial mutations.
```
dvargasfuertes: [nit] I think we could skip this pass. Is there a case you're concerned with?

dvargas92495 (approved):

src/vellum/workflows/nodes/displayable/set_state_node/node.py
vincent0426: `test_set_state_node_rollback_when_second_assignment_fails` will fail because of partial updates, unless we are sure that `setattr` will not throw error? maybe the possibility is very low i think we could remove it now

dvargasfuertes: yea if the `setattr` part fails, then that's a bug on our end that we need to remedy immediately anyway

PR #2889, opened 2025-10-30
Fix VirtualFileLoader to exclude display directories from auto-generation

alex-nork:
## Summary

Fixes the issue raised by @dvargas92495 in PR #2858 where display directories should be excluded from auto-generating empty `__init__.py` files.

## Problem

After merging PR #2858, the VirtualFileLoader would auto-generate empty `__init__.py` files for **any** directory containing Python files, including `display/` directories. This is problematic because:

1. Display directories typically have specific `__init__.py` content like `from .workflow import *`
2. Auto-generating an empty `__init__.py` breaks imports that expect these specific statements
3. The empty auto-generated file replaces what should be meaningful import logic

## Solution

Added a check in `_is_package_directory()` to explicitly exclude `display/` directories from auto-generation. Display directories must provide their own `__init__.py` files with the appropriate content.

## Testing

Added `test_load_from_module__display_directory_not_auto_generated` that:
- ✅ **Fails without the fix** - proves the bug exists
- ✅ **Passes with the fix** - proves the fix works
- Verifies that display directories without explicit `__init__.py` don't get empty ones auto-generated

All existing tests still pass.

## Related

- Addresses concern from @dvargas92495: "2858 can be merged if we exclude 'display' or something like that in the init py resolution logic"
- Follow-up to PR #2858


ee/vellum_ee/workflows/server/virtual_file_loader.py
```diff
@@ -87,6 +87,12 @@ def _is_package_directory(self, fullname: str) -> bool:
         """Check if directory contains .py files that should be treated as a package."""
         directory_prefix = fullname.replace(".", "/") + "/"
 
+        # Exclude display directories from auto-generation as they typically have
+        # specific __init__.py content that shouldn't be replaced with empty files.
+        # Check for both top-level (display/) and nested (.../display/) directories.
+        if directory_prefix == "display/" or "/display/" in directory_prefix:
```
dvargasfuertes: [nit] I _think_ this works, but also there's not really a nested /display/ directory. We just want to ignore the top level display directory from getting hit with this for now

dvargas92495 (approved):

PR #2902, opened 2025-10-31
Add test for falsy state serialization

vincent0426:
this seems not affecting anything?

I set string to hello in workflow inputs and number to 2 but they wont become default here?

Screenshot 2025-10-30 at 9 45 03 PM

### DIFF
--- a/ee/vellum_ee/workflows/display/workflows/tests/test_workflow_display.py
+++ b/ee/vellum_ee/workflows/display/workflows/tests/test_workflow_display.py
@@ -838,6 +838,7 @@ def test_serialize_workflow__empty_rules_indexerror():
 def test_serialize_workflow__input_variables():
     # GIVEN a workflow with inputs
     class Inputs(BaseInputs):
+        empty_string: str = ""
         input_1: str
         input_2: Optional[str]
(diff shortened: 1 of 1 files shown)


ee/vellum_ee/workflows/display/workflows/tests/test_workflow_display.py
```diff
@@ -854,7 +855,17 @@ class MyWorkflow(BaseWorkflow[Inputs, BaseState]):
...
     assert isinstance(input_variables, list)
-    assert len(input_variables) == 4
+    assert len(input_variables) == 5
+
+    empty_string = next(var for var in input_variables if isinstance(var, dict) and var["key"] == "empty_string")
+    assert empty_string == {
+        "id": "09c3e825-8932-4d37-990a-b8a7c2a3bdec",
+        "key": "empty_string",
+        "type": "STRING",
+        "default": None,
```
dvargasfuertes, line 865: [oos] yea I would expect a default value of `{"type": "STRING", "value": ""}`

dvargas92495 (approved):
would love to iterate on this with the proper behavior if straightforward

PR #2912, opened 2025-11-01
Add trigger contexts and getTriggerPathToIdMapping

NgoHarrison:


ee/codegen/src/__test__/project.test.ts
```diff
@@ -313,6 +313,134 @@ describe("WorkflowProjectGenerator", () => {
...
+      vellumApiKey: "<TEST_API_KEY>",
+    });
+
+    await project.generateCode();
+
+    const mapping = project.getTriggerPathToIdMapping();
+    expect(mapping).toEqual({
+      "code.triggers.github-star-event": "trigger-1",
+      "code.triggers.scheduled": "trigger-2",
+    });
```
dvargasfuertes: [blocking] I would love for `.generateCode()` to just start generating the `metadata.json` under the hood instead of adding more public methods on `project`. Thoughts?

ee/codegen/src/project.ts
```diff
@@ -622,6 +626,30 @@ ${errors.slice(0, 3).map((err) => {
...
+      this.workflowContext.triggers.forEach((trigger) => {
+        try {
+          createTriggerContext({
+            workflowContext: this.workflowContext,
+            triggerData: trigger,
+          });
+        } catch (error) {
+          if (error instanceof BaseCodegenError) {
+            this.workflowContext.addError(error);
+          } else {
```
dvargasfuertes: [nit] this part might be redundant as I'm not seeing us raise any uncaught `BaseCodegenError`s in `createTriggerContext`. If this was copy pasted from `createNodeContext`, that method used to have uncaught `BaseCodegenError`s that we should have them all wrapped in `this.workflowContext.addError` at this point I believe

ee/codegen/src/context/workflow-context/workflow-context.ts
```diff
@@ -113,6 +119,11 @@ export class WorkflowContext {
   // new nodes.
   private readonly nodeModuleNames: Set<string> = new Set();
 
+  // Maps trigger IDs to trigger contexts
+  // Tracks local and global contexts in the case of nested workflows.
+  public readonly triggerContextsByTriggerId: TriggerContextsByTriggerId;
+  public readonly globalTriggerContextsByTriggerId: TriggerContextsByTriggerId;
```
dvargasfuertes: [nit] as of right now, I don't believe we will have triggers for nested workflows, so we may get away with just one of these

ee/codegen/src/context/trigger-context/base.ts
```diff
@@ -0,0 +1,41 @@
...
+    this.triggerData = args.triggerData;
+
+    const { moduleName, className, modulePath } = this.getTriggerModuleInfo();
+    this.triggerModuleName = moduleName;
+    this.triggerClassName = className;
+    this.triggerModulePath = modulePath;
+    this.triggerFileName = `${this.triggerModuleName}.py`;
+  }
+
+  protected abstract getTriggerModuleInfo(): {
```
dvargasfuertes, line 30: [oos] would be good to clean up the graph-attribute PR by using this afterwards instead of the disconnected helper it's currently using

ee/codegen/src/__test__/project.test.ts
NgoHarrison: Yeah I am down with this. I was already hesitant with how I was basically stacking new mappings as separate methods to be called. I updated the `generateCode` to generate the metadata now

ee/codegen/src/__test__/project.test.ts
```diff
@@ -314,7 +314,7 @@ describe("WorkflowProjectGenerator", () => {
   });
 
   describe("generateCode", () => {
-    const excludeFilesAtPaths: RegExp[] = [/\.pyc$/];
+    const excludeFilesAtPaths: RegExp[] = [/\.pyc$/, /metadata\.json$/];
```
dvargasfuertes, line 317: [oos] Would actually be good to reinclude this in the future to assert the ids we are moving over

dvargas92495 (approved):

PR #2917, opened 2025-11-01
use trigger path to id mapping from metadata json to generate id for triggers

NgoHarrison:
This PR helps introduce a new pattern for using mappings within our metadata.json to drive serialization of values. In this case it is for workflow trigger id's

### DIFF
--- a/ee/vellum_ee/workflows/display/tests/workflow_serialization/test_scheduled_trigger_serialization.py
+++ b/ee/vellum_ee/workflows/display/tests/workflow_serialization/test_scheduled_trigger_serialization.py
@@ -12,6 +12,7 @@ from vellum.workflows.nodes.bases.base import BaseNode
 from vellum.workflows.state.base import BaseState
 from vellum.workflows.triggers.base import _get_trigger_path_to_id_mapping
 from vellum.workflows.triggers.schedule import ScheduleTrigger
+from vellum.workflows.utils.uuids import uuid4_from_hash
 from vellum_ee.workflows.display.workflows.get_vellum_workflow_display_class import get_workflow_display
 
 
@@ -136,7 +137,10 @@ def test_scheduled_trigger_serialization_without_metadata_json():
     assert trigger["type"] == "SCHEDULED"
 
     # AND the id is the trigger class's default __id__ (hash-based)
-    expected_default_id = str(WeeklyScheduleTrigger.__id__)
+    expected_module_path = (
+        f"{__name__}.test_scheduled_trigger_serialization_without_metadata_json.<locals>.WeeklyScheduleTrigger"
+    )
+    expected_default_id = str(uuid4_from_hash(expected_module_path))
     assert trigger["id"] == expected_default_id
 
     # AND attributes are serialized
--- a/ee/vellum_ee/workflows/display/tests/workflow_serialization/test_scheduled_trigger_serialization.py
+++ b/ee/vellum_ee/workflows/display/tests/workflow_serialization/test_scheduled_trigger_serialization.py
@@ -1,0 +1,145 @@
+from contextlib import contextmanager
+import json
+import os
+import tempfile
+from unittest.mock import patch
+from uuid import UUID
+from typing import Iterator, Tuple
+
+from vellum.workflows import BaseWorkflow
+from vellum.workflows.inputs.base import BaseInputs
+from vellum.workflows.nodes.bases.base import BaseNode
+from vellum.workflows.state.base import BaseState
+from vellum.workflows.triggers.base import _get_trigger_path_to_id_mapping
+from vellum.workflows.triggers.schedule import ScheduleTrigger
+from vellum_ee.workflows.display.workflows.get_vellum_workflow_display_class import get_workflow_display
+
+
+@contextmanager
+def mock_metadata_json(trigger_class_name: str, trigger_id: UUID) -> Iterator[Tuple[str, UUID]]:
+    """
+    Context manager that mocks metadata.json with trigger ID mapping.
+
+    Args:
+        trigger_class_name: Name of the trigger class (e.g., "DailyScheduleTrigger")
+        trigger_id: The UUID to map the trigger to
+
+    Yields:
+        Tuple of (trigger_module_path, trigger_id)
+    """
+    # Create a temporary directory to simulate a workflow module
+    with tempfile.TemporaryDirectory() as tmpdir:
+        # Create metadata.json with trigger mapping
+        metadata_path = os.path.join(tmpdir, "metadata.json")
+        trigger_module_path = f"{__name__}.{trigger_class_name}"
+
+        with open(metadata_path, "w") as f:
+            json.dump(
+                {
+                    "trigger_path_to_id_mapping": {
+                        trigger_module_path: str(trigger_id),
+                    }
+                },
+                f,
+            )
+
+        # Mock the workflow root finder to return our temp directory
+        # Mock os.path.exists to return True for our metadata.json
+        with patch("vellum.workflows.triggers.base._find_workflow_root_with_metadata", return_value=tmpdir):
+            with patch("vellum.workflows.triggers.base.os.path.exists", return_value=True):
+                with patch("vellum.workflows.triggers.base.virtual_open", open):
+                    # Clear the LRU cache to ensure fresh read
+                    _get_trigger_path_to_id_mapping.cache_clear()
+
+                    yield trigger_module_path, trigger_id
+
+
+def test_scheduled_trigger_serialization_with_metadata_json():
+    """ScheduleTrigger uses mapped ID from metadata.json when available."""
+
+    # GIVEN we have an expected trigger ID
+    expected_trigger_id = UUID("12345678-1234-1234-1234-123456789abc")
+
+    # AND we have metadata.json set up with the trigger mapping
+    # Note: Use the full qualified name including the test function name and <locals>
+    with mock_metadata_json(
+        "test_scheduled_trigger_serialization_with_metadata_json.<locals>.DailyScheduleTrigger",
+        expected_trigger_id,
+    ):
+        # Define the trigger class AFTER setting up mocks
+        # This ensures the metaclass reads from our mocked metadata.json
+        class DailyScheduleTrigger(ScheduleTrigger):
+            class Config:
+                cron = "0 9 * * *"  # Every day at 9am
+                timezone = "America/New_York"
+
+        class ProcessNode(BaseNode):
+            class Outputs(BaseNode.Outputs):
+                timestamp = DailyScheduleTrigger.current_run_at
+
+            def run(self) -> Outputs:
+                return self.Outputs()
+
+        # AND a workflow that has the scheduled trigger as the entrypoint
+        class TestWorkflow(BaseWorkflow[BaseInputs, BaseState]):
+            graph = DailyScheduleTrigger >> ProcessNode
+
+        # WHEN we serialize the workflow
+        # The trigger ID should be automatically loaded from metadata.json by the metaclass
+        result: dict = get_workflow_display(workflow_class=TestWorkflow).serialize()
+
+        # THEN we get the expected trigger
+        assert "triggers" in result
+        triggers = result["triggers"]
+        assert isinstance(triggers, list)
+        assert len(triggers) == 1
+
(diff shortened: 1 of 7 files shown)


ee/vellum_cli/pull.py
```diff
@@ -317,8 +318,6 @@ def pull_command(
                     if file_name == ERROR_LOG_FILE_NAME:
                         error_content = content
                         continue
-                    if file_name == METADATA_FILE_NAME:
-                        continue
```
dvargasfuertes, line 321: we're going the metadata.json route!!

ee/vellum_cli/push.py
```diff
@@ -130,12 +130,22 @@ def push_command(
...
+            metadata_json = json.load(f)
+            trigger_path_to_id_mapping = metadata_json.get("trigger_path_to_id_mapping", {})
+
     # Remove this once we could serialize using the artifact in Vembda
     # https://app.shortcut.com/vellum/story/5585
     serialization_result = BaseWorkflowDisplay.serialize_module(
         workflow_config.module,
         client=client,
         dry_run=dry_run or False,
+        trigger_path_to_id_mapping=trigger_path_to_id_mapping,
```
dvargasfuertes: [blocking] Oh this might scale poorly and involve a kwarg for every kind of id mapping. If we already know how to get the mapping based on `workflow_config.module`, then this kwarg is redundant and could be calculated internally from `workflow_config.module` instead

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -127,7 +128,7 @@ def __init__(
         self._serialized_files = []
         self._dry_run = dry_run
 
-    def serialize(self) -> JsonObject:
+    def serialize(self, trigger_path_to_id_mapping: Optional[dict] = None) -> JsonObject:
```
dvargasfuertes: [blocking] same here - would not expect to need a kwarg

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -383,9 +387,12 @@ def serialize(self) -> JsonObject:
             )
 
         # Add edges from trigger/entrypoint to first nodes
-        if has_only_integration_trigger:
```
dvargasfuertes: [oos] this logic doesn't look right, but it's not related to this PR so we don't have to address here. It's saying that we are treating an integration trigger as the entrypoint node which we don't want to do. Will put up a test separately to break this

dvargasfuertes: [In progress](https://app.devin.ai/sessions/52f110685d524a06ae84c1307b3d9e69)

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -536,21 +548,27 @@ def _serialize_workflow_trigger(self) -> Optional[JsonArray]:
...
                 )
 
         # Return as a list with a single trigger object matching Django schema
-        trigger_id = trigger_class.__id__
+        # Use the mapped ID if available, otherwise use the class's default ID
+        trigger_module_path = f"{trigger_class.__module__}.{trigger_class.__name__}"
+        if trigger_path_to_id_mapping and trigger_module_path in trigger_path_to_id_mapping:
+            trigger_id = UUID(trigger_path_to_id_mapping[trigger_module_path])
+        else:
+            trigger_id = trigger_class.__id__
```
dvargasfuertes: [blocking] Ok so the way `Base[Node|Workflow]Display` works is that when they are defined, they automatically edit these id fields directly: https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/vellum_ee/workflows/display/nodes/base_node_display.py#L91-L93

This makes it so `node.__id__` already equals the new id.

We want a similar approach to ids loaded from the `metadata.json`. Here's a proposal:
- In the base trigger metaclass, we try to read from the `metadata.json`: https://github.com/vellum-ai/vellum-python-sdks/blob/main/src/vellum/workflows/triggers/base.py#L35
- The method that tries to read from it is cached. We also want to use a special kind of `open` that takes into account how vembda has the files loaded via the VirtualFileLoader: https://github.com/vellum-ai/vellum-python-sdks/blob/main/ee/vellum_ee/workflows/display/utils/expressions.py#L68
    - This one will need a quick refactor in order to be moved to non-ee code. In progress [here](https://app.devin.ai/sessions/115d9267940c4e02bde83bdbc29f60cf)
- The trigger metaclass tries to lookup the id from the metadata.json, otherwise falls back to default that's currently there
- In tests below, you'll need a fixture that loads the metadata.json into a VirtualFileLoader or mocking the `open` method.

Thoughts?

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -576,12 +594,12 @@ def _serialize_workflow_trigger(self) -> Optional[JsonArray]:
             "attributes": trigger_attributes,
         }
 
-        # For INTEGRATION triggers, include class name, module path, and source_handle_id
+        # For INTEGRATION and SCHEDULED triggers, include class name, module path, and source_handle_id
         # ManualTrigger doesn't need source_handle_id because edges come from ENTRYPOINT node
-        if trigger_type == WorkflowTriggerType.INTEGRATION:
+        if trigger_type in (WorkflowTriggerType.INTEGRATION, WorkflowTriggerType.SCHEDULED):
             trigger_data["class_name"] = trigger_class.__name__
             trigger_data["module_path"] = cast(Json, trigger_class.__module__.split("."))
```
dvargasfuertes, line 578: [oos] this will go in `definition` field

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -576,12 +594,12 @@ def _serialize_workflow_trigger(self) -> Optional[JsonArray]:
...
-            trigger_data["source_handle_id"] = str(trigger_id)  # IntegrationTrigger acts as edge source
+            trigger_data["source_handle_id"] = str(trigger_id)  # Trigger acts as edge source
```
dvargasfuertes, line 579: [oos] this is redundant for now and I think not needed

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -576,12 +594,12 @@ def _serialize_workflow_trigger(self) -> Optional[JsonArray]:
             "attributes": trigger_attributes,
         }
 
-        # For INTEGRATION triggers, include class name, module path, and source_handle_id
+        # For INTEGRATION and SCHEDULED triggers, include class name, module path, and source_handle_id
         # ManualTrigger doesn't need source_handle_id because edges come from ENTRYPOINT node
-        if trigger_type == WorkflowTriggerType.INTEGRATION:
+        if trigger_type in (WorkflowTriggerType.INTEGRATION, WorkflowTriggerType.SCHEDULED):
```
dvargasfuertes, line 576: [nit] I think we want this no matter the trigger type

ee/vellum_cli/push.py
dvargasfuertes: [nit] I'm also tempted to just call this `path_to_id_mapping` as the place to store _all_ the ids we currently have in display in the future, but will leave up to you if you want to keep the `trigger_` prefix to keep focused for now and prove out the pattern

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
NgoHarrison: Removed

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
NgoHarrison: Yup this checks. Updated to address some of these points

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_scheduled_trigger_serialization.py
```diff
@@ -0,0 +1,145 @@
...
+from vellum.workflows.inputs.base import BaseInputs
+from vellum.workflows.nodes.bases.base import BaseNode
+from vellum.workflows.state.base import BaseState
+from vellum.workflows.triggers.base import _get_trigger_path_to_id_mapping
+from vellum.workflows.triggers.schedule import ScheduleTrigger
+from vellum_ee.workflows.display.workflows.get_vellum_workflow_display_class import get_workflow_display
+
+
+@contextmanager
+def mock_metadata_json(trigger_class_name: str, trigger_id: UUID) -> Iterator[Tuple[str, UUID]]:
```
dvargasfuertes, line 20: [oos] can avoid the additional nesting by using `pytest.fixture`'s not worth prioritizing until post tracer bullet

ee/vellum_ee/workflows/display/tests/workflow_serialization/test_scheduled_trigger_serialization.py
```diff
@@ -0,0 +1,145 @@
...
+    assert isinstance(triggers, list)
+    assert len(triggers) == 1
+
+    trigger = triggers[0]
+    assert isinstance(trigger, dict)
+    assert trigger["type"] == "SCHEDULED"
+
+    # AND the id is the trigger class's default __id__ (hash-based)
+    expected_default_id = str(WeeklyScheduleTrigger.__id__)
+    assert trigger["id"] == expected_default_id
```
dvargasfuertes: [nit] this assertion would be true even in the metadata.json case bc we are updating the `__id__` here. I would set the expected to be the hardcoded value here

src/vellum/workflows/triggers/base.py
```diff
@@ -27,13 +31,85 @@ def _is_annotated(cls: Type, name: str) -> bool:
...
         cls = super().__new__(mcs, name, bases, dct)
         trigger_class = cast(Type["BaseTrigger"], cls)
 
+        # Set default ID based on module and class name
         trigger_class.__id__ = uuid4_from_hash(f"{trigger_class.__module__}.{trigger_class.__qualname__}")
 
+        # Try to override with ID from metadata.json if available
+        # This approach is similar to how BaseNodeDisplay automatically sets node IDs
+        workflow_root = _find_workflow_root_with_metadata(trigger_class.__module__)
+        if workflow_root:
```
dvargasfuertes: [nit] I can see this logic being in `_get_trigger_path_to_id_mapping` for when we want to use it for nodes/workflows in the future

src/vellum/workflows/triggers/base.py
```diff
@@ -27,13 +31,85 @@ def _is_annotated(cls: Type, name: str) -> bool:
...
+        metadata_path = os.path.join(module_dir, "metadata.json")
+
+        # Check if metadata.json exists
+        if not os.path.exists(metadata_path):
+            return {}
+
+        # Use virtual_open to support both regular and virtual environments
+        with virtual_open(metadata_path) as f:
+            metadata_json = json.load(f)
+            trigger_mapping = metadata_json.get("trigger_path_to_id_mapping", {})
```
dvargasfuertes, line 89: [oos] how do we foresee the ids from nodes, node outputs, and workflows being moved to in the future?

src/vellum/workflows/utils/files.py
```diff
@@ -0,0 +1,29 @@
...
+def virtual_open(file_path: str, mode: str = "r"):
+    """
+    Open a file, checking BaseWorkflowFinder instances first before falling back to regular open().
+
+    This function is used to support reading files in both regular and virtual environments
+    (e.g., when using VirtualFileLoader in serverless/Vembda deployments).
+
+    Args:
+        file_path: Path to the file to open
+        mode: File open mode (default: "r")
```
dvargasfuertes: [nit] ah damn sorry, I left this arg here. I would remove the arg and use it in `open` below

dvargas92495 (approved):

src/vellum/workflows/triggers/base.py
NgoHarrison: I think metadata.json can have that outer key like `mappings` or whatever and inside have nested mapping keys like trigger ids, node ids, etc.

PR #2922, opened 2025-11-02
Fix default falsy serilaization

vincent0426:


ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -141,7 +141,7 @@ def serialize(self) -> JsonObject:
         for workflow_input_reference, workflow_input_display in self.display_context.workflow_input_displays.items():
             default = (
                 primitive_to_vellum_value(workflow_input_reference.instance)
-                if workflow_input_reference.instance
+                if workflow_input_reference.instance is not None and workflow_input_reference.instance is not undefined
```
dvargasfuertes, line 144: [oos] we should work towards just `is not undefined`, but there are some unknown cases today where we are incorrectly returning `None` here, which is a valid value `null`

dvargas92495 (approved):

PR #2935, opened 2025-11-03
Add cache to MCP tool list

vincent0426:


vincent0426:
@codex review

src/vellum/workflows/utils/functions.py
```diff
@@ -277,6 +278,7 @@ def get_mcp_tool_name(tool_def: MCPToolDefinition) -> str:
     return f"{server_name}__{tool_def.name}"
 
 
+@lru_cache(maxsize=1)
 def compile_mcp_tool_definition(server_def: MCPServer) -> List[MCPToolDefinition]:
```
dvargasfuertes, line 282: [blocking] I'm a bit worried about this since the whole point of MCP is that it can return you different tools based on different circumstances, which may or may not be encoded in the url/name. it's also a bit sus to have a hash that includes api header kay values in it. is there a way to save us on the calls to this method without adding the `lru_cache`, within the context of just the tool calling node's execution?

vincent0426: wouldn't each new thread have fresh cache? don't think MCP list tools return different tools on each run under same thread

dvargasfuertes: list tools is controlled by the server, so they in theory can do whatever they want

I'm not sure if each thread has a fresh cache actually. If so, it would downgrade this to a `[nit]`

PR #2940, opened 2025-11-03
Refactor integration trigger and introduce base trigger abstract class

NgoHarrison:
Set up for scheduled trigger and the display data generation (will be shared among both)

### DIFF
--- a/ee/codegen/src/generators/triggers/base-trigger.ts
+++ b/ee/codegen/src/generators/triggers/base-trigger.ts
@@ -1,0 +1,111 @@
+import { python } from "@fern-api/python-ast";
+
+import {
+  GENERATED_TRIGGERS_MODULE_NAME,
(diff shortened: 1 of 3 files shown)


ee/codegen/src/generators/triggers/base-trigger.ts
```diff
@@ -0,0 +1,111 @@
...
+   */
+  protected abstract getTriggerClassBody(): AstNode[];
+
+  /**
+   * Optionally create a Display class for this trigger.
+   * Override this method to provide a custom Display class.
+   * @returns The Display class AstNode, or undefined if no Display class is needed
+   */
+  protected createDisplayClass(): AstNode | undefined {
+    return undefined;
```
NgoHarrison, line 60: Will implement in follow up PRs

PR #2951, opened 2025-11-04
[APO-2050][Part 2] Generate display data for triggers

NgoHarrison:


ee/codegen/src/__test__/__snapshots__/generate-code.test.ts.snap
```diff
@@ -133,6 +133,13 @@ class SlackMessageTrigger(IntegrationTrigger):
         type = "COMPOSIO"
         integration_name = "slack"
         slug = "SLACK_MESSAGE_TRIGGER"
+
+    class Display(IntegrationTrigger.Display):
+        label = "Slack Message Trigger"
+        position = {"x": 100, "y": 200}
```
dvargasfuertes, line 139: [nit] thoughts on just having `x` and `y` top level so that they are typed? or reuse the pydantic data model we already have `Position`

dvargas92495 (commented):
✅

PR #2952, opened 2025-11-04
Harrison/codegen scheduled trigger

NgoHarrison:


ee/codegen/src/__test__/__snapshots__/generate-code.test.ts.snap
```diff
@@ -120,6 +120,43 @@ class InnerMapNode(MapNode):
...
 
+exports[`generateCode > should generate code for %1 scheduled-trigger.ts > triggers/scheduled.py 1`] = `
+"from datetime import datetime
+
+from vellum.workflows.triggers import BaseTrigger
+
+
+class ScheduleTrigger(BaseTrigger):
+    current_run_at: datetime
+    next_run_at: datetime
```
NgoHarrison, line 131: [q] Do we want these two fields? I can see `next_run_at` being exposed but don't think `current_run_at` should be here. Thoughts @dvargas92495? In any case, I need to expose these two as these aren't currently exposed in the sandbox UI json

dvargasfuertes: seems like we resolved in followup PRs

bc they are defined in the base class, we won't need them here

dvargas92495 (commented):
✅

PR #2955, opened 2025-11-04
[APO-1967] Codegen setstate node

vincent0426:


ee/codegen/src/__test__/nodes/__snapshots__/set-state-node.test.ts.snap
```diff
@@ -0,0 +1,54 @@
...
+
+
+class SetStateNode(BaseSetStateNode[State]):
+    operations = {
+        "chat_history": State.chat_history.concat(MyCustomNode.Outputs.chat_history),
+        "counter": State.counter + 1,
+    }
+
+    class Outputs(BaseSetStateNode.Outputs):
+        result: Any
```
dvargasfuertes: [nit] Can we add similar default output prevention here?

dvargas92495 (approved):

PR #2960, opened 2025-11-04
Harrison/fix import of triggers

NgoHarrison:


ee/codegen/src/__test__/__snapshots__/generate-code.test.ts.snap
```diff
@@ -123,18 +123,18 @@ class InnerMapNode(MapNode):
...
 "from datetime import datetime
 
-from vellum.workflows.triggers import BaseTrigger
+from vellum.workflows.triggers import ScheduleTrigger as TriggersScheduleTrigger
 
 
-class ScheduleTrigger(BaseTrigger):
+class ScheduleTrigger(TriggersScheduleTrigger):
     current_run_at: datetime
     next_run_at: datetime
```
dvargasfuertes, line 131: [oos] we should be able to now avoid generating these since it's redundant with the base class

ee/codegen/src/utils/triggers.ts
dvargasfuertes, line 15: [oos] we should be able to delete this helper in favor of our new context's

dvargas92495 (commented):
✅

PR #2965, opened 2025-11-04
Harrison/fix attribute for scheduled trigger init

NgoHarrison:

PR #2969, opened 2025-11-05
Use module + class name for trigger path key mapping

NgoHarrison:
Addressed these two follow up points:
1. [Use classname to future proof multiple triggers in trigger path mapping](https://github.com/vellum-ai/vellum-python-sdks/pull/2965#discussion_r2492567345)
2. [Have helpers](https://github.com/vellum-ai/vellum-python-sdks/pull/2965#discussion_r2492558520)

### DIFF
--- a/src/vellum/workflows/triggers/base.py
+++ b/src/vellum/workflows/triggers/base.py
@@ -31,6 +31,57 @@ def _is_annotated(cls: Type, name: str) -> bool:
     return False
 
 
+def _convert_to_relative_module_path(absolute_module_path: str, workflow_root: str) -> str:
+    """
+    Convert an absolute module path to a relative path from the workflow root.
+
+    Args:
+        absolute_module_path: The full module path (e.g., "workflow_id.triggers.scheduled")
+        workflow_root: The workflow root module (e.g., "workflow_id")
+
+    Returns:
+        Relative module path with leading dot (e.g., ".triggers.scheduled")
+    """
+    if not absolute_module_path.startswith(workflow_root):
+        return ""
+
+    remaining_path = absolute_module_path[len(workflow_root) :]
+    if remaining_path.startswith("."):
+        return remaining_path
+    else:
+        return "." + remaining_path
+
+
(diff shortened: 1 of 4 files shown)


src/vellum/workflows/triggers/base.py
```diff
@@ -117,20 +168,11 @@ def __new__(mcs, name: str, bases: Tuple[Type, ...], dct: Dict[str, Any]) -> Any
...
-        # Convert absolute module path to relative path for lookup
-        workflow_root = _find_workflow_root_with_metadata(trigger_class.__module__)
-        if workflow_root and trigger_class.__module__.startswith(workflow_root):
-            remaining_path = trigger_class.__module__[len(workflow_root) :]
-            if remaining_path.startswith("."):
-                relative_trigger_module_path = remaining_path
-            else:
-                relative_trigger_module_path = "." + remaining_path
+        # Try to override with ID from metadata.json if available
+        metadata_id = _get_trigger_id_from_metadata(trigger_class.__module__, trigger_class.__qualname__)
```
dvargasfuertes: [nit] seems like we could just pass in `trigger_class`

src/vellum/workflows/triggers/base.py
```diff
@@ -31,6 +31,57 @@ def _is_annotated(cls: Type, name: str) -> bool:
...
+    workflow_root = _find_workflow_root_with_metadata(trigger_module)
+    if not workflow_root:
+        return None
+
+    trigger_path_to_id_mapping = _get_trigger_path_to_id_mapping(trigger_module)
+    if not trigger_path_to_id_mapping:
+        return None
+
+    # Convert module path to relative path and append class name
+    # e.g., "workflow_id.triggers.scheduled" + "ScheduleTrigger" -> ".triggers.scheduled.ScheduleTrigger"
```
dvargasfuertes: [q] `root_module` instead of `workflow_id` correct? I guess in vembda it looks like an id

dvargas92495 (approved):

src/vellum/workflows/triggers/base.py
NgoHarrison: Ah yeah, cursor got confused. Let me update

PR #2971, opened 2025-11-05
Change attributes from nodeattribute to vellum variable for triggers

NgoHarrison:
https://github.com/vellum-ai/vellum/blob/9acd820f24566acc2986955845e5f9b349a3e92b/django/app/domain_models/types/workflows.py#L2133

### DIFF
--- a/ee/codegen/src/serializers/vellum.ts
+++ b/ee/codegen/src/serializers/vellum.ts
@@ -2308,7 +2308,7 @@ export const WorkflowOutputValueSerializer: ObjectSchema<
 const ManualTriggerSerializer = objectSchema({
   id: stringSchema(),
   type: stringLiteralSchema("MANUAL"),
-  attributes: listSchema(NodeAttributeSerializer),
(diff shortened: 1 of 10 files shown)


ee/codegen/src/serializers/vellum.ts
```diff
@@ -2308,7 +2308,7 @@ export const WorkflowOutputValueSerializer: ObjectSchema<
 const ManualTriggerSerializer = objectSchema({
   id: stringSchema(),
   type: stringLiteralSchema("MANUAL"),
-  attributes: listSchema(NodeAttributeSerializer),
+  attributes: listSchema(VellumVariableSerializer),
```
dvargasfuertes, line 2311: [q] Are we changing this on the django side too?

dvargas92495 (commented):
✅

ee/codegen/src/serializers/vellum.ts
NgoHarrison: It's already vellum variable on django side

PR #2972, opened 2025-11-05
Update type to provider in integration trigger config

NgoHarrison:


ee/codegen/src/__test__/__snapshots__/generate-code.test.ts.snap
```diff
@@ -115,7 +116,7 @@ class LinearCommentEventTrigger(IntegrationTrigger):
     url: str
 
     class Config(IntegrationTrigger.Config):
-        type = "COMPOSIO"
+        provider = VellumIntegrationProviderType.COMPOSIO
         integration_name = "LINEAR"
```
NgoHarrison: I am guessing type === provider? We have a type of composio in the json but we called this provider here in the config. cc @dvargas92495 @emmiekehoe

dvargas92495 (approved):

PR #2986, opened 2025-11-06
Regenerate node definition

vincent0426:


ee/vellum_ee/assets/node-definitions.json
```diff
@@ -309,29 +309,17 @@
...
-          "id": "76f813d6-0025-4733-9d03-848c49c93d66",
+          "id": "40f6daf8-16dd-41ee-8f85-4ed717db24be",
           "name": "ml_model",
-          "value": {
-            "type": "CONSTANT_VALUE",
-            "value": {
-              "type": "JSON",
-              "value": null
-            }
-          }
```
vincent0426, line 320: is this intentional?

dvargasfuertes: Yea I think so

dvargas92495 (commented):
✅

PR #2994, opened 2025-11-06
[APO-2041] Codegen default factory for empty list and dict

vincent0426:
I'm assuming our python sdk package will install pydantic right?

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -1295,6 +1295,22 @@ class WeatherFunction(BaseWorkflow[Inputs, BaseState]):
 "
 `;
 
+exports[`WorkflowProjectGenerator > include sandbox > should generate correct inputs for empty list and dict defaults > code/inputs.py 1`] = `
+"from pydantic import Field
+from typing import Any, List, Optional
+
+from vellum import VellumValue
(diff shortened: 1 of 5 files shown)


NgoHarrison (approved):

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -1295,6 +1295,22 @@ class WeatherFunction(BaseWorkflow[Inputs, BaseState]):
 "
 `;
 
+exports[`WorkflowProjectGenerator > include sandbox > should generate correct inputs for empty list and dict defaults > code/inputs.py 1`] = `
+"from pydantic import Field
```
dvargasfuertes, line 1299: yea pydantic is a nested dep

ee/codegen/src/generators/vellum-variable.ts
```diff
@@ -70,12 +70,71 @@ export class VellumVariable extends AstNode {
...
+            name: "default_factory",
+            value: python.reference({
+              name: "dict",
+            }),
+          }),
+        ],
+      });
+    }
+
+    // For non-empty defaults, use the regular VellumValue
```
dvargasfuertes, line 133: [oos] for non-empty lists and objects, would still expect `Field(default_factory=lambda:{"non": "empty"})`

dvargas92495 (approved):

PR #2995, opened 2025-11-07
Remove `openai` as Dependency

noanflaherty:
We no longer have any code that depends on the openai package. Let's remove it entirely.


vincent0426 (approved):

dvargas92495:
✅

NgoHarrison (commented):
✅

PR #2997, opened 2025-11-07
[APO-2104] Add null check on weight value for search nodes

NgoHarrison:


ee/codegen/src/generators/nodes/search-node.ts
```diff
@@ -127,29 +128,44 @@ export class SearchNode extends BaseNode<
     return bodyStatements;
   }
 
-  private getSearchWeightsRequest(): ClassInstantiation {
+  private getSearchWeightsRequest(): AstNode {
     const weightsRule =
       this.findNodeInputByName("weights")?.nodeInputData?.value.rules[0];
     if (!weightsRule || weightsRule.type !== "CONSTANT_VALUE") {
-      throw new NodeAttributeGenerationError("weights input is required");
+      throw new NodeAttributeGenerationError("weights input is required", "WARNING");
```
dvargasfuertes: [nit] `this.workflowContext.addError(`

ee/codegen/src/generators/nodes/search-node.ts
```diff
@@ -127,29 +128,44 @@ export class SearchNode extends BaseNode<
...
+    }
+
+    // Accept null/empty JSON values and add a warning instead of throwing
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const rawValue: any = (weightsRule as any).data?.value;
```
dvargasfuertes: [nit] we shouldn't need the `any` imo. we should properly treat `weightsRule.data.value` as the `unknown` as we are being told and ensure it's not null and is an object. either that or pass it into a zod schema to safe parse

ee/codegen/src/generators/nodes/search-node.ts
```diff
@@ -127,29 +128,44 @@ export class SearchNode extends BaseNode<
...
     // TODO: Determine what we want to cast JSON values to
     //  https://app.shortcut.com/vellum/story/5459
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const { semantic_similarity, keywords } = weightsRule.data.value as Record<
+    const { semantic_similarity, keywords } = rawValue as Record<
       string,
       unknown
     >;
     if (typeof semantic_similarity !== "number") {
       throw new NodeAttributeGenerationError(
```
dvargasfuertes: [nit] let's throw `this.workflowContext.addError(` on all of these too

dvargas92495 (approved):

ee/codegen/src/generators/nodes/search-node.ts
NgoHarrison: Updated

ee/codegen/src/generators/nodes/search-node.ts
NgoHarrison: Updated

ee/codegen/src/generators/nodes/search-node.ts
NgoHarrison: Updated

PR #3001, opened 2025-11-07
Serialize cron and timezone for scheduled triggers

NgoHarrison:


ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -575,28 +575,46 @@ def _serialize_workflow_trigger(self) -> Optional[JsonArray]:
...
+                trigger_data = {
+                    "id": str(trigger_id),
+                    "type": trigger_type.value,
+                    "cron": cron_value,
+                    "timezone": timezone_value,
+                    "attributes": [],
+                }
+            else:
+                # For other triggers, serialize attributes from attribute_references
+                attribute_references = trigger_class.attribute_references().values()
```
dvargasfuertes, line 596: [oos] would love to support `__iter__` on trigger_class then reserving a special name `attribute_references`

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -575,28 +575,45 @@ def _serialize_workflow_trigger(self) -> Optional[JsonArray]:
...
+                cron_value = getattr(config_class, "cron", None)
+                timezone_value = getattr(config_class, "timezone", None)
+
+                trigger_data: JsonObject = {
+                    "id": str(trigger_id),
+                    "type": trigger_type.value,
+                    "cron": cron_value,
+                    "timezone": timezone_value,
+                    "attributes": [],
+                }
```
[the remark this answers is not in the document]
dvargasfuertes: [oos] similar to this vibe, ideally we don't need special logic for `SCHEDULED` in `base_workflow_display`. Ideally, we always store the attributes on the current class that aren't defined in the base class

ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -575,28 +575,46 @@ def _serialize_workflow_trigger(self) -> Optional[JsonArray]:
...
+                trigger_attributes: JsonArray = cast(
+                    JsonArray,
+                    [
+                        cast(
+                            JsonObject,
+                            {
+                                "id": str(reference.id),
+                                "name": reference.name,
+                                "type": primitive_type_to_vellum_variable_type(reference),
+                                "value": None,
```
dvargasfuertes, line 606: [blocking] we'll need to update our serialization paths now that we are using `VellumVariable`s. Understand this was an issue in the previous implementation too, just marking blocking to indicate urgency, particularly before AB starts generating

dvargas92495 (commented):
✅

PR #3040, opened 2025-11-11
Downgrade api node error

vincent0426:
I think we don't need to throw error during codegen?
sentry https://vellum.sentry.io/issues/6938937705/?alert_rule_id=15639644&alert_type=issue&notification_uuid=cd1f655b-54d7-48fa-bc1f-1c5f90b36071&project=4508362703634432&referrer=slack


dvargas92495 (approved):
we should continue to downgrade all of these we see

PR #3041, opened 2025-11-11
Fix vellum resolver ids

dvargas92495:
This bug was leading to all messages in the thread resolving to the same previous_span_id. This PR fixes

### DIFF
--- a/src/vellum/workflows/resolvers/resolver.py
+++ b/src/vellum/workflows/resolvers/resolver.py
@@ -78,8 +78,8 @@ class VellumResolver(BaseWorkflowResolver):
 
         return LoadStateResult(
             state=state,
-            previous_trace_id=response.previous_trace_id,
-            previous_span_id=response.previous_span_id,
+            previous_trace_id=response.trace_id,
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/resolvers/resolver.py
```diff
@@ -62,24 +62,10 @@ def load_state(self, previous_execution_id: Optional[Union[UUID, str]] = None) -
...
-                previous_trace_id=response.trace_id,
-                previous_span_id=response.span_id,
-                root_trace_id=response.trace_id,
-                root_span_id=response.span_id,
-            )
-
         return LoadStateResult(
             state=state,
-            previous_trace_id=response.previous_trace_id,
-            previous_span_id=response.previous_span_id,
```
dvargasfuertes, line 82: we likely can remove these two fields from the API since we don't actually need them

dvargas92495:
@codex review

siddseethepalli (commented):
‼️

marlenc3:
✅

PR #3044, opened 2025-11-11
Fix Python 3.13 compatibility: safely access __annotations__

alex-nork:
The purpose of this PR is to fix Python 3.13 compatibility by safely accessing the `__annotations__` attribute. I was seeing a lot of [acid_test](https://vellum-ai.github.io/vellum-copilot/#suites/5cdebc5f8728009af88397c02c46cdcc/414c8149680982e7/) failures like this 

`E       - run_workflow:split_into_chunks: type object 'object' has no attribute '__annotations__'` 

so hopefully fixing this will help us better understand why these tests are failing.

## Changes Made
- Update error handling in workflows/types/utils.py to use getattr() with default value when accessing `__annotations__`
- Add Python 3.13 compatibility comment explaining the change

## Problem
In Python 3.13+, the built-in object class no longer has the __annotations__ attribute by default. The existing code in the TypeError exception handler tried to access object_.__annotations__[attr_name] directly, causing:
```
AttributeError: type object 'object' has no attribute '__annotations__'
```

## Solution
Use `getattr(object_, '__annotations__', {})` to safely access annotations with a fallback to an empty dict, preventing the AttributeError while maintaining backward compatibility with all Python versions.

## Testing
- Tested with Python 3.13.5 in vellum-copilot workflow execution
- Verified workflows execute successfully without __annotations__ errors
- Pre-commit hooks passed

---

### DIFF
--- a/src/vellum/workflows/types/utils.py
+++ b/src/vellum/workflows/types/utils.py
@@ -122,8 +122,12 @@ def infer_types(object_: Type, attr_name: str, localns: Optional[Dict[str, Any]]
 
         raise AttributeError(f"Failed to infer type from attribute {attr_name} on {object_.__name__}")
     except TypeError:
+        # Python 3.13+: object class doesn't have __annotations__ by default
+        # Use getattr with default to safely access annotations
+        annotations = getattr(object_, "__annotations__", {})
+        annotation_value = annotations.get(attr_name, "<unknown>")
         raise AttributeError(
-            f"Found 3.9+ typing syntax for field '{attr_name}' on class '{object_.__name__}' – {object_.__annotations__[attr_name]}. Type annotations must be compatible with python version 3.8. "  # noqa: E501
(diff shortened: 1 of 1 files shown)


vincent0426 (approved):

src/vellum/workflows/types/utils.py
```diff
@@ -122,8 +122,12 @@ def infer_types(object_: Type, attr_name: str, localns: Optional[Dict[str, Any]]
 
         raise AttributeError(f"Failed to infer type from attribute {attr_name} on {object_.__name__}")
     except TypeError:
+        # Python 3.13+: object class doesn't have __annotations__ by default
+        # Use getattr with default to safely access annotations
+        annotations = getattr(object_, "__annotations__", {})
+        annotation_value = annotations.get(attr_name, "<unknown>")
```
dvargasfuertes, line 128: [nit] would be nice to use either `None` or our `undefined` singleton by default in the future since `"<unknown>"` is technically a valid type

dvargas92495 (commented):
✅

PR #3055, opened 2025-11-11
Construct trigger definition with context

NgoHarrison:
[BLOCKING] Have to rebase based on previous comments in related PR


src/vellum/workflows/runner/runner.py
```diff
@@ -262,6 +271,20 @@ def __init__(
...
         self._execution_context = init_execution_context or get_execution_context()
+        self._trigger = trigger
+
+        if trigger:
+            cls = trigger.__class__
+            if issubclass(cls, IntegrationTrigger):
+                self._execution_context.parent_context = IntegrationTriggerContext(
+                    span_id=uuid4(),
+                    trigger_id=cls.__id__,
+                )
```
dvargasfuertes: [blocking] we should never need to update the parent context. In this case in particular, this context is already constructed in django before reaching here

tests/workflows/integration_trigger_execution/tests/test_integration_trigger_execution.py
```diff
@@ -171,3 +171,32 @@ def test_workflow_with_only_manual_trigger_accepts_manual_trigger_instance():
     # THEN it should execute successfully
     assert result.name == "workflow.execution.fulfilled"
     assert result.outputs.result == "No trigger workflow"
+
+
+def test_workflow_initiated_event_has_integration_trigger_parent_context():
```
dvargasfuertes: [nit] we should just need to verify that the initiated event has the trigger field

NgoHarrison: Updated

src/vellum/workflows/runner/runner.py
NgoHarrison: Removed

dvargas92495 (approved):

PR #3057, opened 2025-11-11
Codegen function file by finding name

vincent0426:
Screenshot 2025-11-11 at 2 49 27 PM

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -497,6 +497,48 @@ def get_current_weather(location: str, unit: str) -> str:
 "
 `;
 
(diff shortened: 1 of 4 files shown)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -487,6 +487,48 @@ exports[`WorkflowProjectGenerator > function > should generate <function_name>.p
 from ...inputs import Inputs
```
dvargasfuertes, line 487: [q] why does the screenshot have `import list`?

dvargas92495 (approved):

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
vincent0426: oh im using import list to test in screenshot

PR #3058, opened 2025-11-11
Serialize comment on workflow triggers

NgoHarrison:


ee/vellum_ee/workflows/display/workflows/base_workflow_display.py
```diff
@@ -690,6 +690,13 @@ def _serialize_workflow_trigger(self) -> Optional[JsonArray]:
             if hasattr(display_class, "color") and display_class.color is not None:
                 display_data["color"] = display_class.color
 
+            # Add comment if present
+            if hasattr(display_class, "comment") and display_class.comment is not None:
```
dvargasfuertes, line 694: [oos] we likely also want to support the class' docstring as the comment, but post GA

dvargas92495 (commented):
✅

PR #3065, opened 2025-11-12
Follow-up: Address comment from PR #3044

alex-nork:
## Overview
This PR addresses a minor feedback item (nit) from PR #3044.

### From PR #3044
- **File**: src/vellum/workflows/types/utils.py:129
  - **Change**: Replace `"<unknown>"` with `undefined` singleton as default value
  - **Rationale**: `"<unknown>"` is technically a valid type, so using the `undefined` singleton is cleaner and more semantically correct
  - **Comment**: https://github.com/vellum-ai/vellum-python-sdks/pull/3044#discussion_r2515880496
  - **Reviewer**: @dvargas92495

## Details

The reviewer suggested using `None` or the `undefined` singleton instead of the string `"<unknown>"` as a default value when annotations are not found. I chose to use `undefined` because:

1. It's semantically more appropriate for representing "no value found"
2. It's consistent with the codebase's use of the `undefined` singleton elsewhere
3. It avoids potential conflicts with actual type strings

## Testing
- All existing tests pass
- Type checking with mypy passes
- No functional changes made

## Related PRs
- #3044

---

### DIFF
--- a/src/vellum/workflows/types/utils.py
+++ b/src/vellum/workflows/types/utils.py
@@ -22,6 +22,7 @@ from typing import (
 )
 
 from vellum import ArrayVellumValue, ArrayVellumValueRequest, ChatMessagePromptBlock
+from vellum.workflows.constants import undefined
 from vellum.workflows.descriptors.base import BaseDescriptor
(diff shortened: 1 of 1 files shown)


dvargas92495 (commented):
✅

PR #3077, opened 2025-11-13
Skip metadata json for subworkflows

NgoHarrison:


NgoHarrison:
@codex review

ee/codegen/src/project.ts
```diff
@@ -1096,6 +1096,12 @@ ${errors.slice(0, 3).map((err) => {
      *            target_node_path  := "<module_path>.<ClassName>.Trigger" for the target node
      *   Value: "<ui_edge_id>"
      **/
+
+    // Skip metadata json for subworkflows for now
+    if (this.workflowContext.parentNode) {
+      return;
+    }
```
dvargasfuertes, line 1103: [oos] there may be an existing subworkflow test where we can assert that the file doesn't get generated as a way to validate the change

dvargas92495 (approved):

PR #3084, opened 2025-11-13
Support default factory inputs

vincent0426:


src/vellum/workflows/inputs/base.py
```diff
@@ -90,6 +91,21 @@ def __init__(self, **kwargs: Any) -> None:
             value = kwargs.get(name)
```
dvargasfuertes, line 91: [nit] I think we should update this to `value = kwargs.get(name, undefined)`. `None/null` is a valid default value, so this will allows us to update the checks below to `if value is undefined`

dvargas92495 (approved):

PR #3091, opened 2025-11-14
[APO-2113] Codegen dict ref definition

vincent0426:


vincent0426:
@codex review

ee/codegen/src/__test__/workflow-value-descriptor-reference/__snapshots__/dictionary-workflow-reference.test.ts.snap
```diff
@@ -15,6 +15,14 @@ exports[`DictionaryWorkflowReference > should generate correct AST for dictionar
 "
 `;
 
+exports[`DictionaryWorkflowReference > should generate correct AST for dictionary with definition 1`] = `
+"{
+    "chat_history": State.chat_history
+    + ChatMessage(text=Inputs.text, role="USER", content=None, source=None),
```
dvargasfuertes: [nit] can we filter out the `=None`? or would that require hardcoding?

If the latter, we can [oos] it

dvargas92495 (approved):

dvargas92495 (approved):

PR #3094, opened 2025-11-14
Create Vellum File Read Utilities

noanflaherty:
This is the first of a few PRs that help aim to improve the DevExp of interacting with vellum file values within workflows.

Specifically, it exposes two new utilities:
– A higher level `read_vellum_file`
- A lower level `stream_vellum_file`

These will later be used to implement the optimal API, which I view to be:

```python
vellum_document = VellumDocument(...)
contents = vellum_document.read()
content_iterator = vellum_document.stream()
```

Additional API's I want to support
```python
vellum_document = VellumDocument(...)
vellum_document.upload()  # Lazily creates an UploadedFile entity in Vellum and returns a new VellumFile instance with an updated `src` value
```

--------------

Here are some examples of Custom Nodes that we could support in Workflows if we had all the above.

```python
class OCRNode(BaseNode):
    """Example of a custom node that needs to read the contents of a Vellum Value that represents a file"""

    file: VellumDocument | VellumImage  # Note that `src` within could be base64, a public url, signed url, or `vellum:uploaded-files:<id>`

        text: str

    def run(self):
        contents = self.file.read()

        # Send the contents to an OCR api...
        text = ...
        return self.Outputs(text=text)
```

```python
class ImageGenerationNode(BaseNode):
    """Example of a custom node that needs to save a generated artifact so it can be viewed later"""

        image: VellumImage

    def run(self):
        prompt = "Generate an image of a hotdog"

        # Hit an image generation API that returns one of base64, public url, or signed url with expiry...
        result = ...
        
        image = VellumImage(src=result)

        # Note: If we didn't want to hit the Vellum backend and were cool with just passing around a base64/public/signed url
        #     then we'd just do this:
        # return self.Outputs(image=image)

        # If we want to reduce the size of the value that's being passed between nodes (i.e. base64's) and/or
        #     make it easy to view the generated artifact later (i.e. support viewing beyond original expiry)
        #     then we'd want to do this:
        uploaded_image = image.upload()
        return self.Outputs(image=uploaded_image)
```

Current limitations as of 11/19/2025
1. The Vellum UI renders suboptimal representations of these multi-modal variable types in the Workflow Sandbox Console and probably other UI surfaces
2. Workflows do not yet support these multi-modal types as workflow-level inputs, start variables, or outputs. Crucially, this means you cannot yet output say, a `VellumImage` as a Workflow Output (even though a Custom Node might produce one as its output)

### DIFF
--- a/src/vellum/utils/files/read.py
+++ b/src/vellum/utils/files/read.py
@@ -1,0 +1,39 @@
+"""Convenience utilities for reading Vellum files."""
+
+from typing import Optional
+
+from vellum.client import Vellum as VellumClient
+from vellum.utils.files.stream import stream_vellum_file
+from vellum.utils.files.types import VellumFileTypes
+
+
+def read_vellum_file(vellum_file: VellumFileTypes, client: Optional[VellumClient] = None) -> bytes:
+    """
+    Convenience function that reads the entire file content into memory.
+
+    This function is implemented using stream_vellum_file() under the hood.
+    For large files (e.g., videos, large datasets), consider using
+    stream_vellum_file() directly to avoid memory issues.
+
+    Args:
+        vellum_file: A VellumDocument, VellumImage, VellumAudio, or VellumVideo instance
+        client: An optional Vellum client instance. If not provided, a default client will be created.
+
+    Returns:
+        bytes: The complete file content
+
+    Example:
+        ```python
+        from vellum import VellumDocument
+        from vellum.utils.files import read_vellum_file
+
+        document = VellumDocument(src="data:text/plain;base64,SGVsbG8gV29ybGQ=")
+        content = read_vellum_file(document)
+        print(content.decode('utf-8'))  # "Hello World"
+        ```
+    """
+    chunks = []
+    with stream_vellum_file(vellum_file, client=client) as chunk_iter:
+        for chunk in chunk_iter:
+            chunks.append(chunk)
+    return b"".join(chunks)
--- a/src/vellum/utils/files/stream.py
+++ b/src/vellum/utils/files/stream.py
@@ -88,9 +88,10 @@ def stream_vellum_file(
                 raise FileNotFoundError(f"Uploaded file not found: {vellum_uploaded_file_id}") from e
             raise FileRetrievalError(f"Failed to retrieve uploaded file: {vellum_uploaded_file_id}") from e
 
-        file_url = uploaded_file.file_url
-        if not file_url:
+        if not uploaded_file.file_url:
             raise FileRetrievalError(f"Uploaded file has no accessible URL: {vellum_uploaded_file_id}")
+
+        file_url = uploaded_file.file_url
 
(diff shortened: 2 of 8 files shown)


src/vellum/utils/files/read.py
```diff
@@ -0,0 +1,39 @@
...
+        bytes: The complete file content
+
+    Example:
+        ```python
+        from vellum import VellumDocument
+        from vellum.utils.files import read_vellum_file
+
+        document = VellumDocument(src="data:text/plain;base64,SGVsbG8gV29ybGQ=")
+        content = read_vellum_file(document)
+        print(content.decode('utf-8'))  # "Hello World"
```
dvargasfuertes, line 32: [oos] Note for @vincent0426 - we want this to be implemented as part of of a new `VellumFile(UniversalBaseModel)` class, then we postprocess the fern generated types to extend the new class

dvargas92495 (approved):

src/vellum/utils/files/stream.py
```diff
@@ -0,0 +1,130 @@
...
+    if match:
+        vellum_uploaded_file_id = match.group(1)
+        vellum_client = client or create_vellum_client()
+
+        try:
+            uploaded_file = vellum_client.uploaded_files.retrieve(vellum_uploaded_file_id)
+        except ApiError as e:
+            if e.status_code == 404:
+                raise FileNotFoundError(f"Uploaded file not found: {vellum_uploaded_file_id}") from e
+            raise FileRetrievalError(f"Failed to retrieve uploaded file: {vellum_uploaded_file_id}") from e
```
noanflaherty, line 89: @siddseethepalli here was one intended API call to the Vellum backend from within Workflows.

Being able to read Vellum Uploaded Files is probably the highest leverage API and hardest to OOS piece, since today you can upload a file to a scenario in the Vellum UI and then it's represented in the workflow as `VellumDocument(src="vellum:uploaded-files:<id>")`.

One goal is is to make it easier for the `run` method of a Custom Node to operate on the file's contents (but at least in this scenario-based case, it only has "vellum:uploaded-files:<id>").

As the PR description notes, I was intending on also supporting a `write` API, which would require an API in Vellum's backend to create/update files. This could be more easily punted if we're cool with continuing to pass around base64's and signed urls with expiries.

noanflaherty:
@dvargas92495 @vincent0426 @awlevin  just h/u I'm going to pump the brakes on this one until we mitigate some good infra concerns.

siddseethepalli (approved):

noanflaherty:
Chatted with @siddseethepalli offline and we're good to proceed with this. We'll block GA on:
- Moving Vellum uploaded files API to `predict` service
- Cutting a new sdk release so that the client is hitting `predict.vellum.ai`

We will defer and not block on:
- Moving the internal store for uploaded files from postgres to something else
- Migrating predict from cloudrun to k8s (to handle larger file upload sizes)

PR #3107, opened 2025-11-17
Add exclude from console to node definition

vincent0426:
excluding router node and else node in console drawer

### DIFF
--- a/src/vellum/workflows/events/tests/test_event.py
+++ b/src/vellum/workflows/events/tests/test_event.py
@@ -30,6 +30,7 @@ from vellum.workflows.events.workflow import (
 from vellum.workflows.exceptions import WorkflowInitializationException
 from vellum.workflows.inputs.base import BaseInputs
 from vellum.workflows.nodes.bases.base import BaseNode
+from vellum.workflows.nodes.displayable.tool_calling_node.utils import ElseNode, RouterNode
 from vellum.workflows.outputs.base import BaseOutput
 from vellum.workflows.references.vellum_secret import VellumSecretReference
 from vellum.workflows.state.base import BaseState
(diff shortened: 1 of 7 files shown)


src/vellum/workflows/events/tests/test_event.py
```diff
@@ -532,3 +533,37 @@ def test_workflow_event_generator_stream_initialization_exception():
...
+        span_id=UUID("123e4567-e89b-12d3-a456-426614174001"),
+        body=NodeExecutionInitiatedBody(
+            node_definition=RouterNode,
+            inputs={},
+        ),
+    )
+
+    serialized = router_node_event.model_dump(mode="json")
+    assert "exclude_from_console" in serialized["body"]["node_definition"]
+    assert serialized["body"]["node_definition"]["exclude_from_console"] is True
```
dvargasfuertes: [nit] let's not include `console` in the name here, since this flag will be used across our monitoring views

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -159,7 +159,7 @@ def run(self) -> Iterator[BaseOutput]:
 class RouterNode(BaseNode[ToolCallingState]):
     """Router node that handles routing to function nodes based on outputs."""
 
-    pass
+    __vellum_exclude_from_console__: bool = True  # Exclude from console drawer display
```
dvargasfuertes: [nit] I think we could remove the `vellum_` prefix

src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -320,6 +320,8 @@ def run(self) -> Iterator[BaseOutput]:
 class ElseNode(BaseNode[ToolCallingState]):
     """Node that executes when no function conditions match."""
 
+    __vellum_exclude_from_console__: bool = True  # Exclude from console drawer display
```
dvargasfuertes: [nit] let's add this flag with a default of `False` on BaseNode

dvargasfuertes: [oos] Starting to wonder if this should actually go in the nested `class Display` since the goal is for this to affect display...

dvargas92495 (approved):

dvargas92495 (approved):

PR #3118, opened 2025-11-18
Introduce Util for Uploading File to Vellum

noanflaherty:
This introduces the last utility that we'll need to start exposing nice APIs on our generated types.

Note: I changed my mind and decided to name this `upload_vellum_file` and the api we make on generated types should be `upload` (instead of `write_vellum_file` and `write`)


dvargas92495 (commented):
✅

PR #3119, opened 2025-11-18
Improve Call Signature of File Upload Utils

noanflaherty:
A quick improvement to the call signature of our file upload helpers. Namely:
1. We use the more explict/consistent `vellum_client` kwarg instead of `vellum`.
2. We enforce the use of named args


dvargas92495 (commented):
✅

PR #3122, opened 2025-11-18
File api extension

vincent0426:


src/vellum/client/types/vellum_file_extensions.py
```diff
@@ -0,0 +1,81 @@
...
+# This file is not generated by Fern and can be freely edited
+
+from typing import Any, ContextManager, Iterator, Optional
+
+from vellum.utils.files.read import read_vellum_file
+from vellum.utils.files.stream import stream_vellum_file
+from vellum.utils.files.upload import upload_vellum_file
+
+
+class VellumFileMixin:
```
dvargasfuertes: [nit] If we extend `UniversalBaseModel`, we could flat out replace the base class's for VellumDocument, etc.

.fernignore
```diff
@@ -13,6 +13,7 @@ src/vellum/prompts
 src/vellum/utils
 src/vellum/workflows
 src/vellum/client/utils.py
+src/vellum/client/types/vellum_file_extensions.py
```
dvargasfuertes: [blocking] this has risk of colliding with one of our generated type. Let's keep the new class in the `vellum.utils.files` module

noanflaherty (commented):
✅

PR #3125, opened 2025-11-18
Add trigger attribute workflow reference type

NgoHarrison:


dvargas92495 (commented):
✅

dvargas92495:
[oos] we may also be missing support in the legacy node input value pointer flow

PR #3131, opened 2025-11-18
Break import cycle for file mixin

vincent0426:
while testing locally with type change

### DIFF
--- a/src/vellum/utils/files/stream.py
+++ b/src/vellum/utils/files/stream.py
@@ -4,16 +4,17 @@ import base64
 from contextlib import contextmanager
 from io import BytesIO
 import re
-from typing import Generator, Iterator, Optional
+from typing import TYPE_CHECKING, Generator, Iterator, Optional
 
 import requests
 
-from vellum.client import Vellum as VellumClient
+if TYPE_CHECKING:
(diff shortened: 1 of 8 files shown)


src/vellum/utils/files/stream.py
```diff
@@ -4,24 +4,25 @@
...
 
-from vellum.client import Vellum as VellumClient
+if TYPE_CHECKING:
+    from vellum.client import Vellum as VellumClient
+
 from vellum.client.core.api_error import ApiError
 from vellum.utils.files.constants import BASE64_DATA_URL_PATTERN, URL_PATTERN, VELLUM_FILE_SRC_PATTERN
 from vellum.utils.files.exceptions import FileNotFoundError, FileRetrievalError, InvalidFileSourceError
 from vellum.utils.files.types import VellumFileTypes
-from vellum.workflows.vellum_client import create_vellum_client
```
dvargasfuertes, line 16: [blocking] can we move this method to `vellum.utils` and reexport from `vellum.workflows.vellum_client`? then we can work to migrate all consumers to eventually import from `vellum.utils`, starting with the ones in this pr that cause the cycle.

This should allow us to avoid the nested imports

PR #3134, opened 2025-11-18
Break mixin cycle

vincent0426:
vellum.client.types.vellum_document (client type) imports VellumFileMixin from vellum.utils.files.mixin
vellum.utils.files.mixin imports read_vellum_file from vellum.utils.files.read
vellum.utils.files.read imports stream_vellum_file from vellum.utils.files.stream
vellum.utils.files.stream imports create_vellum_client from vellum.utils.vellum_client
vellum.utils.vellum_client imports Vellum from vellum.client
vellum.client imports vellum.client.types.vellum_document (back to step 1)
This creates a cycle: vellum.client → vellum.client.types.vellum_document → vellum.utils.files.* → vellum.client

### DIFF
--- a/src/vellum/utils/files/types.py
+++ b/src/vellum/utils/files/types.py
@@ -1,7 +1,8 @@
 """Type definitions for Vellum files."""
 
-from typing import Union
+from typing import TYPE_CHECKING, Union
 
-from vellum import VellumAudio, VellumDocument, VellumImage, VellumVideo
+if TYPE_CHECKING:
+    from vellum.client import VellumAudio, VellumDocument, VellumImage, VellumVideo
 
-VellumFileTypes = Union[VellumDocument, VellumImage, VellumVideo, VellumAudio]
+VellumFileTypes = Union["VellumDocument", "VellumImage", "VellumVideo", "VellumAudio"]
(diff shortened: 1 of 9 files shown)


src/vellum/utils/files/upload.py
```diff
@@ -4,17 +4,18 @@
...
 import logging
 import re
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 import requests
 
-from vellum.client import Vellum as VellumClient
 from vellum.client.core.api_error import ApiError
 from vellum.client.core.file import File
```
dvargasfuertes, line 12: [oos] The dangerous part of the circ is `vellum.utils` trying to import from `vellum.client`, which we should probably lint against in some future

src/vellum/utils/files/types.py
```diff
@@ -1,7 +1,8 @@
...
 
-from typing import Union
+from typing import TYPE_CHECKING, Union
 
-from vellum import VellumAudio, VellumDocument, VellumImage, VellumVideo
+if TYPE_CHECKING:
+    from vellum import VellumAudio, VellumDocument, VellumImage, VellumVideo
 
-VellumFileTypes = Union[VellumDocument, VellumImage, VellumVideo, VellumAudio]
+VellumFileTypes = Union["VellumDocument", "VellumImage", "VellumVideo", "VellumAudio"]
```
dvargasfuertes, line 8: [nit] Now that we have the base type, I could see us not needing this union

dvargas92495 (commented):
✅

PR #3135, opened 2025-11-19
[APO-2194] Raise error on json serialization errors for outputs

NgoHarrison:


src/vellum/workflows/runner/runner.py
```diff
@@ -565,6 +572,16 @@ def initiate_node_streaming_output(
...
+                        try:
+                            default_serializer(output)
+                        except (TypeError, ValueError) as exc:
+                            raise NodeException(
+                                message=(
+                                    f"Node {node.__class__.__name__} produced output: "
+                                    f"'{output.name}' that could not be serialized to JSON: {exc}"
+                                ),
+                                code=WorkflowErrorCode.INVALID_OUTPUTS,
+                            ) from exc
```
dvargasfuertes, line 602: [blocking] I would rather support than add this overhead on every chunk from a node. Should be as easy as updating `VellumJsonEncoder` I believe

dvargas92495 (approved):

PR #3136, opened 2025-11-19
None default vellum variable should generate None as default

vincent0426:
Screenshot 2025-11-18 at 5 53 12 PM

### DIFF
--- a/ee/codegen/src/generators/vellum-variable.ts
+++ b/ee/codegen/src/generators/vellum-variable.ts
@@ -31,12 +31,14 @@ export declare namespace VellumVariable {
 
 export class VellumVariable extends AstNode {
   private readonly field: python.Field;
+  private readonly defaultRequired?: boolean;
 
   constructor({ variable, defaultRequired }: VellumVariable.Args) {
     super();
 
     const baseType = getVellumVariablePrimitiveType(variable.type);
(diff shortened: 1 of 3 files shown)


vincent0426:
@codex review

NgoHarrison (approved):
Thanks! 🙌

ee/codegen/src/generators/vellum-variable.ts
```diff
@@ -31,12 +31,14 @@ export declare namespace VellumVariable {
...
 export class VellumVariable extends AstNode {
   private readonly field: python.Field;
+  private readonly defaultRequired?: boolean;
 
   constructor({ variable, defaultRequired }: VellumVariable.Args) {
     super();
 
     const baseType = getVellumVariablePrimitiveType(variable.type);
     const name = variable.name ?? variable.key;
+    this.defaultRequired = defaultRequired;
```
dvargasfuertes, line 41: [oos] I'm not sure we want to maintain this in the future, not worth removing now

dvargas92495 (approved):

ee/codegen/src/generators/vellum-variable.ts
vincent0426: yeah i didn't add that at first and it works but GPT point it out

PR #3138, opened 2025-11-19
Handle chat hisotry default factory

vincent0426:


ee/codegen/src/generators/vellum-variable.ts
```diff
@@ -78,13 +78,25 @@ export class VellumVariable extends AstNode {
         : undefined;
     }
 
-    // Check if the default type is ARRAY or JSON
-    // Use Field(default_factory=...) for ARRAY and JSON types to avoid mutable default issues
+    // Check if the default type is ARRAY, JSON, or CHAT_HISTORY
+    // Use Field(default_factory=...) for list types to avoid mutable default issues
     const isArrayType = variable.default.type === "ARRAY";
     const isJsonType = variable.default.type === "JSON";
+    const isChatHistoryType = variable.default.type === "CHAT_HISTORY";
```
dvargasfuertes, line 85: [oos] I think we'll have a similar story with SEARCH_RESULTS, but that could be deferred

dvargas92495 (approved):

PR #3140, opened 2025-11-19
Infer File Extensions as Part of File Upload

noanflaherty:
The Vellum file upload API requires a file name with a valid extension. To supply this, we'll infer the extension as needed from the mimetype of the content.

In current state, the API returns a 400 when the `upload` api is used.


vincent0426 (approved):

dvargas92495 (commented):
✅

PR #3149, opened 2025-11-19
Expose New `get_signed_url` API on `VellumFileMixin`

noanflaherty:
It's common to want to pass a url to an external service rather than needing to read a byte stream, turn it into a base64 encoded string, and _then_ send it to the external service.

This PR introduces a new high-level API for easily fetching that url. This is another one that we can teach AB how to use within custom nodes.

### DIFF
--- a/src/vellum/utils/files/urls.py
+++ b/src/vellum/utils/files/urls.py
@@ -4,7 +4,9 @@ import logging
 import re
 from typing import TYPE_CHECKING, Optional
 
+from vellum.client.core.api_error import ApiError
 from vellum.utils.files.constants import VELLUM_FILE_SRC_PATTERN
+from vellum.utils.files.exceptions import FileRetrievalError, InvalidFileSourceError
 from vellum.utils.files.types import VellumFileTypes
 from vellum.utils.files.upload import upload_vellum_file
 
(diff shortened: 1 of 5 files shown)


vincent0426 (approved):

src/vellum/utils/files/urls.py
```diff
@@ -0,0 +1,61 @@
...
+    Args:
+        vellum_file: A VellumDocument, VellumImage, VellumAudio, or VellumVideo instance
+        vellum_client: An optional Vellum client instance. If not provided, a default client will be created.
+
+    Returns:
+        str: A signed URL for accessing the uploaded file
+    """
+
+    # Upload the file to Vellum if it isn't already
+    uploaded_file = upload_vellum_file(vellum_file, vellum_client=vellum_client)
```
dvargasfuertes, line 40: [q] does the api not return `file_url`? or do we only generate it on the `retrieve` endpoint?

src/vellum/utils/files/urls.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # Parse out its reference ID
+    vellum_uploaded_file_id: Optional[str] = None
+    src = uploaded_file.src
+    match = re.match(VELLUM_FILE_SRC_PATTERN, src, re.IGNORECASE)
+    if match:
+        vellum_uploaded_file_id = match.group(1)
+
+    if not vellum_uploaded_file_id:
+        raise ValueError("Failed to determine id of uploaded file.")
```
dvargasfuertes: [nit] am partial to custom exceptions over `ValueError`s

dvargas92495 (approved):

src/vellum/utils/files/urls.py
noanflaherty: It does, but this helper only lazily uploads the file and in the event that’s only been uploaded, we don’t have the file object yet,

PR #3161, opened 2025-11-20
Add Support for `expiry_seconds` to `get_signed_url` Helpers

noanflaherty:
This makes use of the newly available `expiry_seconds` parameter from the Vellum API when retrieving the signed url for a VellumFile.

### DIFF
--- a/src/vellum/utils/files/mixin.py
+++ b/src/vellum/utils/files/mixin.py
@@ -1,3 +1,4 @@
+import typing
 from typing import Any, ContextManager, Iterator, Optional
 
 from vellum.client.core.pydantic_utilities import UniversalBaseModel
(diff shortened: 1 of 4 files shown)


src/vellum/utils/files/mixin.py
```diff
@@ -89,6 +90,7 @@ def upload(
     def get_signed_url(
         self: Any,
         *,
+        expiry_seconds: typing.Optional[int] = 7 * 24 * 60 * 60,  # 7 days
```
vincent0426, line 93: Do we allow expiry_seconds to be None here if user pass "None" explicitly (indicated by Optional)? or maybe could remove optional which will default to 7 days)

noanflaherty: Ah good point. I merged before seeing this but will address rq before cutting a release

dvargas92495 (commented):
✅

PR #3185, opened 2025-11-22
Parallel test

vincent0426:
15-20s faster


Makefile
```diff
@@ -35,10 +35,10 @@ setup-fern:
 
 file ?= .
 test:
-	poetry run pytest -rEf -s -vv $(file)
+	poetry run pytest -rEf -s -vv -n 4 $(file)
```
dvargasfuertes, line 38: [nit] can we add a version without parallelization (like `make test-serial` from django)?

workers are slower when you are running just 1 test, which is useful for debugging the feature branch

dvargas92495 (approved):

PR #3189, opened 2025-11-22
Use zero diffs workflow to catch list[str] bug

dvargas92495:
Was able to repro the bug preventing `list[str]` from working end to end and solved in PR using the zero diff test

### DIFF
--- a/tests/workflows/builtin_list_str/nodes/the_end.py
+++ b/tests/workflows/builtin_list_str/nodes/the_end.py
@@ -3,4 +3,7 @@ from vellum.workflows import BaseState, FinalOutputNode
 
 class TheEnd(FinalOutputNode[BaseState, list[str]]):
     class Outputs(FinalOutputNode.Outputs):
-        value = ["hello", "world"]
+        value = [
+            "hello",
+            "world",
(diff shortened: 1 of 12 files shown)


tests/workflows/builtin_list_str/nodes/the_end.py
```diff
@@ -3,4 +3,7 @@
 
 class TheEnd(FinalOutputNode[BaseState, list[str]]):
     class Outputs(FinalOutputNode.Outputs):
-        value = ["hello", "world"]
+        value = [
+            "hello",
+            "world",
+        ]
```
dvargasfuertes, line 9: catching formatting diffs will be far into the future

tests/workflows/trivial/nodes/start.py
```diff
@@ -0,0 +1,5 @@
+from vellum.workflows import BaseNode
+
+
+class StartNode(BaseNode):
+    pass
```
dvargasfuertes, line 5: this highlighted a bug! we're not generating `pass` in these cases

vincent0426 (commented):
✅

PR #3192, opened 2025-11-22
Fix function node display class

vincent0426:


src/vellum/workflows/nodes/displayable/tool_calling_node/utils.py
```diff
@@ -567,12 +567,18 @@ def wrapper(self, **kwargs):
             wrapper.__name__ = func.__name__
             return wrapper
 
+        display_class = type(
+            f"FunctionNodeDisplay_{function.__name__}",
+            (BaseNode.Display,),
```
dvargasfuertes: [nit] more consistent to extend from `FunctionNode.Display` here

dvargas92495 (approved):

PR #3225, opened 2025-11-24
File upload example

vincent0426:
Screenshot 2025-11-24 at 11 12 41 AM


examples/workflows/file_uploads/nodes/upload_file_node.py
```diff
@@ -0,0 +1,58 @@
+from typing import List
+
+from vellum import VellumImage
+from vellum.utils.files import upload_vellum_file
```
dvargasfuertes: [blocking] we want to be able to use the fields that are now accessible directly from the `VellumImage` class instead of importing the helper directly

examples/workflows/file_uploads/sandbox.py
```diff
@@ -0,0 +1,38 @@
...
+four_pillars_image_base64_src = f"data:image/png;base64,{four_pillars_image_base64}"
+
+runner = WorkflowSandboxRunner(
+    workflow=Workflow(),
+    dataset=[
+        Inputs(
+            images=[
+                VellumImage(src=receipt_image_url),
+                VellumImage(src=four_pillars_image_base64_src),
+            ],
```
dvargasfuertes, line 49: [nit] let's add a third example here of an Image that is already using a vellum uploaded uri, and show that we have access to the contents from within the workflow

dvargas92495 (approved):

PR #3240, opened 2025-11-25
[APO-2233] Remove use tool inputs from src

vincent0426:
after push, code should not have `@use_tool_inputs`

Screenshot 2025-11-25 at 10 42 10 AM

### DIFF
--- a/ee/vellum_ee/workflows/display/utils/expressions.py
+++ b/ee/vellum_ee/workflows/display/utils/expressions.py
@@ -1,3 +1,4 @@
+import ast
 from dataclasses import asdict, is_dataclass
 import inspect
 from uuid import UUID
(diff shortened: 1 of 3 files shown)


ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -504,6 +564,8 @@ def serialize_value(executable_id: UUID, display_context: "WorkflowDisplayContex
         if source_path is not None:
             with virtual_open(source_path) as f:
                 source_code = f.read()
+            # Remove @use_tool_inputs decorator since it will be regenerated during codegen
+            source_code = _remove_use_tool_inputs_decorator(source_code, name)
```
dvargasfuertes, line 568: This seems spooky to do on serialization side, since we are in effect changing what the user generated. can we do this check on the codegen side?

dvargasfuertes: I also feel that serialization should stand on its own merits, and not make decisions bc what codegen does or doesn't do

vincent0426: hmm I think function code will have @use tool inputs if user pull with inputs defined. However, we don’t want to show @use tool inputs as code explicitly in function code editor as it’s displayed by inputs sidebar. During serialization, if we don’t filter out, tool editor will contain @ use tool inputs and it will duplicate again in codegen

dvargasfuertes: >  it will duplicate again in codegen

is there a way to prevent this?

vincent0426: but frontend uses `tool.src` to show the code too so if user doesn't define `@use_tool_inputs` in code. after pull -> push, if we don't filter out, the code will add `@use_tool_inputs` while still having inputs sidebar on the left, wouldn't this be more confusing?

dvargasfuertes: We should likely let AB drive this decision. I would raise it in the `proj-agent-builder` channel to see what the agent would prefer.

More and more, we should treat AB as the highest leverage SDK user

PR #3304, opened 2025-12-05
Add datetime to str vellum variable conversion

NgoHarrison:
This is in the context of ScheduleTriggers. They have two fields on the base config of datetime. Therefore, need to serialize this as type STRING


NgoHarrison:
@codex review

vincent0426 (approved):

PR #3308, opened 2025-12-05
Update workflow sandbox mock then outputs generation

NgoHarrison:


NgoHarrison:
@codex review

ee/codegen/src/types/vellum.ts
```diff
@@ -898,10 +898,15 @@ type WorkflowSandboxInput =
...
+export interface WorkflowSandboxDatasetRowMockOutput {
+  output_id: string;
+  value: WorkflowValueDescriptor;
+}
+
 export interface WorkflowSandboxDatasetRowMock {
   node_id: string;
   when_condition?: WorkflowValueDescriptor;
-  then_outputs: Record<string, unknown>;
+  then_outputs: Record<string, unknown> | WorkflowSandboxDatasetRowMockOutput[];
```
dvargasfuertes, line 909: [blocking] this should never be the case from django... would love to see where this is coming from

ee/codegen/src/__test__/__snapshots__/workflow-sandbox.test.ts.snap
```diff
@@ -1,5 +1,42 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`Workflow Sandbox > write > should generate DatasetRow with mocks using array format for then_outputs 1`] = `
```
dvargasfuertes, line 3: Even if this was a bug that we wanted to handle, the `sandbox.py` file shouldn't be used in the executions path at all and would be surprised if it was blocking the customer

PR #3334, opened 2025-12-09
Add chatbot_set_state example

vincent0426:
https://github.com/user-attachments/assets/92c0f0a2-a155-4d96-b0fc-f08c4568bc7c

### DIFF
--- a/examples/workflows/chatbot_set_state/nodes/__init__.py
+++ b/examples/workflows/chatbot_set_state/nodes/__init__.py
@@ -1,0 +1,11 @@
+from .agent import Agent
+from .append_assistant_message import AppendAssistantMessage
+from .append_user_message import AppendUserMessage
+from .final_output import FinalOutput
+
+__all__ = [
+    "Agent",
+    "AppendAssistantMessage",
+    "AppendUserMessage",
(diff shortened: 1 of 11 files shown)


examples/workflows/chatbot_set_state/nodes/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .agent import Agent
```
dvargasfuertes, line 1: [blocking] could we replace the existing `chatbot` example with this one? I would like for us to have 1 authoritative example of how to build chatbots on Vellum

dvargas92495 (approved):

examples/workflows/chatbot_set_state/nodes/__init__.py
vincent0426: I’m thinking to remove the previous one after this one get merged since that one is pretty hard to setup

vincent0426: and rename this one to chatbot

dvargasfuertes: on board

PR #3355, opened 2025-12-11
[APO-2325] Do not use inputs class when trigger based workflow is ran

NgoHarrison:
Context: This came up from trigger based workflows having scenarios with input variables marked as required

### DIFF
--- a/src/vellum/workflows/runner/runner.py
+++ b/src/vellum/workflows/runner/runner.py
@@ -214,7 +214,6 @@ class WorkflowRunner(Generic[StateType]):
             if state:
                 self._initial_state = deepcopy(state)
                 self._initial_state.meta.workflow_inputs = default_inputs
-                self._initial_state.meta.workflow_inputs_skipped = default_inputs is None
                 self._initial_state.meta.span_id = execution_id or uuid4()
                 self._initial_state.meta.workflow_definition = self.workflow.__class__
             else:
--- a/src/vellum/workflows/runner/runner.py
+++ b/src/vellum/workflows/runner/runner.py
@@ -220,7 +220,7 @@ class WorkflowRunner(Generic[StateType]):
                 self._initial_state = self.workflow.get_default_state(
                     default_inputs,
                     execution_id,
-                    fallback_to_default_inputs=False,
+                    trigger_attributes={},
                 )
                 self._should_emit_initial_state = False
 
--- a/src/vellum/workflows/runner/runner.py
+++ b/src/vellum/workflows/runner/runner.py
@@ -214,6 +214,7 @@ class WorkflowRunner(Generic[StateType]):
             if state:
                 self._initial_state = deepcopy(state)
                 self._initial_state.meta.workflow_inputs = default_inputs
+                self._initial_state.meta.workflow_inputs_skipped = default_inputs is None
                 self._initial_state.meta.span_id = execution_id or uuid4()
                 self._initial_state.meta.workflow_definition = self.workflow.__class__
             else:
--- a/src/vellum/workflows/runner/runner.py
+++ b/src/vellum/workflows/runner/runner.py
@@ -64,6 +64,7 @@ from vellum.workflows.events.workflow import (
     WorkflowExecutionStreamingBody,
 )
 from vellum.workflows.exceptions import NodeException, WorkflowInitializationException
+from vellum.workflows.inputs.base import BaseInputs
 from vellum.workflows.nodes.bases import BaseNode
 from vellum.workflows.nodes.bases.base import NodeRunResponse
 from vellum.workflows.nodes.mocks import MockNodeExecutionArg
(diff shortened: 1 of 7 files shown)


NgoHarrison:
@codex review

NgoHarrison:
@codex review

NgoHarrison:
@codex review

NgoHarrison:
@codex review

NgoHarrison:
@codex review

src/vellum/workflows/runner/runner.py
```diff
@@ -209,14 +210,19 @@ def __init__(
...
+                self._initial_state.meta.workflow_inputs_skipped = default_inputs is None
                 self._initial_state.meta.span_id = execution_id or uuid4()
                 self._initial_state.meta.workflow_definition = self.workflow.__class__
             else:
-                self._initial_state = self.workflow.get_default_state(normalized_inputs, execution_id)
+                self._initial_state = self.workflow.get_default_state(
+                    default_inputs,
+                    execution_id,
+                    fallback_to_default_inputs=False,
+                )
```
NgoHarrison: Would love feedback here - This is the crux of the reported issue I think. `get_default_state()` was calling `self.get_default_inputs()` which is going to have pydantic validate the Inputs class that we always generate. When `Inputs.some_attr` was marked required, it was missing a value since we don't pass in actual values when it's a trigger based workflow. I am thinking we do this or we introduce something else in PWS to use here

src/vellum/workflows/runner/runner.py
```diff
@@ -209,14 +210,19 @@ def __init__(
             self._entrypoints = self.workflow.get_entrypoints()
         elif trigger:
             # When trigger is provided, set up default state and filter entrypoints by trigger type
-            normalized_inputs = deepcopy(inputs) if inputs else self.workflow.get_default_inputs()
+            default_inputs = deepcopy(inputs) if inputs else None
             if state:
                 self._initial_state = deepcopy(state)
-                self._initial_state.meta.workflow_inputs = normalized_inputs
+                self._initial_state.meta.workflow_inputs = default_inputs
+                self._initial_state.meta.workflow_inputs_skipped = default_inputs is None
```
dvargasfuertes: [blocking] why do we need this flag? seems inferred from `.meta.workflow_inputs is None`

Adding to state we should be careful about due to it being emitted to clickhouse

src/vellum/workflows/runner/runner.py
```diff
@@ -938,7 +944,7 @@ def _initiate_workflow_event(self) -> WorkflowExecutionInitiatedEvent:
             span_id=self._initial_state.meta.span_id,
             body=WorkflowExecutionInitiatedBody(
                 workflow_definition=self.workflow.__class__,
-                inputs=self._initial_state.meta.workflow_inputs,
+                inputs=self._initial_state.meta.workflow_inputs or BaseInputs(),
                 initial_state=deepcopy(self._initial_state) if self._should_emit_initial_state else None,
                 trigger=self._trigger.__class__ if self._trigger else None,
```
dvargasfuertes, line 948: [oos] damn this is interesting. Ideally we have the trigger's attributes appear as `inputs`. I think what that means is that we eventually remove `BaseInputs` in the type definition of `workflow_initiated_event.body.inputs`

dvargasfuertes: In the same breath, we likely have the full trigger instance on the state object eventually

src/vellum/workflows/runner/runner.py
NgoHarrison: I was trying to differentiate the case between optional inputs and trigger based workflows. For example, if I just default to None, then `test_base_workflow__deserialize_state_with_optional_inputs` these situations wouldf fail since it's hard to know. Alternatively, this could be solved by your other comment of just having the trigger instance in the meta. That could essentially serve the dual purpose of "flag" and "we already have inputs object and trigger attributes in meta so just include this instance as well"

src/vellum/workflows/runner/runner.py
```diff
@@ -209,14 +210,18 @@ def __init__(
...
-                self._initial_state.meta.workflow_inputs = normalized_inputs
+                self._initial_state.meta.workflow_inputs = default_inputs
                 self._initial_state.meta.span_id = execution_id or uuid4()
                 self._initial_state.meta.workflow_definition = self.workflow.__class__
             else:
-                self._initial_state = self.workflow.get_default_state(normalized_inputs, execution_id)
+                self._initial_state = self.workflow.get_default_state(
+                    default_inputs,
+                    execution_id,
+                    fallback_to_default_inputs=False,
```
dvargasfuertes: [blocking] again, instead of the flag, we should pass in the default trigger attributes consistent with the default inputs. I believe this will make it easier to cutover fully to the latter when ready

dvargas92495 (approved):
last blocker is quick i believe

PR #3392, opened 2025-12-16
Expose `vellum_client` for inline code exec

clopen-set:
Make `vellum_client`  accessible when we're
1. using the built-in code execution node
2. using Python 3.11 with no custom packages or secrets

I'm introducing this as a global so users don't have to modify their function signatures and so we don't have to modify the inputs we pass in


NgoHarrison (approved):

dvargas92495 (approved):
dope

dvargas92495:
test failure was a flake and should be unrelated. we should in general resolve it sooner or later

clopen-set:
@dvargas92495 possible fix for the flaky test: https://github.com/vellum-ai/vellum-python-sdks/pull/3393

PR #3393, opened 2025-12-16
Fix flaky streaming node pipeline test

clopen-set:
I assume we want to see the nodes' outputs interleaved in `final_events.outputs` 
https://github.com/vellum-ai/vellum-python-sdks/blob/ff6e8f1ec3b3391f0ef31b9dac9641c00ae4fafe/tests/workflows/streaming_node_pipeline/tests/test_workflow.py#L17-L24
to show that we stream one node's output to another, so we don't want to just sort the output for our asserts

Let's use an event to coordinate work between the nodes instead of sleeps. Assumes nodes run in separate threads: https://github.com/vellum-ai/vellum-python-sdks/blob/ff6e8f1ec3b3391f0ef31b9dac9641c00ae4fafe/src/vellum/workflows/runner/runner.py#L793

### DIFF
--- a/tests/workflows/streaming_node_pipeline/workflow.py
+++ b/tests/workflows/streaming_node_pipeline/workflow.py
@@ -1,4 +1,4 @@
-import time
+import threading
 from typing import Iterator, List, Set
 
 from vellum.workflows.inputs import BaseInputs
@@ -8,6 +8,8 @@ from vellum.workflows.outputs.base import BaseOutputs
 from vellum.workflows.ports.port import Port
 from vellum.workflows.state import BaseState
 from vellum.workflows.workflows import BaseWorkflow
+
+consumption_event = threading.Event()
 
 
 class Inputs(BaseInputs):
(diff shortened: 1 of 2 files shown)


tests/workflows/streaming_node_pipeline/workflow.py
```diff
@@ -39,9 +41,14 @@ def __lt__(self, output: BaseOutput) -> Set[Port]:
...
     def run(self) -> Iterator[BaseOutput]:
+        consumption_event.set()
+
         reversed_fruits = []
-        for index, fruit in enumerate(self.fruits):
-            time.sleep(0.01 * (index + 1))
+        for fruit in self.fruits:
+            # Block until SecondNode to consumes the output
+            consumption_event.wait()
+            consumption_event.clear()
```
dvargasfuertes, line 50: [q] This would still fail if our sdk didn't support streaming outputs from one node to the next bc it would hang right?

dvargas92495 (approved):

tests/workflows/streaming_node_pipeline/workflow.py
clopen-set: Yeah this should hang if we try to synchronously resolve `FirstNode`'s outputs. I'll add a timeout so we fail quickly if there's a deadlock

PR #3423, opened 2025-12-18
Raise error when constructing outgoing edge from final output node

clopen-set:
We added logic to guard against this a while back:
https://github.com/vellum-ai/vellum-python-sdks/blob/d38341c35b9726f1213c2de4bb82680271aac75d/src/vellum/workflows/graph/graph.py#L187-L192

We often didn't meet this condition because we often dropped `FInalOutputNodes` from `Graph._terminals` altogether during `Graph` construction. We instead need to preserve them as `NoPortsNode`

### DIFF
--- a/src/vellum/workflows/graph/graph.py
+++ b/src/vellum/workflows/graph/graph.py
@@ -232,12 +232,19 @@ class Graph:
             return self
 
         if hasattr(other, "Ports"):
+            # This is a workaround for circular imports
+            from vellum.workflows.nodes.bases.base import BaseNode
+
             for final_output_node in self._terminals:
                 if isinstance(final_output_node, NoPortsNode):
                     continue
                 subgraph = final_output_node >> other
                 self._extend_edges(subgraph.edges)
-            self._terminals = {port for port in other.Ports}
+
+            other_ports = {port for port in other.Ports}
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/graph/graph.py
```diff
@@ -207,8 +211,10 @@ def __rshift__(self, other: GraphTarget) -> "Graph":
                     elif hasattr(elem, "Ports"):
                         midgraph = final_output_node >> elem
                         self._extend_edges(midgraph.edges)
-                        for other_terminal in elem.Ports:
-                            new_terminals.add(other_terminal)
+                        other_ports = {port for port in elem.Ports}
+                        if isinstance(elem, type) and issubclass(elem, BaseNode) and not other_ports:
+                            other_ports = {NoPortsNode(elem)}
```
clopen-set, line 216: Should we filter specifically for subclasses of `FinalOutputNode`? I'm not sure if we allow users to define other classes without Ports

dvargasfuertes: I'd rather not reference specific Node classes in base generic logic. We don't allow users to define other classes without Ports, but if we did, I think it's likely we would want this logic to apply to them as well

dvargas92495 (approved):

PR #3445, opened 2025-12-22
Support workflow output reference serialization

vincent0426:
chat message trigger will reference workflow output from a node

### DIFF
--- a/ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_attributes_serialization.py
+++ b/ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_attributes_serialization.py
@@ -190,6 +190,43 @@ def test_serialize_node__lazy_reference_with_string():
     ]
 
 
+def test_serialize_node__lazy_reference_workflow_output():
+    """Test that LazyReference can resolve to workflow output references during serialization."""
+
+    # GIVEN a node with a LazyReference to a workflow output
+    class NodeWithWorkflowOutputReference(BaseNode):
+        workflow_output_ref = LazyReference(lambda: TestWorkflow.Outputs.final_result)
+
+    class TestNode(BaseNode):
+        class Outputs(BaseNode.Outputs):
+            result: str = "test result"
+
+    # AND a workflow that defines an output
+    class TestWorkflow(BaseWorkflow):
+        graph = NodeWithWorkflowOutputReference >> TestNode
+
+        class Outputs(BaseWorkflow.Outputs):
+            final_result = TestNode.Outputs.result
+
+    # WHEN the node is serialized in the context of the workflow
+    workflow_display = get_workflow_display(workflow_class=TestWorkflow)
+    serialized_workflow: dict = workflow_display.serialize()
+
+    # THEN the node should properly serialize the workflow output reference
+    node_with_lazy_reference = next(
+        node
+        for node in serialized_workflow["workflow_raw_data"]["nodes"]
+        if node["id"] == str(NodeWithWorkflowOutputReference.__id__)
+    )
+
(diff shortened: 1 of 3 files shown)


ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -326,6 +326,15 @@ def serialize_value(executable_id: UUID, display_context: "WorkflowDisplayContex
...
     if isinstance(value, OutputReference):
+        if issubclass(value.outputs_class, BaseWorkflow.Outputs):
+            # Only resolve workflow output references that point to node outputs
+            if isinstance(value.instance, OutputReference) and issubclass(
+                value.instance.outputs_class, BaseNode.Outputs
+            ):
+                return serialize_value(executable_id, display_context, value.instance)
+            # For other instances (constants, other workflow outputs), raise an error
+            raise InvalidOutputReferenceError(f"Reference to outputs '{value.outputs_class.__qualname__}' is invalid.")
+
```
vincent0426, line 331: only allow node output to reference workflow output and still raise error when workflow output reference workflow output with this test `test_serialize_workflow__workflow_outputs_reference_non_node_outputs`

ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_attributes_serialization.py
```diff
@@ -190,6 +190,43 @@ class Workflow(BaseWorkflow):
     ]
 
 
+def test_serialize_node__lazy_reference_workflow_output():
+    """Test that LazyReference can resolve to workflow output references during serialization."""
+
+    # GIVEN a node with a LazyReference to a workflow output
+    class NodeWithWorkflowOutputReference(BaseNode):
+        workflow_output_ref = LazyReference(lambda: TestWorkflow.Outputs.final_result)
```
dvargasfuertes, line 198: [nit] Wait we only care about this because of `BaseTrigger.Config` correct? not as a node attribute?

ee/vellum_ee/workflows/display/utils/expressions.py
```diff
@@ -326,6 +326,15 @@ def serialize_value(executable_id: UUID, display_context: "WorkflowDisplayContex
         }
 
     if isinstance(value, OutputReference):
+        if issubclass(value.outputs_class, BaseWorkflow.Outputs):
+            # Only resolve workflow output references that point to node outputs
+            if isinstance(value.instance, OutputReference) and issubclass(
+                value.instance.outputs_class, BaseNode.Outputs
+            ):
```
dvargasfuertes: [nit] why discriminate here? why not just always try to serialize?

I can see us returning workflow outputs that point to one of 3 prompt nodes that coalesce to each other, for example

dvargas92495 (changes requested):

dvargas92495 (approved):
whoops, fat finger

ee/vellum_ee/workflows/display/tests/workflow_serialization/generic_nodes/test_attributes_serialization.py
vincent0426: yes this is just testing serialization for workflow output lazy reference

ee/vellum_ee/workflows/display/utils/expressions.py
vincent0426: doesn't this only block workflow output referencing workflow output that is tested here `test_serialize_workflow__workflow_outputs_reference_non_node_outputs`

dvargasfuertes: `value.instance` could be all sorts of things, including a lazy reference, coalesce expression, etc...

vincent0426:
@codex review

PR #3454, opened 2025-12-23
Add a hacky `_is_vellum_value_subtype` helper

clopen-set:
Used in https://github.com/vellum-ai/vellum-python-sdks/pull/3452

We only want to infer a type of `ArrayVellumValue` if the array elements are subtypes of `VellumValue`.

## Thoughts
1. Our current implementation of `_is_subtype` is broken and it also doesn't account for unions. I decided to implement a helper that was more restricted in scope.
```python
[ins] In [4]: from vellum.workflows.utils.vellum_variables import _is_subtype

[ins] In [5]: from typing import List

[ins] In [6]: _is_subtype(List[str], List)
Out[6]: False

[ins] In [7]: _is_subtype(List[str], list)
Out[7]: True

[ins] In [8]: _is_subtype(list[str], list)
Out[8]: True
```
2. CI should test against all supported versions of Python
3. I don't like this approach, and (not) handling forward refs makes me like it even less. Would be great to move away from handrolled type introspection, and lean towards native python / JSON types

### DIFF
--- a/src/vellum/workflows/utils/tests/test_vellum_variables.py
+++ b/src/vellum/workflows/utils/tests/test_vellum_variables.py
@@ -1,10 +1,12 @@
 import pytest
 from datetime import datetime
-from typing import List, Optional
+from typing import List, Optional, Union
 
 from vellum import ChatMessage, SearchResult, VellumAudio, VellumDocument, VellumImage, VellumValue
+import vellum.client.types as vellum_types
 from vellum.workflows.types.core import Json
 from vellum.workflows.utils.vellum_variables import (
+    _is_vellum_value_subtype,
     primitive_type_to_vellum_variable_type,
     vellum_variable_type_to_openapi_type,
 )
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/utils/tests/test_vellum_variables.py
```diff
@@ -64,3 +66,48 @@ def test_primitive_type_to_vellum_variable_type(type_, expected):
...
+        (vellum_types.ErrorVellumValue, True),
+        (vellum_types.ArrayVellumValue, True),
+        (vellum_types.ChatHistoryVellumValue, True),
+        (vellum_types.SearchResultsVellumValue, True),
+        (vellum_types.ThinkingVellumValue, True),
+        # Misc union types
+        (Union[vellum_types.StringVellumValue, vellum_types.NumberVellumValue], True),
+        (_STRING_OR_NUMBER_VELLUM_VALUE, True),
+        (Union[vellum_types.StringVellumValue], True),
+        (Union["vellum_types.StringVellumValue"], False),
```
clopen-set, line 99: this is a limitation of not resolving forward refs in our helper

src/vellum/workflows/utils/vellum_variables.py
```diff
@@ -213,3 +214,50 @@ def _builtin_list_to_vellum_type(type_: Type) -> Union[str, None]:
             return "ARRAY"
 
     return None
+
+
+def _is_vellum_value_subtype(type_: Type) -> bool:
```
dvargasfuertes, line 219: >  and lean towards native python / JSON types

[q] What would this look like?

dvargas92495 (approved):

PR #3456, opened 2025-12-24
[test pollution] use pytest `monkeypatch` for file operations + setting env variables

clopen-set:
Introducing a new test in https://github.com/vellum-ai/vellum-python-sdks/pull/3454 changed test execution order, which caused unrelated tests to fail

There was some messy combination of file system operations that we didn't clean up (temp dir + chdir) and orphaned `.env` files that our tests relied on. We can clean this up a little with pytest's `monkeypatch` and `tmp_path` fixtures:
- https://docs.pytest.org/en/stable/how-to/monkeypatch.html
- https://docs.pytest.org/en/stable/how-to/tmp_path.html

We should consider using https://github.com/pytest-dev/pytest-randomly to catch these more proactively

ref APO-2447

### DIFF
--- a/ee/vellum_cli/tests/conftest.py
+++ b/ee/vellum_cli/tests/conftest.py
@@ -1,8 +1,6 @@
 import pytest
 from dataclasses import dataclass
-import os
-import shutil
-import tempfile
+import pathlib
 from uuid import uuid4
 from typing import Any, Callable, Dict, Generator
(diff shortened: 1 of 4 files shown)


vincent0426 (commented):
✅

dvargas92495 (commented):
✅

PR #3481, opened 2025-12-31
Introduce `WorkflowContext.emit_log_event`

clopen-set:
First pass at emitting log events to monitoring API

### DIFF
--- a/src/vellum/workflows/state/context.py
+++ b/src/vellum/workflows/state/context.py
@@ -3,10 +3,12 @@ from functools import cached_property
 import json
 from queue import Queue
 from uuid import UUID, uuid4
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Type
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type
 
 from vellum import Vellum, __version__
+from vellum.client.types import SeverityEnum
 from vellum.workflows.context import ExecutionContext, get_execution_context, set_execution_context
+from vellum.workflows.events.node import NodeExecutionLogBody, NodeExecutionLogEvent
 from vellum.workflows.events.types import ExternalParentContext
 from vellum.workflows.nodes.mocks import MockNodeExecution, MockNodeExecutionArg
 from vellum.workflows.outputs.base import BaseOutputs
@@ -18,6 +20,7 @@ from vellum.workflows.vellum_client import create_vellum_client
 
 if TYPE_CHECKING:
     from vellum.workflows.events.workflow import WorkflowEvent
+    from vellum.workflows.nodes.bases import BaseNode
     from vellum.workflows.state.base import BaseState
     from vellum.workflows.workflows.base import BaseWorkflow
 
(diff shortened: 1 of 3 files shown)


src/vellum/workflows/state/context.py
```diff
@@ -146,6 +149,35 @@ def _get_ui_url_from_api_url(self, api_url: str) -> str:
...
+        """Emit a log event for a particular node.
+
+        This is in active development and may have breaking changes.
+        """
+        if self._event_queue is None:
+            return
+
+        execution_context = self.execution_context
+        if execution_context.parent_context is None:
+            span_id = uuid4()
```
clopen-set, line 164: `parent_context` should always be defined because of the fallback logic in the init method

tests/workflows/emit_log_event/workflow.py
```diff
@@ -0,0 +1,26 @@
...
+class LoggingNode(BaseNode):
+    class Outputs(BaseOutputs):
+        result: str
+
+    def run(self) -> Outputs:
+        self._context.emit_log_event(
+            severity="INFO",
+            message="Custom log message",
+            attributes={"key": "value", "count": 42},
+            node=self,
```
clopen-set, line 17: Including `node_definition` on the node log event makes this a little clunky to use, since we need to include a reference to the class definition for each call

tests/workflows/emit_log_event/tests/test_workflow.py
```diff
@@ -0,0 +1,71 @@
...
+
+from tests.workflows.emit_log_event.workflow import EmitLogEventWorkflow, LoggingNode
+
+
+def test_emit_log_event__happy_path():
+    # GIVEN a workflow with a node that emits a log event
+    workflow = EmitLogEventWorkflow()
+
+    # WHEN
+    events = list(workflow.stream(event_filter=all_workflow_event_filter))
```
clopen-set, line 15: By default, only workflow level events are streamed back. We'll still emit the log events to monitoring if the appropriate emitter is registered

src/vellum/workflows/state/context.py
NgoHarrison: [q] Why would this be a breaking change?

NgoHarrison (approved):

src/vellum/workflows/state/context.py
clopen-set: This by itself isn't a breaking change, but we might introduce breaking changes to the function signature as we iterate on this

tests/workflows/emit_log_event/tests/test_workflow.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # THEN
+    log_events = [e for e in events if e.name == "node.execution.log"]
+    assert len(log_events) == 1
+
+    log_event = log_events[0]
```
dvargasfuertes, line 21: [blocking] Can we assert that this event has the same span id as its node initiated and node fulfilled event? I believe I saw a bug in that logic where only the root span id was used

tests/workflows/emit_log_event/workflow.py
dvargasfuertes: [nit] can we make it such that `self._context` here knows which node it's being run a part of? would love to be able to remove this kwarg.

I'd take a look at `get_execution_context()` instead of `self._context.execution_context` for both this and the span id feedback

clopen-set: To clarify, was the suggestion to track the node class as part of the `execution_context` context manager?

The other approach I considered was to traverse the parent contexts until I found a `NodeParentContext` with a `node_definition` to use

Neither of those felt quite right to me since logging would fail if we missed updating context somewhere or if we had a bug in setting parent context

dvargasfuertes: Ideally it's doable from the parent context which could be found on `execution_context`. It should be the first immediate parent in the parent context.

absent that, I believe there are call stack hijinks we could do to trace up to the node class who's run method called this one.

> a bug in setting parent context

This kind of bug would have poor ripple effects in our monitoring layer that would be worse than this method's dx, so I'm bullish on us being incentivized to have proper parent context setting well tested

PR #3482, opened 2026-01-01
Add parallel test execution

vincent0426:


tests/workflows/tool_calling_node_parallel_execution/tests/test_workflow.py
```diff
@@ -0,0 +1,241 @@
...
+    terminal_event = workflow.run()
+    end_time = time.time()
+    total_time = end_time - start_time
+
+    # THEN the workflow is executed successfully
+    assert terminal_event.name == "workflow.execution.fulfilled"
+    assert terminal_event.outputs.text == "All three slow tools executed successfully."
+
+    # AND the execution took approximately 0.5 seconds (parallel: max(0.5, 0.5, 0.5))
+    assert total_time >= 1.5
```
vincent0426, line 105: this should be < 1.5 when its parallel

dvargasfuertes: [oos] We should inline Followup linear tickets with TODO comments if we expect to iterate on test assertions

dvargas92495 (commented):
✅

PR #3517, opened 2026-01-05
Improve tests for log event trace/span ID

clopen-set:
Addressing https://github.com/vellum-ai/vellum-python-sdks/pull/3481#discussion_r2658464725


dvargas92495 (commented):
✅

PR #3521, opened 2026-01-05
Derive node class from parent context when emitting node log events

clopen-set:
Technically a breaking change since we've removed an argument, but this feature is still in development

### DIFF
--- a/src/vellum/workflows/state/context.py
+++ b/src/vellum/workflows/state/context.py
@@ -9,7 +9,7 @@ from vellum import Vellum, __version__
 from vellum.client.types import SeverityEnum
 from vellum.workflows.context import ExecutionContext, get_execution_context, set_execution_context
 from vellum.workflows.events.node import NodeExecutionLogBody, NodeExecutionLogEvent
-from vellum.workflows.events.types import ExternalParentContext
+from vellum.workflows.events.types import ExternalParentContext, NodeParentContext
 from vellum.workflows.nodes.mocks import MockNodeExecution, MockNodeExecutionArg
 from vellum.workflows.outputs.base import BaseOutputs
 from vellum.workflows.references.constant import ConstantValueReference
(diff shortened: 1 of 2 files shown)


src/vellum/workflows/state/context.py
```diff
@@ -160,17 +157,23 @@ def emit_log_event(
...
-            span_id = uuid4()
-        else:
-            span_id = execution_context.parent_context.span_id
+        parent_context = execution_context.parent_context
+        while parent_context:
+            if isinstance(parent_context, NodeParentContext):
+                break
+            parent_context = parent_context.parent
+
+        if not isinstance(parent_context, NodeParentContext):
```
clopen-set, line 171: The actual type is `NodeParentContext | None`, so this is equivalent to
```suggestion
        if not parent_context:
```
but we need to convince mypy of this. We could also introduce a separate variable to help with this

src/vellum/workflows/state/context.py
```diff
@@ -160,17 +157,23 @@ def emit_log_event(
...
+            return
+
+        node_class = parent_context.node_definition.decode()
```
clopen-set: `CodeResourceDefinition.decode` is currently only used in deserialization, and I don't like introducing this dependency here.

It's inconsistent with the other `NodeExecutionXBody` classes, but it'd be simpler to have `NodeExecutionLogBody` directly accept `node_definition: CodeResourceDefinition`

dvargasfuertes: [oos] Oh wow, I thought we had the SDK version of these parent context classes pointing to the node/workflow class itself already... would ideally one day prefer that direction since SDK users (including AB) don't and shouldn't care about `CodeResourceDefinition`. That class is an implementation detail of how we represent them in the Vellum app

dvargasfuertes: [nit] we probably want to wrap a try/except around this `decode` method just in case so that we don't let a bad log event kill the whole node run

dvargas92495 (approved):

clopen-set:
@codex another review

PR #3529, opened 2026-01-06
[APO-2663] Remove codegen chat message trigger message attribute

vincent0426:


ee/codegen/src/generators/triggers/chat-message-trigger.ts
```diff
@@ -38,6 +38,11 @@ export class ChatMessageTrigger extends BaseTrigger<ChatMessageTriggerType> {
     return "ChatMessageTrigger";
   }
 
+  protected createAttributeFields(): AstNode[] {
+    // Don't generate attribute fields for ChatMessageTrigger (message is handled internally)
+    return [];
+  }
```
dvargasfuertes, line 44: [oos] long term, would love to use the default definitions json to drive this sort of exclusion similar to base nodes

dvargas92495 (approved):

PR #3588, opened 2026-01-13
Fix subworkflow node shadowing from star imports

clopen-set:
## Changes
Migrate display `__init__` files away from star imports of the sibling `workflow` modules
```diff
-from .workflow import *
+from . import workflow
```
to prevent shadowing explicitly imported symbols

## Problem
Say we have a workflow structured like this, where `MySubworkflowNode` is an inline subworkflow node:
```
Workflow
|--MySubworkflowNode (InlineSubworkflowNode)
    |--MyNodeWorkflow
        |--MySubworkflowNode (some custom node)
```
See https://github.com/vellum-ai/vellum-python-sdks/pull/3588/commits/2d2d06d334ebbf034cb13e8a05245c22e18e4386#diff-6449b80fb80f0c5e78e24829dc820952afa0379ebdd4e2326359d119fe924302 for directory structure.

Currently we generate a `display` directory for the inner subworkflow node whose `__init__.py` file looks something like
```python
from uuid import UUID

from vellum_ee.workflows.display.editor import NodeDisplayComment, NodeDisplayData, NodeDisplayPosition
from vellum_ee.workflows.display.nodes import BaseInlineSubworkflowNodeDisplay
from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplayOverrides

from ....nodes.node import MySubworkflowNode  # parent node, InlineSubworkflowNode
from .nodes import *
from .workflow import *


class MySubworkflowNodeDisplay(BaseInlineSubworkflowNodeDisplay[MySubworkflowNode]):
    ...
```

and where the sibling `workflow.py` file looks something like
```python

from .nodes.node import MySubworkflowNode  # some child custom node


class Workflow(BaseWorkflow):
    graph = MySubworkflowNode

        result = MySubworkflowNode.Outputs.result
```

The star import from the `workflow` module pulls in the child node, which shadows the intended import of the parent `MySubworkflowNode`, which causes serialization to fail.

## Thoughts
1. Alternatively / additionally, we could define `__all__ = ["MyNodeWorkflowDisplay"]` in `workflow.py` to limit the symbols pulled in by star imports.
2. I think the various `from .nodes import *` imports are also no-ops as of https://github.com/vellum-ai/vellum-python-sdks/pull/3214. Importing the sibling `workflow` module that defines the subclass of `BaseWorkflowDisplay` should register the node classes.
3. Separately, do we want to update codegen to deduplicate node names between a parent workflow and subworkflow?

### DIFF
--- a/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
+++ b/ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
@@ -563,8 +563,8 @@ class GetCurrentWeatherNodeDisplay(BaseNodeDisplay[GetCurrentWeatherNode]):
 exports[`WorkflowProjectGenerator > function > should generate inline workflow tool 'no packages' > code/display/nodes/tool_call_get_current_weather_node/weather_function/__init__.py 1`] = `
 "# flake8: noqa: F401, F403
 
+from . import workflow
 from .nodes import *
-from .workflow import *
 "
 `;
 
@@ -884,8 +884,8 @@ class GetCurrentWeatherNodeDisplay(BaseNodeDisplay[GetCurrentWeatherNode]):
 exports[`WorkflowProjectGenerator > function > should generate inline workflow tool 'with packages' > code/display/nodes/tool_call_get_current_weather_node/weather_function/__init__.py 1`] = `
 "# flake8: noqa: F401, F403
 
+from . import workflow
 from .nodes import *
-from .workflow import *
 "
 `;
 
--- a/ee/codegen/src/__test__/nodes/__snapshots__/inline-subworkflow-node.test.ts.snap
+++ b/ee/codegen/src/__test__/nodes/__snapshots__/inline-subworkflow-node.test.ts.snap
@@ -22,9 +22,9 @@ from vellum_ee.workflows.display.editor import NodeDisplayData, NodeDisplayPosit
 from vellum_ee.workflows.display.nodes import BaseInlineSubworkflowNodeDisplay, BaseRetryNodeDisplay
 from vellum_ee.workflows.display.nodes.types import NodeOutputDisplay, PortDisplayOverrides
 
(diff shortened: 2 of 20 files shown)


ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
```diff
@@ -563,8 +563,8 @@ class GetCurrentWeatherNodeDisplay(BaseNodeDisplay[GetCurrentWeatherNode]):
 exports[`WorkflowProjectGenerator > function > should generate inline workflow tool 'no packages' > code/display/nodes/tool_call_get_current_weather_node/weather_function/__init__.py 1`] = `
 "# flake8: noqa: F401, F403
 
+from . import workflow
 from .nodes import *
```
dvargasfuertes, line 567: [q] This works bc the collision is between both of these star imports correct?

dvargas92495 (approved):
Cool with this as the stop gap solution for now.

Instead of investing more in the serialization/codegen of the display directory, I'd rather make more forward progress on nuking the `display` directory

ee/codegen/src/__test__/__snapshots__/project.test.ts.snap
clopen-set: No, IIUC the
```
from .nodes import *
```
import is a no-op. Previously the `nodes/__init__.py` file re-exported specific display nodes. But we stopped generating the display `nodes/__init__.py` file, so a star import on the package shouldn't do anything.

I've added an example of the conflict to the PR description. The collision happens if we
1. Explicitly import a Node in this `__init__` file for use in a custom node display (e.g. `from ....nodes.node import MySubworkflowNode`)
2. Implicitly import a Node with a conflicting name from a subworkflow via `from workflow import *`, shadowing the correct node from (1)

PR #3622, opened 2026-01-21
Handle scenarios with null file input

clopen-set:
There's a mismatch between (1) our webapp scenario serializers and (2) the serializers used to validate datasets for codegen. (1) Allows null file inputs, but (2) doesn't and marks the inputs as invalid during codegen.

### DIFF
--- a/ee/codegen/src/serializers/vellum.ts
+++ b/ee/codegen/src/serializers/vellum.ts
@@ -20,12 +20,12 @@ import {
 import {
   ChatMessageRole as ChatMessageRoleSerializer,
   LogicalOperator as LogicalOperatorSerializer,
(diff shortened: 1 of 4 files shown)


ee/codegen/src/serializers/vellum.ts
```diff
@@ -2646,7 +2646,7 @@ export declare namespace WorkflowDisplayDataSerializer {
 }
 
 export const WorkflowSandboxInputsSerializer = listSchema(
-  WorkflowInputSerializer
+  NamedTestCaseVariableValueRequestSerializer
```
clopen-set, line 2649: This is the hackier approach that doesn't require us to allow null values in the deployment `WorkflowInputSerializer`

dvargas92495:
discussed offline about moving this over to codegen

PR #3625, opened 2026-01-21
Fix resolve_node_ref suffix matching for dynamically loaded workflows

awlevin:
## Summary
- Fixes `resolve_node_ref` to use suffix matching when resolving node references in dynamically loaded workflows
- When workflow files are loaded via VirtualFileFinder, nodes get a UUID namespace prefix (e.g., `{uuid}.nodes.my_node.MyNode`)
- The API passes node references without the prefix (e.g., `nodes.my_node.MyNode`)
- Previously, the fallback path did exact string matching which failed
- Now uses suffix matching: `full_path.endswith(f".{node_ref}")`

## Test plan
- [x] Added test `test_resolve_node_ref__suffix_matching_with_namespace_prefix` that verifies:
  - Workflow loads with UUID namespace
  - Node's `__module__` has namespace prefix
  - `run_node("nodes.my_node.MyNode")` resolves correctly via suffix matching
  - Node executes and produces expected output

## Context
This fixes the `workflows.execute_node` API endpoint which was returning "Node Execution failed to terminate properly" for all node executions. The issue was that the node reference couldn't be resolved because of the namespace prefix mismatch.


src/vellum/workflows/workflows/base.py
```diff
@@ -686,7 +686,9 @@ def resolve_node_ref(cls, node_ref: Union[Type[BaseNode], UUID, str]) -> Type[Ba
                 )
             except (ValueError, ModuleNotFoundError, AttributeError):
                 for node in candidate_nodes:
-                    if f"{node.__module__}.{node.__name__}" == node_ref:
+                    full_path = f"{node.__module__}.{node.__name__}"
+                    # Check exact match OR suffix match (for dynamically loaded workflows with namespace prefix)
+                    if full_path == node_ref or full_path.endswith(f".{node_ref}"):
                         return node
```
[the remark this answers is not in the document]
awlevin: yeah, had considered this, but unsure right fix. maybe instead we should just omit the uuid module name and do an exact match on the remainder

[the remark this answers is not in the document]
dvargasfuertes: We likely want to move `src/vellum/workflows/utils/uuids.py::_normalize_module_path` into its own util file and reuse here

tests/workflows/test_base_workflow.py
```diff
@@ -434,3 +435,87 @@ def test_deserialize_trigger__raises_error_when_trigger_name_not_found():
...
     # AND the error message should list available trigger names
     assert "Available trigger names" in str(exc_info.value)
+
+
+@pytest.fixture
+def virtual_file_loader():
+    """Fixture to manage VirtualFileFinder registration and cleanup."""
+    try:
+        from vellum_ee.workflows.server.virtual_file_loader import VirtualFileFinder
+    except ImportError:
```
dvargasfuertes: [nit] let's move this test to the `ee/vellum_ee` directory

tests/workflows/test_base_workflow.py
```diff
@@ -434,3 +435,87 @@ def test_deserialize_trigger__raises_error_when_trigger_name_not_found():
...
+        return finder
+
+    yield _register
+
+    for finder in reversed(finders):
+        if finder in sys.meta_path:
+            sys.meta_path.remove(finder)
+
+
+def test_resolve_node_ref__suffix_matching_with_namespace_prefix(virtual_file_loader):
```
dvargasfuertes: This test should perfectly repro

ee/vellum_ee/workflows/tests/test_resolve_node_ref.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # AND the last event should be a fulfilled event with the correct output
+    fulfilled_events = [e for e in events if e.name == "node.execution.fulfilled"]
+    assert len(fulfilled_events) == 1
+
+    outputs = fulfilled_events[0].outputs
+    assert outputs.greeting == "Hello, World!"
+
+
+def test_resolve_node_ref__node_with_relative_imports(virtual_file_loader):
```
dvargasfuertes, line 118: [nit] Not following what these second two tests verify that isn't captured by the first

dvargas92495 (approved):

ee/vellum_ee/workflows/tests/test_resolve_node_ref.py
awlevin: the workflow had a relative import in the first but i wanted to ensure that run_node works correctly when the node itself has relative imports, and also when the node's attributes/inputs are specified from relative imports

awlevin: likely overkill, can remove if it would be cleaner

PR #3630, opened 2026-01-22
Add validation guarding against parent node output references

clopen-set:
Prevent users from defining custom `Outputs` classes where an output references something from the parent class
```python
class MyPromptNode(InlinePromptNode):
    class Outputs(InlinePromptNode.Outputs):
        custom_output: str = InlinePromptNode.Outputs.text
```

Here `custom_output` doesn't get resolved during execution and remains `undefined`. Separately, I think custom outputs for an `InlinePromptNode` will just get dropped during serialization and possibly should be forbidden

### DIFF
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -75,17 +75,16 @@ def _validate_no_parent_output_references(node_cls: Type["BaseNode"]) -> None:
     """
     Validates that the node does not reference parent class outputs.
     """
-    class_outputs = vars(node_cls.Outputs)
-
-    for name, value in class_outputs.items():
-        if not isinstance(value, OutputReference):
+    for node_output in node_cls.Outputs:
+        node_value = node_output.instance
+        if not isinstance(node_value, OutputReference):
             continue
 
-        parent_node_class = value.outputs_class.__parent_class__
+        parent_node_class = node_value.outputs_class.__parent_class__
         if parent_node_class in node_cls.__mro__:
             raise ValueError(
-                f"'{node_cls.Outputs.__qualname__}.{name}' references parent class output "
-                f"'{value.outputs_class.__qualname__}.{value.name}'. "
+                f"'{node_cls.Outputs.__qualname__}.{node_output.name}' references parent class output "
+                f"'{node_value.outputs_class.__qualname__}.{node_value.name}'. "
                 "Referencing outputs from a node's parent class is not allowed."
             )
 
--- a/src/vellum/workflows/nodes/bases/base.py
+++ b/src/vellum/workflows/nodes/bases/base.py
@@ -71,6 +71,25 @@ def _is_annotated(cls: Type, name: str) -> Any:
     return None
 
 
+def _validate_no_parent_output_references(node_cls: Type["BaseNode"]) -> None:
+    """
+    Validates that the node does not reference parent class outputs.
+    """
+    class_outputs = vars(node_cls.Outputs)
+
+    for name, value in class_outputs.items():
+        if not isinstance(value, OutputReference):
+            continue
+
+        parent_node_class = value.outputs_class.__parent_class__
+        if parent_node_class in node_cls.__mro__:
+            raise ValueError(
+                f"Node {node_cls.__name__} references parent class output "
+                f"'{value.outputs_class.__qualname__}.{name}'. "
+                "Referencing parent class outputs is not allowed."
+            )
+
+
 class BaseNodeMeta(ABCMeta):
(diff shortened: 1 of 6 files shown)


src/vellum/workflows/nodes/displayable/bases/inline_prompt_node/node.py
```diff
@@ -565,6 +586,8 @@ def __validate__(cls) -> None:
         Raises:
             jsonschema.exceptions.SchemaError: If the JSON schema structure is invalid
         """
+        _validate_no_parent_output_references(cls)
```
clopen-set: In general, do we expect referencing a parent class's `Outputs` to work? Poked around at the output resolution logic and I'd expect the answer is no

I initially restricted this validation to `InlinePromptNode` to play it safe. Expanding this to run for all nodes causes a single test to fail:
https://github.com/vellum-ai/vellum-python-sdks/blob/e14d1ded3f4852bbd2c48f8f61e7803f043271c3/src/vellum/workflows/nodes/displayable/final_output_node/tests/test_node.py#L100-L106
and I'm not sure if this is a valid or expected usage pattern

dvargasfuertes: That test looks off and I would expect this to run for all nodes. The linked test seems to have the goal of referencing an upstream node that has a `Dict` type, so we should reformat it as such

clopen-set: Fixed the test and moved validation to `BaseNode`

CI on https://github.com/vellum-ai/vellum-python-sdks/pull/3630/commits/0f19fd0376d83275a5f1a2ec60710afc22eb98f1 failed because of a test flake. I'll put up a fix separately

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -71,6 +71,25 @@ def _is_annotated(cls: Type, name: str) -> Any:
     return None
 
 
+def _validate_no_parent_output_references(node_cls: Type["BaseNode"]) -> None:
+    """
+    Validates that the node does not reference parent class outputs.
+    """
+    class_outputs = vars(node_cls.Outputs)
+
+    for name, value in class_outputs.items():
```
dvargasfuertes: [nit] we support iterating over the node class: `for reference, value in node_cls.Outputs:`

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -591,4 +610,4 @@ def __validate__(cls) -> None:
 
         Default implementation performs no validation.
```
dvargasfuertes: [nit] seems like we can nuke this line now

dvargas92495 (approved):

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -71,6 +71,28 @@ def _is_annotated(cls: Type, name: str) -> Any:
...
+    Validates that the node does not reference parent class outputs.
+    """
+    errors = []
+    for node_output in node_cls.Outputs:
+        node_value = node_output.instance
+        if not isinstance(node_value, OutputReference):
+            continue
+
+        parent_node_class = node_value.outputs_class.__parent_class__
+        if parent_node_class in node_cls.__mro__:
```
clopen-set, line 85: Is it valid to reference a sibling value in the same `Outputs` class? e.g.
```python
class CustomNode(BaseNode):
    class Outputs:
        a: str = SomeOtherNode.Outputs.str_output
        b: str = a
    ...
```

From some quick testing the `node_value` for `b` points to `SomeOtherNode.Outputs.str_output` and it is correctly marked as valid

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -71,6 +71,28 @@ def _is_annotated(cls: Type, name: str) -> Any:
     return None
 
 
+def _validate_no_parent_output_references(node_cls: Type["BaseNode"]) -> None:
+    """
+    Validates that the node does not reference parent class outputs.
+    """
+    errors = []
+    for node_output in node_cls.Outputs:
```
clopen-set, line 79: re: https://github.com/vellum-ai/vellum-python-sdks/pull/3630#discussion_r2718678549

we're iterating on a subclass of `BaseOutputs` and not an instance, so we need to fetch the values separately

src/vellum/workflows/nodes/bases/base.py
dvargasfuertes: Yea that's just standard output referencing of upstream nodes and how final output nodes work

dvargas92495 (approved):

PR #3632, opened 2026-01-22
Fix flaky `multi_branch_merge_loop` test

clopen-set:
This is an attempt to fix flakiness in `tests/workflows/multi_branch_merge_loop/tests/test_workflow.py`

We run nodes in different threads, so there's no guarantee on when they'll be scheduled and `time.sleep` isn't actually reliable. We can add a short `time.sleep(0.0145)` before or after
https://github.com/vellum-ai/vellum-python-sdks/blob/c74400d17c41c96b2860607593b7fe84f9c41d5b/tests/workflows/multi_branch_merge_loop/workflow.py#L57
to simulate a pause in `LoopNode` execution and we'll sporadically see 3 loop iterations. We can use threading events to coordinate and make the test deterministic

### DIFF
--- a/tests/workflows/multi_branch_merge_loop/workflow.py
+++ b/tests/workflows/multi_branch_merge_loop/workflow.py
@@ -1,4 +1,4 @@
-import time
+import threading
 
 from vellum.workflows import BaseWorkflow
 from vellum.workflows.inputs.base import BaseInputs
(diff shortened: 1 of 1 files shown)


dvargas92495 (approved):

NgoHarrison (commented):
✅

PR #3697, opened 2026-02-03
[APO-2918] Handle chat message trigger string type

vincent0426:
Preserves string messages as-is in `ChatMessageTrigger.__init__` instead of converting them to `[StringChatMessageContent(...)]`. Nodes that annotate trigger attributes as `str` now receive the raw string directly from state. Serialization still outputs array format via an overridden `to_trigger_attribute_values()`, and `bind_to_state()` stores the raw value so downstream resolution preserves the original type.

---

## Review & Testing Checklist for Human

- [ ] `_message_as_array` helper is defined but appears unused in the diff — confirm it's needed or remove it
- [ ] `bind_to_state` duplicates iteration logic from `BaseTrigger` to skip the `to_trigger_attribute_values` array conversion — verify this stays in sync if `BaseTrigger.bind_to_state` changes
- [ ] When `message` is a string, `__on_workflow_initiated__` creates a `ChatMessage` with `text=` instead of `content=` — confirm downstream consumers of `state.chat_history` handle both shapes correctly
- [ ] Test a workflow where a node references `ChatMessageTrigger.message` as `List[ArrayChatMessageContentItem]` with a string input to confirm it receives the raw string (not an array), and that this is the desired behavior
- [ ] Verify push/pull serialization round-trips still produce `[{"type": "STRING", "value": "..."}]` for string messages in dataset rows

### Notes

- Requested by: @vincent0426
- Session: https://app.devin.ai/sessions/cf614a304d93425e851ad3545530d45c

### DIFF
--- a/src/vellum/workflows/descriptors/utils.py
+++ b/src/vellum/workflows/descriptors/utils.py
@@ -1,7 +1,7 @@
 from collections.abc import Mapping
 import dataclasses
 import inspect
-from typing import Any, Dict, Optional, Sequence, Set, Tuple, Type, TypeVar, Union, cast, overload
+from typing import Any, Dict, Optional, Sequence, Set, Type, TypeVar, Union, cast, overload
 from typing_extensions import TypeGuard
 
 from pydantic import BaseModel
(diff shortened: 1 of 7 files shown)


src/vellum/workflows/triggers/chat_message.py
```diff
@@ -45,10 +45,7 @@ def __init__(self, **kwargs: Any):
         # Convert message from VellumValue format to ChatMessageContent format if needed
         if "message" in kwargs:
             message = kwargs["message"]
-            # Handle string messages by converting to a list with a single StringChatMessageContent
-            if isinstance(message, str):
-                kwargs["message"] = [StringChatMessageContent(value=message)]
-            elif isinstance(message, list):
+            if isinstance(message, list):
                 converted_message = []
```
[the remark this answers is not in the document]
vincent0426: @dvargas92495 we could define a custom reference for chat message trigger and resolve to the string value if its a single string list?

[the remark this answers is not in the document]
dvargasfuertes: I would rather tackle this on the execution side rather than on the serialization side.

Meaning, if Chat Message Trigger sees a `[{type:STRING, value: ""}]`, but the `message` attribute is annotated as just `str`, then we treat as just `str`

src/vellum/workflows/descriptors/utils.py
```diff
@@ -40,7 +63,12 @@ def resolve_value(
         return cast(_T, value)
 
     if isinstance(value, BaseDescriptor):
-        resolved_value = value.resolve(state)
+        from vellum.workflows.references.trigger import TriggerAttributeReference
+
+        if isinstance(value, TriggerAttributeReference):
+            resolved_value = value.resolve(state, expected_types)
```
vincent0426: hmm don't think we prefer this pattern

vincent0426:
@codex review

vincent0426:
@codex review

src/vellum/workflows/descriptors/utils.py
vincent0426: don't think we want to make this change? but we somehow need to get type annotation during `resolve_value`

dvargasfuertes: Is there a way we could just check in the chat trigger side? meaning that we always resolve the array, but in the parse method of the chat trigger we cast to str if we're being told to

vincent0426:
@codex review

vincent0426:
@codex review

vincent0426:
@codex review

src/vellum/workflows/nodes/bases/base.py
```diff
@@ -563,6 +564,12 @@ def __init__(
                 continue
 
             resolved_value = resolve_value(descriptor.instance, self.state, path=descriptor.name, memo=inputs_memo)
+            if isinstance(descriptor.instance, TriggerAttributeReference) and descriptor.types:
+                coerced_value = descriptor.instance.coerce_resolved_value(resolved_value, descriptor.types)
+                if coerced_value is not resolved_value:
+                    resolved_value = coerced_value
+                    if descriptor.name in inputs_memo:
+                        inputs_memo[descriptor.name] = resolved_value
```
dvargasfuertes: Ah I was hoping this just happened in the trigger's init instead of in downstream nodes. This won't work if we have a binary expression of trigger references, etc

vincent0426: Hmm i don't think we can get descriptor type in trigger init though

dvargasfuertes: let's close this PR when we can

PR #3719, opened 2026-02-05
Integrate workflow dependency type

vincent0426:


ee/vellum_ee/workflows/display/nodes/vellum/inline_prompt_node.py
```diff
@@ -73,11 +75,10 @@ def serialize(
         model = display_context.ml_models_map.get(ml_model)
```
dvargasfuertes, line 75: [nit] let's also update the `hosted_by` in `ml_models_map` to use the same enum as the one from `WorkflowModelProviderDependency`, bc otherwise we will have providers fail to deserialize since we are using a hard coded integration dependency there

dvargas92495 (approved):

vincent0426:
@codex review

ee/vellum_ee/workflows/display/utils/dependencies.py
```diff
@@ -26,4 +26,12 @@ class MLModel(UniversalBaseModel):
...
     name: str
-    hosted_by: MLModelHostingInterface
+    hosted_by: MlModelHostingInterface
+
+    @field_validator("hosted_by", mode="before")
+    @classmethod
+    def validate_hosted_by(cls, value: str, info: ValidationInfo) -> str:
+        """Validate that hosted_by is one of the valid literal values."""
+        if value not in VALID_HOSTING_INTERFACES:
+            raise ValueError(f"Invalid hosting interface: {value}. Must be one of {sorted(VALID_HOSTING_INTERFACES)}")
```
dvargasfuertes, line 36: [oos] Pydantic already fails this case by default. Would maybe be neat to default to some value, or remove the check if continues to be redundant

dvargas92495 (approved):
✅

ee/vellum_ee/workflows/display/utils/dependencies.py
vincent0426: fern generates `typing.Any` here so the test
- test_parse_ml_models__skips_invalid_models_and_returns_valid_ones
- test_parse_ml_models__returns_empty_list_when_all_models_invalid

fails https://github.com/vellum-ai/vellum-python-sdks/blob/78791bebfc2a13a4fcbc647596a56acfc598f55a/src/vellum/client/types/ml_model_hosting_interface.py#L5-L39

PR #3722, opened 2026-02-05
Migrate last direct usage of `MypyTypeAlias` -> `_create_type_alias`

clopen-set:
Building off https://github.com/vellum-ai/vellum-python-sdks/pull/3590, I noticed some crashes when testing mypy 1.19.1

ref APO-2945


dvargas92495 (commented):
✅

PR #3723, opened 2026-02-06
Emit log events for code exec nodes

clopen-set:
ref: APO-2949

Integrate the logs emitted by code exec with the recently introduced log events infra

Screenshot 2026-02-06 at 10 38 46 AM
Screenshot 2026-02-06 at 10 38 50 AM

### DIFF
--- a/src/vellum/workflows/nodes/displayable/code_execution_node/node.py
+++ b/src/vellum/workflows/nodes/displayable/code_execution_node/node.py
@@ -105,7 +105,7 @@ class CodeExecutionNode(BaseNode[StateType], Generic[StateType, _OutputType], me
         code, filepath = self._resolve_code()
         if not self.packages and self.runtime == "PYTHON_3_11_6" and not self._has_secrets_in_code_inputs():
             logs, result = run_code_inline(code, self.code_inputs, output_type, filepath, self._context.vellum_client)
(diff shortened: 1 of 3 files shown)


clopen-set:
@codex review again

src/vellum/workflows/nodes/displayable/code_execution_node/node.py
```diff
@@ -130,7 +130,14 @@ def run(self) -> Outputs:
                     message=f"Expected an output of type '{expected_output_type}', received '{actual_type}'",
                 )
 
-            return self.Outputs(result=code_execution_result.output.value, log=code_execution_result.log)
+            logs = code_execution_result.log
+            outputs = self.Outputs(result=code_execution_result.output.value, log=logs)
+
+        self._context.emit_log_event(
+            severity="INFO",
+            message=logs,
```
clopen-set, line 138: The node concatenates these logs into one giant string. I don't think we can distinguish between discrete `print` statements and `print` statements that happen to contain new line

For readability in UI / convenience we may want just split on new lines and emit an event for each line

There's also an edge case where we can't distinguish between code exec printing a single empty string vs no output at all

dvargas92495 (approved):

PR #3735, opened 2026-03-05
Fix codegen incorrectly relativizing imports sharing top-level package

vincent0426:
## Summary
- When a workflow has a multi-segment module name (e.g., `org.app.support.my_workflow`), the codegen's `writeImports` only checked the first path element (`path[0]`) to decide whether to relativize an import
- This caused imports from sibling sub-packages (e.g., `org.shared.utils`) to be incorrectly converted to relative imports with too many dots (e.g., `......shared.utils`), producing "attempted relative import beyond top-level package" errors in the sandbox
- Adds `rootModulePath` to `PythonFile` to define the workflow's module boundary — only imports starting with this prefix are relativized
- When `rootModulePath` is not provided, the old first-element check is preserved for backward compatibility

## Test plan
- [x] Added unit test that reproduces the bug (deep module path + sibling sub-package import)
- [x] Added unit test for existing behavior (same module tree imports still relativized)
- [x] Full codegen test suite passes (736 tests, 0 failures)

### DIFF
--- a/ee/codegen/src/__test__/python-file-imports.test.ts
+++ b/ee/codegen/src/__test__/python-file-imports.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from "vitest";
+
+import { PythonFile } from "src/generators/extensions/protected-python-file";
+import { Reference } from "src/generators/extensions/reference";
+
+describe("PythonFile import writing", () => {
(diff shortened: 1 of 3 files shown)


dvargas92495 (commented):
✅ thank you!

PR #3741, opened 2026-07-26
Exclude unpopulated values from get_mean_metric_output denominator

arthi-arumugam-git:
## Problem

`VellumTestSuiteRunResults.get_mean_metric_output` filters `None` out of the numerator but divides by the length of the unfiltered list:

```python
return sum(cast(Iterable[float], filter(lambda o: isinstance(o, float), output_values))) / len(output_values)
```

`get_numeric_metric_output_values` is declared `-> List[float | None]`, and `TestSuiteRunMetricNumberOutput.value` is `Optional[float]`, so `None` is a legal and reachable element. Any test suite run where some executions produced no score returns a mean that is silently too low. There is no exception and no warning, so the caller has no way to notice.

## Reproduction

```python
results.get_mean_metric_output()  # executions scored [1.0, None, 1.0, None]
```

Before: `0.5`
After: `1.0`

When every value is `None` the old code returned `0.0`, which reads as a real score of zero. It now raises `TestSuiteRunResultsException` instead.

## Fix

Build the populated list once and use it for both the sum and the count.

## Test

`src/vellum/evaluations/tests/test_resources.py` covers the mixed case. I confirmed it fails on `main` with `assert 0.5 == 1.0` and passes with this change. `black`, `isort`, `flake8` and `mypy` are clean on both touched files.

## Related, not included here

`get_min_metric_output` and `get_max_metric_output` use the same `isinstance(o, float)` filter, and on an all-`None` set they leak a bare `ValueError: min() iterable argument is empty` rather than `TestSuiteRunResultsException`. I left that out to keep this to one change and one test. Happy to open a follow-up if you want it.


arthi-arumugam-git:
A note while this waits, in case it helps triage.

No workflow run has started on this one, which I think is the fork approval gate rather than anything wrong with the branch, so there is no CI signal here either way. Happy to run anything locally and paste the output if that is easier than approving the run.

The behaviour: `get_mean_metric_output` divided by the count of all values rather than the count of populated ones, so any unpopulated entry pulled the mean toward zero. A metric that is missing for half a dataset reported as though it had scored zero on that half. The test in the PR fails on main.

